Issue #435, opened 2014-10-10
Golang support


kirs:
Hey,
I'm willing to add golang support to Hound.

What do the maintainers think about it?

salbertson:
Done, https://github.com/thoughtbot/hound/pull/852 and https://github.com/thoughtbot/hound-go.

@kirs thanks for the initial push for Go!

kirs:
Wow, thanks!

shurcooL:
This issue being closed leads me to think that there may be Go support (yay!), but the homepage does not:

![image](https://cloud.githubusercontent.com/assets/1924134/10229930/6f258512-682f-11e5-8d7f-d6a2e1f4ba5f.png)

Is Go officially supported?

teoljungberg:
@shurcooL It's currently in beta, and you can activate it by adding:

```yaml
go:
  enabled: true
```

To your `.hound.yml`. Once it's considered stable, we'll officially document it

gylaz:
Go has moved out of beta, and it should be reviewed by default.

croaky:
Do we have icons/illustrations for the homepage for the new lingers out of beta? I think Lydia or Devon were working on them?

PR #448, opened 2014-10-16
Go linter

kirs:
As we discussed in #435
Requires `go` and `golint` installed on server.


gylaz:
@croaky @srt32 @bernerdschaefer Do you think that it would be useful in having Hound support Go?

srt32:
@gylaz I think now is not a great time for us to begin Go support. Given our focus on stability and languages for which we know we have user demand perhaps we should revisit this idea in a few months. Regardless, @kirs thanks for putting together this PR!

croaky:
> Given our focus on stability and languages for which we know we have user demand perhaps we should revisit this idea in a few months. 

I think we've addressed most stability issues at this point? This thread is also showing the user demand for this language?

gylaz:
> This thread is also showing the user demand for this language?

Can you expand on this more, @croaky? The only person outside of our team to comment on this thread is the original PR creator.

salbertson:
@kirs, I opened a new PR with a couple updates. https://github.com/thoughtbot/hound/pull/678

Issue #472, opened 2014-11-05
Idea: provide gem dependent on Hound's RuboCop dependency


dnrce:
It would be nice for developers to have a way to keep their local RuboCop version in line with Hound's, in order for local results to be the same as Hound results. Plus, new RuboCop versions aren't always 100% configuration-compatible.

Suggestion: publish a hound-rubocop (or otherwise-named) gem that is empty except for a dependency on a particular version of RuboCop. Hound itself would depend on the latest version of this gem. Developers could also depend on it to ensure a matching RuboCop version.

PR #488, opened 2014-11-14
Move rubocop dependency and configuration to hound-rubocop gem

theodorton:
Attempt to solve #472 and #480

The gem is located at http://github.com/theodorton/hound-rubocop - I was thinking in the lines of:
- Version always matches `rubocop` version, and using a 4th decimal for patches, e.g. `0.25.0.1`
- Developers can add `gem 'hound-rubocop', group: :development` to their `Gemfile` and be sure they're using the same configuration as `hound`- no surprises when opening the PR and no headache trying to get the configuration right

I haven't published the gem to Rubygems, I think it's approperiate that someone at thoughtbot does the publishing.


salbertson:
I would rather not move the configuration out of Hound's codebase while adding another dependency.

e2:
@salbertson - I think you misunderstand: the dependency is the point - because projects using Hound can lock to it. This way the same version of RuboCop used by Hound could be always be used locally in the project. 

Otherwise everyone would have to manually track which version of RuboCop is used by Hound and lock to than in *every* project.

Unless you're declaring Hound will forever support only Rubocop 0.25.0...

Otherwise it would at least be fair for Hound to warn if the given project depends on a conflicting version of Rubocop.

e2:
I do think having the style in moved outside Hound is pretty useless - since Hound doesn't support inheritance anyway (which means users cannot "override" parts of Hound's style - they'd have to copy the whole file into their project and add/override what they need).

PR #495, opened 2014-11-22
Add support for PHP.

blainesch:
This supports allowing you to create a ruleset for your php projects.

This does not support everything:
1) Custom violations.
2) Include/exclude directories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ cache: bundler
 rvm: 2.1.5
 
 before_install:
+  - sudo apt-get install php5
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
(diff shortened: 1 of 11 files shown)


nodesocket:
Any update on this? Is php going to be supported in hound? We'd love to use it.

gylaz:
Eventually, yes. There are other linters we are focusing on right now. It's a risk for us to support languages we don't use internally, thus priority for such languages is lower.

salbertson:
@blainesch, we are moving to a service based approach for reviewing languages that don't have native Ruby libraries. In the case of PHP, when we add support it will most likely be written in PHP and be run in an external service.

https://github.com/thoughtbot/hound/pull/678

Thank you for your time and the work you put into this change, it definitely helped prove a need for PHP support. Thanks again.

Issue #394, opened 2014-09-18
Have Hound comment on poorly formatted commit messages


linduxed:
Tim Pope has [a guide for how to write proper commit messages](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) which is by many projects considered to be the *required* way to write commit messages. thoughtbot's style guide includes a link to this article as well.

While I haven't dug deeply into whether this is easily achieved with the current code of hound, I think it would be good if hound could comment inside offending commits, encouraging better commit messages.

PR #468, opened 2014-11-03
[WIP] Comment on commits

bartoszkopinski:
This is addressing #271 - let me know your thoughts.

PR #499, opened 2014-12-03
WIP: Comment on commit messages

Koronen:
This code implements (parts of) the functionality discussed in #394. The discussion ended rather quickly, but I decided to go ahead and implement it.

I've been sitting on this code for a couple of months now, and I though I might as well share it. While the pre-squash-and-rebase version of it has been running smoothly on a separate instance, there are some things I'd like to fix with the current implementation. One such thing is to make this functionality opt-in (as mentioned in #394).

I'd love some more feedback on this. Mainly, is there sufficient interest in seeing this feature completed?

### DIFF
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -1,6 +1,7 @@
 class Commit
   pattr_initialize :repo_name, :sha, :github
   attr_reader :repo_name, :sha
+  attr_accessor :message
 
   def file_content(filename)
     contents = @github.file_contents(repo_name, filename, sha)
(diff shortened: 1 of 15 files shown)


salbertson:
@Koronen, a few users have mentioned they don't use Pull Requests and they would like Hound to review commits that aren't part of a Pull Request.

Is that what this change does?

According to existing users it's a good idea. @gylaz, thoughts?

Koronen:
@salbertson no, I believe that would be #468. This PR adds support for linting commit messages, like checking to see if the subject line exceeds the [recommended maximum of 50 characters](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

salbertson:
@Koronen, ah, ok. Thanks for clarifying.

I wouldn't want Hound to comment on poorly formatted commit messages. For one, the commit message doesn't live on in the source code and wouldn't decrease readability like inconsistent code style can.

Second, I'm not sure there is enough consensus around what makes a "good" commit message. That means we would have to disable it by default, which I would rather not do.

@Koronen, thanks for the PR but I don't think we will be merging this.

@gylaz, thoughts?

gylaz:
@Koronen Thanks for the PR. Lets revisit this later when we support more languages. I'm going to close it for now.

PR #543, opened 2014-12-30
Display failure status when config is invalid

malandrina:
This PR is a continuation of https://github.com/thoughtbot/hound/pull/525.

* Hound users currently receive no feedback when their Ruby, JavaScript,
  and CoffeeScript configuration files are invalid. This commit
  introduces a new feature to address this disconnect:
  When a config file is empty or can't be parsed, the GitHub status will
  be changed to "failure".
* Validate repo config in `BuildRunner#run`. If config is invalid,
  create failure status and create build record.
* Add `RepoConfig#invalid` and `RepoConfig#load_style_guides`
* Add `GithubApi#create_failure_status`
* Refactor constructor of StyleChecker to accept repo config
* Remove null constraint and presence validation for violation filename
* https://trello.com/c/JLt3FVUb/397-use-status-api-to-tell-user-when-custom-config-is-broken

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -8,7 +8,11 @@ class RepoConfig
   HOUND_CONFIG_FILE = ".hound.yml"
   STYLE_GUIDES = %w(ruby coffee_script java_script)
 
-  pattr_initialize :commit
+  attr_reader :errors
+
+  pattr_initialize :commit do
+    @errors = []
+  end
 
   def enabled_for?(style_guide_name)
     style_guide_name == "ruby" && legacy_config? ||
@@ -36,6 +40,15 @@ class RepoConfig
       ignore_file_content.split("\n")
     else
       []
+    end
+  end
+
+  def validate
+    @errors = invalid_style_guides.map do |style_guide|
+      I18n.t(
+        "invalid_config",
+        config_file_name: config_path_for(style_guide)
+      )
     end
   end
 
@@ -93,5 +106,11 @@ class RepoConfig
     JSON.parse(content)
   rescue JSON::ParserError
     {}
+  end
+
+  def invalid_style_guides
+    STYLE_GUIDES.select do |style_guide|
+      enabled_for?(style_guide) && self.for(style_guide).empty?
+    end
   end
 end
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -8,11 +8,7 @@ class RepoConfig
   HOUND_CONFIG_FILE = ".hound.yml"
   STYLE_GUIDES = %w(ruby coffee_script java_script)
 
-  attr_reader :errors
-
-  pattr_initialize :commit do
-    @errors = []
-  end
+  pattr_initialize :commit
 
   def enabled_for?(style_guide_name)
     style_guide_name == "ruby" && legacy_config? ||
@@ -20,15 +16,17 @@ class RepoConfig
   end
 
   def for(style_guide_name)
-    if style_guide_name == "ruby" && legacy_config?
-      hound_config
-    else
-      config_file_path = config_path_for(style_guide_name)
-
-      if config_file_path
-        load_file(config_file_path, FILE_TYPES[style_guide_name])
+    style_guides[style_guide_name] ||= begin
+      if style_guide_name == "ruby" && legacy_config?
+        hound_config
       else
-        {}
+        config_file_path = config_path_for(style_guide_name)
+
+        if config_file_path
+          load_file(config_file_path, FILE_TYPES[style_guide_name])
+        else
+          {}
+        end
       end
     end
   end
@@ -43,13 +41,16 @@ class RepoConfig
     end
   end
 
-  def validate
-    @errors = invalid_style_guides.map do |style_guide|
-      I18n.t(
-        "invalid_config",
-        config_file_name: config_path_for(style_guide)
-      )
+  def load_style_guides
+    @invalid = false
+
+    STYLE_GUIDES.each do |style_guide|
+      self.for(style_guide)
     end
+  end
+
+  def invalid?
+    @invalid == true
   end
 
   private
@@ -99,18 +100,18 @@ class RepoConfig
   def parse_yaml(content)
     YAML.load(content)
   rescue Psych::SyntaxError
+    @invalid = true
     {}
   end
 
   def parse_json(content)
     JSON.parse(content)
   rescue JSON::ParserError
+    @invalid = true
     {}
   end
 
-  def invalid_style_guides
-    STYLE_GUIDES.select do |style_guide|
-      enabled_for?(style_guide) && self.for(style_guide).empty?
-    end
+  def style_guides
+    @style_guides ||= {}
   end
 end
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -8,7 +8,11 @@ class RepoConfig
   HOUND_CONFIG_FILE = ".hound.yml"
   STYLE_GUIDES = %w(ruby coffee_script java_script)
 
-  pattr_initialize :commit
+  attr_reader :style_guides
+
+  pattr_initialize :commit do
+    @style_guides = {}
+  end
 
   def enabled_for?(style_guide_name)
     style_guide_name == "ruby" && legacy_config? ||
@@ -109,9 +113,5 @@ class RepoConfig
   rescue JSON::ParserError
     @invalid = true
     {}
-  end
-
-  def style_guides
-    @style_guides ||= {}
   end
 end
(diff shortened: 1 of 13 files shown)


app/models/repo_config.rb
```diff
@@ -86,12 +100,18 @@ def load_javascript_ignore
...
   def parse_json(content)
     JSON.parse(content)
   rescue JSON::ParserError
+    @invalid = true
     {}
   end
+
+  def style_guides
+    @style_guides ||= {}
+  end
```
gylaz: How about moving this into the initializer?
```rb
pattr_initialize :commit do
  @style_guides = {}
end
```

app/models/style_checker.rb
```diff
@@ -2,9 +2,10 @@
 # Builds style guide based on file extension.
 # Delegates to style guide for line violations.
 class StyleChecker
-  def initialize(pull_request)
+  def initialize(pull_request, repo_config)
     @pull_request = pull_request
     @style_guides = {}
+    @config = repo_config
```
gylaz: Should we use the same variable name for parity? Just `config`?

Also, it might be worth converting this to use `attr_extras` (allows to remove line 17):
```rb
pattr_initialize :pull_request, :config do
  @style_guides = {}
end
```

config/locales/en.yml
```diff
@@ -2,8 +2,9 @@
...
 en:
-  sync_repos: 'Refresh repo list'
-  syncing_repos: 'Loading repos...'
-  authenticate: 'Sign In with GitHub'
   active_repos: 'Active repos'
+  authenticate: 'Sign In with GitHub'
+  invalid_config: "Invalid configuration syntax"
   search_placeholder: 'search by repo or organization name'
+  sync_repos: 'Refresh repo list'
+  syncing_repos: 'Loading repos...'
```
gylaz: Convert touched to double quotes?

app/services/build_runner.rb
```diff
@@ -21,12 +19,46 @@ def relevant_pull_request?
...
+    repo.builds.create!(
+      violations: violations,
+      pull_request_number: payload.pull_request_number,
+      commit_sha: payload.head_sha
+    )
+  end
+
+  def run_build
+    create_build(style_checker_violations)
+    create_pending_status
```
gylaz: Should we move this out... to above line 6?

app/models/repo_config.rb
malandrina: Good idea. Done: https://github.com/thoughtbot/hound/commit/6945eb3b07e5d17bfb7e724c5e997bdc67709c02

app/models/style_checker.rb
malandrina: Yes, good call. Done: https://github.com/thoughtbot/hound/commit/6945eb3b07e5d17bfb7e724c5e997bdc67709c02

config/locales/en.yml
malandrina: Done: https://github.com/thoughtbot/hound/commit/6945eb3b07e5d17bfb7e724c5e997bdc67709c02

spec/models/repo_config_spec.rb
```diff
@@ -133,6 +133,174 @@
...
+        EOS
+        commit = stub_commit(
+          hound_config: hound_config,
+          file_name => invalid_config
+        )
+        config = RepoConfig.new(commit)
+
+        config.load_style_guides
+
+        expect(config.instance_variable_get("@invalid")).to eq true
```
gylaz: Will this work?
```rb
expect(config).to be_invalid
```

spec/models/repo_config_spec.rb
```diff
@@ -133,6 +133,174 @@
...
+          {
+            "predef": ["myGlobal"]
+          }
+        EOS
+        commit = stub_commit(
+          hound_config: hound_config,
+          "config/rubocop.yml" => valid_config,
+          "coffeelint.json" => valid_config,
+          "javascript.json" => valid_config
+        )
```
gylaz: Why is `valid_config` valid for ruby or coffeescript? How about just testing one language here to avoid confusion?

spec/models/repo_config_spec.rb
```diff
@@ -133,6 +133,174 @@
...
+        file_name = "javascript.json"
+        hound_config = <<-EOS
+          java_script:
+            enabled: true
+            config_file: #{file_name}
+        EOS
+        commit = stub_commit(
+          hound_config: hound_config,
+          file_name => invalid_config
+        )
```
gylaz: Can we use `config_for_file` here?

spec/models/repo_config_spec.rb
```diff
@@ -133,6 +133,174 @@
...
+    context "when repo config is valid" do
+      it "returns false" do
+        commit = double("Commit")
+        config = RepoConfig.new(commit)
+        config.instance_variable_set("@invalid", false)
+
+        expect(config).not_to be_invalid
+      end
+    end
+  end
```
gylaz: Not really sure if there's value in these test, if we use `invalid?` in the above contexts.

spec/models/style_checker_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 describe StyleChecker, "#violations" do
   it "returns a collection of computed violations" do
+    config = double("RepoConfig", enabled_for?: true, "for" => {})
```
gylaz: `for: {}`?

app/services/build_runner.rb
malandrina: The current order came out of some discussion on my previous PR: https://github.com/thoughtbot/hound/pull/525#discussion_r22133021

The reasoning is that if `create_pending_status` fails, we will at least have a `build` record in our DB. Thoughts?

spec/models/style_checker_spec.rb
malandrina: Changed in https://github.com/thoughtbot/hound/commit/dc0467a715e581067d620abfab23d9f84b904c74.

spec/models/repo_config_spec.rb
malandrina: Yes, that works. I guess I'd wanted to avoid exercising `#invalid` initially, but coming back to it now I can't think of a compelling reason to do so. Changed in https://github.com/thoughtbot/hound/commit/dc0467a715e581067d620abfab23d9f84b904c74.

spec/models/repo_config_spec.rb
malandrina: Agreed. Removed: https://github.com/thoughtbot/hound/commit/dc0467a715e581067d620abfab23d9f84b904c74

spec/models/repo_config_spec.rb
malandrina: All three need to be stubbed because 'RepoConfig#load_style_guides` calls `RepoConfig#for` with each style guide.

spec/models/repo_config_spec.rb
malandrina: Maybe -- I'll give it a try.

app/services/build_runner.rb
gylaz: The reason I mentioned it, is that ideally we would set a pending status as soon as we receive the request to review a build -- maybe even in the controller before we schedule the job.

It seems there's no real value in `create_pending_status` at this point. As the only things that separate it and `create_success_status` are comment creations. Should we remove this altogether because Hound is so fast at doing its thing?

malandrina: > Should we remove this altogether because Hound is so fast at doing its thing?

I'm not sure. I'd be open to removing the status updates, but I view that as a change that should be made in a separate PR.

Before deciding to axe the status updates I'd want to have a broader conversation about what kind of feedback is most useful to the user during the build process and how GitHub status updates fit into that user experience.

gylaz: I wasn't proposing nixing the status updates altogether, just the pending one. There's not a lot of value in setting it here, since creating the actual comments only takes a split moment. The other problem is that the failed configuration status will be set, before and without `pending`. My main point being that there isn't much usefulness for `pending` status. I'm okay with making a separate card for this.

spec/models/repo_config_spec.rb
malandrina: I believe we would have to change the signature of `config_for_file` in order to use it here. We want `enabled` to be set to `false` for Ruby and CoffeeScript for this scenario, but currently `enabled` is hardcoded as `true`.

app/services/build_runner.rb
malandrina: >  My main point being that there isn't much usefulness for `pending` status.

Got it. That seems reasonable to me. I think it should be a separate card.

gylaz:
Closing in favor of new config settings.

PR #547, opened 2015-01-06
Add thoughtbot-specific SCSS style guide

croaky:
Fixes https://app.getsentry.com/hound/production/group/49280598/


gylaz:
The link seems to be wrong. LGTM otherwise.

PR #548, opened 2015-01-06
Fix line numbers for SCSS linter

croaky:


thorncp:
Looks good to me. Though, a little troubling that a spec wasn't blowing up beforehand.

PR #549, opened 2015-01-06
Improve local Hound instructions

croaky:
* Use ngrok for GitHub app.
* Get Stripe tokens automatically from staging in `bin/setup`.

### DIFF
--- a/.sample.env
+++ b/.sample.env
@@ -1,14 +1,10 @@
+CHANGED_FILES_THRESHOLD=300
 ENABLE_HTTPS=no
 GITHUB_CLIENT_ID=github_client_id
 GITHUB_CLIENT_SECRET=github_client_secret
-SECRET_KEY_BASE=secret_key_base
-CHANGED_FILES_THRESHOLD=300
+HOUND_GITHUB_TOKEN=your_github_token
+HOUND_GITHUB_USERNAME=your_github_username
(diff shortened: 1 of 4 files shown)


thorncp:
I was able to get this working locally with these instructions. I did have to modify my *.env* file:

* Add a `RESQUE_ADMIN_PASSWORD` to see the queue locally. For some reason this didn't work if the password was nil, and it's not included in *.sample.env*. Perhaps it should be?
* Add the houndci's key for `HOUND_GITHUB_TOKEN`. This was potentially not the right thing to do, since it could have a negative impact on houndci's api use. The `HOUND_GITHUB_USERNAME` value in *.sample.env* is `houndci`, but maybe it be a placeholder, and the readme can instruct users to fill in their personal info?

croaky:
> Add a RESQUE_ADMIN_PASSWORD to see the queue locally. For some reason this didn't work if the password was nil, and it's not included in .sample.env. Perhaps it should be?

Yup, done.

> Add the houndci's key for HOUND_GITHUB_TOKEN. This was potentially not the right thing to do, since it could have a negative impact on houndci's api use. The HOUND_GITHUB_USERNAME value in .sample.env is houndci, but maybe it be a placeholder, and the readme can instruct users to fill in their personal info?

Yup, done.

PR #551, opened 2015-01-07
Use active job

arunagw:
As we are on rails 4.2 we can use `ActiveJob` and remove our custom `JobQueue` code.

This PR needs to be review carefully and needs more testing on staging before merging into master.


app/jobs/repo_synchronization_job.rb
```diff
@@ -1,18 +1,24 @@
...
+class RepoSynchronizationJob < ActiveJob::Base
+  class AlreadyRefresing < StandardError; end
   extend Retryable
 
-  @queue = :high
+  queue_as :high
 
-  def self.before_enqueue(user_id, github_token)
-    User.set_refreshing_repos(user_id)
+  before_perform do |job|
```
salbertson: `before_perform` sounds different than `before_enqueue`. Is there a difference?

salbertson:
@arunagw, the before and after code looks very similar. What do we get from `ActiveJob` that makes it worth changing?

app/jobs/repo_synchronization_job.rb
```diff
@@ -1,18 +1,24 @@
...
+    unless User.set_refreshing_repos(job.arguments.first)
+      raise AlreadyRefresing
+    end
```
gylaz: Can we move this into the controller, to check if the user is refreshing before we enqueue, instead of raising/rescuing.

gylaz:
This is looking good. We'll need to give a good test run on staging.

arunagw:
@salbertson @gylaz 

The main problem I can see here is that we are not utilizing the `ActiveJob`. We are going to have some custom code inside every job class. Like `rescue Resque::TermException` and I couldn't find any workaround for that.

Do you have any idea how we can solve this? I don't want to put `rescue Resque::TermException` in every job class because I cannot switch to another adapter easily with this custom code.

gylaz:
Does `ActiveJob` now wrap errors like `Resque::TermException` into some `ActiveJob` related error?

Also, what is the main benefit of switching to `ActiveJob`?

lunks:
The main benefit of using ActiveJob is making it easier for other people to run Hound with other job queues. All one would have to do is changing the ActiveJob queue and remove the `rescue` specific to `Resque`.

app/jobs/repo_synchronization_job.rb
lunks: We'd be queuing jobs and only skipping them after they are ready to be performed. There is a `before_enqueue` callback but I agree with @gylaz and would prefer this check at the controller level.

gylaz:
Since we're not really planning on moving off of Rescue, I'm wondering about the value of this work. I like the general re-factoring/abstraction, but if some things are any gotchas, or risks in deploying this, I'm not sure this is worth merging.

app/jobs/repo_synchronization_job.rb
salbertson: Yea, I'd rather not enqueue jobs if we know they shouldn't be run.

salbertson: @arunagw thoughts on this?

app/jobs/repo_synchronization_job.rb
```diff
@@ -1,18 +1,24 @@
...
     user = User.find(user_id)
     synchronization = RepoSynchronization.new(user, github_token)
     synchronization.start
     user.update_attribute(:refreshing_repos, false)
   rescue Resque::TermException
-    Resque.enqueue(self, user_id, github_token)
+    retry_job
+  end
+
+  rescue_from(AlreadyRefresing) do |_e|
```
salbertson: Is there a way we can avoid raising just to rescue and do nothing? Does it make more sense to move the "refreshing repos" check out of the job or is there a way to move it to "before enqueue" hook and not raise?

spec/jobs/org_invitation_job_spec.rb
```diff
@@ -5,22 +5,25 @@
...
     allow(GithubApi).to receive(:new).and_return(github)
 
-    OrgInvitationJob.perform
+    OrgInvitationJob.perform_now
 
     expect(github).to have_received(:accept_pending_invitations)
   end
 
   it "retries when Resque::TermException is raised" do
     allow(GithubApi).to receive(:new).and_raise(Resque::TermException.new(1))
```
salbertson: Should all references to `Resque` be removed?

spec/jobs/repo_synchronization_job_spec.rb
```diff
@@ -48,17 +65,15 @@
 
     it "retries when Resque::TermException is raised" do
       allow(User).to receive(:find).and_raise(Resque::TermException.new(1))
```
salbertson: Same here, should we be referencing `Resque` at all now with using `ActiveJob`?

app/jobs/repo_synchronization_job.rb
arunagw: Yes. This needs to be fixed

arunagw:
There is a problem using `ActiveJob` right now that we can't use `ActiveJob` reference everywhere and we have to use `Resque` at few places. Specially capturing `Resque::TermException`.

Also right now we can't really know which adapter we are using so we need to keep `Resque` references in test as well to run jobs.

I would suggest we can close this PR fir now and wait for fixes or wrap around `ActiveJob` and write some code to get it working.

lunks:
What exactly needs to be fixed, could you clarify? I'd like to help.

It bothers me as well to have `Resque` exceptions being rescued when adding a layer that should abstract `Resque` completely, but it looks to me more like a `Resque` problem than anything else i.e. if we replaced `Resque` for `Sidekiq` we wouldn't have to add any `Sidekiq` specific code in our jobs.

arunagw:
@lunks 
The only problem I feel is that we have remaining trace of `Resque` in every job and in test helper. 
We can also remove this in second attempt.
I will clean this PR for the first attempt and will keep the `Resque` specific code for now.

If you want we can work on this together in your fork.

lunks:
No problem using yours, let's hope I get some free time to work on it. :)

Is it a possible solution to ditch `Resque` and use `Sidekiq`?

gylaz:
I lie of all the issues we've been having with Resque recently, I'd be willing to explore switching to Sidekiq.

lunks:
After reviewing we could merge this PR and create another one to replace `Resque` with `Sidekiq`.

salbertson:
Thanks @lunks and @arunagw for moving this along! I think if we can use ActiveJob to its strengths and void littering the code with Resque references then this will be a nice change.

PR #554, opened 2015-01-07
Restrict commenting to a maximum per build

thorncp:
Pull requests with many style violations pose two problems:

* users feel overwhelmed by the amount of comments made by hound
* hound approaches or reaches the GitHub api rate limit

This change is an attempt to alleviate both problems, by only submitting
a maximum number of comments per build. Potential downsides could be:

* users continue to feel overwhelmed when they fix the first batch of
  issues, and receive another batch
* no indication of how many violations remain
* grabbing the first 10 violations is naive, and could lead to more than 10 comments on a PR. e.g. if a new commit is pushed that adds violations that occur before the existing 10, they will be commented

Potential todo items:

* indicate in the PR how many violations there are, if it's more than
  the maximum
* be "smart" about which violations are commented on. currently the
  first 10 are taken, but we can probably come up with something better
  without much effort


spec/services/build_runner_spec.rb
```diff
@@ -44,6 +44,29 @@
         with(style_checker.violations)
     end
 
+    it "comments a maximum number of times" do
+      max_comments = ENV["MAX_COMMENTS"].to_i
```
gylaz: How about setting `MAX_COMMENTS` to `1`, having `2` violations, and making sure only the one is reported? Not sure what the value of creating more test data in the setup than necessary.

spec/services/build_runner_spec.rb
```diff
@@ -44,6 +44,29 @@
...
+        )
+      end
+      stubbed_style_checker(violations: violations)
+      commenter = Commenter.new(stubbed_pull_request)
+      allow(Commenter).to receive(:new).and_return(commenter)
+      stubbed_github_api
+
+      build_runner.run
+
+      expect(commenter).to have_received(:comment_on_violations) do |arg|
```
gylaz: Can this be:
```rb
expect(commenter).to have_received(:comment_on_violations).exactly(n).times
```

spec/services/build_runner_spec.rb
thorncp: That makes sense.

spec/services/build_runner_spec.rb
thorncp: Not with the current implementation of `comment_on_violations`. It expects an array of all violations, and does the looping itself, so it's only called once no matter how many violations there are.

gylaz: Gotcha. How about this?
```
expect(commenter).to have_received(:comment_on_violations).with(violations.take(ENV["MAX_COMMENTS"]))
```

spec/services/build_runner_spec.rb
```diff
@@ -44,6 +44,29 @@
         with(style_checker.violations)
     end
 
+    it "comments a maximum number of times" do
+      ENV["MAX_COMMENTS"] = "1"
```
gylaz: You probably want to use:
```rb
allow(ENV).to receive(:[]).with("MAX_COMMENTS").and_return("1")
```
to avoid changing ENV for other tests.

spec/services/build_runner_spec.rb
```diff
@@ -44,6 +44,29 @@
...
+      build_runner = make_build_runner
+      stubbed_commenter
+      violations = 2.times.map do |i|
+        build(:violation,
+              patch_position: i,
+              line_number: i,
+              messages: ["error #{i}"],
+        )
+      end
```
gylaz: We simplify this block to just: `violations = build_list(:violation, 2)`

spec/services/build_runner_spec.rb
thorncp: My thinking was that it will eventually do something more interesting than taking the first N violations. For example, if the violations are ordered by number of messages before being restricted to the first N, then this spec would need to have the knowledge of how that ordering happens in order to set up the correct expectation. Maybe it should, but it seems like that would be better covered by another spec (when that implementation lands). Thoughts?

spec/services/build_runner_spec.rb
thorncp: Oh, didn't see that one.

spec/services/build_runner_spec.rb
gylaz: I'd keep it as simple as possible for now, and wait until we do something more interesting to modify this spec. The block makes it a bit harder to grok, then `with`, IMO.

thorncp: That makes sense.

gylaz:
:+1:

PR #556, opened 2015-01-08
Create team with push permissions

thorncp:
In order to create/update commit statuses for repositories belonging to an organization, the *Services* team needs **push** access.  No change is required for user owned repositories, because *houndci* is added as a collaborator, which includes push access.

Does this warrant an update to the [security] doc? The relevant section mentions the *Services* team, but it doesn't call out the access level.

https://trello.com/c/Z3GpdyNn

[security]: https://github.com/thoughtbot/hound/blob/b8a1169f0f25a33f04cbb5c75fb83d2edbf89d9d/doc/SECURITY.md#what-happens-when-you-enable-hound-on-your-github-repository

### DIFF
--- a/doc/SECURITY.md
+++ b/doc/SECURITY.md
@@ -102,10 +102,11 @@ The database is hosted by [Redis to Go].
 [Redis to Go]: http://redistogo.com
 
 We use your GitHub token to add the [@houndci] GitHub user to your repository
-via the [GitHub collaborator API][api1].
-@houndci is added to a "Services" team within your organization
-and creates the "Services" team if it doesn't exist.
-Your GitHub user will need admin privileges for that repository.
+via the [GitHub collaborator API][api1]. @houndci will be added to a team that
+has access to the enabled repository. If an existing team cannot be found, we'll
+create a "Services" team with *push* access to the enabled repository. This is
+necessary for @houndci to see pull requests, make comments, and update pull
+request statuses.
 
 [@houndci]: https://github.com/houndci
 [api1]: https://developer.github.com/v3/repos/collaborators/#add-collaborator
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -176,7 +176,7 @@ class GithubApi
     team_options = {
       name: name,
       repo_names: [repo.full_name],
-      permission: "pull"
+      permission: "push"
     }
     client.create_team(repo.organization.login, team_options)
   end
--- a/spec/models/github_user_spec.rb
+++ b/spec/models/github_user_spec.rb
@@ -33,7 +33,7 @@ describe GithubUser, '#has_admin_access_through_team?' do
         api = GithubApi.new
         user = GithubUser.new(api)
         team_id = 4567
-        teams = [double(permission: 'pull', id: 4567)]
+        teams = [double(permission: "push", id: 4567)]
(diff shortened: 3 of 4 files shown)


lib/github_api.rb
```diff
@@ -176,7 +176,7 @@ def create_team(name, repo)
     team_options = {
       name: name,
       repo_names: [repo.full_name],
-      permission: "pull"
+      permission: "push"
```
gylaz: Does `push` include `pull`, or do we need to specify both?

gylaz:
@croaky Thoughts on this?

@thorncp Could you please add the relevant trello ticket to the commit and the PR description?

thorncp:
Updated with trello link.

lib/github_api.rb
thorncp: It includes it. The permission value can only be one of `pull`, `push`, or `admin`.

https://developer.github.com/v3/orgs/teams/#create-team

thorncp:
Updated SECURITY.md.

doc/SECURITY.md
```diff
@@ -104,11 +104,14 @@ The database is hosted by [Redis to Go].
 We use your GitHub token to add the [@houndci] GitHub user to your repository
 via the [GitHub collaborator API][api1].
 @houndci is added to a "Services" team within your organization
-and creates the "Services" team if it doesn't exist.
+and creates the "Services" team if it doesn't exist. The "Services" team will
```
gylaz: I would re-phrase this a bit, to something like:
```
@houndci will be added to a team that has access to the enabled repository. If an existing team cannot be found, we'll create a "Services" team with *push* access to the enabled repository. This is necessary for ...
```

doc/SECURITY.md
```diff
@@ -104,11 +104,16 @@ The database is hosted by [Redis to Go].
 We use your GitHub token to add the [@houndci] GitHub user to your repository
 via the [GitHub collaborator API][api1].
 @houndci is added to a "Services" team within your organization
```
gylaz: Remove this line?

doc/SECURITY.md
```diff
@@ -104,11 +104,16 @@ The database is hosted by [Redis to Go].
...
-and creates the "Services" team if it doesn't exist.
-Your GitHub user will need admin privileges for that repository.
+and creates the "Services" team if it doesn't exist. @houndci will be added to a
+team that has access to the enabled repository. If an existing team cannot be
+found, we'll create a "Services" team with *push* access to the enabled
+repository. This is necessary for @houndci to use the [GitHub statuses
+API][api3] to update the status of your pull request.  Your GitHub user will
```
gylaz: Two spaces before `Your`.

doc/SECURITY.md
```diff
@@ -104,11 +104,16 @@ The database is hosted by [Redis to Go].
...
+need admin privileges for that repository.
```
gylaz: Actually, the GitHub user will need rights to be able to add users to a team, which is not the same as admin privileges for a repo.

Thoughts on removing this last sentence, as it could potentially only confuse further?

thorncp: That sounds fine to me. I think it's implied that you need the correct permission to do all of this stuff anyway.

gylaz:
:+1:

PR #557, opened 2015-01-08
Ensure uniqueness of repository name

gylaz:
* add a uniquness constraint on `full_github_name`
* attempt to find the repo based on name first, then by id, then update it

https://trello.com/c/2Kl8oLFV

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -10,7 +10,7 @@ class Repo < ActiveRecord::Base
   delegate :type, :price, to: :plan, prefix: true
   delegate :price, to: :subscription, prefix: true
 
-  validates :full_github_name, presence: true
+  validates :full_github_name, uniqueness: true, presence: true
   validates :github_id, uniqueness: true, presence: true
 
   def self.active
@@ -18,8 +18,12 @@ class Repo < ActiveRecord::Base
   end
 
   def self.find_or_create_with(attributes)
-    repo = where(github_id: attributes[:github_id]).first_or_initialize
-    repo.update_attributes(attributes)
+    repo = find_by(full_github_name: attributes[:full_github_name]) ||
+      find_by(github_id: attributes[:github_id]) ||
+      Repo.new
+
+    repo.update!(attributes)
+
     repo
   end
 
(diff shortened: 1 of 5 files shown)


app/services/repo_synchronization.rb
```diff
@@ -20,10 +20,11 @@ def start
   private
 
   def repo_attributes(attributes)
-    attributes.slice(:private).merge(
+    {
+      private: attributes[:private],
```
thorncp: Does hound update subscriptions when paid/private repos become public? If so, does this have an effect on it?

app/models/repo.rb
```diff
@@ -18,8 +18,12 @@ def self.active
   end
 
   def self.find_or_create_with(attributes)
-    repo = where(github_id: attributes[:github_id]).first_or_initialize
-    repo.update_attributes(attributes)
+    repo = find_by(full_github_name: attributes[:full_github_name]) ||
+      find_by(github_id: attributes[:github_id]) ||
+      Repo.new
+
+    repo.update!(attributes)
```
croaky: Does this continue our issue of having `NULL` timestamps?

app/models/repo.rb
```diff
@@ -18,8 +18,12 @@ def self.active
   end
 
   def self.find_or_create_with(attributes)
-    repo = where(github_id: attributes[:github_id]).first_or_initialize
-    repo.update_attributes(attributes)
+    repo = find_by(full_github_name: attributes[:full_github_name]) ||
+      find_by(github_id: attributes[:github_id]) ||
```
bernerdschaefer: Do we actually care about the github id any more? Or just the name?

app/models/repo.rb
gylaz: This will touch `updated_at`.

app/models/repo.rb
gylaz: Good point. We can likely move over to just using `full_github_name`, but sounds like a job for another task. `github_id` is referenced a lot in our code, but `full_github_name` can replace that.

app/services/repo_synchronization.rb
gylaz: Good catch. Since this is an existing issue and requires a lot more thought/work, I've made a card for it: https://trello.com/c/lh32IXb1

app/models/repo.rb
gylaz: An update: we actually do need `github_id` in case the repo was renamed.

thorncp:
:+1:

PR #558, opened 2015-01-08
Handle GitHub error when file is too large to fetch

gylaz:
GitHub will error when a file is over 1MB.

https://app.getsentry.com/hound-1/production/group/49620125/

### DIFF
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -11,5 +11,11 @@ class Commit
     end
   rescue Octokit::NotFound
     ""
+  rescue Octokit::Forbidden => exception
+    if exception.errors.first[:code] == "too_large"
+      ""
+    else
+      raise exception
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/commit.rb
```diff
@@ -11,5 +11,11 @@ def file_content(filename)
     end
   rescue Octokit::NotFound
     ""
+  rescue Octokit::Forbidden => exception
+    if exception.errors.first[:code] == "too_large"
```
thorncp: From the sentry link, it looks like the value of `code` is `"too_large // See: https://developer.github.com/v3/repos/contents/#get-contents"`. Am I seeing that right? Is Octokit adding that in as a helpful message?

gylaz: It's just `too_large` the doc link must be added for printing.
screenshot 2015-01-08 14 31 01

thorncp:
:+1:

PR #561, opened 2015-01-09
Update email link in SECURITY

srt32:
Previously clicking on the `security@` email link would open https://github.com/thoughtbot/hound/blob/master/doc/security@thoughtbot.com, which 404's.  Now it opens a mail client.


croaky:
Tight.

PR #563, opened 2015-01-12
Update pull permission teams to push permission

thorncp:
Teams created before 855e932 will only have *pull* permission, leaving hound unable to update pull request statuses for repositories of that organization. This change adds a check for *pull* only access during repository activation, and will upgrade the Services team to *push* permission.

I updated the existing team fixtures to have "push" access, to avoid having to stub the update team requests for every spec. I tried to keep the names for the new requests/fixtures in-line with what was there, but they feel cumbersome. If anyone has suggestions for better ones, I'd love to hear them.

https://trello.com/c/kdFGneUh

### DIFF
--- a/spec/lib/github_api_spec.rb
+++ b/spec/lib/github_api_spec.rb
@@ -82,6 +82,25 @@ describe GithubApi do
             end
           end
 
+          context "when Services team has pull access" do
+            it "updates permissions to push access" do
+              stub_repo_with_org_request(repo_name, user_token)
+              stub_empty_repo_teams_request(repo_name, user_token)
+              stub_org_teams_with_pull_permission_services_request(
+                organization,
+                user_token
+              )
+              stub_add_repo_to_team_request(repo_name, team_id, user_token)
+              stub_add_user_to_team_request(username, team_id, user_token)
+              update_team_request =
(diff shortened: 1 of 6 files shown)


spec/lib/github_api_spec.rb
```diff
@@ -82,6 +82,24 @@ def stub_common_requests(token)
             end
           end
 
+          context "when Services team has pull access" do
+            it "updates permissions to push access" do
+              stub_repo_with_org_request(repo_name, user_token)
+              stub_empty_repo_teams_request(repo_name, user_token)
+              stub_org_teams_with_pull_permission_services_request(organization,
+                                                                   user_token)
```
gylaz: From style guide:

> If you break up an argument list, keep the arguments on their own lines and closing parenthesis on its own line.

spec/lib/github_api_spec.rb
```diff
@@ -82,6 +82,24 @@ def stub_common_requests(token)
...
+              stub_add_repo_to_team_request(repo_name, team_id, user_token)
+              stub_add_user_to_team_request(username, team_id, user_token)
+
```
gylaz: Can we remove this line, if line 94 is still part of setup?

gylaz:
Ready to merge.

PR #564, opened 2015-01-12
Cleanup db connections for Resque workers

gylaz:
Attempt to fix:
```
could not receive data from client: Connection reset by peer
```

As per [Heroku docs][1].
[1]: https://devcenter.heroku.com/articles/forked-pg-connections#resque-ruby-queuing


thorncp:
Looks good. :facepunch:

PR #565, opened 2015-01-14
Fix SCSS linter creating erroneous violations

### LINKED ISSUE #562: SCSS Bugs
I enabled SCSS on Hound on a project and out of around 50 Hound comments, none have been a proper style violation. I ended up disabling SCSS on Hound for now…

95% of Hound’s comments have been about using `//` Sass-style comments (as opposed to CSS-style comments), but there are no comments in the code there:

comments-1
comments-2

I’ve also noticed Hound commenting about `margin` shorthand, but there is no use of `margin` here:

margin

thorncp:
The SCSS linter stores violations in an [instance variable][omg-state], and doesn't clear them between runs or provide a public interface for clearing them. This led to erroneous violations being reported on files when a violation in another file happened to be on the same line number that's included in the diff.

This change uses a newly created linter for every file that's checked, in order to avoid the responsibility of clearing the state between each run.

I think the name/phrasing in the spec need some work, but I'm unable to come up with something better at the moment. Anyone have any suggestions?

Will close #562

https://trello.com/c/qObHGpDB

[omg-state]: https://github.com/causes/scss-lint/blob/8319483e3686e1956613069e6a02a56e62db74bc/lib/scss_lint/runner.rb#L12

### DIFF
--- a/spec/models/style_guide/scss_spec.rb
+++ b/spec/models/style_guide/scss_spec.rb
@@ -65,15 +65,33 @@ describe StyleGuide::Scss do
         end
       end
     end
+
+    context "over multiple runs" do
+      it "it reports errors only for the given file" do
+        style_guide = build_style_guide
+        bad_content = ".a { .b { .c { background: #000; } } }"
+        good_content = ".a { margin: 0.5em; }\n"
+
+        bad_run = style_guide.violations_in_file(build_file(bad_content))
+        good_run = style_guide.violations_in_file(build_file(good_content))
+
+        expect(bad_run).not_to be_empty
+        expect(good_run).to be_empty
+      end
+    end
   end
 
   private
 
   def violations_in(content, config = nil)
+    style_guide = build_style_guide(config)
+    style_guide.violations_in_file(build_file(content)).flat_map(&:messages)
+  end
+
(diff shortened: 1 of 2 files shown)


spec/models/style_guide/scss_spec.rb
```diff
@@ -65,15 +65,32 @@
         end
       end
     end
+
+    context "over multiple runs" do
+      it "does not carry over state" do
+        style_guide = build_style_guide
+        bad_content = ".a { .b { .c { background: #000; } } }"
+        good_content = ".a { margin: 0.5em; }\n"
+        bad_run = style_guide.violations_in_file(build_file(bad_content))
```
gylaz: Newline before 74 to separate setup from execution?

spec/models/style_guide/scss_spec.rb
```diff
@@ -65,15 +65,32 @@
         end
       end
     end
+
+    context "over multiple runs" do
+      it "does not carry over state" do
```
gylaz: Maybe something like `it "reports errors for the given run...`?

gylaz:
A few minor suggestions. LGTM.

croaky:
Nice work, @thorncp.

PR #566, opened 2015-01-14
Deactivate repos deleted on github

thorncp:
Adds a check to the synchronization process that identifies deleted repos and deactivates them.

The added spec is quite big, but I think the assertions it makes are important (I'm paranoid about unintentionally deactivating repos). Maybe it should be broken into smaller specs? Also, because I'm paranoid, I'd appreciate multiple eyes on the logic for determining which repos where deleted. :eyes: 

https://trello.com/c/SBKDsHjl

### DIFF
--- a/app/services/repo_synchronization.rb
+++ b/app/services/repo_synchronization.rb
@@ -9,12 +9,10 @@ class RepoSynchronization
   end
 
   def start
-    user.repos.clear
-
-    api.repos.each do |resource|
-      attributes = repo_attributes(resource.to_hash)
-      user.repos << Repo.find_or_create_with(attributes)
-    end
+    existing_repos = user.repos.to_a
+    rebuild_user_repos
+    deleted_repos = existing_repos - user.repos
+    deactivate_repos(deleted_repos)
   end
 
   private
@@ -26,5 +24,25 @@ class RepoSynchronization
       full_github_name: attributes[:full_name],
       in_organization: attributes[:owner][:type] == ORGANIZATION_TYPE
     }
+  end
+
+  def rebuild_user_repos
+    user.repos.clear
+
+    api.repos.each do |resource|
+      attributes = repo_attributes(resource.to_hash)
+      repo = Repo.find_or_create_with(attributes)
+      user.repos << repo
+    end
+  end
+
+  def deactivate_repos(deleted_repos)
+    deleted_repos.each do |repo|
+      repo.deactivate
+
+      if repo.subscription
+        RepoSubscriber.unsubscribe(repo, user)
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/repo_synchronization.rb
```diff
@@ -27,4 +25,24 @@ def repo_attributes(attributes)
...
+      user.repos << repo
+    end
+  end
+
+  def deactivate_repos(deleted_repos)
+    deleted_repos.each do |repo|
+      repo.deactivate
+
+      if repo.subscription
+        RepoSubscriber.unsubscribe(repo, user)
```
gylaz: If a user was removed from a repo, then we will unsubscribe that repo -- it doesn't necessarily mean that the repo was deleted. I think the same applies for deactivation.

thorncp: Good point! Would it be enough to simply check for other memberships of the repo? The only edge case I can think of at the moment is if the repo wasn't deleted, but the user that was removed from it was the only user that has logged into hound (meaning they enabled it). I think for that case it makes sense to deactivate it anyway, but are there any others?

gylaz: I think that if we have no associated memberships then it makes sense to deactivate and unsubscribe. Though, that won't cover all of the cases -- if the repo was deleted, it got associated to two users in Hound, then repo was deleted, but only one of those users logs into Hound.

Maybe we should always show the repo if it was subscribed to by a certain user. Then even if it was deleted, they can still see it and unsubscribe?

app/services/repo_synchronization.rb
```diff
@@ -27,4 +25,24 @@ def repo_attributes(attributes)
       in_organization: attributes[:owner][:type] == ORGANIZATION_TYPE
     }
   end
+
+  def rebuild_user_repos
+    user.repos.clear
+
+    api.repos.each do |resource|
```
srt32: Could we call this block parameter `repo` so that it is the singular of the collection over which we're iterating?

app/services/repo_synchronization.rb
```diff
@@ -27,4 +25,24 @@ def repo_attributes(attributes)
...
+      attributes = repo_attributes(resource.to_hash)
+      repo = Repo.find_or_create_with(attributes)
+      user.repos << repo
+    end
+  end
+
+  def deactivate_repos(deleted_repos)
```
srt32: This method is really deactivating and unsubscribing (not just deactivating). What if we named this two-step process? Maybe something like `expire_subscription` or something along those lines?

app/services/repo_synchronization.rb
gylaz: Perhaps even better to call it `github_repo` since this could be confused with our representation of `repo` (which also happens to be used within the block.

srt32: :+1:

thorncp:
Closing this to take a different approach: during repo sync, take note of repos that the user no longer has access to, and displaying them differently in the UI.

PR #567, opened 2015-01-15
Move logic of adding Hound to an repo

gylaz:
Extract the logic out of the `GithubApi` into a service. `GithubApi` should
only be a wrapper over Octokit methods and not handle complex logic.
This allows us to test that logic easier in isolation without stubbing
external requests.

### DIFF
--- a/app/services/add_hound_to_repo.rb
+++ b/app/services/add_hound_to_repo.rb
@@ -1,0 +1,86 @@
+# Adds @hounci user to the a private repository as collaborator,
+# or to an organization team that has access to the repository
+# This is necessary for making comments as the houndci user
+class AddHoundToRepo
+  SERVICES_TEAM_NAME = "Services"
+
+  pattr_initialize :repo_name, :github
+
+  def self.run(repo_name, github)
+    new(repo_name, github).run
+  end
+
+  def run
+    if repo.organization
+      add_user_to_org
+    else
+      add_collaborator_to_repo
+    end
+  end
+
+  private
+
+  def repo
+    @repo ||= github.repo(repo_name)
+  end
+
+  def add_collaborator_to_repo
+    github.add_collaborator(repo_name, github_username)
+  end
+
+  def add_user_to_org
+    if team_with_admin_access
+      github.add_user_to_team(github_username, team_with_admin_access.id)
+    else
+      add_user_and_repo_to_services_team
+    end
+  end
+
+  def team_with_admin_access
+    @team_with_admin_access ||= begin
+      repo_teams = github.repo_teams(repo_name)
+      token_bearer = GithubUser.new(github)
+
+      repo_teams.detect do |repo_team|
+        token_bearer.has_admin_access_through_team?(repo_team.id)
+      end
+    end
+  end
+
+  def add_user_and_repo_to_services_team
+    team = find_services_team
+
+    if team
+      ensure_push_permission(team)
+      github.add_repo_to_team(team.id, repo_name)
+    else
+      team = github.create_team(
+        team_name: SERVICES_TEAM_NAME,
+        org_name: org_name,
+        repo_name: repo_name
+      )
+    end
+
+    github.add_user_to_team(github_username, team.id)
+  end
+
+  def find_services_team
+    github.org_teams(org_name).detect do |team|
+      team.name.downcase == SERVICES_TEAM_NAME.downcase
+    end
+  end
+
+  def ensure_push_permission(team)
+    if team.permission == "pull"
+      github.update_team(team.id, permission: "push")
+    end
+  end
+
+  def org_name
+    repo.organization.login
+  end
+
+  def github_username
+    @github_username ||= ENV.fetch("HOUND_GITHUB_USERNAME")
+  end
+end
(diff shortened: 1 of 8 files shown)


app/services/repo_activator.rb
```diff
@@ -32,8 +32,7 @@ def change_repository_state_quietly
   end
 
   def add_hound_to_repo
-    github_username = ENV.fetch("HOUND_GITHUB_USERNAME")
-    github.add_user_to_repo(github_username, repo.full_github_name)
+    AddHoundToRepo.run(repo.full_github_name, github)
```
srt32: Is there any sort of explicit error handling we want to have around this call?

thorncp: Looks like it's being handled the same way: https://github.com/thoughtbot/hound/blob/f9022d95160e95e68f1dbac59828d166c02ce284/app/services/repo_activator.rb#L21-L32

I assume that's been working well, but maybe something more specific would be beneficial?

gylaz: Yes, we already handle errors at the parent method. Did you have anything more specific in mind, @srt32?

srt32: There seem to be many potential error cases based on the amount of different api calls we could be making. If we trapped errors we could present nicer messages to the users when something is wrong. I don't have any specifics though...

gylaz: Right, so as per @thorncp's link `change_repository_state_quietly` will trap all of the Octokit errors.

thorncp:
LGTM

app/services/add_hound_to_repo.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    github.add_user_to_team(github_username, team.id)
+  end
+
+  def find_services_team
+    github.org_teams(org_name).detect do |team|
+      team.name.downcase == SERVICES_TEAM_NAME.downcase
+    end
+  end
```
salbertson: @gylaz, what do you think about moving some of these private methods back to `GithubApi` but add arguments to them so team name can be passed in? This would make the methods public and testable and keep GitHub API logic in one place and leave our business logic here.

app/services/add_hound_to_repo.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def ensure_push_permission(team)
+    if team.permission == "pull"
+      github.update_team(team.id, permission: "push")
+    end
+  end
+
+  def org_name
+    repo_name.split("/").first
```
salbertson: Could we use the `repo` object and make `organization` public to get the organization name?

app/services/add_hound_to_repo.rb
```diff
@@ -0,0 +1,86 @@
+# Adds @hounci user to the a private repository as collaborator,
+# or to an organization team that has access to the repository
+# This is necessary for making comments as the houndci user
+class AddHoundToRepo
```
salbertson: Would having two of these service classes help at all? We would use one if the repo is under an org and the other if it's not.

app/services/add_hound_to_repo.rb
gylaz: The reason I pulled these methods out is that I considered them to be our business logic, and `GitHub` api should be a thin wrapper over `octokit`. Is there some point I'm missing here?

app/services/add_hound_to_repo.rb
gylaz: Do you mean to suggest passing `repo` instead of just `repo_name`? I can see that being beneficial.

I'm not sure I followed the part about making `organization` public.

app/services/add_hound_to_repo.rb
gylaz: Interesting idea. I will experiment.

gylaz: It turns out that 99% percent of the logic in this service is for adding hound to an organization.
All that's needed for adding hound to a repo is:
```rb
github.add_collaborator(repo_name, "houndci")
```
We could add an `if/else` to `RepoActivator#add_hound_to_repo` but I don't think it's worth it.

lib/github_api.rb
```diff
@@ -125,68 +110,46 @@ def create_success_status(full_repo_name, sha, description)
...
-    add_user_to_team(username, team.id)
+  def add_repo_to_team(team_id, repo_name)
+    client.add_team_repository(team_id, repo_name)
   end
 
   def add_user_to_team(username, team_id)
     with_preview_client do |preview_client|
       preview_client.add_team_membership(team_id, username)
     end
-  rescue Octokit::NotFound
```
srt32: Did we purposefully remove this `rescue`?

spec/services/add_hound_to_repo_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    github_repo = double("GithubRepo", organization: double(login: "foo"))
+    double(
+      "GithubApi",
+      repo: github_repo,
+      user_teams: [github_team],
+      repo_teams: [github_team],
+    )
+  end
+
+  def hound
```
srt32: Could we call `hound` `hound_github_username` for clarity's sake?

lib/github_api.rb
gylaz: Yes, this should fail. And if it does we should know about it -- raise.

spec/services/add_hound_to_repo_spec.rb
gylaz: Good idea.

PR #569, opened 2015-01-15
Update Octokit

gylaz:
Remove `with_preview_client` blocks, but leave the method because we'll need
to use the preview client in the future. I've udpated the `PREVIEW_MEDIA_TYPE`
to the current [suggested][github] string.

[github]:
https://developer.github.com/changes/2015-01-07-prepare-for-organization-permissions-changes/


thorncp:
:+1:

Issue #522, opened 2014-12-18
Hound ignoring path-based excludes with wildcards


prsimp:
Given this configuration:

```yml
# ./.hound.yml
ruby:
  enabled: true
  config_file: .rubocop.yml

# ./.rubocop.yml
Style/Blocks:
  Exclude:
    - spec/**/*
```

And a spec file with this sort of syntax:
```ruby
it 'changes something' do
  expect {
    SomeClassName.a_method_that_changes_the_count
  }.to change { SomeClassName.count }
end
```

Locally, Rubocop 0.25.0 does not complain about this block syntax, however, Hound will comment on the line with the `Avoid using {...} for multi-line blocks.` message.

This occurs with other cops (i.e. Style/PredicateName, Style/RescueModifier), but seems to be limited to cops that are using a wildcard pattern (though this is an intuition, not a confirmed fact).

The configuration *is* being detected and loaded, as other rules are respected (particularly when they differ from the default).

Hound is very useful, but it's usefulness is greatly diminished when you're unable to stop it from commenting on non-issues.

Let me know if there's any other info I can add to help diagnose. The repo in question isn't public, but I can create a public dummy project to demonstrate the issue if needed. Thanks in advance!

salbertson:
@prsimp, this has been fixed!

prsimp:
@salbertson thanks for the heads up!

djiang:
Did this fix every get merged in and deployed? I'm trying almost the exact same config as the OP:

```ruby
Style/Blocks:
  Exclude:
    - 'spec/**/*'
```

djiang:
@prsimp @salbertson Are you guys still this issue?

djiang:
Update: It looks like rubocop changed the name to `BlockDelimiters`, so the following now works as expected:

```yml
Style/BlockDelimiters:
  Exclude:
    - 'spec/**/*`
```

gylaz:
Thanks for the follow-up @djiang! Hopefully others will find this useful.

PR #571, opened 2015-01-16
Individual cops can be excluded for files

brittballard:
* Pass the file's path into ProcessedSource initializer
* This path is later used to determine if a specific Cop is applicable
* https://github.com/thoughtbot/hound/issues/522

### DIFF
--- a/spec/models/style_guide/ruby_spec.rb
+++ b/spec/models/style_guide/ruby_spec.rb
@@ -420,6 +420,24 @@ end
 
         expect(violations).to be_empty
       end
+
+      it 'allows a cop to be excluded for specific files or directories' do
+        config = {
+          "StringLiterals" => {
+            "EnforcedStyle" => "single_quotes"
+          },
+          "HashSyntax" => {
+            "Exclude" => ["lib/**/*"]
+          }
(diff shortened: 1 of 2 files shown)


brittballard:
This is the line that does the check for the specific file: https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/commissioner.rb#L56.

brittballard:
Do you guys really want me to shorten those strings in the specs?

spec/models/style_guide/ruby_spec.rb
```diff
@@ -420,6 +420,23 @@ def initialize(name: , age: )
...
+          },
+          "HashSyntax" => {
+            "Exclude" => ["lib/**/*"]
+          }
+        }
+
+        violations = violations_with_config(config)
+
+        expect(violations).to eq [
+          "Prefer single-quoted strings when you don't need string interpolation or special symbols."
```
gylaz: Yeah, I'd say just break it up with a `+`. We can't preach what we don't practice.

gylaz:
Ready to merge once the hound comments are handled!

PR #573, opened 2015-01-19
Fix Mixpanel aliasing

croaky:
The order of events should be:

```js
alias('12345');
identify('12345');
track('Signed Up');
```

https://segment.com/docs/integrations/mixpanel/#alias


gylaz:
Looks good.

PR #575, opened 2015-01-20
Use Stripe Checkout to update credit card

salbertson:
This isn't fully built, designed or tested but I wanted to get some early feedback on the approach before getting too far.

update card

### DIFF
--- a/app/assets/javascripts/controllers/account_controller.coffee
+++ b/app/assets/javascripts/controllers/account_controller.coffee
@@ -1,0 +1,18 @@
+App.controller "accountController", [
+  "$scope",
+  "$window",
+  "StripeCheckout",
+  "CreditCard",
+  ($scope, $window, StripeCheckout, CreditCard) ->
+    updateCustomer = (stripeToken) ->
+      user = new CreditCard(card_token: stripeToken.id)
+      user.$update().catch(->
+        $window.alert("There was an issue updating your card. We have been notified and will contact you soon.")
+      )
+
+    $scope.showStripeForm = ->
+      StripeCheckout.open(
+        buttonText: "Update Card",
+        updateCustomer
+      )
+]
--- a/app/models/payment_gateway_customer.rb
+++ b/app/models/payment_gateway_customer.rb
@@ -23,6 +23,11 @@ class PaymentGatewayCustomer
     end
   end
 
+  def update_card(card_token)
+    customer.card = card_token
+    customer.save
+  end
+
   private
 
   def default_card
--- a/app/views/accounts/show.haml
+++ b/app/views/accounts/show.haml
@@ -1,9 +1,15 @@
 - content_for :page_title, "Your Account"
 
 = render 'shared/header'
+= render 'settings'
 
 %section.account-section
-  %h2.label-text Your monthly cost
+  %h2.label-text
+    Your monthly cost
+
+    %span.section-action
+      %a{ 'ng-controller' => 'accountController', 'ng-click' => 'showStripeForm()' }
+        Update Credit Card
 
   .account-content.account-breakdown
     %table
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -1,3 +1,5 @@
+ENV["SECRET_KEY_BASE"] = "supersecret"
+
 Houndapp::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
   config.eager_load = false
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,15 +1,16 @@
 Houndapp::Application.routes.draw do
   mount Resque::Server, at: "/queue"
   mount StripeEvent::Engine => "/stripe-event"
+  mount JasmineRails::Engine => "/specs" if defined?(JasmineRails)
 
   get "/auth/github/callback", to: "sessions#create"
   get "/sign_out", to: "sessions#destroy"
   get "/configuration", to: "application#configuration"
-
   get "/faq", to: "pages#show", id: "faq"
 
   resource :account, only: [:show]
   resources :builds, only: [:create]
+  resource :credit_card, only: [:update]
 
   resources :repos, only: [:index] do
     resource :activation, only: [:create]
--- a/spec/controllers/credit_cards_controller_spec.rb
+++ b/spec/controllers/credit_cards_controller_spec.rb
@@ -1,0 +1,42 @@
+require "spec_helper"
+
+describe CreditCardsController, "#update" do
+  context "when credit card is updated" do
+    it "responds with 200" do
+      card_token = "cardtoken"
+      customer = double("PaymentGatewayCustomer", update_card: true)
+      allow(PaymentGatewayCustomer).to receive(:new).and_return(customer)
+      stub_sign_in(create(:user))
+
+      put :update, card_token: card_token, format: :json
+
+      expect(response).to have_http_status(:success)
+      expect(customer).to have_received(:update_card).with(card_token)
+    end
+  end
+
+  context "when credit card fails to update" do
+    it "responds with 422" do
+      request_credit_card_update_and_fail
+
+      expect(response).to have_http_status(:unprocessable_entity)
+    end
+
+    it "captures exception" do
+      allow(Raven).to receive(:capture_exception)
+
+      request_credit_card_update_and_fail
+
+      expect(Raven).to have_received(:capture_exception).
+        with(kind_of(CreditCardsController::CreditCardUpdateFailed))
+    end
+  end
+
+  def request_credit_card_update_and_fail
+    customer = double("PaymentGatewayCustomer", update_card: false)
+    allow(PaymentGatewayCustomer).to receive(:new).and_return(customer)
+    stub_sign_in(create(:user))
+
+    put :update, card_token: "cardtoken", format: :json
+  end
+end
--- a/spec/javascripts/controllers/account_controller_spec.coffee
+++ b/spec/javascripts/controllers/account_controller_spec.coffee
@@ -1,0 +1,44 @@
+#= require spec_helper
+
+describe "accountController", ->
+  [scope, http, window, stripeCheckout] = []
+
+  beforeEach(inject ($rootScope, $httpBackend, $controller, CreditCard) ->
+    scope = $rootScope.$new()
+    http = $httpBackend
+    window = {
+      alert: ->
+    }
+    stripeCheckout = {
+      open: ->
+    }
+    $controller("accountController",
+      $scope: scope,
+      $window: window,
+      StripeCheckout: stripeCheckout,
+      CreditCard: CreditCard
+    )
+  )
+
+  afterEach ->
+    http.verifyNoOutstandingExpectation()
+    http.verifyNoOutstandingRequest()
+    http.resetExpectations()
+
+  describe "#showStripeForm", ->
+    describe "when update fails", ->
+      it "alerts the user", ->
+        spyOn(stripeCheckout, "open")
+        spyOn(window, "alert")
+        stripeToken = {
+          id: "tokenid"
+        }
+
+        scope.showStripeForm()
+
+        http.expectPUT("/credit_card", card_token: stripeToken.id).respond(422)
+        callback = stripeCheckout.open.calls.argsFor(0)[1]
+        callback(stripeToken)
+        http.flush()
+
+        expect(window.alert).toHaveBeenCalled()
--- a/spec/javascripts/spec_helper.coffee
+++ b/spec/javascripts/spec_helper.coffee
@@ -1,0 +1,3 @@
+#= require angular-mocks
+
+beforeEach(module("Hound"))
--- a/spec/support/helpers/stripe_api_helper.rb
+++ b/spec/support/helpers/stripe_api_helper.rb
@@ -1,0 +1,128 @@
+module StripeApiHelper
+  def stripe_customer_id
+    "cus_2e3fqARc1uHtCv"
+  end
+
+  def stripe_subscription_id
+    "sub_488ZZngNkyRMiR"
+  end
+
+  def stub_customer_create_request(user)
+    stub_request(
+      :post,
+      "https://api.stripe.com/v1/customers"
+    ).with(
+      body: {
+        "card" => "cardtoken",
+        "metadata" => { "user_id" => "#{user.id}" }
+      },
+      headers: { "Authorization" => "Bearer #{ENV["STRIPE_API_KEY"]}" }
+    ).to_return(
+      status: 200,
+      body: File.read("spec/support/fixtures/stripe_customer_create.json"),
+    )
+  end
+
+  def stub_customer_find_request
+    stub_request(
+      :get,
+      "https://api.stripe.com/v1/customers/#{stripe_customer_id}"
+    ).with(
+      headers: { "Authorization" => "Bearer #{ENV["STRIPE_API_KEY"]}" }
+    ).to_return(
+      status: 200,
+      body: File.read("spec/support/fixtures/stripe_customer_find.json"),
+    )
+  end
+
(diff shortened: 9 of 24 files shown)


config/routes.rb
```diff
@@ -19,6 +19,7 @@
 
   resources :repo_syncs, only: [:index, :create]
   resource :user, only: [:show]
+  resource :credit_card, only: [:update]
```
croaky: `:sort`

croaky:
Very cool. The UI looks great and the interaction seems simple. I say push forward on it.

salbertson:
@croaky @gylaz, let me know what you think. I added our first JavaScript tests too.

app/assets/javascripts/controllers/account_controller.coffee
```diff
@@ -0,0 +1,13 @@
+App.controller 'accountController', ['$scope', '$window', 'StripeCheckout', 'CreditCard', ($scope, $window, StripeCheckout, CreditCard) ->
+  updateCustomer = (stripeToken) ->
+    user = new CreditCard(card_token: stripeToken.id)
+    user.$update().catch(->
+      $window.alert('There was an issue updating your card. We have been notified and will contact you soon.')
```
croaky: The way we're currently notified is by looking in the Stripe and Baremetrics dashboards? Not sure if we have an item in the backlog for it, but we've had good luck with Churnbusters on another project.

croaky: I see the notification in the controller now.

config/routes.rb
```diff
@@ -1,15 +1,16 @@
 Houndapp::Application.routes.draw do
   mount Resque::Server, at: "/queue"
   mount StripeEvent::Engine => "/stripe-event"
+  mount JasmineRails::Engine => "/specs" if defined?(JasmineRails)
```
croaky: Should we put this at the bottom of the block so there's slightly less processing and chance for this code to cause problems for user-facing routes?

spec/controllers/credit_cards_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "spec_helper"
+
+describe CreditCardsController, "#update" do
+  context "when credit card is updated successfully" do
```
croaky: Can drop "successfully" and description will still read correctly.

https://github.com/thoughtbot/guides/tree/master/style/testing#acceptance-tests

spec/controllers/credit_cards_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      stub_sign_in(create(:user))
+
+      put :update, card_token: "cardtoken", format: :json
+
+      expect(response).to have_http_status(:unprocessable_entity)
+    end
+
+    it "captures exception" do
+      customer = double("PaymentGatewayCustomer", update_card: false)
+      allow(PaymentGatewayCustomer).to receive(:new).and_return(customer)
```
croaky: Any desire to move these duplicated lines to a before block? Or flatten the `it` blocks into one "responds with 422 and captures exception"?

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -0,0 +1,45 @@
...
+    $window: @window,
+    StripeCheckout: @stripeCheckout,
+    CreditCard: CreditCard
+  )
+)
+
+afterEach ->
+  @http.verifyNoOutstandingExpectation()
+  @http.verifyNoOutstandingRequest()
+  @http.resetExpectations()
```
croaky: Are lines 3 and 22-25 specific to this account controller spec? If they're generic, should they be moved to `spec_helper`?

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -0,0 +1,45 @@
...
+beforeEach(inject ($rootScope, $controller, CreditCard, $httpBackend) ->
+  @scope = $rootScope.$new()
+  @window = {
+    alert: ->
+  }
+  @http = $httpBackend
+  @stripeCheckout = {
+    open: ->
+  }
+  @controller = $controller('accountController',
```
croaky: Double quotes in here.

https://github.com/thoughtbot/guides/tree/master/style/coffeescript#coffeescript

app/models/payment_gateway_customer.rb
```diff
@@ -23,6 +23,11 @@ def customer
     end
   end
 
+  def update_card(card_token)
+    customer.card = card_token
+    customer.save
+  end
```
croaky: Do we want a spec for this method?

croaky:
I like it.

croaky:
Anything we need to add to the `Rakefile` so the JavaScript specs are run on CI?

config/routes.rb
salbertson: I don't think it's necessary since the engine is not mounted if `JasmineRails` doesn't exist and the gem is only included in the development and test groups.

spec/controllers/credit_cards_controller_spec.rb
salbertson: :+1:

spec/controllers/credit_cards_controller_spec.rb
salbertson: I moved the common code to a method. Let me know what you think.

https://github.com/thoughtbot/hound/commit/816e1ce87d98e258f7d1108c6aed487e1d791b18

spec/javascripts/controllers/account_controller_spec.coffee
salbertson: Some of this can definitely be moved, good call.

Here's what I moved, https://github.com/thoughtbot/hound/commit/8b39f387484b022655bb03105b78b9bbcfef96f2 and https://github.com/thoughtbot/hound/commit/8ea4f999b405778c1f91107f6ab5a025bbae7719.

spec/javascripts/controllers/account_controller_spec.coffee
salbertson: Done and I created this, https://github.com/thoughtbot/hound/pull/583.

spec/controllers/credit_cards_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    it "captures exception" do
+      allow(Raven).to receive(:capture_exception)
+
+      request_credit_card_update_and_fail
+
+      expect(Raven).to have_received(:capture_exception).
+        with(kind_of(CreditCardsController::CreditCardUpdateFailed))
+    end
```
croaky: Nice. This one's really easy to read, now.

croaky:
Maybe get a review of the Angular code from @gylaz. Other than maybe a unit test for `update_card` and any `Rakefile` / JS-tests-on-CI changes, this looks ready to merge to me.

spec/controllers/credit_cards_controller_spec.rb
salbertson: Sweet!

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -0,0 +1,125 @@
...
+      body: File.read("spec/support/fixtures/stripe_subscription_create.json"),
+    )
+  end
+
+  def stub_subscription_find_request(subscription)
+    stub_request(
+      :get,
+      "https://api.stripe.com/v1/customers/#{stripe_customer_id}/subscriptions/#{subscription.stripe_subscription_id}"
+    ).with(
+      headers: { "Authorization" => "Bearer #{ENV["STRIPE_API_KEY"]}",}
```
salbertson: Thanks Hound!

salbertson:
@croaky, I tested `update_card` and created test helper class for testing interaction with Stripe API.

salbertson:
@croaky, I thought `jasmine-rails` would magically add the rake task as a dependency to the default one, guess not. I added the JavaScript test suite task to the default rake task, https://github.com/thoughtbot/hound/commit/b9a13a87e4b2952d657721d544aeb63bb4087a7c.

app/models/payment_gateway_customer.rb
salbertson: Good idea, done!

app/views/accounts/show.haml
```diff
@@ -2,8 +2,19 @@
 
 = render 'shared/header'
 
+:javascript
+  Settings = {
```
gylaz: Aren't capitalized words usually used for constructor functions? Should we namespace this?

config/environments/test.rb
```diff
@@ -1,3 +1,5 @@
+ENV["SECRET_KEY_BASE"] = "supersecret"
```
gylaz: Where is this used? Should we move this into `spec_helper` instead?

spec/controllers/credit_cards_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def request_credit_card_update_and_fail
```
gylaz: What do you think about passing the `update_card` result and the `token` to this function, so it's re-usable in the first test?

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -0,0 +1,36 @@
+#= require spec_helper
+
+beforeEach(inject ($controller, CreditCard) ->
```
gylaz: How about moving this into the first `describe`?

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -0,0 +1,36 @@
...
+  @window = {
```
gylaz: Creating `@` instances will make them visible to other test files. The best convention is to do something like this:
```coffee
describe "SomeController", ->
  [window, stripeCheckout, $controller] = []

  beforeEach ->
    inject ($controller, CreditCard) ->
      $controller(
        "accountController",
        $scope: @scope,
        ...
```

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -0,0 +1,36 @@
...
+    StripeCheckout: @stripeCheckout,
+    CreditCard: CreditCard
+  )
+)
+
+describe "accountController", ->
+  describe "#showStripeForm", ->
+    describe "when update fails", ->
+      it "alerts the user", ->
+        spyOn(@stripeCheckout, "open")
```
gylaz: Consider doing this in `beforeEach` instead:
```coffee
stripeCheckout = jasmine.createSpyObj("stripeCheckout", ["open"])
```
You can add more simple method spies to that array.

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -0,0 +1,36 @@
...
+        spyOn(@window, "alert")
```
gylaz: Same here:
```coffee
$window = jasmine.createSpyObj("$window", ["alert"])
```

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -0,0 +1,36 @@
...
+        stripeToken = {
+          id: "tokenid"
+        }
+        @scope.showStripeForm()
+        @http.whenPUT("/credit_card").respond(422)
+        callback = @stripeCheckout.open.calls.argsFor(0)[1]
+
+        @http.expectPUT("/credit_card", card_token: stripeToken.id)
```
gylaz: You can combine the `whenPUT` and `expectPut` into just:
```coffee
$http.expectPUT("/credit_card", card_token: stripeToken.id).respond(422)
```

spec/javascripts/spec_helper.coffee
```diff
@@ -0,0 +1,13 @@
+#= require angular-mocks
+
+beforeEach(module("Hound"))
+
+beforeEach(inject ($rootScope, $httpBackend) ->
+  @scope = $rootScope.$new()
+  @http = $httpBackend
```
gylaz: Why do we need this global `beforeEach`? What if some tests don't need `scope` or `httpBackend`?

spec/javascripts/spec_helper.coffee
```diff
@@ -0,0 +1,13 @@
...
+)
+
+afterEach ->
+  @http.verifyNoOutstandingExpectation()
+  @http.verifyNoOutstandingRequest()
+  @http.resetExpectations()
```
gylaz: It's not ideal that this will run for every tests. What value does this provide?

app/views/accounts/show.haml
salbertson: I guess constructor functions or "constants". What were you thinking as far as namespacing? Something like this?

```
Hound = {
  settings: {
    ...
  }
}
```

salbertson: I namespaced using a handy little [library](https://github.com/salbertson/namespaced), https://github.com/thoughtbot/hound/commit/e3072f9efc32b9f0b8b74ed6f2b68320d8128a1e.

config/environments/test.rb
salbertson: Maybe, the Jasmine tests wouldn't run without it.

spec/controllers/credit_cards_controller_spec.rb
salbertson: I made it reusable and it turned out to be too messy; I had to pass customer, card token and whether or not to update card successfully.

spec/javascripts/controllers/account_controller_spec.coffee
salbertson: Sure, done.

spec/javascripts/controllers/account_controller_spec.coffee
salbertson: I prefer the spies closer to the exercise and expectations.

spec/javascripts/controllers/account_controller_spec.coffee
salbertson: Good call, done.

spec/javascripts/controllers/account_controller_spec.coffee
salbertson: Nice, done.

spec/javascripts/spec_helper.coffee
salbertson: It's just there to make writing tests easier. What is the downside to injecting these dependencies when they aren't actually needed?

spec/javascripts/spec_helper.coffee
salbertson: Which lines? Everything in the `afterEach`?

spec/javascripts/spec_helper.coffee
gylaz: The downside are: setup far removed from the test, unnecessary operations, global variables.

spec/javascripts/spec_helper.coffee
gylaz: Yes, this `afterEach` performs operations on `http` for every test. Many tests aren't going to need `httpBackend`. Could we just move them into the one test that needs it for now?

spec/javascripts/spec_helper.coffee
salbertson: Ok, sounds like it's worth moving. We can create functions to share any of the behavior, maybe `httpBackend` stuff, later when we need it for other tests.

salbertson: Done, https://github.com/thoughtbot/hound/commit/d0cb36e489b95d13eba9b5837b1142d4c659738e.

spec/javascripts/spec_helper.coffee
salbertson: This has been moved, https://github.com/thoughtbot/hound/commit/d0cb36e489b95d13eba9b5837b1142d4c659738e.

salbertson:
@croaky and @gylaz, thanks for the review!

PR #577, opened 2015-01-23
Deactivate repos the user can't see anymore

thorncp:
If the user cannot see a repo and the repo has no other memberships. It
will be inactivated and the memberships will be destroyed.

https://trello.com/c/SBKDsHjl

### DIFF
--- a/app/services/repo_synchronization.rb
+++ b/app/services/repo_synchronization.rb
@@ -9,15 +9,54 @@ class RepoSynchronization
   end
 
   def start
-    user.repos.clear
-
-    api.repos.each do |resource|
-      attributes = repo_attributes(resource.to_hash)
-      user.repos << Repo.find_or_create_with(attributes)
-    end
+    update_user_repos_from_api
+    deactivate_orphaned_repos
   end
 
   private
+
+  def user_repos
+    user.repos
+  end
+
+  def deactivate_orphaned_repos
+    user_repos.each do |repo|
+      if repo_not_visible_to_anyone?(repo)
+        deactivate_repo(repo)
+      end
+    end
+  end
+
+  def deactivate_repo(repo)
+    repo.deactivate
+    repo.memberships.destroy_all
+    if repo.subscription
+      RepoSubscriber.unsubscribe(repo, user)
+    end
+  end
+
+  def api_repos
+    @api_repos ||= api.repos.map do |resource|
+      find_or_create_repo(resource)
+    end
+  end
+
+  def update_user_repos_from_api
+    api_repos.each do |repo|
+      unless user_repos.include?(repo)
+        user.repos << repo
+      end
+    end
+  end
+
+  def repo_not_visible_to_anyone?(repo)
+    !api_repos.include?(repo) && repo.memberships.count == 1
+  end
+
+  def find_or_create_repo(resource)
+    attributes = repo_attributes(resource.to_hash)
+    Repo.find_or_create_with(attributes)
+  end
 
   def repo_attributes(attributes)
     {
--- a/app/services/repo_synchronization.rb
+++ b/app/services/repo_synchronization.rb
@@ -11,6 +11,7 @@ class RepoSynchronization
   def start
     update_user_repos_from_api
     deactivate_orphaned_repos
+    destroy_stale_memberships
   end
 
   private
@@ -32,6 +33,14 @@ class RepoSynchronization
     repo.memberships.destroy_all
     if repo.subscription
       RepoSubscriber.unsubscribe(repo, user)
+    end
+  end
+
+  def destroy_stale_memberships
+    user.memberships.each do |membership|
+      if !api_repos.include?(membership.repo)
+        membership.destroy
+      end
     end
   end
 
--- a/spec/services/repo_synchronization_spec.rb
+++ b/spec/services/repo_synchronization_spec.rb
@@ -118,6 +118,21 @@ describe RepoSynchronization do
         expect(repo).to be_active
       end
 
+      it "removes the user's membership when there are other memberships" do
+        removed_membership = create(:membership)
+        repo = removed_membership.repo
+        repo.update(active: true)
+        create(:membership, repo: repo)
+        github_token = "githubtoken"
+        stub_api_repos(repos: [])
+        synchronization =
+          RepoSynchronization.new(removed_membership.user, github_token)
+
+        synchronization.start
+
+        expect(repo.memberships).not_to include(removed_membership)
+      end
+
       it "destroys the users membership when a repo gets deactivated" do
         repo = create(:repo, :active)
         repo.memberships.destroy_all
--- a/spec/services/repo_synchronization_spec.rb
+++ b/spec/services/repo_synchronization_spec.rb
@@ -12,8 +12,7 @@ describe RepoSynchronization do
         }
       }
       resource = double(:resource, to_hash: attributes)
-      api = double(:github_api, repos: [resource])
-      allow(GithubApi).to receive(:new).and_return(api)
+      stub_api_repos(repos: [resource])
       user = create(:user)
       github_token = 'token'
       synchronization = RepoSynchronization.new(user, github_token)
@@ -33,8 +32,7 @@ describe RepoSynchronization do
         }
       }
       resource = double(:resource, to_hash: attributes)
-      api = double(:github_api, repos: [resource])
-      allow(GithubApi).to receive(:new).and_return(api)
+      stub_api_repos(repos: [resource])
       user = create(:user)
       github_token = 'token'
       synchronization = RepoSynchronization.new(user, github_token)
@@ -42,31 +40,6 @@ describe RepoSynchronization do
       synchronization.start
 
       expect(user.repos.first).to be_in_organization
-    end
-
-    it 'replaces existing repos' do
-      attributes = {
-        full_name: 'user/newrepo',
-        id: 456,
-        private: false,
-        owner: {
-          type: 'User'
-        }
-      }
-      resource = double(:resource, to_hash: attributes)
-      github_token = 'token'
-      membership = create(:membership)
-      user = membership.user
-      api = double(:github_api, repos: [resource])
-      allow(GithubApi).to receive(:new).and_return(api)
-      synchronization = RepoSynchronization.new(user, github_token)
-
-      synchronization.start
-
-      expect(GithubApi).to have_received(:new).with(github_token)
-      expect(user.repos.size).to eq(1)
-      expect(user.repos.first.full_github_name).to eq 'user/newrepo'
-      expect(user.repos.first.github_id).to eq 456
     end
 
     it 'renames an existing repo if updated on github' do
@@ -82,9 +55,7 @@ describe RepoSynchronization do
       }
       resource = double(:resource, to_hash: attributes)
       github_token = 'githubtoken'
-
-      api = double(:github_api, repos: [resource])
-      allow(GithubApi).to receive(:new).and_return(api)
+      stub_api_repos(repos: [resource])
       synchronization = RepoSynchronization.new(membership.user, github_token)
 
       synchronization.start
@@ -108,9 +79,8 @@ describe RepoSynchronization do
         }
         resource = double(:resource, to_hash: attributes)
         github_token = 'githubtoken'
+        stub_api_repos(repos: [resource])
         second_user = create(:user)
-        api = double(:github_api, repos: [resource])
-        allow(GithubApi).to receive(:new).and_return(api)
         synchronization = RepoSynchronization.new(second_user, github_token)
 
         synchronization.start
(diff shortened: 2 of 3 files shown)


app/services/repo_synchronization.rb
```diff
@@ -9,15 +9,47 @@ def api
   end
 
   def start
-    user.repos.clear
+    update_user_repos_from_api
+    deactivate_invisble_repos
```
malandrina: Typo: "invisible"

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +93,70 @@
         expect(second_user.reload.repos.size).to eq(1)
       end
     end
+
+    describe "deleted repos" do
```
malandrina: Could we make this more descriptive? `describe "when repo has been deleted"`, maybe?

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +93,70 @@
...
+      it "inactivates a repo when there are no other memberships" do
```
malandrina: Prefer "deactivate" to be consistent with the rest of the code.

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +93,70 @@
...
+        expect(repo).to be_active
+      end
+
+      it "destroys the users membership when a repo gets inactivated" do
+        repo = create(:repo, :active)
+        repo.memberships.destroy_all
+        membership = create(:membership, repo: repo)
+        user = membership.user
+        github_token = "githubtoken"
+
```
malandrina: Style nitpick: Looks like lines 151-153 are part of the setup phase, so this newline can be removed.

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +93,70 @@
...
+        expect(membership.user.repos).to include(second_repo)
+      end
+
+      it "does not inactivate a repo when there are other memberships" do
+        first_membership = create(:membership)
+        repo = first_membership.repo
+        repo.update(active: true)
+        membership = create(:membership, repo: repo)
+        github_token = "githubtoken"
+
```
malandrina: See below comment about removing the newline.

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +93,70 @@
         expect(second_user.reload.repos.size).to eq(1)
       end
     end
+
+    describe "deleted repos" do
+      it "inactivates a repo when there are no other memberships" do
+        repo = create(:repo, :active)
+        repo.memberships.destroy_all
```
malandrina: Is this necessary?

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +93,70 @@
...
+          full_name: second_repo.full_github_name,
+          id: second_repo.github_id,
+          private: true,
+          owner: {
+            type: "User"
+          }
+        }
+        resource = double(:resource, to_hash: attributes)
+        github_token = "githubtoken"
+        api = double(:github_api, repos: [resource])
```
malandrina: Consider extracting lines 114-115 to a private method that can be called here and on lines 134-135 (and 151-152).

spec/services/repo_synchronization_spec.rb
thorncp: This should probably be something like `"when a user no longer has access to a repo"`, because we don't know that the repository was deleted, only that the user can no longer see it. Jake and I tried out several terms/phrases for this, and settled on `invisible` in one of the method names. I don't know if that's the most clear way to say that though. Any thoughts?

app/services/repo_synchronization.rb
```diff
@@ -9,15 +9,47 @@ def api
...
     end
   end
 
-  private
+  def deactivate_repo(repo)
+    repo.deactivate
+    repo.memberships.destroy_all
+  end
+
+  def update_user_repos_from_api
```
malandrina: Looks like there's a lot going on here. What do you think about moving the memoization of repos returned by the API to a separate method? Something like this:

```
def api_repos
  @api_repos ||= api.repos.map do |resource|
    find_or_create_repo(resource)
  end
end

def update_user_repos_from_api
  api_repos.each do |repo|
    unless user_repos.include?(repo)
      user.repos << repo
    end
  end
end
```

spec/services/repo_synchronization_spec.rb
thorncp: We added that because the [after create hook](https://github.com/thoughtbot/hound/blob/1c86a7bb55907ce522a769bc54dfe309fc3c0c30/spec/factories.rb#L24-L28) in the factory was causing us to have more memberships than we wanted in this spec. It's definitely a bandaid, but the only way we thought of to fix that was to remove the hook completely. We didn't want to do that in this PR though, since it would likely require updating other specs.

Is there a more sensible approach? I feel like the factories shouldn't have any hooks at all. It took us awhile to track down where the extra records were being creating, which was pretty frustrating.

spec/services/repo_synchronization_spec.rb
malandrina: `"when a user no longer has access to a repo"` sounds good to me!

spec/services/repo_synchronization_spec.rb
malandrina: Got it. In that case I think it's fine. We could move the hook to a trait (`:with_user` ?), but that's a refactoring and should be done in a separate PR.

app/services/repo_synchronization.rb
thorncp: I like it.

spec/services/repo_synchronization_spec.rb
thorncp: :+1: for the trait

thorncp:
Needs re-review

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +88,71 @@
...
+        stub_api_repos(repos: [])
+        synchronization = RepoSynchronization.new(membership.user, github_token)
+
+        synchronization.start
+        repo.reload
+
+        expect(repo).to be_active
+      end
+
+      it "destroys the users membership when a repo gets inactivated" do
```
malandrina: "inactivated" => "deactivated"

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +88,71 @@
...
+        synchronization.start
+        repo.reload
+        second_repo.reload
+
+        expect(repo).not_to be_active
+        expect(second_repo).to be_active
+        expect(membership.user.repos).to include(second_repo)
+      end
+
+      it "does not inactivate a repo when there are other memberships" do
```
malandrina: "inactivate" => "deactivate"

spec/services/repo_synchronization_spec.rb
thorncp: Apparently my grep needs work :confounded:

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +88,71 @@
...
+        stub_api_repos(repos: [resource])
+
+        synchronization = RepoSynchronization.new(membership.user, github_token)
+
+        synchronization.start
+        repo.reload
+        second_repo.reload
+
+        expect(repo).not_to be_active
+        expect(second_repo).to be_active
```
malandrina: I'm not sure that lines 118-119 are useful. What is the value of these assertions?

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +88,71 @@
...
+        expect(membership.user.repos).to include(second_repo)
+      end
+
+      it "does not deactivate a repo when there are other memberships" do
+        first_membership = create(:membership)
+        repo = first_membership.repo
```
malandrina: Lines 123-124 could be condensed to `repo = create(:membership).repo`.

app/services/repo_synchronization.rb
```diff
@@ -9,15 +9,51 @@ def api
   end
 
   def start
-    user.repos.clear
+    update_user_repos_from_api
+    deactivate_invisible_repos
```
malandrina: Here's a thought: I'm not sure the word "invisible" adds much here because it is kind of ambiguous, so this could become simply `deactivate_repos`. Thoughts?

spec/services/repo_synchronization_spec.rb
thorncp: I think we added these because our implementation was deactivating *all* repos. However, this should be covered by other specs, and I think it was only meant to be temporary. I'll dig into it further.

app/services/repo_synchronization.rb
thorncp: I think that would be fine. I believe our thought process was something like "but *what* repos would be deactivated?". Perhaps we simply overthought it, given that it's a private method. I agree that "invisible" really has no meaning here.

spec/services/repo_synchronization_spec.rb
thorncp: Yep, totally superfluous.

malandrina: :cool:

malandrina:
Might want to get another set of eyes on this to validate the business logic, but it looks good to me.

app/services/repo_synchronization.rb
```diff
@@ -9,15 +9,51 @@ def api
...
+      if repo_not_visible_to_anyone?(repo)
+        deactivate_repo(repo)
+      end
     end
   end
 
-  private
+  def deactivate_repo(repo)
+    repo.deactivate
+    repo.memberships.destroy_all
```
gylaz: Should we remove subscription in this case too?

app/services/repo_synchronization.rb
```diff
@@ -9,15 +9,51 @@ def api
...
+  def update_user_repos_from_api
+    api_repos.each do |repo|
+      unless user_repos.include?(repo)
+        user.repos << repo
+      end
+    end
+  end
+
+  def repo_not_visible_to_anyone?(repo)
+    !api_repos.include?(repo) && repo.memberships.count == 1
```
gylaz: If the repo is not part of this user's repos, but some other user does have a membership to it, this logic would be flawed...?

app/services/repo_synchronization.rb
thorncp: Yes, yes we should.

spec/services/repo_synchronization_spec.rb
```diff
@@ -118,5 +88,55 @@
         expect(second_user.reload.repos.size).to eq(1)
       end
     end
+
+    describe "when a user no longer has access to a repo" do
+      it "deactivates a repo when there are no other memberships" do
+        repo = create(:repo, :active)
+        repo.memberships.destroy_all
+        membership = create(:membership, repo: repo)
```
gylaz: Instead of lines 94-95 what do you think about grabbing the repo from `memberships`?

app/services/repo_synchronization.rb
thorncp: This method is called inside a loop over the user's repos, so I don't think that will ever be the case.

spec/services/repo_synchronization_spec.rb
thorncp: The default repo factory sets active to false. So, if we use the membership's repo, we'd have to add a line to set that repo to active. Unless there's some way to create a membership and pass the `:active` trait through to the repo? Even there isn't, updating the repo's active flag would probably more clear than what's here.

app/services/repo_synchronization.rb
```diff
@@ -9,15 +9,51 @@ def api
...
   def start
-    user.repos.clear
+    update_user_repos_from_api
+    deactivate_repos
+  end
+
+  private
+
+  def user_repos
+    @user_repos ||= user.repos
```
gylaz: I don't think this memoization is necessary. `user.repos` will cache the db call.

app/services/repo_synchronization.rb
gylaz: Gotcha, the `repo.memberships.count == 1` assumes only the current user's membership. Makes sense now.

gylaz:
Looks good to me.

app/services/repo_synchronization.rb
thorncp: Yep. I'm definitely open to feedback for making this more clear, though.

thorncp:
Updated to kill the subscription if necessary.

gylaz:
Once tested on staging, ready to be merged.

app/services/repo_synchronization.rb
```diff
@@ -9,15 +9,54 @@ def api
...
+  private
+
+  def user_repos
+    user.repos
+  end
 
-    api.repos.each do |resource|
-      attributes = repo_attributes(resource.to_hash)
-      user.repos << Repo.find_or_create_with(attributes)
+  def deactivate_repos
```
salbertson: Would `deactivate_orphaned_repos` make more sense?

salbertson:
@thorncp, is it possible to have a repo in our database without any memberships that hasn't been deleted on GitHub?

I'm wondering if we really want to disable Hound and unsubscribe, if there's a subscription, whenever there are no memberships.

thorncp:
@salbertson that's possible if the user who enabled the repo was the only person with access to it that has logged into hound, and then they are removed from that repo on github. We certainly don't have to disable and unsubscribe if a subscription exists, but that seems like an aggressive approach to me.

app/services/repo_synchronization.rb
thorncp: That's more clear. I like it.

thorncp:
I discovered a case where memberships weren't being destroyed when the user can no longer see a repo, and other memberships still exist. Fixed in  0814da6

thorncp:
Per Trello discussion, closing this until a real world scenario occurs that forces a solution.

PR #578, opened 2015-01-24
Document SCSS support

croaky:
We have addressed the reported SCSS bugs during the internal beta with
thoughtbot. It seems like we are ready to open it to the public.


gylaz:
LGTM.

PR #582, opened 2015-01-26
Impliment new header layout and visual style

frechg:
screen shot 2015-01-25 at 8 25 03 pm

### DIFF
--- a/app/views/shared/_header.haml
+++ b/app/views/shared/_header.haml
@@ -1,14 +1,17 @@
 %header
-  = link_to root_path do
-    %h1.logo Hound
+  %ul.pull-left
+    %li
+      = link_to repos_path, class: "logo" do
+        %h1 Hound
+    %li
+      = link_to "Docs", configuration_path, class: "docs"
+    %li
+      = link_to "FAQ", faq_path, class: "faq"
   - if signed_in?
-    %ul.nav
-      %li.dropdown
-        %input#account-menu-toggle{type: 'checkbox'}
-        %label.toggle{for: 'account-menu-toggle'} #{current_user}
-        .menu
-          = link_to 'Your Repos', repos_path, class: 'repo-link'
-          = link_to "Your Account", account_path
-          = link_to 'Configure Hound', configuration_path
-          = link_to 'FAQ', faq_path
-          = link_to 'Sign Out', sign_out_path
+    %ul.pull-right
+      %li
+        = link_to account_path, class: "account" do
+          .avatar
+          %span #{current_user}
+      %li
+        = link_to "Sign Out", sign_out_path, class: "sign-out"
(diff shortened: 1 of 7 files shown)


croaky:
Code looks good. Tests are failing because email_address is nil in some cases. Might need to define email_address method on Guest returning "". Should also check the production database to see if there are any User records with nil email addresses. User.where(email_address: nil)

croaky:
Do we need the "sign out" link showing at all times? Current version is nice and clean with it hidden behind a click of the username. If we kept just the sign out hidden, it might also give us more room across the header for users with longer names.

salbertson:
What about putting the signout link somewhere else, like a simple footer?

croaky:
I had the same thought but my list of repos is crazy long. It also conceptually makes sense to me to group authentication with my name/avatar.

frechg:
Yeah, sign out link could move to a footer. I am just progressively adding the new UI right now and adding a footer will come later if it still makes sense.

frechg:
@croaky what do you think about only showing the avatar right now if the current_user has an email?

app/views/shared/_header.haml
```diff
@@ -1,14 +1,18 @@
...
-        .menu
-          = link_to 'Your Repos', repos_path, class: 'repo-link'
-          = link_to "Your Account", account_path
-          = link_to 'Configure Hound', configuration_path
-          = link_to 'FAQ', faq_path
-          = link_to 'Sign Out', sign_out_path
+    %ul.pull-right
+      %li
+        = link_to account_path, class: "account" do
+          - unless current_user.email_address.nil?
```
graysonwright: I think we should try to avoid a double-negative here. How about `- if current_user.email_address`

salbertson: You could do something like this to support blank email addresses, not only nil ones:

```
- if current_user.email_address.present?
...
```

salbertson:
One comment, looks good to me. :+1:

app/views/shared/_header.haml
```diff
@@ -1,14 +1,18 @@
...
-          = link_to "Your Account", account_path
-          = link_to 'Configure Hound', configuration_path
-          = link_to 'FAQ', faq_path
-          = link_to 'Sign Out', sign_out_path
+    %ul.pull-right
+      %li
+        = link_to account_path, class: "account" do
+          - if current_user.email_address.present?
+            .avatar(style="background-image: url('#{avatar_url(current_user)}')")
+          %span #{current_user.github_username}
```
gylaz: How about `%span = current_user.github_username` instead for consistency.

croaky:
> @croaky what do you think about only showing the avatar right now if the current_user has an email?

Works for me.

PR #583, opened 2015-01-26
Enable CoffeeScript

salbertson:


.hound.yml
```diff
@@ -0,0 +1,5 @@
+ruby:
+  enabled: true
+
+coffee_script:
+  enabled: true
```
croaky: Also SCSS?

croaky:
Good to merge.

gylaz:
Why CoffeeScript? Why not all the languages?

salbertson:
@gylaz, because it came up [here](https://github.com/thoughtbot/hound/pull/575#diff-for-comment-23513806). We can enable other stuff in a different PR.

PR #584, opened 2015-01-26
Upgrade scss-lint gem

salbertson:
Trying to get rid of `undefined method bubbles?` exception.

https://github.com/causes/scss-lint/issues/278

This worked locally so we'll see if it runs successfully on TravisCI.

screenshot 2015-01-26 08 51 48

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -265,7 +265,7 @@ GEM
     rufus-scheduler (3.0.9)
       tzinfo
     safe_yaml (1.0.4)
-    sass (3.4.9)
+    sass (3.4.10)
     sass-rails (5.0.1)
       railties (>= 4.0.0, < 5.0)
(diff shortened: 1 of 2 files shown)


gylaz:
Have we tried upgrading the gem to see if the errors would go away? @thorncp Had a reproducible seed on a branch we can test against.

salbertson:
@gylaz, that's a good idea, ha. I haven't tried.

What branch? I'll try out both potential fixes.

salbertson:
That didn't work.

https://travis-ci.org/thoughtbot/hound/builds/48377766

screenshot 2015-01-26 11 07 30

thorncp:
The branch I snapshotted is `bubbles`, and the seed is `1496`.

https://trello.com/c/8jnTDeQk

PR #586, opened 2015-01-27
Adds owners tables and 	assigns owner in repo creation / update

srt32:
This PR updates the `RepoSynchronization` process and `Build`'s to assign the
`repo` to its parent `owner` (which is either a user or an organization). Tying
the `repo` to the `owner` is a preliminary step to having `owner` level
configuration.

New `repo`'s and updated `repo`'s (updated via the user clicking 'Refresh Repo
List' or via a build) will have their `owner`s populated.  We'll need to
backfill the `repo`'s that are not new and are not refreshed. I propose we merge
this PR as is and then have follow up work that adds an idempotent rake task
(which requires calls to GitHub) that will backfill `repo`'s missing
`owner_id`'s. Once the data is backfilled, we can apply a `null: false`
constraint to `repos.owner_id`.

Part of
https://trello.com/c/V0jY5LWs/452-migrate-existing-config-files-into-database

### DIFF
--- a/app/jobs/buildable.rb
+++ b/app/jobs/buildable.rb
@@ -3,6 +3,11 @@ require 'resque'
 module Buildable
   def perform(payload_data)
     payload = Payload.new(payload_data)
+    Owner.upsert(
+      github_id: payload.repository_owner_id,
+      name: payload.repository_owner_name,
+      organization: payload.repository_owner_is_organization?
+    )
     build_runner = BuildRunner.new(payload)
     build_runner.run
   rescue Resque::TermException
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -1,0 +1,11 @@
+class Owner < ActiveRecord::Base
+  has_many :repos
+
+  def self.upsert(github_id:, name:, organization:)
+    owner = find_or_initialize_by(github_id: github_id)
+    owner.name = name
+    owner.organization = organization
+    owner.save!
+    owner
+  end
+end
--- a/app/models/payload.rb
+++ b/app/models/payload.rb
@@ -1,5 +1,3 @@
-require 'json'
-
 class Payload
   pattr_initialize :unparsed_data
 
@@ -35,8 +33,16 @@ class Payload
     data['zen']
   end
 
-  def repository_owner
+  def repository_owner_id
+    repository["owner"]["id"]
+  end
+
+  def repository_owner_name
     repository["owner"]["login"]
+  end
+
+  def repository_owner_is_organization?
+    repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
   private
--- a/app/models/pull_request.rb
+++ b/app/models/pull_request.rb
@@ -21,8 +21,8 @@ class PullRequest
     )
   end
 
-  def repository_owner
-    payload.repository_owner
+  def repository_owner_name
+    payload.repository_owner_name
   end
 
   def opened?
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -1,9 +1,9 @@
 class Repo < ActiveRecord::Base
-  has_many :memberships
-  has_many :users, through: :memberships
   has_many :builds
-
+  has_many :memberships
+  belongs_to :owner
   has_one :subscription
+  has_many :users, through: :memberships
 
   alias_attribute :name, :full_github_name
 
--- a/app/models/style_guide/scss.rb
+++ b/app/models/style_guide/scss.rb
@@ -53,7 +53,10 @@ module StyleGuide
     end
 
     def default_config_file
-      DefaultConfigFile.new(DEFAULT_CONFIG_FILENAME, repository_owner).path
+      DefaultConfigFile.new(
+        DEFAULT_CONFIG_FILENAME,
+        repository_owner_name
+      ).path
     end
   end
 end
--- a/app/services/repo_synchronization.rb
+++ b/app/services/repo_synchronization.rb
@@ -1,12 +1,6 @@
 class RepoSynchronization
-  ORGANIZATION_TYPE = 'Organization'
-
   pattr_initialize :user, :github_token
   attr_reader :user
-
-  def api
-    @api ||= GithubApi.new(github_token)
-  end
 
   def start
     user.repos.clear
@@ -19,12 +13,27 @@ class RepoSynchronization
 
   private
 
+  def api
+    @api ||= GithubApi.new(github_token)
+  end
+
   def repo_attributes(attributes)
+    owner = upsert_owner(attributes[:owner])
+
     {
       private: attributes[:private],
       github_id: attributes[:id],
       full_github_name: attributes[:full_name],
-      in_organization: attributes[:owner][:type] == ORGANIZATION_TYPE
+      in_organization: attributes[:owner][:type] == GithubApi::ORGANIZATION_TYPE,
+      owner: owner,
     }
+  end
+
+  def upsert_owner(owner_attributes)
+    Owner.upsert(
+      github_id: owner_attributes[:id],
+      name: owner_attributes[:login],
+      organization: owner_attributes[:type] == GithubApi::ORGANIZATION_TYPE
+    )
   end
 end
--- a/db/migrate/20141211224528_create_owners.rb
+++ b/db/migrate/20141211224528_create_owners.rb
@@ -1,0 +1,13 @@
+class CreateOwners < ActiveRecord::Migration
+  def change
+    create_table :owners do |t|
+      t.timestamps null: false
+      t.integer :github_id, null: false
+      t.string :name, null: false
+      t.boolean :organization, default: false, null: false
+
+      t.index :github_id, unique: true
+      t.index :name, unique: true
+    end
+  end
+end
--- a/db/migrate/20150123224104_add_owner_to_repos.rb
+++ b/db/migrate/20150123224104_add_owner_to_repos.rb
@@ -1,0 +1,6 @@
+class AddOwnerToRepos < ActiveRecord::Migration
+  def change
+    add_reference :repos, :owner, index: true
+    add_foreign_key :repos, :owners
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -40,6 +40,17 @@ ActiveRecord::Schema.define(version: 20150130051749) do
   add_index "memberships", ["repo_id"], name: "index_memberships_on_repo_id", using: :btree
   add_index "memberships", ["user_id"], name: "index_memberships_on_user_id", using: :btree
 
+  create_table "owners", force: :cascade do |t|
+    t.datetime "created_at",                   null: false
+    t.datetime "updated_at",                   null: false
+    t.integer  "github_id",                    null: false
+    t.string   "name",                         null: false
+    t.boolean  "organization", default: false, null: false
+  end
+
+  add_index "owners", ["github_id"], name: "index_owners_on_github_id", unique: true, using: :btree
+  add_index "owners", ["name"], name: "index_owners_on_name", unique: true, using: :btree
+
   create_table "repos", force: :cascade do |t|
     t.integer  "github_id",                                    null: false
     t.boolean  "active",                       default: false, null: false
@@ -49,11 +60,13 @@ ActiveRecord::Schema.define(version: 20150130051749) do
     t.datetime "updated_at"
     t.boolean  "private"
     t.boolean  "in_organization"
+    t.integer  "owner_id"
   end
 
   add_index "repos", ["active"], name: "index_repos_on_active", using: :btree
   add_index "repos", ["full_github_name"], name: "index_repos_on_full_github_name", unique: true, using: :btree
   add_index "repos", ["github_id"], name: "index_repos_on_github_id", using: :btree
+  add_index "repos", ["owner_id"], name: "index_repos_on_owner_id", using: :btree
 
   create_table "subscriptions", force: :cascade do |t|
     t.datetime "created_at",                                                               null: false
@@ -92,4 +105,5 @@ ActiveRecord::Schema.define(version: 20150130051749) do
 
   add_index "violations", ["build_id"], name: "index_violations_on_build_id", using: :btree
 
+  add_foreign_key "repos", "owners"
 end
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -4,8 +4,9 @@ require "base64"
 require "active_support/core_ext/object/with_options"
 
 class GithubApi
-  SERVICES_TEAM_NAME = "Services"
+  ORGANIZATION_TYPE = "Organization"
   PREVIEW_MEDIA_TYPE = "application/vnd.github.moondragon-preview+json"
+  SERVICES_TEAM_NAME = "Services"
 
   pattr_initialize :token
 
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,4 +1,7 @@
 FactoryGirl.define do
+  sequence(:github_id)
+  sequence(:github_name) { |n| "github_name#{n}" }
+
   factory :build do
     repo
 
@@ -17,13 +20,13 @@ FactoryGirl.define do
     end
 
     sequence(:full_github_name) { |n| "user/repo#{n}" }
-    sequence(:github_id) { |n| n }
+    github_id
     private false
     in_organization false
   end
 
   factory :user do
-    sequence(:github_username) { |n| "github#{n}" }
+    github_username { generate(:github_name) }
 
     ignore do
       repos []
@@ -52,5 +55,10 @@ FactoryGirl.define do
     patch_position 1
     line_number 42
     messages ["Trailing whitespace detected."]
+  end
+
+  factory :owner do
+    github_id
+    name { generate(:github_name) }
   end
 end
--- a/spec/jobs/buildable_spec.rb
+++ b/spec/jobs/buildable_spec.rb
@@ -1,7 +1,4 @@
-require 'fast_spec_helper'
-require 'app/jobs/buildable'
-require 'app/models/payload'
-require 'app/services/build_runner'
+require "spec_helper"
 
 describe Buildable do
   class TestJob
@@ -10,11 +7,16 @@ describe Buildable do
 
   describe '.perform' do
     it 'runs build runner' do
-      payload_data = double(:payload_data)
-      payload = double(:payload)
       build_runner = double(:build_runner, run: nil)
+      payload = double(
+        "Payload",
+        repository_owner_id: 1,
+        repository_owner_name: "test",
+        repository_owner_is_organization?: true
+      )
       allow(Payload).to receive(:new).and_return(payload)
       allow(BuildRunner).to receive(:new).and_return(build_runner)
+      allow(Owner).to receive(:upsert)
 
       TestJob.perform(payload_data)
 
@@ -26,11 +28,42 @@ describe Buildable do
     it 'retries when Resque::TermException is raised' do
       allow(Payload).to receive(:new).and_raise(Resque::TermException.new(1))
       allow(Resque).to receive(:enqueue)
-      payload_data = double(:payload_data)
 
       TestJob.perform(payload_data)
 
       expect(Resque).to have_received(:enqueue).with(TestJob, payload_data)
     end
+
+    it "upserts repository owner" do
+      github_id = "2345"
+      name = "thoughtbot"
+      payload_data = payload_data(
+        github_id: github_id,
+        name: name
+      )
+      build_runner = double("BuildRunner", run: true)
+      allow(BuildRunner).to receive(:new).and_return(build_runner)
+      allow(Owner).to receive(:upsert)
+
+      TestJob.perform(payload_data)
+
+      expect(Owner).to have_received(:upsert).with(
+        github_id: github_id,
+        name: name,
+        organization: true
+      )
+    end
+  end
+
+  def payload_data(github_id: 1234, name: "test")
+    {
+      "repository" => {
+        "owner" => {
+          "id" => github_id,
+          "login" => name,
+          "type" => "Organization"
+        }
+      }
+    }
   end
 end
--- a/spec/models/owner_spec.rb
+++ b/spec/models/owner_spec.rb
@@ -1,0 +1,39 @@
+require "spec_helper"
+
+describe Owner do
+  it { should have_many(:repos) }
+
+  describe ".upsert" do
+    context "when owner does not exist" do
+      it "creates owner" do
+        github_id = 1234
+        name = "thoughtbot"
+        organization = true
+
+        new_owner = Owner.upsert(
+          github_id: github_id,
+          name: name,
+          organization: organization
+        )
+
+        expect(new_owner).to be_persisted
+      end
+    end
+
+    context "when owner exists" do
+      it "updates owner" do
+        owner = create(:owner)
+        new_name = "ralphbot"
+
+        updated_owner = Owner.upsert(
+          github_id: owner.github_id,
+          name: new_name,
+          organization: true
+        )
+
+        expect(updated_owner.name).to eq new_name
+        expect(updated_owner.organization).to eq true
+      end
+    end
+  end
+end
--- a/spec/models/payload_spec.rb
+++ b/spec/models/payload_spec.rb
@@ -1,6 +1,6 @@
-require 'fast_spec_helper'
+require "fast_spec_helper"
 require "attr_extras"
-require 'app/models/payload'
+require "app/models/payload"
 
 describe Payload do
   describe '#changed_files' do
@@ -61,7 +61,7 @@ describe Payload do
     end
   end
 
-  describe "#repository_owner" do
+  describe "#repository_owner_name" do
     it "returns the owner of the repo's name" do
       data = {
         "repository" => {
@@ -73,7 +73,57 @@ describe Payload do
 
       payload = Payload.new(data)
 
-      expect(payload.repository_owner).to eq "thoughtbot"
+      expect(payload.repository_owner_name).to eq "thoughtbot"
+    end
+  end
+
+  describe "#repository_owner_id" do
+    it "returns the owner of the repo's ID" do
+      data = {
+        "repository" => {
+          "owner" => {
+            "id" => 1
+          }
+        }
+      }
+
+      payload = Payload.new(data)
+
+      expect(payload.repository_owner_id).to eq 1
+    end
+  end
+
+  describe "#repository_owner_is_organization?" do
+    context "when the repository owner is a user" do
+      it "returns false" do
+        payload_json = {
+          "repository" => {
+            "owner" => {
+              "id" => 1,
+              "type" => "User"
+            }
+          }
+        }
+        payload = Payload.new(payload_json)
+
+        expect(payload.repository_owner_is_organization?).to be false
+      end
+    end
+
+    context "when the repository owner is an organization" do
+      it "returns true" do
+        payload_json = {
+          "repository" => {
+            "owner" => {
+              "id" => 1,
+              "type" => "Organization"
+            }
+          }
+        }
+        payload = Payload.new(payload_json)
+
+        expect(payload.repository_owner_is_organization?).to be true
+      end
     end
   end
 end
--- a/spec/models/repo_spec.rb
+++ b/spec/models/repo_spec.rb
@@ -1,10 +1,11 @@
 require "spec_helper"
 
 describe Repo do
-  it { should have_many(:users).through(:memberships) }
   it { should have_many :builds }
   it { should validate_presence_of :full_github_name }
   it { should validate_presence_of :github_id }
+  it { should belong_to :owner }
+  it { should have_many(:users).through(:memberships) }
 
   it "validates uniqueness of github_id" do
     create(:repo)
--- a/spec/models/style_guide/coffee_script_spec.rb
+++ b/spec/models/style_guide/coffee_script_spec.rb
@@ -78,7 +78,7 @@ describe StyleGuide::CoffeeScript do
         spy_on_file_read
         config_file = thoughtbot_configuration_file(StyleGuide::CoffeeScript)
 
-        violations_in("var foo = 'bar'", repository_owner: "thoughtbot")
+        violations_in("var foo = 'bar'", repository_owner_name: "thoughtbot")
 
         expect(File).to have_received(:read).
           with(config_file)
@@ -93,7 +93,7 @@ describe StyleGuide::CoffeeScript do
         spy_on_file_read
         config_file = default_configuration_file(StyleGuide::CoffeeScript)
 
-        violations_in("var foo = 'bar'", repository_owner: "not_thoughtbot")
+        violations_in("var foo = 'bar'", repository_owner_name: "foo")
 
         expect(File).to have_received(:read).
           with(config_file)
@@ -104,9 +104,12 @@ describe StyleGuide::CoffeeScript do
 
     private
 
-    def violations_in(content, repository_owner: "ralph")
+    def violations_in(content, repository_owner_name: "ralph")
       repo_config = double("RepoConfig", enabled_for?: true, for: {})
-      style_guide = StyleGuide::CoffeeScript.new(repo_config, repository_owner)
+      style_guide = StyleGuide::CoffeeScript.new(
+        repo_config,
+        repository_owner_name
+      )
       style_guide.violations_in_file(build_file(content)).flat_map(&:messages)
     end
 
--- a/spec/services/repo_synchronization_spec.rb
+++ b/spec/services/repo_synchronization_spec.rb
@@ -1,91 +1,67 @@
-require 'spec_helper'
+require "spec_helper"
 
 describe RepoSynchronization do
-  describe '#start' do
-    it 'saves privacy flag' do
-      attributes = {
-        full_name: 'user/newrepo',
-        id: 456,
-        private: true,
-        owner: {
-          type: 'User'
-        }
-      }
-      resource = double(:resource, to_hash: attributes)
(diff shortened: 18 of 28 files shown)


spec/factories.rb
```diff
@@ -1,4 +1,7 @@
 FactoryGirl.define do
+  sequence(:github_id) { |n| n }
```
thorncp: Is this block necessary?

thorncp:
This looks good to me. Though, I regret not finishing up #577 yet. Conflicts abound :boom:

croaky:
Does an owner mean an organization or a user? Should `repos.in_organization` get moved to `owners.organization`?

app/models/repo.rb
```diff
@@ -1,8 +1,8 @@
 class Repo < ActiveRecord::Base
+  belongs_to :owner
```
croaky: > Order ActiveRecord associations alphabetically by attribute name.

https://github.com/thoughtbot/guides/tree/master/style/rails#rails

db/migrate/20150123224104_add_owner_to_repos.rb
```diff
@@ -0,0 +1,6 @@
+class AddOwnerToRepos < ActiveRecord::Migration
+  def change
+    add_reference :repos, :owner, index: true
+    add_foreign_key :repos, :owners
```
croaky: Interesting. Inspired by http://robots.thoughtbot.com/referential-integrity-with-foreign-keys ? I still haven't done this myself but I think I like it.

spec/factories.rb
```diff
@@ -29,7 +32,7 @@
   end
 
   factory :user do
-    sequence(:github_username) { |n| "github#{n}" }
+    github_username { |n| "github#{n}" }
```
croaky: The block looks like repetition of the `github_name` sequence.

```ruby
github_username { generate(:github_name) }
```

db/migrate/20141211224528_create_owners.rb
```diff
@@ -0,0 +1,12 @@
+class CreateOwners < ActiveRecord::Migration
+  def change
+    create_table :owners do |t|
+      t.timestamps null: false
+      t.integer :github_id, null: false
+      t.string :github_name, null: false
```
croaky: Is it worth naming this column `github_username` to match the existing `users.github_username`?

app/models/owner.rb
```diff
@@ -0,0 +1,14 @@
+class Owner < ActiveRecord::Base
+  has_many :repos, dependent: :destroy
+
+  def self.upsert(github_id:, github_name:)
```
croaky: How very Mongo of you.

app/models/owner.rb
```diff
@@ -0,0 +1,14 @@
...
+    owner = find_or_initialize_by(github_id: github_id)
+    owner.github_name = github_name
+
+    if owner.changed?
```
croaky: Does a spec fail if this `if` statement is deleted? Is it needed?

app/services/repo_synchronization.rb
```diff
@@ -19,12 +15,24 @@ def start
...
+  end
+
   def repo_attributes(attributes)
     {
       private: attributes[:private],
       github_id: attributes[:id],
       full_github_name: attributes[:full_name],
-      in_organization: attributes[:owner][:type] == ORGANIZATION_TYPE
+      in_organization: attributes[:owner][:type] == ORGANIZATION_TYPE,
+      owner: upsert_owner(attributes[:owner])
```
croaky: > Use a trailing comma after each item in a multi-line list, including the last item.

https://github.com/thoughtbot/guides/tree/master/style/ruby
https://github.com/thoughtbot/guides/blob/master/style/ruby/sample.rb#L49

spec/models/repo_spec.rb
```diff
@@ -1,6 +1,7 @@
 require "spec_helper"
 
 describe Repo do
+  it { should belong_to :owner }
```
croaky: After the associations are sorted by attribute name,

> Order validation, association, and method tests in the same order that they appear in the class.

https://github.com/thoughtbot/guides/tree/master/style/testing

spec/services/repo_synchronization_spec.rb
```diff
@@ -1,91 +1,85 @@
...
       attributes = {
-        full_name: 'user/newrepo',
+        full_name: "user/newrepo",
         id: 456,
         private: true,
         owner: {
-          type: 'User'
+          type: "User",
+          id: 1,
+          login: "thoughtbot"
```
croaky: `:sort`, trailing comma

croaky: Same for other `owner` hashes later

spec/services/repo_synchronization_spec.rb
```diff
@@ -103,20 +97,79 @@
...
+          attributes = {
+            full_name: "thoughtbot/newrepo",
+            id: repo_github_id,
+            private: true,
+            owner: {
+              type: "Organization",
+              id: owner.github_id,
+              login: owner.github_name
+            }
+          }
```
croaky: Looks like a lot of repeated setup. I wonder if it'd be any easier to read if it were pulled out into a method. Don't know exactly what it'd look like, would need to extract then compare. Whatever you think.

app/models/owner.rb
```diff
@@ -0,0 +1,14 @@
+class Owner < ActiveRecord::Base
+  has_many :repos, dependent: :destroy
```
croaky: If deleting the owner deletes the repo, does it continue to cascade? Are the memberships, builds, and subscription deleted? Do they also need `dependent: :destroy` and foreign key constraints?

srt32: Good question! We don't currently destroy `owner`s anywhere via code. Assuming we do want to FK all the things, perhaps we should remove the `dependent: :destroy` here for now. If we do want to figure out the deletion use cases (cascading through all the associations) we can do that in another PR and apply all the FK's and `dependent: :destroys` consistently.

Thoughts?

croaky: My experience has been that if we don't consider the deletion case when we add the associations, we tend to forget to do it later and risk creating orphaned records. Even if we don't expose the UI, sometimes we'll accidentally do it through a rake task or console or something.

Doesn't matter to me whether that consideration comes in the form of a Trello card for later or handled in this pull request if we're confident that it should happen a certain way. My gut is that we probably want to remove the `dependent: :destroy` on `Owner has_many :repos` but definitely want to delete the subscription when a repo is deleted, so separate Trello card for that case.

app/models/owner.rb
srt32: Nice, removed in cbb7dfa

app/models/owner.rb
srt32: Ok, `dependent: :destroy` removed in e704e26 and a card for tracking here: https://trello.com/c/uTBir4jX/474-handle-deletion-cases

db/migrate/20141211224528_create_owners.rb
srt32: I think `username` could be confusing since an `owner` is either a `user` or an `organization`.

db/migrate/20150123224104_add_owner_to_repos.rb
srt32: Yup! And also inspired from `dependent: :destroys` not always working as well as giving us some more confidence when running SQL queries that delete / update data.

app/models/repo.rb
srt32: Done in 8f54898

app/services/repo_synchronization.rb
srt32: Done in 8f54898

spec/factories.rb
srt32: Nice! Updated in a843dff

spec/factories.rb
srt32: Done in a843dff

spec/models/repo_spec.rb
srt32: Done in e4a1fc6

spec/services/repo_synchronization_spec.rb
srt32: Done in e4a1fc6

spec/services/repo_synchronization_spec.rb
srt32: @croaky I took a stab in 04a1a0f.  Thoughts?

salbertson:
@srt32, will all team members be able to update the configuration with this change and "owner level configuration"?

db/migrate/20141211224528_create_owners.rb
gylaz: How about `github_login`, since that would closer match what GitHub calls it?

spec/services/repo_synchronization_spec.rb
```diff
@@ -94,29 +70,80 @@
...
+
+          synchronization.start
+
+          owner = Owner.find_by(github_id: owner.github_id)
+          expect(owner.repos.map(&:github_id)).to include(repo_github_id)
+        end
+      end
+    end
+
+    def stub_github_api_repos(repo_id:, owner_id:, owner_name:, private_repo: true, repo_name: "thoughtbot/newrepo")
```
gylaz: How about?
```rb
def stub_github_api_repos(
  repo_id:,
  owner_id:,
  owner_name:,
  private_repo: true,
  repo_name: "thoughtbot/newrepo"
)
  attributes = ...
```

db/migrate/20141211224528_create_owners.rb
```diff
@@ -0,0 +1,12 @@
...
+
+      t.index :github_name, unique: true
+      t.index :github_id, unique: true
+    end
```
gylaz: Should we also add a `organization` boolean field to this table?

app/models/owner.rb
gylaz: I think the idea was to skip the trip to the db if the record didn't change, right?

srt32:
@salbertson not exactly. This PR is step one for going down that road. This PR introduces owners (orgs or users) so that later we could associate style guides with those owenrs.

salbertson:
@srt32, when would a style guide be associated with a user?

db/migrate/20141211224528_create_owners.rb
srt32: I do think it makes sense to add that column here. Does that mean we would drop it from `repos`?

srt32:
Good question. I was thinking that a `user` would have a style guide the same way an `organization` would have a style guide.

db/migrate/20141211224528_create_owners.rb
gylaz: Yeah, that's what I was thinking.

db/migrate/20141211224528_create_owners.rb
srt32: Maybe just `login`?

spec/services/repo_synchronization_spec.rb
srt32: That approach is much nicer ;)

srt32: Updated in 6882062

db/migrate/20141211224528_create_owners.rb
srt32: Done in dfa719f

srt32:
I added the `owner.organization` column in dfa719f. My next commit will update that column.

db/schema.rb
```diff
@@ -40,6 +39,17 @@
   add_index "memberships", ["repo_id"], name: "index_memberships_on_repo_id", using: :btree
   add_index "memberships", ["user_id"], name: "index_memberships_on_user_id", using: :btree
 
+  create_table "owners", force: :cascade do |t|
+    t.datetime "created_at",                   null: false
+    t.datetime "updated_at",                   null: false
+    t.integer  "github_id",                    null: false
+    t.string   "github_login",                 null: false
```
salbertson: Should we name this `github_username` to stay consistent with the users table?

app/jobs/buildable.rb
```diff
@@ -3,6 +3,11 @@
 module Buildable
   def perform(payload_data)
     payload = Payload.new(payload_data)
+    Owner.upsert(
+      github_id: payload.repository_owner_id,
+      github_login: payload.repository_owner_name,
+      organization: payload.repository_owner_is_organization?
+    )
```
salbertson: Why is the upsert done during a build and when repos are refreshed? Do we need to do it in both places?

app/models/payload.rb
```diff
@@ -1,4 +1,5 @@
-require 'json'
+require "json"
+require_relative "../../lib/github_api"
```
salbertson: Why are these needed?

app/models/payload.rb
```diff
@@ -35,10 +36,18 @@ def ping?
     data['zen']
   end
 
-  def repository_owner
+  def repository_owner_id
+    repository["owner"]["id"]
+  end
+
+  def repository_owner_name
```
salbertson: `repository_owner_username` to stay consistent?

db/schema.rb
srt32: We went with `login` because we thought `username` might be confusing since an `owner` is either a `user` or an `org`.

https://github.com/thoughtbot/hound/pull/586#discussion_r23898835

app/models/pull_request.rb
```diff
@@ -22,7 +22,7 @@ def comment_on_violation(violation)
   end
 
   def repository_owner
```
salbertson: `repository_owner_name` or `repository_owner_username` to make it clear a name is returned and not an owner object or something?

app/jobs/buildable.rb
srt32: I believe we do for 1) legacy reasons (we'll get backfilled data for free) and 2) we'll keep the active `owners` up to date this way (less stale data). Do you think the updates are unnecessary?

app/jobs/buildable.rb
```diff
@@ -3,6 +3,11 @@
 module Buildable
   def perform(payload_data)
     payload = Payload.new(payload_data)
+    Owner.upsert(
+      github_id: payload.repository_owner_id,
+      github_login: payload.repository_owner_name,
+      organization: payload.repository_owner_is_organization?
```
gylaz: `is_` usually comes first, in ruby `is_` is bypassed in favor of `?`. I'm not sure about `repository_owner_organization?`.

app/services/repo_synchronization.rb
```diff
@@ -19,12 +13,25 @@ def start
...
+    @api ||= GithubApi.new(github_token)
+  end
+
   def repo_attributes(attributes)
     {
       private: attributes[:private],
       github_id: attributes[:id],
       full_github_name: attributes[:full_name],
-      in_organization: attributes[:owner][:type] == ORGANIZATION_TYPE
+      in_organization: attributes[:owner][:type] == GithubApi::ORGANIZATION_TYPE,
```
salbertson: This comparison is done in a couple places. Can we get this from the payload object or return "GitHub repo" objects from `api.repos` that respond to `in_organization?`?

app/services/repo_synchronization.rb
```diff
@@ -19,12 +13,25 @@ def start
...
+      owner: upsert_owner(attributes[:owner]),
     }
   end
+
+  def upsert_owner(owner_attributes)
+    Owner.upsert(
+      github_id: owner_attributes[:id],
+      github_login: owner_attributes[:login],
+      organization: owner_attributes[:type] == GithubApi::ORGANIZATION_TYPE
```
salbertson: Yea, it would be nice if we could wrap these hash like data structures with an object with a nice interface.

app/models/payload.rb
srt32: Running `rspec spec/models/payload_spec.rb` without the `require_relative` causes `  uninitialized constant GithubApi`. Is it worth keeping the requires so we can run individual testS?

app/services/repo_synchronization.rb
srt32: `in_organization` on the `repo` will go away once we have all the data backfilled on `owners` so I'd rather not invest in that change now. Thoughts?

app/services/repo_synchronization.rb
srt32: Do you mean something like `OwnerParameters` or something like it?

app/jobs/buildable.rb
srt32: I see what you mean. I enjoy the symmetry of these methods starting with `repositoy_owner_` but do find `repository_owner_organization?` funny to read. Maybe keep it as is?

app/models/payload.rb
gylaz: Or maybe `repository_owner_login` to follow GitHub's naming?

spec/services/repo_synchronization_spec.rb
```diff
@@ -94,29 +70,86 @@
...
+        synchronization = RepoSynchronization.new(second_user, "githubtoken")
 
         synchronization.start
 
         expect(second_user.reload.repos.size).to eq(1)
       end
     end
+
+    describe "repo owners" do
+      context "when the owner doesn't exit" do
```
salbertson: "exist"?

app/services/repo_synchronization.rb
```diff
@@ -19,12 +13,25 @@ def start
...
+  end
+
   def repo_attributes(attributes)
     {
       private: attributes[:private],
       github_id: attributes[:id],
       full_github_name: attributes[:full_name],
-      in_organization: attributes[:owner][:type] == ORGANIZATION_TYPE
+      in_organization: attributes[:owner][:type] == GithubApi::ORGANIZATION_TYPE,
+      owner: upsert_owner(attributes[:owner]),
```
gylaz: Feels like this functionality is kind of hidden in the hash here. What do you think about calling it before the hash construction?

salbertson:
@srt32, couple of comments/suggestions, looks good to me. :+1:

spec/factories.rb
```diff
@@ -29,7 +33,7 @@
   end
 
   factory :user do
-    sequence(:github_username) { |n| "github#{n}" }
+    github_username { generate(:github_name) }
```
gylaz: Should we just use `github_login` generator here and kill the `github_name` one?

spec/jobs/buildable_spec.rb
```diff
@@ -1,7 +1,8 @@
-require 'fast_spec_helper'
-require 'app/jobs/buildable'
-require 'app/models/payload'
-require 'app/services/build_runner'
+require "spec_helper"
```
gylaz: Why do we need `spec_helper` in this test?

spec/jobs/buildable_spec.rb
```diff
@@ -1,7 +1,8 @@
...
+require "app/jobs/buildable"
+require "app/models/owner"
+require "app/models/payload"
+require "app/services/build_runner"
```
gylaz: If we truly need `spec_helper` then these should be already loaded.

spec/factories.rb
srt32: hmmm, good question. `github_username` should be a subset of `github_login` so I think we should make this change.

lib/github_api.rb
```diff
@@ -3,8 +3,10 @@
 require "active_support/core_ext/object/with_options"
 
 class GithubApi
-  SERVICES_TEAM_NAME = "Services"
+  ORGANIZATION_TYPE = "Organization"
   PREVIEW_MEDIA_TYPE = "application/vnd.github.moondragon-preview+json"
+  SERVICES_TEAM_NAME = "Services"
```
gylaz: Where is this used?

lib/github_api.rb
```diff
@@ -3,8 +3,10 @@
...
+  USER_TYPE = "User"
```
gylaz: Do we need this? If it's only used in test, how about just hard-coding it there?

spec/models/payload_spec.rb
```diff
@@ -73,7 +73,57 @@
...
+  end
+
+  describe "#repository_owner_is_organization?" do
+    context "when the repository owner is an user" do
+      it "returns false" do
+        payload_json = {
+          "repository" => {
+            "owner" => {
+              "id" => 1,
+              "type" => GithubApi::USER_TYPE
```
gylaz: Like, this can just be `"Not organization"`.

app/models/payload.rb
srt32: Nice, done in 0c74b51

spec/services/repo_synchronization_spec.rb
```diff
@@ -94,29 +70,86 @@
...
+            owner_id: owner_github_id,
+            owner_login: owner_login
+          )
+          synchronization = RepoSynchronization.new(user, "githubtoken")
+
+          synchronization.start
+
+          owner = Owner.find_by(github_id: owner_github_id)
+          expect(owner.github_login).to eq(owner_login)
+          expect(owner.repos.map(&:github_id)).to include(repo_github_id)
```
gylaz: Is it safe to say `expect(owner.repos.map(&:github_id)).to eq([repo_github_id])`?

spec/services/repo_synchronization_spec.rb
```diff
@@ -94,29 +70,86 @@
...
+            repo_id: repo_github_id,
+            owner_id: owner.github_id,
+            owner_login: owner.github_login
+          )
+          synchronization = RepoSynchronization.new(user, "githubtoken")
+
+          synchronization.start
+
+          owner = Owner.find_by(github_id: owner.github_id)
+          expect(owner.repos.map(&:github_id)).to include(repo_github_id)
```
gylaz: Same here.

spec/services/repo_synchronization_spec.rb
```diff
@@ -94,29 +70,86 @@
...
+        id: repo_id,
+        private: private_repo,
+        owner: {
+          id: owner_id,
+          login: owner_login,
+          type: "Organization",
+        }
+      }
+      resource = double(:resource, to_hash: attributes)
+      api = double(:github_api, repos: [resource])
```
gylaz: `:github_api` -> `"GithubApi"`

app/models/payload.rb
gylaz: The `require` should be in the test.

app/models/pull_request.rb
srt32: Good idea, done in 94c22f6. This change touched a lot of files...

lib/github_api.rb
srt32: I was just moving things around here. Looks like it's in `add_hound_to_repo` and 'github_api` (https://github.com/thoughtbot/hound/search?utf8=%E2%9C%93&q=SERVICES_TEAM_NAME). Perhaps keep it in `GithubApi`?

lib/github_api.rb
srt32: Nice, done in c50f2fb

app/services/repo_synchronization.rb
srt32: Cool, done in c50f2fb

app/models/payload.rb
srt32: Superb, done in c50f2fb

app/services/repo_synchronization.rb
gylaz: What if passed the `owner_attributes` into `upsert` and had it slice the right parameters out? Just a thought.

spec/jobs/buildable_spec.rb
srt32: `Owner` inherits from AR::Base so we need `spec_helper`. Removed the extra `requires` in 63d3952

spec/services/repo_synchronization_spec.rb
srt32: Done in 22db10d

spec/services/repo_synchronization_spec.rb
srt32: Nice, yup! 22db10d

app/services/repo_synchronization.rb
srt32: Thoughts on a potential fast follow refactor if we want to go that route? This PR is becoming large and long enough on its own ;)

db/schema.rb
salbertson: `github_name`? Can be used in context of org or user.

app/services/repo_synchronization.rb
salbertson: Sure, makes sense.

app/services/repo_synchronization.rb
salbertson: Makes sense.

spec/factories.rb
salbertson: Haha, naming is hard.

db/schema.rb
srt32: Agreed. Done in bb46f64 and e1828de

srt32:
I squashed a bunch of intermediate commits to clean things up a bit but left the most recent commits there to make seeing the name changes easier. Another review would be appreciated cc/ @thorncp @croaky @gylaz @salbertson

db/schema.rb
```diff
@@ -15,7 +15,6 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "pg_stat_statements"
```
srt32: Hmmm, should this line have been removed?

spec/jobs/buildable_spec.rb
salbertson: @srt32, can we use `fast_spec_helper` here or is there something that's easier to test using `spec_helper`?

spec/models/payload_spec.rb
```diff
@@ -73,7 +74,57 @@
...
+      }
+
+      payload = Payload.new(data)
+
+      expect(payload.repository_owner_id).to eq 1
+    end
+  end
+
+  describe "#repository_owner_is_organization?" do
+    context "when the repository owner is an user" do
```
salbertson: "a user"?

db/schema.rb
salbertson: I'm not sure, probably not.

salbertson:
@srt32, looks good to me! :+1:

db/schema.rb
croaky: I think you can bring it back with:

```
rails dbconsole
CREATE EXTENSION pg_stat_statements;
\q
rake db:migrate
```

teoljungberg:
Looks good to me aswell, this makes the refactoring to use the user's api token more easier. :+1:

srt32:
Squashed and rebased in 67f6fa9

app/models/repo.rb
```diff
@@ -1,9 +1,9 @@
 class Repo < ActiveRecord::Base
-  has_many :memberships
-  has_many :users, through: :memberships
   has_many :builds
-
+  has_many :memberships
+  belongs_to :owner
   has_one :subscription
+  has_many :users, through: :memberships
```
gylaz: `has_many`s Should be together?

db/migrate/20141211224528_create_owners.rb
gylaz: How about just `name`? I've actually aliased Repo's `full_github_name` to `name`, and always wonder, why we didn't just go with `name`.

app/models/repo.rb
srt32: From https://github.com/thoughtbot/guides/tree/master/style/rails#rails

> Order ActiveRecord associations alphabetically by attribute name.

Am I misunderstanding?

db/migrate/20141211224528_create_owners.rb
srt32: I enjoy the symmetry with `github_id` but am fine with whatever.

spec/jobs/buildable_spec.rb
gylaz: I'm also curious why we need Rails for this spec.

spec/models/owner_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        owner = create(:owner)
+        new_github_name = "ralphbot"
+
+        updated_owner = Owner.upsert(
+          github_id: owner.github_id,
+          github_name: new_github_name,
+          organization: true
+        )
+
+        expect(updated_owner.github_name).to eq new_github_name
```
gylaz: Do we need to call `updated_owner.reload` before the verification? What happens if `upsert` fails?

spec/models/payload_spec.rb
```diff
@@ -1,6 +1,7 @@
-require 'fast_spec_helper'
+require "fast_spec_helper"
 require "attr_extras"
-require 'app/models/payload'
+require "app/models/payload"
+require "lib/github_api"
```
gylaz: This is a heavy requirement just for `GithubApi::ORGANIZATION_TYPE`. I'd favor just using a string.

app/models/repo.rb
gylaz: When in doubt, change the style guide! :wink: https://github.com/thoughtbot/guides/pull/304

Let's keep it as it is for now.

srt32: you sly dog you.

spec/jobs/buildable_spec.rb
srt32: This spec is dependent on `Owner`, which inherits from `AR::Base`, so running the spec on its own breaks without Rails. We could create a `faksOwner` to get around this issue but it feels a bit heavy handed. Are there other stubbing techniques you'd recommend?

spec/models/owner_spec.rb
srt32: Adding the `reload` has no impact. `Owner.upsert` calls `save` under the hood and then returns the `owner`, so whatever calls this code could handle errors however it likes. Does that make sense?

spec/models/payload_spec.rb
srt32: Cool, updated in 2a293ec. Did you also mean to use the string within `Payload`?

spec/models/owner_spec.rb
gylaz: So, `save` might fail silently. If `save` fails, should the `github_name` on the returned object still be set? I'm guessing so. Thus, `reload` would ensure it actually got updated in the db correctly. Maybe I'm overthinking this?

srt32: We changed `Owner.upsert` to use `save!` in 155085a. It makes sense to raise on failures because these cases should truly be exceptional (we are trying to insert a duplicate record where we should really be updating the record). 

@gylaz does this approach make sense to you?

db/migrate/20141211224528_create_owners.rb
srt32: We went with `name` in f89eccd

app/models/style_guide/scss.rb
```diff
@@ -53,7 +53,10 @@ def default_options
     end
 
     def default_config_file
-      DefaultConfigFile.new(DEFAULT_CONFIG_FILENAME, repository_owner).path
+      DefaultConfigFile.new(
+        DEFAULT_CONFIG_FILENAME,
+        repository_owner_name
+      ).path
```
salbertson: Should this method be moved to `Base`?

spec/factories.rb
```diff
@@ -53,4 +56,10 @@
     line_number 42
     messages ["Trailing whitespace detected."]
   end
+
+  factory :owner do
+    github_id
+    name { generate(:github_name) }
+    organization false
```
salbertson: What do you think about removing this since the default is false?

spec/jobs/buildable_spec.rb
salbertson: Fine with me.

salbertson:
@srt32, I really like the way this turned out, thanks for sticking with it. I have a couple questions, other than that is looks great. :+1:

app/models/payload.rb
```diff
@@ -1,4 +1,5 @@
-require 'json'
+require "json"
+require "lib/github_api"
```
gylaz: Both of these get auto-loaded by Rails. For tests, we should move these to tests :wink:

app/models/repo.rb
gylaz: I'm betting my option will win out!

app/models/style_guide/scss.rb
gylaz: :+1:

spec/models/style_guide/coffee_script_spec.rb
```diff
@@ -93,7 +93,10 @@ class FooBar
         spy_on_file_read
         config_file = default_configuration_file(StyleGuide::CoffeeScript)
 
-        violations_in("var foo = 'bar'", repository_owner: "not_thoughtbot")
+        violations_in(
+          "var foo = 'bar'",
+          repository_owner_name: "not_thoughtbot"
```
gylaz: Shorten `not_thougthbot` to `foo`, and one line it?

gylaz: The spec description/context should be enough maybe.

gylaz:
Blam! :+1: 

P. S. I really want to see the comment count of this PR to go to 200

spec/factories.rb
srt32: Great! Done in a619894

app/models/payload.rb
srt32: Excellent. Done in a619894

spec/models/style_guide/coffee_script_spec.rb
srt32: Sounds good. a619894

app/models/style_guide/scss.rb
srt32: @gylaz @salbertson thoughts on the implementation in 0d2cca3? Is there a better way to handle the subclasses' constants?

srt32: In the name of getting this PR done, I'm going to :boot: on this one. I'll make a card to track this refactoring separately.

PR #592, opened 2015-01-28
Fix bad params in configuration email

croaky:
https://twitter.com/just3ws/status/560526236045565952


graysonwright:
It looks like we lose the UTM tracking data. Not sure if that's worth holding this up, though

graysonwright:
the code looks good, merge if you're okay with ^

croaky:
@graysonwright Good call, pushed a fix to maintain them.

PR #593, opened 2015-01-29
Remove code duplication from Report

salbertson:


gylaz:
Looks good. What do you think about limiting the amount of weeks to the last 12 or so? That report is getting pretty lengthy. Probably worth of a separate PR though.

app/models/report.rb
```diff
@@ -1,80 +1,72 @@
...
-        generate_output(weekly_activity_sql, week)
-      end
+    generate_report("North Star") do |week|
+      <<-SQL
+        SELECT COUNT(distinct repos.id)
+        FROM repos
+        JOIN builds ON builds.repo_id = repos.id
+        WHERE builds.created_at >= '#{week}'
+        AND builds.created_at < '#{week + 7.days}'
+      SQL
```
croaky: Could probably delete this report. It's a dup of https://tbot.io/ns/hound

gylaz: How does that spreadsheet pull our data?

croaky: magic

salbertson: ha, we can remove it in another PR, this is just to remove duplication.

salbertson:
Thanks dudes.

PR #594, opened 2015-01-29
Don't prompt user for credit card if one has already been set up

salbertson:
This change skips Stripe Checkout when a card has already been entered. I also added some view logic to not show the "Update Credit Card" link if no Stripe customer exists.

### DIFF
--- a/app/assets/javascripts/directives/repo.js.coffee.erb
+++ b/app/assets/javascripts/directives/repo.js.coffee.erb
@@ -1,4 +1,4 @@
-App.directive 'repo', ['Subscription', 'StripeCheckout', (Subscription, StripeCheckout) ->
+App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription, StripeCheckout, User) ->
   scope: true
   templateUrl: '/templates/repo'
   link: (scope, element, attributes) ->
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/directives/repo.js.coffee.erb
```diff
@@ -16,15 +16,34 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', (Subscription, StripeCh
         .catch(-> alert('Your repo failed to deactivate.'))
         .finally(-> scope.processing = false)
 
-    createSubscription = (stripeToken) ->
+    createSubscription = (user) ->
+      if scope.repo.price_in_cents > 0
+        if user.stripe_customer_id
+          createSubscriptionWithoutCard()
```
gylaz: This naming implies that we are creating a subscription without card, which seems incorrect. How about something like `createSubscriptionWithExistingCard`?

app/assets/javascripts/directives/repo.js.coffee.erb
```diff
@@ -16,15 +16,34 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', (Subscription, StripeCh
...
+        else
+          StripeCheckout.open(
+            name: scope.repo.full_plan_name,
+            amount: scope.repo.price_in_cents,
+            createSubscriptionWithCard
+          )
```
gylaz: Thoughts on moving this to a function like `promptForCardInfo`?

app/serializers/user_serializer.rb
```diff
@@ -1,3 +1,3 @@
 class UserSerializer < ActiveModel::Serializer
-  attributes :id, :github_username, :refreshing_repos
+  attributes :id, :github_username, :refreshing_repos, :stripe_customer_id
```
gylaz: Are there any security issues with exposing `stripe_customer_id`?

app/services/repo_subscriber.rb
```diff
@@ -11,8 +11,7 @@ def self.unsubscribe(repo, user)
 
   def subscribe
     customer = if user.stripe_customer_id
-      payment_gateway_customer.card = card_token
-      payment_gateway_customer.save
+      payment_gateway_customer
```
gylaz: One way to get around this is to extract the `if/else` to a method.

app/assets/javascripts/directives/repo.js.coffee.erb
salbertson: I like that better, done.

app/assets/javascripts/directives/repo.js.coffee.erb
salbertson: Sounds good, done.

app/serializers/user_serializer.rb
salbertson: Probably not, but it's not worth it. Good call. I updated the serializer to not return the ID.

salbertson:
@frechg, see any issues with skipping the CC prompt after the user has entered their card info once? I think it's clear enough that they will be subscribing, and paying, when the repo is private.

salbertson:
@gylaz, thanks for the review!

PR #595, opened 2015-01-30
Remove Stripe event listener

salbertson:
The webhook was never set up in Stripe. I think we can handle failed charges by hand until it becomes an issue.


croaky:
Red everywhere. As far as the eye can see.

croaky:
Looks good to me.

cpytel:
I recommend using churnbuster to help with this. We use it on Upcase and FormKeep.

gylaz:
What is the process of handling this by hand? Would be helpful to document this somewhere.

PR #596, opened 2015-01-30
Add index for user_id to memberships

salbertson:
This should make queries in `Analytics` faster.

screenshot 2015-01-29 21 23 24

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,10 +11,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150108003045) do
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -37,6 +38,7 @@
   end
 
   add_index "memberships", ["repo_id", "user_id"], name: "index_memberships_on_repo_id_and_user_id", using: :btree
```
croaky: How about:

```ruby
remove_index :memberships, [:repo_id, :user_id]
add_index :memberships, :repo_id
add_index :memberships, :user_id
```

salbertson: Good idea, done.

PR #600, opened 2015-01-30
Hide credit card link if Stripe customer hasn't been set up

salbertson:
This avoids a potential issue where a users without subscriptions, or Stripe Customer set up, updates their card info for some reason.

I think it happened once already.

https://twitter.com/IanWalter/status/561265511133351936

### DIFF
--- a/app/views/accounts/show.haml
+++ b/app/views/accounts/show.haml
@@ -7,9 +7,10 @@
   %h2.label-text
     Your monthly cost
 
-    %span.section-action
-      %a{ 'ng-controller' => 'accountController', 'ng-click' => 'showStripeForm()' }
-        Update Credit Card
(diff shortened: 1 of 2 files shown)


app/views/accounts/show.haml
```diff
@@ -7,9 +7,10 @@
   %h2.label-text
     Your monthly cost
 
-    %span.section-action
-      %a{ 'ng-controller' => 'accountController', 'ng-click' => 'showStripeForm()' }
-        Update Credit Card
+    - if current_user.stripe_customer_id
```
srt32: Perhaps `if current_user.stripe_customer_id.present?`

salbertson: Good idea.

salbertson: Done.

PR #602, opened 2015-02-04
Only enqueue org invitation job when repo is in Org

arunagw:


salbertson:
@arunagw, how did this come up? Was there an exception or did you just run across the code and realize there was a problem?

arunagw:
There is no exception for this. I was reading code and before that I saw a card for rate limiting so I thought it might reduce a bit.

Also there were no spec for this call so wrote that as well.

spec/services/repo_activator_spec.rb
```diff
@@ -2,9 +2,35 @@
...
+        repo = create(:repo, in_organization: true)
+        stub_github_api
+        activator = build_activator(repo: repo)
+
+        activator.activate
+
+        expect(JobQueue).to have_received(:push).with(OrgInvitationJob)
+      end
+
+      it "will not enqueue org invitation job if repo is not a part of Org" do
```
salbertson: What do you think about creating a context for "when repo is not part of an organization"?

spec/services/repo_activator_spec.rb
```diff
@@ -2,9 +2,35 @@
 
 describe RepoActivator do
   describe '#activate' do
+
+    context "Org invitation" do
```
salbertson: What do you think about "when repo is part of an organization" or "with org repo"?

salbertson:
Looks good to me. :+1:

PR #603, opened 2015-02-05
Deactivation cleanup

arunagw:
Remove hound from repo and org when user deactivate repo.


salbertson:
@arunagw, what happens when Hound is enabled for multiple repos in an organization? Should Hound only be removed from an organization if no other repos are being reviewed?

app/services/manage_hound.rb
```diff
@@ -0,0 +1,33 @@
...
+  SERVICES_TEAM_NAME = "Services"
+
+  pattr_initialize :repo_name, :github
+
+  def self.run(repo_name, github)
+    new(repo_name, github).run
+  end
+
+  def run
+    raise "Must be implemented in child class"
```
gylaz: What do you think about doing the same as in `style_guide/base.rb` class?
```ruby
raise NotImplementedError.new("must implement ##{__method__}")
```

app/services/repo_activator.rb
```diff
@@ -24,6 +22,12 @@ def activate_repo
     end
   end
 
+  def deactivate_repo
+    change_repository_state_quietly do
+      remove_hound_from_repo && delete_webhook && repo.deactivate
+    end
```
gylaz: Why not just have this block in `deactivate`?

app/services/remove_hound_from_repo.rb
```diff
@@ -0,0 +1,37 @@
...
+      team_id = team.id
+      remove_repo_from_team(team_id)
+      delete_team(team_id)
+    end
+  end
+
+  def delete_team(team_id)
+    team_repos = github.team_repos(team_id)
+
+    if team_repos.count == 0
```
gylaz: `team_repos.empty?`

lib/github_api.rb
```diff
@@ -150,12 +165,14 @@ def user_repos
     authorized_repos(client.repos)
   end
 
-  def org_repos
-    repos = orgs.flat_map do |org|
-      client.org_repos(org[:login])
+  def all_org_repos
```
gylaz: What do you think about naming this `repos_from_all_orgs`? A bit more descriptive, as `all_org_repos` reads as in "all repos from org".

gylaz:
Thoughts on not removing the "Services" team? What is the benefit of removing it?

arunagw:
@gylaz @salbertson please have a look again. 

cc/ @samphippen

spec/lib/github_api_spec.rb
```diff
@@ -328,4 +321,73 @@
...
+        team_id, username, token
+      )
+
+      api.remove_user_from_team(team_id, username)
+
+      expect(request).to have_been_requested
+    end
+  end
+
+  private
```
teoljungberg: Our styleguide says that we should avoid having `private` in specs

app/services/manage_hound.rb
```diff
@@ -0,0 +1,33 @@
...
+  SERVICES_TEAM_NAME = "Services"
+
+  pattr_initialize :repo_name, :github
+
+  def self.run(repo_name, github)
+    new(repo_name, github).run
+  end
+
+  def run
+    raise NotImplementedError.new("must implement ##{__method__}")
```
teoljungberg: How about writing
```ruby
raise NotImplementedError, "must implement ##{__method__}"
```
Instead of instantiating the error class with the message?

app/services/remove_hound_from_repo.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def remove_hound_from_repo
+    github.remove_collaborator(repo_name, github_username)
+  end
+
+  def remove_services_team_from_repo
+    team = find_services_team
+
+    if team
+      team_id = team.id
```
teoljungberg: How about inlining this local variable instead?

arunagw:
This PR is in WIP again. Will ping back once ready to review

arunagw:
@gylaz I think we should not remove team "Services" as user or any other services might have added something to it and we don't want to mess with that.

Remove repo and user from that team seems fine.

app/models/github_team.rb
```diff
@@ -0,0 +1,63 @@
+class GithubTeam
+  def initialize(team, github)
```
teoljungberg: We have `attr_extras` in the project, how about using `pattr_initialize here?

app/services/manage_hound.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  private
+
+  def repo
+    @repo ||= github.repo(repo_name)
+  end
+
+  def github_username
+    @github_username ||= ENV.fetch("HOUND_GITHUB_USERNAME")
```
teoljungberg: Maybe this does not need to be memoized?

app/services/manage_hound.rb
```diff
@@ -0,0 +1,43 @@
+require "app/models/github_team"
+
+class ManageHound
+  SERVICES_TEAM_NAME = "Services"
```
teoljungberg: Maybe this can be renamed from `GITHUB_TEAM_NAME`?

teoljungberg:
Few comments, and fixed a few of them. But besides that, LGTM

spec/lib/github_api_spec.rb
gylaz: It does? Mind pasting the rule, or linking to it?

teoljungberg: https://github.com/thoughtbot/guides/tree/master/style/testing, First
bullet-point

app/services/repo_activator.rb
arunagw: I think it's good to keep the same as `activate` if you look into this file it's now same as we are doing in `activate`

app/models/github_team.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def has_pull_permission?
+    team.permission == "pull"
+  end
+
+  def remove_user_from_team(github_username)
+    team_repos = github.team_repos(team.id)
+
+    if team_repos.empty?
```
arunagw: Should we have this check here or in `RemoveHoundFromRepo` class?

gylaz: Seems good here.

arunagw:
This one looks good to merge? 

What we are doing here is in case of org repo.
1. Removing repo from "Services" team.
2. Removing Hound User from team if no more repo in "Services" team

app/models/github_team.rb
```diff
@@ -0,0 +1,52 @@
+class GithubTeam
+  pattr_initialize :team, :github
+
+  def add_repo(repo_name)
+    ensure_push_permission
+    add_repo_to_team(repo_name)
+  end
+
+  def remove_repo(repo_name)
+    remove_repo_from_team(repo_name)
```
gylaz: I'm curious about this coding approach -- `remove_repo_from_team` contains just one line. Why abstract it to yet another method?

app/models/github_team.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def remove_user(github_username)
+    remove_user_from_team(github_username)
+  end
+
+  def add_user(github_username)
+    add_user_to_team(github_username)
```
gylaz: Same question here. This approach generates more code, without an apparent benefit. `github.add_repo_to_team(team.id, repo_name)` is pretty readable to me. Am I missing something?

app/services/manage_hound.rb
```diff
@@ -0,0 +1,43 @@
+require "app/models/github_team"
```
gylaz: Why is this necessary?

app/services/manage_hound.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def github_username
+    ENV.fetch("HOUND_GITHUB_USERNAME")
+  end
+
+  def decorated_services_team
+    if find_services_team
+      GithubTeam.new(find_services_team, github)
+    else
+      nil
```
gylaz: `else` isn't necessary here, technically.

app/services/remove_hound_from_repo.rb
```diff
@@ -0,0 +1,24 @@
+require "app/services/manage_hound"
```
gylaz: Why is this necessary?

gylaz: If to make the tests happy, then this `require` should be in the test, no?

app/services/remove_hound_from_repo.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+  end
+
+  private
+
+  def remove_hound_from_repo
+    github.remove_collaborator(repo_name, github_username)
+  end
+
+  def remove_services_team_from_repo
```
gylaz: Should this be `remove_repo_from_services_team`?

app/models/github_team.rb
gylaz: Looking at this a bit more, I'm torn. On one hand we want to use the "Tell don't ask" principal, on the other hand, reading [this line of code](https://github.com/thoughtbot/hound/pull/603/files#diff-153e44c17166a5b3c6f96a2a1988860eR21) it would be surprising to me if that didn't remove the user from the team.

spec/features/repo_list_spec.rb
```diff
@@ -150,4 +155,8 @@ def with_job_delay
   ensure
     Resque.inline = true
   end
+
+  def username
+    "houndci"
+  end
```
gylaz: Should this be a `let`?

spec/lib/github_api_spec.rb
```diff
@@ -114,7 +108,7 @@
     it "returns changed files in a pull request" do
       hound_token = ENV["HOUND_GITHUB_TOKEN"]
       api = GithubApi.new(hound_token)
-      pull_request = double(:pull_request, full_repo_name: "thoughtbot/hound")
+      pull_request = double(:pull_request, full_repo_name: full_repo_name)
```
gylaz: Please use the string version of double names. `double("PullRequest"...`

spec/lib/github_api_spec.rb
```diff
@@ -133,15 +127,14 @@
...
     it "adds comment to GitHub pull request" do
       api = GithubApi.new("authtoken")
-      repo_name = "test/repo"
       pull_request_number = 2
       comment = "test comment"
       commit_sha = "commitsha"
       file = "test.rb"
       patch_position = 123
-      commit = double(:commit, repo_name: repo_name, sha: commit_sha)
+      commit = double(:commit, repo_name: full_repo_name, sha: commit_sha)
```
gylaz: `double("Commit"...`

spec/lib/github_api_spec.rb
gylaz: Since we are touching this line, can we make that update as well? It's a nice way to prep ourselves for RSpec 3.1 upgrade.

spec/models/github_team_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  describe "#add_user" do
+    it "calls github with the team id and passed username" do
+      id = 1234
+      team = double("OctoKitTeam", id: id, permission: non_pull_permission)
+      github = double(
+        "Github",
+        add_user_to_team: true
+      )
```
gylaz: Seems like this could fit on one line.

spec/models/github_team_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    "pull"
+  end
+
+  def repo_name
+    "foo/bar"
+  end
+
+  def username
+    "houndci"
+  end
```
gylaz: These all look like good candidates for `let`s.

gylaz:
Left a few comments/suggestions. Looks good to me. Merge at your discretion.

spec/models/github_team_spec.rb
arunagw: It seems that we don't want to use `let` here  https://github.com/thoughtbot/guides/tree/master/best-practices#testing

I have seen that we have used `let` in `builds_spec`. 

What do you think? Should we add?

gylaz: We don't want to use `let`, because it has higher-level code issues. We shouldn't just re-implement `let`s because of that rule.

PR #604, opened 2015-02-05
Check if Octokit exception contains errors

gylaz:
It appears that sometimes there are no errors associated with an Octokit 
exception as show by:

https://app.getsentry.com/hound-1/production/group/53642463/

This fix checks for presense of an error before fetching `:code` from it.

### DIFF
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -12,7 +12,7 @@ class Commit
   rescue Octokit::NotFound
     ""
   rescue Octokit::Forbidden => exception
-    if exception.errors.first[:code] == "too_large"
+    if exception.errors.any? && exception.errors.first[:code] == "too_large"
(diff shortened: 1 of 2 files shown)


app/models/commit.rb
```diff
@@ -12,7 +12,7 @@ def file_content(filename)
   rescue Octokit::NotFound
     ""
   rescue Octokit::Forbidden => exception
-    if exception.errors.first[:code] == "too_large"
+    if exception.errors.any? && exception.errors.first[:code] == "too_large"
```
srt32: Worth pulling `"too_large"` into a method or constant?

salbertson:
:+1:

salbertson:
@gylaz, this issue came up again.

PR #607, opened 2015-02-07
Track "Build Started" and "Build Completed" events

croaky:
* "Reviewed Repo" is not reporting correctly to Segment/Mixpanel.
* We already have build totals in the North Star Metric.
* Since "Reviewed Repo" is not useful to us, right now,
 we can try a more constrained event tracking approach.
* Track only private (subscribed) repos.
* Track both a "Build Started" and a "Build Completed" event
 to see if there is some code in between that is preventing
 the "Reviewed Repo" (now called "Build Completed") event
 from triggering.

### DIFF
--- a/app/models/analytics.rb
+++ b/app/models/analytics.rb
@@ -10,49 +10,44 @@ class Analytics
     track(event: "Signed In")
   end
 
-  def track_activated(repo)
+  def track_repo_activated(repo)
     track(
-      event: "Activated Public Repo",
-      properties: {
-        name: repo.full_github_name
-      }
-    )
-  end
-
-  def track_deactivated(repo)
-    track(
-      event: "Deactivated Public Repo",
-      properties: {
-        name: repo.full_github_name
-      }
-    )
-  end
-
-  def track_reviewed(repo)
-    track(
-      event: "Reviewed Repo",
-      properties: {
-        name: repo.full_github_name
-      }
-    )
-  end
-
-  def track_subscribed(repo)
-    track(
-      event: "Subscribed Private Repo",
+      event: "Repo Activated",
       properties: {
         name: repo.full_github_name,
-        revenue: repo.plan_price
+        private: repo.private,
+        revenue: repo.plan_price,
       }
     )
   end
 
-  def track_unsubscribed(repo)
+  def track_repo_deactivated(repo)
     track(
-      event: "Unsubscribed Private Repo",
+      event: "Repo Deactivated",
       properties: {
         name: repo.full_github_name,
-        revenue: -repo.plan_price
+        private: repo.private,
+        revenue: -repo.plan_price,
+      }
+    )
+  end
+
+  def track_build_started(repo)
+    track(
+      event: "Build Started",
+      properties: {
+        name: repo.full_github_name,
+        private: repo.private,
+      }
+    )
+  end
+
+  def track_build_completed(repo)
+    track(
+      event: "Build Completed",
+      properties: {
+        name: repo.full_github_name,
+        private: repo.private,
       }
     )
   end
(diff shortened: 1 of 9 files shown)


app/models/analytics.rb
```diff
@@ -28,21 +28,30 @@ def track_deactivated(repo)
...
   end
 
-  def track_reviewed(repo)
+  def track_subscribed(repo)
     track(
-      event: "Reviewed Repo",
+      event: "Subscribed Private Repo",
+      properties: {
+        name: repo.full_github_name,
+        revenue: repo.plan_price
```
salbertson: Is this a `RepoSerializer` object?

app/services/build_runner.rb
```diff
@@ -46,10 +47,19 @@ def repo
...
-      analytics.track_reviewed(repo)
+      analytics.track_build_started(repo)
+    end
+  end
+
+  def track_build_completed
+    if repo.subscription
+      user = repo.subscription.user
+      analytics = Analytics.new(user)
+      analytics.track_build_completed(repo)
```
salbertson: Could we track all builds and pass some data along that would let us filter by subscribed repo builds and free repo builds in Segment?

salbertson: If not should we name these methods to indicate that we are only tracking builds for subscribed repos?

croaky: I would like to track all builds but we can't figure out why the "Reviewed
Repo" event wasn't firing.

salbertson:
Couple of questions, looks good to me. :+1:

app/services/build_runner.rb
croaky: What if we sent `private` as a property? 9065f3a

gylaz: I like the idea of just passing `private`.

app/services/build_runner.rb
```diff
@@ -11,7 +12,7 @@ def run
       )
       commenter.comment_on_violations(priority_violations)
       create_success_status
-      track_reviewed_repo_for_each_user
+      track_build_completed
```
salbertson: What do you think about naming this method so it's clear that only builds for repos with subscriptions get tracked? Or, the conditional could be moved out of the method, not great though.

spec/controllers/activations_controller_spec.rb
```diff
@@ -17,9 +17,13 @@
       expect(activator).to have_received(:activate)
       expect(RepoActivator).to have_received(:new).
         with(repo: repo, github_token: token)
-      expect(analytics).to have_tracked("Activated Public Repo").
+      expect(analytics).to have_tracked("Repo Activated").
         for_user(membership.user).
-        with(properties: { name: repo.full_github_name })
+        with(properties: {
+          name: repo.full_github_name,
```
salbertson: Do these not fit on a single line?

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -25,9 +25,13 @@
         with(repo: repo, github_token: token)
       expect(RepoSubscriber).to have_received(:subscribe).
         with(repo, membership.user, "cardtoken")
-      expect(analytics).to have_tracked("Subscribed Private Repo").
+      expect(analytics).to have_tracked("Repo Activated").
```
salbertson: These expectations are piling up. Think it's worth breaking up the cases into separate `it` blocks? We could pull the shared setup out into a method.

salbertson:
We should probably fix the indentation Hound is complaining about. Looks good to me. :+1:

spec/controllers/subscriptions_controller_spec.rb
croaky: I just attempted to breaking this into separate `it` blocks and it felt more confusing to read with more indirection and mystery guests and lines of code. I'd like to leave it as-is.

PR #608, opened 2015-02-07
Track "Clicked Sign In" event

croaky:
Find out how many people click, but then drop out of the funnel because they
think the GitHub permissions are too aggressive.


salbertson:
Works for me. :+1:

PR #612, opened 2015-02-08
Don't have Foreman start Redis

salbertson:


croaky:
Looks good.

salbertson:
@croaky, updated readme too.

gylaz:
It would be great to include an explanation in the commit message of what the issue with starting Redis via foreman was.

salbertson:
@frechg ran into an issue over the weekend, which I've also run into before, where starting Hound with Foreman fails because Redis server is already running.

PR #615, opened 2015-02-09
Enable all languages by default

salbertson:
This enables all languages by default. Some users have used Hound for months without realizing they need to enable newly added languages. If the defaults are an issue we can make them less opinionated.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,5 +1,0 @@
-ruby:
-  enabled: true
-
-coffee_script:
-  enabled: true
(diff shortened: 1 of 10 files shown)


salbertson:
You're right, I got a little carried away. I'll add the ability to disable languages back in.

gylaz:
How did @houndci comment?

croaky:
It's become sentient.

No, I was signed in as @houndci because I was creating a private repo
there. I forgot I was signed in as Hound when I responded.

PR #617, opened 2015-02-10
Review all languages by default

salbertson:


salbertson:
@croaky, let's try this again.

app/views/pages/configuration.haml
```diff
@@ -37,20 +27,19 @@
...
-          enabled: true
           config_file: config/.rubocop.yml
 
-  %h3 CoffeeScript
-
   %p
-    To enable CoffeeScript style checking, add the following to
+    To disable Ruby style checking, add the following to
     %em.code .hound.yml
     in the root of your project
```
croaky: I like this move of the disable text to after the code block.

spec/models/repo_config_spec.rb
```diff
@@ -5,102 +5,62 @@
 
 describe RepoConfig do
   describe "#enabled_for?" do
-    context "with invalid format in Hound config" do
-      it "only returns true for ruby" do
+    context "with invalid config" do
+      it "returns true for all languages" do
```
croaky: "enabled for with invalid config returns true for all languages" - Is there a "why?" that can be added into the test description? We sort of ignore it, I guess?

spec/models/style_checker_spec.rb
```diff
@@ -220,63 +141,33 @@
...
-
-          violations = StyleChecker.new(pull_request).violations
-          messages = violations.flat_map(&:messages)
-
-          expect(messages).to include(
-            "Selector should have depth of applicability no greater than 2, but was 4"
-          )
-        end
-      end
+      it "returns violations" do
```
croaky: Maybe "with violating code it returns violations"? "with violations it returns violations" reads a little funny to me.

croaky:
Solid, I like it. Would be good to get a least one of @gylaz @malandrina @srt32 @thorncp to take a look as well.

We should probably time an Intercom message to go out right after the deploy?

spec/models/repo_config_spec.rb
salbertson: Yea, we treat it similar to a missing config I guess.

spec/models/style_checker_spec.rb
salbertson: I changed it to "with style violations". I think that makes sense, the test reads "#violations with for a SCSS file with style violations returns violations". What do you think?

salbertson:
@croaky, thanks for the review. I think we should definitely send out an announcement to current users and probably get https://github.com/thoughtbot/hound/pull/618 out at the same time.

app/models/repo_config.rb
```diff
@@ -42,9 +41,28 @@ def ignored_javascript_files
 
   private
 
-  def enabled_in_config?(name)
-    config = hound_config[name] || hound_config[name.camelize]
-    config && (config["enabled"] == true || config["Enabled"] == true)
+  def legacy_enabled?(style_guide_name)
+    if legacy_config?
+      option = hound_config[style_guide_name.camelize]
+      return true unless option
```
srt32: It's more verbose but could we unwind this trailing conditional?

app/models/repo_config.rb
```diff
@@ -42,9 +41,28 @@ def ignored_javascript_files
...
+
+      option_is_enabled? option
```
srt32: I find `option_is_enabled?(option)` with the parens more readable. What do you think?

app/models/repo_config.rb
```diff
@@ -42,9 +41,28 @@ def ignored_javascript_files
...
+
+  def enabled?(name)
+    option = hound_config[name] || hound_config[name.camelize]
+    return true unless option
+
+    option_is_enabled? option
+  end
+
+  def option_is_enabled?(option)
+    if option["enabled"] == false || option["Enabled"] == false
```
srt32: Is there a way to avoid checking "enabled" and "Enabled" here and then also having to check ` hound_config[name] || hound_config[name.camelize]` on 54? Is there somewhere upstream we could call `downcase`?

srt32:
I like it! Would it be worth sending a note out prior to the release so that users know these changes will be coming (before the changes arrive)?

app/models/repo_config.rb
salbertson: Good question. I reworked things a bit to combine the logic used to check `enabled` for legacy and the new config style. Let me know what you think.

app/models/repo_config.rb
```diff
@@ -42,26 +42,34 @@ def ignored_javascript_files
...
-  def enabled_in_config?(name)
-    config = hound_config[name] || hound_config[name.camelize]
-    config && (config["enabled"] == true || config["Enabled"] == true)
+  def options_for(language)
+    hound_config[language] || hound_config[language.camelize]
   end
 
-  def legacy_config?
-    (hound_config.keys & STYLE_GUIDES).empty?
+  def disabled?(language)
```
salbertson: nooooo

salbertson:
There's definitely some logic that could be moved to a class to avoid passing around `language` all over, but we can do that in another branch.

PR #618, opened 2015-02-10
Disable quote checking by default

salbertson:
Commenting on double or single quotes has surprised some users. I think it's best to disable the check by default and let each team choose which type of quotes they would like Hound to enforce.

### DIFF
--- a/config/style_guides/javascript.json
+++ b/config/style_guides/javascript.json
@@ -26,7 +26,7 @@
     "inject",
     "module"
   ],
-  "quotmark": true,
+  "quotmark": false,
   "trailing": true,
   "undef": true,
   "unused": true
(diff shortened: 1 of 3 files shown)


salbertson:
@gylaz, thoughts?

salbertson:
I was thinking we should get this in before https://github.com/thoughtbot/hound/pull/617.

srt32:
Is is right that this change will disable quote checking for all active repos
without a custom config overriding the default?

salbertson:
Yes, I don't think it's worth it.

croaky:
I think we want to do this, but not for existing repos, and only after we've moved the server-side config into the db.

srt32:
Agreed!

PR #620, opened 2015-02-10
Backfill stripe metadata

arunagw:


app/models/payment_gateway_customer.rb
```diff
@@ -43,6 +43,16 @@ def email
     def cards
       []
     end
+
+    def subscriptions
+      NoSubscription.new
+    end
```
gylaz: Why is this necessary?

app/services/update_stripe_metadata.rb
```diff
@@ -0,0 +1,20 @@
+class UpdateStripeMetadata
+  def self.run
+    Repo.where(private: true).find_each do |repo|
```
gylaz: How about just `Repo.joins(:subscriptions)`?

app/models/payment_gateway_customer.rb
arunagw: This is required in one of test where customer do not have stripe_id. 

I will see if I can get rid of this

arunagw: Should we remove this spec all together ? https://github.com/thoughtbot/hound/blob/aa-backfill-stripe-data/spec/services/update_stripe_metadata_spec.rb#L37-L51

arunagw:
@gylaz @teoljungberg and I have also added tests for `PaymentGatewayCustomer` class as those were missing.

app/models/payment_gateway_customer.rb
gylaz: Ah, I understand it now. Let's keep the spec.

app/services/update_stripe_metadata.rb
```diff
@@ -0,0 +1,20 @@
+class UpdateStripeMetadata
+  def self.run
+    Repo.joins(:subscription).find_each do |repo|
+      if subscription = repo.subscription
```
gylaz: I don't think we need this `if` anymore. The query from line 3 guarantees a subscription, right?

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
       expect(update_request).to have_been_requested
     end
   end
+
+  describe "#email" do
+    it "will return customer email" do
+      user = build_stubbed :user, stripe_customer_id: stripe_customer_id
```
gylaz: Parens around the arguments here for consistency?

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
...
+
+      customer = payment_gateway_customer.customer
+      expect(payment_gateway_customer.email).to eq customer.email
+    end
+  end
+
+  describe "#card_last4" do
+    context "when a customer has a card in stripe" do
+      it "returns last 4 from the customers card" do
+        user = build_stubbed :user, stripe_customer_id: stripe_customer_id
```
gylaz: Parens.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
       expect(update_request).to have_been_requested
     end
   end
+
+  describe "#email" do
+    it "will return customer email" do
```
gylaz: We've been following the convention of `it "returns ...`.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
...
+      user = build_stubbed :user, stripe_customer_id: stripe_customer_id
+      stub_customer_find_request
+
```
gylaz: This newline is unnecessary.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
...
+    end
+  end
+
+  describe "#card_last4" do
+    context "when a customer has a card in stripe" do
+      it "returns last 4 from the customers card" do
+        user = build_stubbed :user, stripe_customer_id: stripe_customer_id
+        stub_customer_find_request
+
+        payment_gateway_customer = PaymentGatewayCustomer.new(user)
```
gylaz: This should be part of setup. Since we're not testing the constructor, separating this with newlines isn't necessary.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
...
+
+        expect(payment_gateway_customer.card_last4).to eq "4242"
+      end
+    end
+
+    context "when a customer does not have a card in stripe" do
+      it "returns a null object" do
+        user = build_stubbed :user, stripe_customer_id: nil
+
+        payment_gateway_customer = PaymentGatewayCustomer.new(user)
```
gylaz: Same here.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
...
+
+        expect(payment_gateway_customer.card_last4).to eq ""
+      end
+    end
+  end
+
+  describe "#customer" do
+    context "when stripe_customer_id is present" do
+      it "retrieve customer data" do
+        user = build_stubbed :user, stripe_customer_id: stripe_customer_id
```
gylaz: Parens.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
       expect(update_request).to have_been_requested
     end
   end
+
+  describe "#email" do
+    it "will return customer email" do
+      user = build_stubbed :user, stripe_customer_id: stripe_customer_id
+      stub_customer_find_request
```
gylaz: Since this is just testing delegations, what do you think about just stubbing `Stripe::Customer` here?

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,74 @@
...
+      end
+    end
+
+    it "is memoized" do
+      user = build_stubbed :user, stripe_customer_id: nil
+
+      payment_gateway_customer = PaymentGatewayCustomer.new(user)
+      customer = payment_gateway_customer.customer
+
+      expect(customer).to be customer
```
gylaz: I don't understand this test.

spec/services/update_stripe_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    context "updates successfully" do
+      it "updates stripe metadata with repo id" do
+        user = create(:user, stripe_customer_id: stripe_customer_id)
+        repo = create(:repo, private: true)
+        subscription = create(
+          :subscription,
+          stripe_subscription_id: stripe_subscription_id,
+          user: user,
+          repo: repo
+        )
```
gylaz: To cut down on some of the noise in this spec consider:
```rb
subscription = create(:subscription)
```
Then pass the subscription object to `stub_subscription_meta_data_update_request`.

spec/services/update_stripe_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        UpdateStripeMetadata.run
+
+        expect(stripe_update_request).to have_been_requested
+      end
+    end
+
+    context "does not update" do
+      it "will not update if repo does not have subscription" do
+        create(:repo)
+        stripe_customer_find_request = stub_customer_find_request
```
gylaz: I'm not sure this test is very valid. `stub_customer_find_request` could still fire with a different `stripe_customer_id` (other than the global).

spec/services/update_stripe_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        )
+
+        UpdateStripeMetadata.run
+
+        expect(stripe_update_request).to have_been_requested
+      end
+    end
+
+    context "does not update" do
+      it "will not update if repo does not have subscription" do
```
gylaz: `it "does not update ...`

spec/services/update_stripe_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+          repo.id
+        )
+
+        UpdateStripeMetadata.run
+
+        expect(stripe_update_request).to have_been_requested
+      end
+    end
+
+    context "does not update" do
```
gylaz: `context` should provide the condition. In this case this should be:
```rb
context "when repo does not have a subscription" do
  it "does not make a request to Stripe" do
```

spec/services/update_stripe_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      it "will not update if repo does not have subscription" do
+        create(:repo)
+        stripe_customer_find_request = stub_customer_find_request
+
+        UpdateStripeMetadata.run
+
+        expect(stripe_customer_find_request).not_to have_been_requested
+      end
+
+      it "will not update stripe when stripe_customer_id is missing" do
```
gylaz:
```rb
context "when stripe_customer_id is missing" do
  it "does not make a request to Stripe" do
````

File: spec/services/update_stripe_metadata_spec.rb [hunk 18]
```diff
@@ -0,0 +1,54 @@
...
+        repo = create(:repo, private: true)
+        create(
+          :subscription,
+          user: user,
+          repo: repo
+        )
+
+        UpdateStripeMetadata.run
+
+        stripe_customer_find_request = stub_customer_find_request
```
gylaz: The request still could have fired, but with different params. It might be best to verify that no requests to stripe were made:
```rb
stub_request(:any,%r{^https://api\.stripe\.com.+})
```

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -89,10 +91,24 @@ def stub_subscription_delete_request
     )
   end
 
+  def stub_subscription_meta_data_update_request(repo_id)
+    stub_request(
+      :post,
+      "#{stripe_base_url}/#{stripe_customer_id}/"\
+      "subscriptions/#{stripe_subscription_id}"
```
gylaz: Is this missing an indentation?

spec/models/payment_gateway_customer_spec.rb
teoljungberg: Fixed and pushed, the test is supposed to check that the two customer's object_id are the same(hence they are memoized). I blame that it was on a friday afternoon

arunagw:
This PR is ready for final review

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,73 @@
       expect(update_request).to have_been_requested
     end
   end
+
+  describe "#email" do
+    it "returns customer email" do
+      user = build_stubbed(:user, stripe_customer_id: stripe_customer_id)
+      stub_customer_find_request
+
+      payment_gateway_customer = PaymentGatewayCustomer.new(user)
```
gylaz: Why is this separated by newlines? This doesn't seem like the execution. Typically, the execution phase is the call to the method mentioned in the `describe`. It seems that this test has a combined execution and verification phases. I think it isn't helpful to mark this line as execution.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,73 @@
...
+    end
+  end
+
+  describe "#card_last4" do
+    context "when a customer has a card in stripe" do
+      it "returns last 4 from the customers card" do
+        user = build_stubbed(:user, stripe_customer_id: stripe_customer_id)
+        stub_customer_find_request
+
+        payment_gateway_customer = PaymentGatewayCustomer.new(user)
```
gylaz: Same here.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,73 @@
...
+    end
+  end
+
+  describe "#customer" do
+    context "when stripe_customer_id is present" do
+      it "retrieve customer data" do
+        user = build_stubbed(:user, stripe_customer_id: stripe_customer_id)
+        stub_customer_find_request
+
+        payment_gateway_customer = PaymentGatewayCustomer.new(user)
```
gylaz: And here.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,73 @@
...
+
+        expect(payment_gateway_customer.customer.id).to eq stripe_customer_id
+      end
+    end
+
+    context "when stripe_customer_id is not present" do
+      it "return null object" do
+        user = build_stubbed(:user, stripe_customer_id: nil)
+
```
gylaz: Unnecessary newline.

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -14,4 +14,73 @@
...
+        expect(customer.email).to eq ""
+        expect(customer.cards).to eq []
+        expect(customer.subscriptions.retrieve).to eq nil
+      end
+    end
+
+    it "is memoized" do
+      user = build_stubbed(:user, stripe_customer_id: nil)
+
```
gylaz: Unnecessary newline.

spec/services/update_stripe_metadata_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+          create(
+            :subscription,
+            user: user,
+            repo: repo
+          )
+
+          UpdateStripeMetadata.run
+
+          stripe_customer_find_request = stub_customer_find_request
+
```
gylaz: Unnecessary newline.

gylaz:
Left a few comments, then good to merge.

PR #622, opened 2015-02-11
Backfill timestamp data for repos and memberships

arunagw:
https://trello.com/c/XxaSIcNb


db/migrate/20150211092151_backfill_timestamp_for_repos.rb
```diff
@@ -0,0 +1,55 @@
...
+      UPDATE repos
+        SET updated_at = memberships.updated_at
+        FROM memberships
+        WHERE repos.id = memberships.repo_id
+        AND repos.updated_at is NULL;
+    SQL
+
+    ActiveRecord::Base.connection.execute(sql)
+
+    remaining_updates = <<-SQL
```
arunagw: Found few entires with no memberships.

db/migrate/20150211092151_backfill_timestamp_for_repos.rb
```diff
@@ -0,0 +1,55 @@
...
+      UPDATE memberships
+        SET updated_at = (select current_timestamp)
+        WHERE updated_at is NULL;
+    SQL
+
+    ActiveRecord::Base.connection.execute(remaining_updates)
+  end
+
+  def down
+    ActiveRecord::IrreversibleMigration
```
Koronen: Is this line missing a `fail`/`raise` keyword? I would've expected `fail ActiveRecord::IrreversibleMigration`.

arunagw: Yes we need to have raise add here 0fdd1d9140707d4af65597b63c05171dfc97c12a

db/migrate/20150211092151_backfill_timestamp_for_repos.rb
arunagw: @gylaz @salbertson Found few entries in DB for repo which don't have memberships.

Should we delete them? Or just update `created_at` right now?

db/migrate/20150211092151_backfill_timestamp_for_repos.rb
```diff
@@ -0,0 +1,55 @@
+class BackfillTimestampForRepos < ActiveRecord::Migration
+  def up
+    sql = <<-SQL
+      UPDATE memberships
+        SET created_at = users.created_at
```
gylaz: Likely `repos.created_at` is a better candidate.

db/migrate/20150211092151_backfill_timestamp_for_repos.rb
```diff
@@ -0,0 +1,55 @@
...
+        FROM users
+        WHERE memberships.user_id = users.id
+        AND memberships.created_at is NULL;
+
+      UPDATE memberships
+        SET updated_at = users.updated_at
+        FROM users
+        WHERE memberships.user_id = users.id
+        AND memberships.updated_at is NULL;
```
gylaz: Can the two above queries be combined into one?

db/migrate/20150211092151_backfill_timestamp_for_repos.rb
```diff
@@ -0,0 +1,55 @@
...
+        WHERE updated_at is NULL;
+    SQL
+
+    ActiveRecord::Base.connection.execute(remaining_updates)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
```
gylaz: Should this be a rake task instead?

gylaz: Or is it because the following migration adds a constraint?

croaky:
Ahhhh, thank you! Been wanting this for awhile. I have nothing further beyond @gylaz's comments.

db/migrate/20150211101020_change_timestamp_to_null_false.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeTimestampToNullFalse < ActiveRecord::Migration
+  def change
+    change_column_null :repos, :created_at, false
```
srt32: Would it make sense to run these `change_column_null` commands within the previous migration that sets all the NULL values to not NULL so that we can be within a single transaction? I think we could avoid an edge case where a record gets written in between the two migrations. Is that possible / worth considering?

db/migrate/20150211092151_backfill_timestamp_for_repos.rb
arunagw: Yes I wanted to add null constraint so keeping this into migration

db/migrate/20150211092151_backfill_timestamp_for_repos.rb
arunagw: If we are going to remove them then we don't need to update and can remove this code from here.

PR for updated task: https://github.com/thoughtbot/hound/pull/629

arunagw: I have also found in DB that we have few memberships for which user is not exists so we can't really take created_at from users for those.

db/migrate/20150211101020_change_timestamp_to_null_false.rb
arunagw: @srt32 It might be a good idea but only drawback is that it will not have any rollback migration.

I will give a try and see if it allows us to do this.

arunagw:
@gylaz updated PR. Please have a look

db/migrate/20150213132248_backfill_timestamps_add_null_constraints.rb
```diff
@@ -0,0 +1,65 @@
+class BackfillTimestampsAddNullConstraints < ActiveRecord::Migration
+  def up
+    sql = <<-SQL
```
gylaz: What do you think about giving this a descriptive name like you did on line 31?

db/migrate/20150213132248_backfill_timestamps_add_null_constraints.rb
```diff
@@ -0,0 +1,65 @@
...
+        AND repos.created_at is NULL;
+
+      UPDATE repos
+        SET updated_at = memberships.updated_at
+        FROM memberships
+        WHERE repos.id = memberships.repo_id
+        AND repos.updated_at is NULL;
+    SQL
+
+    ActiveRecord::Base.connection.execute(sql)
```
gylaz: Thoughts on moving this right above the other execute (line 49), to group them together?

gylaz:
A couple of comments. Merge at your discretion.

PR #616, opened 2015-02-11
Update rubocop to latest version

camelmasa:
I think I'd like to update the rubocop version on hound-ci to latest version.
What do you think ?

Or, Should we create pull-request for it :question:

### DIFF
--- a/config/style_guides/thoughtbot/ruby.yml
+++ b/config/style_guides/thoughtbot/ruby.yml
@@ -42,6 +42,7 @@ ClassVars:
   Enabled: false
 
 CollectionMethods:
+  Enabled: true
   PreferredMethods:
     find: detect
     inject: reduce
@@ -83,6 +84,9 @@ Encoding:
 
 EvenOdd:
   Enabled: false
+
+ExtraSpacing:
+  Enabled: true
 
 FileName:
   Enabled: false
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ gem "responders", "~> 2.0"
 gem "resque", "~> 1.25.0"
 gem "resque-retry"
 gem "resque-sentry"
-gem "rubocop", "0.25.0"
+gem "rubocop", "0.29.1"
 gem "sass-rails"
 gem "scss-lint", require: false
 gem "sentry-raven"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,6 +45,8 @@ GEM
     angularjs-rails (1.2.22)
     arel (6.0.0)
     ast (2.0.0)
+    astrolabe (1.3.0)
+      parser (>= 2.2.0.pre.3, < 3.0)
     attr_extras (3.1.0)
     bourbon (4.1.0)
       sass (~> 3.3)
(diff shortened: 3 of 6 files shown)


camelmasa:
I don't know a reason of failing ... on this point.
https://github.com/camelmasa/hound/blob/update-rubocop/spec/models/style_guide/ruby_spec.rb#L439

So, I'm checking the reason :watch:
Please somebody comments if knows it.

gylaz:
@salbertson Should we enable the new `SymbolProc`, `Debugger `, `EmptyLinesAroundBlockBody` and `ExtraSpacing` cops by default?

camelmasa:
About failing test.
See. https://travis-ci.org/thoughtbot/hound/builds/50364697

I succeeded to test when change  rubocop version from`0.29.0` to `0.28.0`.
So, I'll check points from`0.28.0` to `0.29.0`.

camelmasa:
@salbertson I finished to fix failed tests. Could you review if you have chance :question: 
Also, we are waiting your response about this https://github.com/thoughtbot/hound/pull/616#issuecomment-73933072 :wink:

salbertson:
@gylaz, yea, I think those all should be on by default.

config/style_guides/thoughtbot/ruby.yml
```diff
@@ -42,6 +42,7 @@ ClassVars:
   Enabled: false
 
 CollectionMethods:
+  Enabled: true
```
salbertson: @camelmasa, why was this enabled?

gylaz:
@camelmasa do you mind enabling those, and throwing in some tests?

config/style_guides/thoughtbot/ruby.yml
camelmasa: @salbertson The CollectionMethods was changed default to disabled from rubocop v0.27.0. So, I added that.

See. https://github.com/bbatsov/rubocop/commit/88f00951660dc1383de468640bb4d1c717ec4243

camelmasa:
@gylaz Yeah. I'll do it :rocket:

camelmasa:
@gylaz I added tests for new cops.

config/style_guides/thoughtbot/ruby.yml
salbertson: Ah, perfect.

salbertson: Actually, the reason they disabled it was because it reports false positives. Thoughts on leaving it disabled and relying on RuboCop default?

salbertson: @gylaz thoughts?

salbertson:
@camelmasa this looks great, thank you! Would you squash the commits before we merge?

camelmasa:
@salbertson OK :smiley: I squashed the commits.

salbertson:
Thanks @camelmasa!

camelmasa:
@salbertson Thanks for merging :octocat:

config/style_guides/thoughtbot/ruby.yml
gylaz: Are the false positives related to this? https://github.com/thoughtbot/hound/issues/540

I think turning it off now for people that rely on this rule, would be an unexpected change.

PR #625, opened 2015-02-12
Use the user's Github token when inviting to org.

teoljungberg:
To avoid hitting Github's API rate limit, let's use the user's api token as
much as we can. And only use Hound's when we're commenting on violations


teoljungberg:
After re-reading the code, we can't use the user's token to accept Hound's invitations to organizations.
I'll open a PR clarifying that instead

PR #626, opened 2015-02-12
Refactor invitation to organization

teoljungberg:
- Rename `OrgInvitationJob` to `AcceptOrgInvitationsJob`
  - Be explicit that it accepts the org invitations, not sending them
- Explicitly provide github token to `GithubApi`

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -80,6 +80,6 @@ class BuildRunner
   end
 
   def github
-    @github ||= GithubApi.new
+    @github ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
   end
 end
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -1,3 +1,4 @@
+require "attr_extras"
 require "octokit"
 require "base64"
 require "active_support/core_ext/object/with_options"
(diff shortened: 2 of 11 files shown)


arunagw:
I like it. One more eye and should be good to merge.

croaky:
Seems alright to me. Can you further explain the benefits in the commit message?

teoljungberg:
Thanks @croaky and @arunagw, merged

app/services/build_runner.rb
```diff
@@ -80,6 +80,6 @@ def create_success_status
   end
 
   def github
-    @github ||= GithubApi.new
+    @github ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
```
gylaz: Why did we feel the need to make this explicit? I think this kind of explanation is good to include in the commit. Instead of what was done, it's better to specify why it was do.

lib/github_api.rb
```diff
@@ -6,12 +7,10 @@ class GithubApi
   SERVICES_TEAM_NAME = "Services"
   PREVIEW_MEDIA_TYPE = "application/vnd.github.moondragon-preview+json"
 
-  def initialize(token = ENV["HOUND_GITHUB_TOKEN"])
-    @token = token
-  end
+  pattr_initialize :token
```
gylaz: Same question here about token. I made the change to default the token to `houndci`'s because I often use `GithubApi` in the console. It was getting cumbersome to type out `api = GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])` every time.

teoljungberg: I found it hard grasp which token we were using in the differect cases. I feel that it's better to be explicit in the code than having it do unexpected things, like setting a value as the default.

PR #628, opened 2015-02-13
Cache file content requested from GitHub

salbertson:
This should help us avoid hitting our rate limit by not making multiple requests to GitHub for the same file content.

### DIFF
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -8,7 +8,12 @@ class GithubApi
   PREVIEW_MEDIA_TYPE = "application/vnd.github.moondragon-preview+json"
   SERVICES_TEAM_NAME = "Services"
 
-  pattr_initialize :token
+  attr_reader :file_cache, :token
+
+  def initialize(token = ENV["HOUND_GITHUB_TOKEN"])
+    @token = token
+    @file_cache = {}
+  end
 
   def client
     @client ||= Octokit::Client.new(access_token: token, auto_paginate: true)
@@ -76,7 +81,8 @@ class GithubApi
   end
 
   def file_contents(full_repo_name, filename, sha)
-    client.contents(full_repo_name, path: filename, ref: sha)
+    file_cache["#{full_repo_name}/#{sha}/#{filename}"] ||=
+      client.contents(full_repo_name, path: filename, ref: sha)
   end
 
   def accept_pending_invitations
--- a/spec/lib/github_api_spec.rb
+++ b/spec/lib/github_api_spec.rb
@@ -17,6 +17,35 @@ describe GithubApi do
     end
   end
 
+  describe "#file_contents" do
+    context "used multiple times with same arguments" do
+      it "requests file content once" do
+        client = double("Octokit::Client", contents: "filecontent")
+        allow(Octokit::Client).to receive(:new).and_return(client)
+        token = "authtoken"
+        github = GithubApi.new(token)
+        repo = "jimtom/wow"
+        filename = ".hound.yml"
+        sha = "abc123"
+
+        contents = github.file_contents(repo, filename, sha)
+        same_contents = github.file_contents(repo, filename, sha)
+
+        expect(contents).to eq "filecontent"
+        expect(same_contents).to eq contents
+        expect(Octokit::Client).to have_received(:new).with(
+          access_token: token,
+          auto_paginate: true
+        )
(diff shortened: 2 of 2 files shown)


lib/github_api.rb
```diff
@@ -76,7 +78,8 @@ def pull_request_files(full_repo_name, number)
   end
 
   def file_contents(full_repo_name, filename, sha)
-    client.contents(full_repo_name, path: filename, ref: sha)
+    @file_cache[Digest.hexencode(full_repo_name + filename + sha)] ||=
```
salbertson: We might want to get this out of memory and cache somewhere else.

teoljungberg:
I like the idea, and I agree that saving it in memory is not the wisest choice. Maybe redis?

lib/github_api.rb
srt32: Thoughts on pulling `Digest.hexencode(full_repo_name + filename + sha)` into a named private method (something like `file_cache_key_for(repo_name, path, sha)`) for readability?

srt32:
Would it be worth trying this change out using the in memory solution as a starting point (perhaps with a a known large repo) and then move to something like redis in another step?

lib/github_api.rb
srt32: Another thought, what if we didn't `hexencode` here and left the key as human readable (easier for debugging)?

spec/lib/github_api_spec.rb
```diff
@@ -17,6 +17,35 @@
     end
   end
 
+  describe "#file_contents" do
+    context "used multiple times with same arguments" do
+      it "requests file content once" do
```
srt32: Could we add a spec that asserts that the same file for a repo but with a different SHA gets re-requested?

gylaz:
The problem at hand was that we would fetch the contents of the config file multiple times. This seems like a heavy handed solution to that problem. Seems like we really only need to cache the `RepoConfig#for` result per language. Why do we think this approach is better?

gylaz:
This also feels like a "premature memoization" to me. We shouldn't be requesting the same ruby/js/coffee/etc file multiple times per build. So memoizing it doesn't give us any wins there, only potential caching problems. We do know that we happen to hit the config files multiple times. I can see two ways in solving that: A) cache the config per build per language. B) Pass parsed config contents into the `StyleGuide::*` constructor.

salbertson:
> Why do we think this approach is better?

My thought was to drop down lower than `RepoConfig` so any future use of `GithubApi#file_contents` would be efficient without having to think about it. Thoughts?

> This also feels like a "premature memoization" to me.

I don't think it's premature, we are flirting with our rate limit multiple times a day.

> So memoizing it doesn't give us any wins there, only potential caching problems.

What potential caching problems are you thinking of?

@gylaz, want to create a PR with the `RepoConfig` changes you are talking about and we can see how that turns out?

gylaz:
> I don't think it's premature, we are flirting with our rate limit multiple times a day.

But the problem isn't that we're fetching other files multiple times. It's more of a logic error, that we're re-fetching the config files.

> What potential caching problems are you thinking of?

Ha, as if we knew what problems we'd run into when we enabled GitHub API caching in Octokit.

> @gylaz, want to create a PR with the RepoConfig changes you are talking about and we can see how that turns out?

@salbertson Done in https://github.com/thoughtbot/hound/pull/636

salbertson:
@gylaz the issues we ran into before with invalidating the cache were because we relied on GitHub to tell us if the cache was invalid using cache headers, not a filename and sha.

salbertson:
> We shouldn't be requesting the same ruby/js/coffee/etc file multiple times per build.

You're right, we shouldn't, but we are. Caching file content in multiple places will fix the currenct problem but we might end up in the same spot if code is added to fetch other file content and it's not cached correctly.

I'm trying to avoid duplicated logic and future issues by having the caching logic at the `GithubApi` level.

lib/github_api.rb
gylaz: How about just:
```rb
@file_contents["#{full_repo_name}/#{sha}/#{filename}"] ||= ...
```
It guarantees uniqueness and is will make it easier to debug the cache if we need to?

salbertson: Sure, works for me.

salbertson:
@gylaz, you closed your PRs that basically did the same thing so I'm guessing you're ok with this approach?

PR #629, opened 2015-02-13
Update `remove_without_memberships` task to only remove inactive repo

arunagw:
Based on this comment https://github.com/thoughtbot/hound/pull/622#discussion_r24605555

### DIFF
--- a/lib/tasks/repo.rake
+++ b/lib/tasks/repo.rake
@@ -24,7 +24,10 @@ namespace :repo do
       WHERE id IN (
         SELECT repos.id FROM repos
         LEFT OUTER JOIN memberships on memberships.repo_id = repos.id
+        LEFT OUTER JOIN subscriptions on subscriptions.repo_id = repos.id
         WHERE memberships.id IS NULL
+        AND subscriptions.id is NULL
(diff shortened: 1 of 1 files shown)


lib/tasks/repo.rake
```diff
@@ -25,7 +25,7 @@ namespace :repo do
         SELECT repos.id FROM repos
         LEFT OUTER JOIN memberships on memberships.repo_id = repos.id
         WHERE memberships.id IS NULL
-      )
```
srt32: Would this change also work if `AND repos.active = false` went inside the `)` so that it was part of the sub-select? I think it would be readable that way but perhaps I'm missing the point of this change.

gylaz:
How about also making sure there isn't an associated subscription? I'm thinking that we'd want to keep those repos for data integrity.

salbertson:
When are these rake tasks used?

arunagw:
@salbertson I think this is going to run once in a while to clean database. Or we can run this daily via cron.

gylaz:
I'm okay with this change as it improves the existing query. I think it's a separate discussion of when, or why we want to run it.

PR #632, opened 2015-02-13
Display detailed error message if activation fails

malandrina:
* Currently when repo activation fails a generic error message
  is displayed to the user. This commit introduces support for
  displaying detailed, actionable error messages when repo activation fails.
* Update `RepoActivator` to add and expose errors when repo activation
  fails due to a GitHub permissions issue.
* Update `ActivationsController#create` to return error messages when
  error messages are present.
* Introduce `ErrorMessageTranslation` to handle translating
  non-user-friendly Octokit error messages into messages that can be
  displayed to the end user.
* Update client-side error handling to display error messages included
  in response body
* https://trello.com/c/KYuTdbXS/405-when-a-user-tries-to-enable-a-repo-and-it-fails-they-get-no-feedback-as-to-why-enabling-the-repo-failed

### DIFF
--- a/app/assets/javascripts/directives/repo.js.coffee.erb
+++ b/app/assets/javascripts/directives/repo.js.coffee.erb
@@ -6,8 +6,14 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription, 
       scope.processing = true
 
       scope.repo.$activate()
-        .catch(-> alert('Your repo failed to activate.'))
+        .catch((response) -> alert(errorMessage(response.data)))
         .finally(-> scope.processing = false)
+
+    errorMessage = (responseData) ->
+      if responseData.errors.length > 0
+        responseData.errors.join("\n")
+      else
+        "Your repo failed to activate"
 
     deactivateRepo = ->
       scope.processing = true
--- a/app/controllers/activations_controller.rb
+++ b/app/controllers/activations_controller.rb
@@ -9,10 +9,9 @@ class ActivationsController < ApplicationController
   def create
     if activator.activate
       analytics.track_repo_activated(repo)
-
       render json: repo, status: :created
     else
-      head 502
+      render json: { errors: activator.errors }, status: 502
     end
   end
 
@@ -25,7 +24,7 @@ class ActivationsController < ApplicationController
   end
 
   def activator
-    RepoActivator.new(repo: repo, github_token: github_token)
+    @activator ||= RepoActivator.new(repo: repo, github_token: github_token)
   end
 
   def repo
--- a/app/services/repo_activator.rb
+++ b/app/services/repo_activator.rb
@@ -1,7 +1,10 @@
 class RepoActivator
+  attr_reader :errors
+
   def initialize(github_token:, repo:)
     @github_token = github_token
     @repo = repo
+    @errors = []
   end
 
   def activate
@@ -27,6 +30,7 @@ class RepoActivator
   def change_repository_state_quietly
     yield
   rescue Octokit::Error => error
+    add_error(error)
     Raven.capture_exception(error)
     false
   end
@@ -67,5 +71,10 @@ class RepoActivator
     else
       "http"
     end
+  end
+
+  def add_error(error)
+    error_message = ErrorMessageTranslation.from_error_response(error)
+    errors.push(error_message).compact!
   end
 end
--- a/lib/error_message_translation.rb
+++ b/lib/error_message_translation.rb
@@ -1,0 +1,12 @@
+module ErrorMessageTranslation
+  ERROR_CODE_AND_MESSAGE = /.*(\d{3})\s\-\s(.*)\s\/\/.*/
+  FORBIDDEN_ERROR_CODE = "403"
+
+  def self.from_error_response(error)
+    matches = error.message.match(ERROR_CODE_AND_MESSAGE)
+
+    if matches.present? && matches.captures[0] == FORBIDDEN_ERROR_CODE
+      matches.captures[1]
+    end
+  end
+end
--- a/spec/lib/error_message_translation_spec.rb
+++ b/spec/lib/error_message_translation_spec.rb
@@ -1,0 +1,48 @@
+require "fast_spec_helper"
+require "lib/error_message_translation"
+
+describe ErrorMessageTranslation do
+  describe ".from_error_response" do
+    context "when error status is 403" do
+      it "returns error message" do
+        error = double("error", message: octokit_403_error_message)
+        expected_message = "You must be an admin to add a team membership."
+
+        result = ErrorMessageTranslation.from_error_response(error)
+
+        expect(result).to eq expected_message
+      end
+    end
+
+    context "when error status is not 403" do
+      it "returns nil" do
+        error = double("error", message: octokit_400_error_message)
+
+        result = ErrorMessageTranslation.from_error_response(error)
+
+        expect(result).to be_nil
+      end
+    end
+
+    context "when error does not adhere to expected formatting" do
+      it "returns nil" do
+        message = "error"
+        error = double("error", message: message)
+
+        result = ErrorMessageTranslation.from_error_response(error)
+
+        expect(result).to be_nil
+      end
+    end
+  end
+
+  private
+
+  def octokit_403_error_message
+    "PUT https://api.github.com/teams/3675/memberships/houndci: 403 - You must be an admin to add a team membership. // See: https://developer.github.com/v3"
+  end
+
+  def octokit_400_error_message
+    "PUT https://api.github.com/teams/3675/memberships/houndci: 400 - Problems parsing JSON. // See: https://developer.github.com/v3"
+  end
+end
--- a/spec/controllers/activations_controller_spec.rb
+++ b/spec/controllers/activations_controller_spec.rb
@@ -6,7 +6,7 @@ describe ActivationsController, "#create" do
       token = "sometoken"
       membership = create(:membership)
       repo = membership.repo
-      activator = double(:repo_activator, activate: true)
+      activator = double(:repo_activator, activate: true, errors: [])
       allow(RepoActivator).to receive(:new).and_return(activator)
       stub_sign_in(membership.user, token)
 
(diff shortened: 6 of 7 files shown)


lib/error_message_translation.rb
```diff
@@ -0,0 +1,17 @@
+# Parses OctoKit error messages
+# Returns human-readable error message
```
salbertson: Hopefully this is clear based on the class and method names. Think we could remove these comments?

spec/lib/error_message_translation_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "spec_helper"
+
+describe ErrorMessageTranslation do
+  describe ".from_error_response" do
+    context "when error status is 403" do
+      it "returns error message" do
+        message = "PUT https://api.github.com/teams/3675/memberships/houndci: 403 - You must be an admin to add a team membership. // See: https://developer.github.com/v3"
```
salbertson: What do you think about putting this in a fixture, similar to other responses?

salbertson:
I like it! :+1: 

What error messages did you run across when testing? Just curious what errors we will see, for customer support.

app/controllers/activations_controller.rb
```diff
@@ -7,12 +7,13 @@ class CannotActivatePaidRepo < StandardError; end
   before_action :check_repo_plan
 
   def create
-    if activator.activate
-      analytics.track_repo_activated(repo)
+    activator.activate
 
-      render json: repo, status: :created
+    if activator.errors.any?
+      render json: { errors: activator.errors }, status: 502
```
srt32: Unrelated to this PR, but would it be interesting to track in analytics, failed repo activations?

malandrina: That sounds like useful information. I made a card in the "Chores" list: https://trello.com/c/RKKrlrbB/487-track-failed-repo-activation-events

app/services/repo_activator.rb
```diff
@@ -27,6 +30,7 @@ def activate_repo
   def change_repository_state_quietly
     yield
   rescue Octokit::Error => error
+    @errors.push(ErrorMessageTranslation.from_error_response(error))
```
srt32: Can this be `errors`?

lib/error_message_translation.rb
malandrina: Sure, I'm happy to get rid of them. Removed in https://github.com/thoughtbot/hound/commit/811f34bf6e87e0a86620cedfa51543eb7e79fc32. I noticed we have descriptive comments for a bunch of classes in `app/models` and wasn't sure if we were still doing that.

spec/lib/error_message_translation_spec.rb
malandrina: Good call. It makes the tests much easier to read. Done in https://github.com/thoughtbot/hound/commit/811f34bf6e87e0a86620cedfa51543eb7e79fc32.

app/services/repo_activator.rb
malandrina: Yep! Good :eyes: . Done: https://github.com/thoughtbot/hound/commit/85c439aee3e8ffa557fbd030d6ca4262a47375ac

malandrina:
@salbertson unfortunately I haven't found any GitHub documentation listing the error messages that their API might return. `"You must be an admin to add a team membership"` is the message I get when I try to activate a thoughtbot repo right now, so I figured I'd start there.

It looks like the Octokit gem builds its errors by looking at the GitHub API response and concatenating the request URL, status code, error message(s), and documentation URL ([reference](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/error.rb)). More research is needed to figure out which GitHub error messages should be displayed to end users (and which ones shouldn't be).

app/assets/javascripts/directives/repo.js.coffee.erb
```diff
@@ -6,9 +6,15 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription,
       scope.processing = true
 
       scope.repo.$activate()
-        .catch(-> alert('Your repo failed to activate.'))
+        .catch((response) -> alert(errorMessage(response.data)))
         .finally(-> scope.processing = false)
 
+    errorMessage = (responseData) ->
+      if responseData.errors
```
thorncp: Will this always be truthy? It seems if `ErrorMessageTranslation` doesn't recognize the error, it will return nil, which gets pushed into the errors collection, resulting in `[nil]`, and serialized to json as `[null]`.

spec/support/fixtures/octokit_400_error.txt
```diff
@@ -0,0 +1 @@
+PUT https://api.github.com/teams/3675/memberships/houndci: 400 - Problems parsing JSON. // See: https://developer.github.com/v3
```
srt32: We get plaintext errors back from Octokit?

app/assets/javascripts/directives/repo.js.coffee.erb
salbertson: What about this?

```
if responseData.errors.length > 0
...
```

spec/support/fixtures/octokit_400_error.txt
malandrina: We get an instance of `Octokit::Forbidden`, which has a `message` method. `message` returns a string. The string is a concatenation of the request URL, error response status, error message from the GitHub API, and the documentation URL.

app/assets/javascripts/directives/repo.js.coffee.erb
malandrina: @thorncp you're right -- in the case you describe the response is `"{ errors: [null] }"`. I'll work on changing that!

gylaz:
```js
[null].length //=> 1
```
But we shouldn't be pushing that `nil` into the array.

app/controllers/activations_controller.rb
```diff
@@ -25,7 +24,7 @@ def check_repo_plan
   end
 
   def activator
-    RepoActivator.new(repo: repo, github_token: github_token)
+    @activator ||= RepoActivator.new(repo: repo, github_token: github_token)
```
gylaz: Why is this necessary?

lib/error_message_translation.rb
```diff
@@ -0,0 +1,14 @@
+class ErrorMessageTranslation
```
gylaz: Should this be `module`, to express that this isn't meant to be instantiated?

lib/error_message_translation.rb
```diff
@@ -0,0 +1,14 @@
...
+  ERROR_CODE_AND_MESSAGE = /.*(\d{3})\s\-\s(.*)\.\s\/\/.*/
```
gylaz: Good lord!

gylaz: I'm not sure why, but I'm not getting match against:
```
GET https://api.github.com/repos/thoughtbot/houd: 404 - Not Found // See: https://developer.github.com/v3
```

gylaz: I think it's because of us assuming there will be a dot `.`.

lib/error_message_translation.rb
```diff
@@ -0,0 +1,14 @@
...
+  FORBIDDEN_ERROR_CODE = "403"
+
+  def self.from_error_response(error)
+    error_code_and_message = Hash.new
+
+    error.message.scan(ERROR_CODE_AND_MESSAGE) do |error_code, error_message|
+      error_code_and_message[error_code] = error_message
+    end
+
+    error_code_and_message.fetch(FORBIDDEN_ERROR_CODE, nil)
```
gylaz: Wouldn't this method be simpler and be more readable as:
```
captures = error.message.match(ERROR_CODE_AND_MESSAGE).captures
if capture[0] == FORBIDDEN_ERROR_CODE
  captures[1]
end
```

spec/lib/error_message_translation_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "spec_helper"
```
gylaz: I think this can be `fast_spec_helper`.

spec/support/fixtures/octokit_400_error.txt
gylaz: Any reason this is in a file, as opposed to just a string in the test?

app/services/repo_activator.rb
```diff
@@ -27,6 +30,7 @@ def activate_repo
   def change_repository_state_quietly
     yield
   rescue Octokit::Error => error
+    errors.push(ErrorMessageTranslation.from_error_response(error))
```
gylaz: How about throwing a `compact` on the end to strip out the `nil`s?

spec/support/fixtures/octokit_400_error.txt
malandrina: Scott and I thought it made the tests easier to read: https://github.com/thoughtbot/hound/pull/632#discussion_r24702535

app/controllers/activations_controller.rb
malandrina: `activator` is now being called twice: on line 10 and again on line 14.

lib/error_message_translation.rb
malandrina: Thanks for testing it out. I changed the regex in https://github.com/thoughtbot/hound/commit/7ff9921756e30e8454dabd5d67825313292288f0.

lib/error_message_translation.rb
malandrina: Sure, sounds good. Done: https://github.com/thoughtbot/hound/commit/7ff9921756e30e8454dabd5d67825313292288f0

lib/error_message_translation.rb
malandrina: Looks simpler to me. Done: https://github.com/thoughtbot/hound/commit/7ff9921756e30e8454dabd5d67825313292288f0

malandrina:
@thorncp @salbertson @srt32 @gylaz ready for re-review.

srt32:
More bubbles. I re-kicked the Travis job.

malandrina:
@srt32 thanks again! We need a :bubble: emoji.

spec/support/fixtures/octokit_400_error.txt
gylaz: The only thing, it's not a response, but just a text representation of the error extracted from the response. I can see that confusing people.

gylaz:
:+1:

spec/support/fixtures/octokit_400_error.txt
malandrina: @gylaz yeah, I agree that it might be confusing. What do you think about putting them in a subdirectory? e.g. `spec/support/fixtures/octokit/403_error.txt`

gylaz: Personally, I feel like moving an arbitrary string to a file is a bit of a premature extraction. If it makes the test ugly because it's long, I'd extract it to a method first. Whatever, you feel is the best approach :smile:.

malandrina: I think the fact that both you and @srt32 found it a little misleading/confusing is a good reason to not extract fixtures here. I'll make the change.

malandrina: Done: https://github.com/thoughtbot/hound/commit/c81a0120614fe5147175b67d72d67647cf77f707

Issue #293, opened 2014-06-18
Idea: Store config in guides


jferris:
We're in a situation now where adding a new style guideline means:

1. Opening a pull request to guides.
2. Waiting a week for discussion.
3. Merging the pull request.
4. Opening a pull request to hound.
5. Reviewing, merging, and deploying that pull request.

In between steps 3 and 5, we get comments from Hound which conflict with our style guide.

It would be nice if:

* We had a config file in the guides repo.
* We could point Hound to it as our default config.
* Hound would somehow fetch updates for that config.

This way, we could update the config as part of pull requests which change the style guidelines. The guides and Hound will never be out of sync.

salbertson:
@jferris, organization level configuration is something we are talking about adding. There has been talk of creating a UI in Hound to manage these style guides.

I think it would be pretty cool if Hound could use a gist or a file in a repo for each language.

PR #635, opened 2015-02-14
Remove disabling of cops for thoughtbot Ruby config

gylaz:
IIRC, the original decision to disable these was because they were not explicitly
specified in our style guide. The default RuboCop configuration is pretty good.
We should only override the things we don't agree with, rather than disabling
the cops that are not in our style guide.


gylaz:
@jferris your thoughts on this?

salbertson:
I like it. :+1:

salbertson:
@gylaz, what's the latest with this?

gylaz:
Wanted to see if @jferris had anything to say about this.

salbertson:
@croaky is there anything that would be enabled by this change that you think is a deal breaker? I want to make sure we aren't enabling something that would surprise most users, in a bad way.

tute:
I was just working with some rubocop output and thought "why did hound never tell me this?". I like this change.

jferris:
I don't think we should do this. I would actually prefer that cops were not added until we enable them proactively.

Every time Hound makes a comment which isn't in line with our guides, it decreases our success with Hound. A pull request submitter will likely choose to ignore the comment (since our guides don't support what Hound is saying), which makes people get more used to ignoring Hound. In turn, this worsens the signal-to-noise ratio for Hound comments.

The extra noise gives ammunition to nay-sayers. Every comment by Hound which doesn't produce value for us creates a situation in which some of us will question whether or not to use Hound in the future.

It's been enough of a battle to get people to agree to almost anything in the style guide. It's been an additional fight to get people to accept Hound enforcing that guide. Having Hound automatically start commenting with new requirements we never agreed to starts yet another fight.

I short, I would prefer to be conservative, because I think these comments will start arguments instead of discussions.

mcmire:
I agree with @jferris on this, good reasoning.

salbertson:
@jferris, sounds good.

Here's the entire [default Hound config](https://github.com/thoughtbot/guides), maybe the thoughtbot version should be part of https://github.com/thoughtbot/guides? That way it's very clear what is enabled and thoughtbot repos won't be affected by changes to RuboCop or Hound defaults.

Let me know if that sounds like a good idea and I'll open a PR.

jferris:
> maybe the thoughtbot version should be part of https://github.com/thoughtbot/guides?

Yes, [I'd love that](https://github.com/thoughtbot/hound/issues/293)!

PR #636, opened 2015-02-14
Cache the language config contents in RepoConfig

gylaz:
`RepoConfig#for` is called multiple times for a particular language, which
causes a fetch of the file contents from GitHub each time. We don't expect the
config to change within a single build. The caching will prevent unnecessary
requests that count against our API rate limit.

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -9,7 +9,9 @@ class RepoConfig
     "scss" => "yaml",
   }
 
-  pattr_initialize :commit
+  pattr_initialize :commit do
+    @language_configs = {}
+  end
 
   def enabled_for?(language)
     options = options_for(language)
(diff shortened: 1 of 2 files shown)


salbertson:
@gylaz is it possible for `commit.file_content(ignore_file)` to be called multiple times in #load_javascript_ignore?

gylaz:
> @gylaz is it possible for commit.file_content(ignore_file) to be called multiple times in #load_javascript_ignore?

Sure is. I think I'm starting to see the light of caching at `Commit` level.

gylaz:
@salbertson What do you think of this approach? https://github.com/thoughtbot/hound/pull/638
It's a bit easier to understand the cache key, and caches in the only place that `file_contents` is called.

PR #638, opened 2015-02-14
Cache file contents in the Commit

gylaz:
Cache the file contents to avoid hitting the GitHub API unnecessarily. This is
particularly to fix the issue of us fetching the configuration files repeatedly
for each file that we process.


app/models/commit.rb
```diff
@@ -1,9 +1,19 @@
...
+    @file_contents = {}
+  end
+
   def file_content(filename)
+    @file_contents[filename] ||= fetch_file(filename)
+  end
+
+  private
+
+  def fetch_file(filename)
```
salbertson: Oh dang.

Issue #515, opened 2014-12-17
houndci.com 404 page


Ashton-W:
Not sure if website issues should be posted here.

`404`s look like this, wether from a mistyped url or a broken link
![screencapture at 17-12-2014-10_25_46](https://cloud.githubusercontent.com/assets/1650398/5464159/9dcc955c-85d8-11e4-983e-01631192368d.png)

This could be improved with a styled page with links to the homepage, etc.

salbertson:
@frechg @donokuda, thoughts?

frechg:
Yeah we could use some styled error pages.

srt32:
Maybe something like this?

![image](https://cloud.githubusercontent.com/assets/2181356/6086143/58bbb4b8-adf4-11e4-8671-fa3cac16bec5.png)

frechg:
hahaha nice!

frechg:
Also kind of sad.

frechg:
But I like the idea of building on that theme.

Ashton-W:
No, that's perfect 🐶

salbertson:
@frechg I'm merging https://github.com/thoughtbot/hound/pull/639, feel free to make the pages look nice or add some personality, ha.

salbertson:
@Ashton-W thanks for opening an issue for this! Nicer error pages should be out today.

PR #639, opened 2015-02-15
Show error pages content in normal layout

salbertson:
This change gives us more control over the look of error pages.

http://wearestac.com/blog/dynamic-error-pages-in-rails

https://github.com/thoughtbot/hound/issues/515


salbertson:
https://github.com/thoughtbot/hound/issues/515

arunagw:
Looks ok to me. I have never done any customization for error page via rails. 

For example will it serve 500 if DB is down ?

srt32:
Would it be possible to add a spec around the error page routing?

teoljungberg:
I would like to see a request spec for this aswell

config/routes.rb
```diff
@@ -20,5 +20,9 @@
   resources :repo_syncs, only: [:index, :create]
   resource :user, only: [:show]
 
+  %w( 404 422 500 ).each do |status_code|
```
gylaz: no spaces after `(` and before `)`.

salbertson:
@srt32 @teoljungberg good call on tests, done. I added a simple test that makes sure error pages render correctly and avoided testing the way Rails handles errors.

config/routes.rb
salbertson: Why didn't Hound comment?

PR #641, opened 2015-02-16
Add spring with binstubs

teoljungberg:
Create binstubs for `rails`, `rake`, and `rspec`


arunagw:
Looks fine to me

gylaz:
Why do we need Spring?

teoljungberg:
Having spring makes the rails tests feel as fast as the non-rails tests, which always gives me a warm little feeling inside.
It's not like we're in a dire need of `spring`, but it just feels little better

gylaz:
We've been pretty conscious about keeping Hound's test suite fast. `spring` tends to "hide" the fact that the tests suite is slow -- if it's an obvious pain then we should work to improve the speed of the tests. Also, on the minor note, there are certain issues with `spring` that can waste hours of time debugging, because it didn't pick up the latest change.

Right now, on my machine, the test suite runs in under 30 seconds.

I'm :-1: on this, as I see `spring` as a crutch, and I'd like to keep it out.

salbertson:
I agree, let's avoid including spring unless we need it.

teoljungberg:
Thanks for weighing in @gylaz, @salbertson. I'll close this and we'll revisit this once we feel the need for spring

PR #642, opened 2015-02-16
Timeout long-running jobs

teoljungberg:
I'm unsure if 60 seconds is enough though, thoughts?


config/initializers/resque.rb
```diff
@@ -22,3 +23,5 @@
 Resque.after_fork do
   defined?(ActiveRecord::Base) && ActiveRecord::Base.establish_connection
 end
+
+Resque::Plugins::Timeout.timeout = 60
```
arunagw: I think this needs to be a ENV variable as well and defaults to 60.

teoljungberg: I agree, I'll fix that

config/initializers/resque.rb
```diff
@@ -22,3 +23,5 @@
 Resque.after_fork do
   defined?(ActiveRecord::Base) && ActiveRecord::Base.establish_connection
 end
+
+Resque::Plugins::Timeout.timeout = ENV["RESQUE_TIMEOUT"]
```
arunagw: How about having a fallback here if we do not set in ENV. Like we have in unicorn.rb `timeout (ENV["UNICORN_TIMEOUT"] || 15).to_i`

teoljungberg: Absolutely, I like it

teoljungberg:
I'm still unsure if 60seconds is a good value, what do you think @arunagw ?

arunagw:
Not sure about the number.. Need to know what are current production stats may be check newrelic ?

May be @gylaz can tell us more about this

Gemfile
```diff
@@ -26,6 +26,7 @@ gem "responders", "~> 2.0"
 gem "resque", "~> 1.25.0"
 gem "resque-retry"
 gem "resque-sentry"
+gem "resque-timeout"
```
srt32: Is this gem coming from [here](https://github.com/jeffgran/resque-timeout)? It seems not under active development but perhaps that's okay?

config/initializers/resque.rb
```diff
@@ -22,3 +23,5 @@
 Resque.after_fork do
   defined?(ActiveRecord::Base) && ActiveRecord::Base.establish_connection
 end
+
+Resque::Plugins::Timeout.timeout = (ENV["RESQUE_TIMEOUT"] || 60).to_i
```
croaky: I think it would need to be less than 30s? The app is running on Heroku, which has a 30s timeout.

srt32: Background jobs can go longer, no?

Gemfile
teoljungberg: I didn't notice that it was that unmaintained, wow.
I think it's okey since it's a pretty small gem. If we notice any issues with it we can just fork it and take over the development of it.

config/initializers/resque.rb
teoljungberg: It's only web-processes that have a 30s timeout iirc, worker processes can live longer

croaky: @teoljungberg Yeah, seems like the 30s timeout is only for Heroku web processes, and is an HTTP request timeout through their routing mesh (the H12 error). Looks like there is no timeout for other dyno timeouts, although the docs cover things like dynos get restarted once per day.

Should this ENV variable be differently, or more specifically named? It's not Resque's boot timeout, for instance. Maybe `RESQUE_JOB_TIMEOUT`?

srt32: Or be more library agnostic and call it something like `BACKGROUND_JOB_MAX_TIME` or something...

teoljungberg: `RESQUE_JOB_TIMEOUT` sounds like great fit!

.sample.env
```diff
@@ -8,3 +8,4 @@ MAX_COMMENTS=10
 QUEUE=high,medium,low
 RESQUE_ADMIN_PASSWORD=password
 SECRET_KEY_BASE=secret_key_base
+RESQUE_JOB_TIMEOUT=60
```
croaky: Maybe `:sort` this file.

croaky:
Looks good to me.

gylaz:
I don't have any stats around how long jobs take, but 60 seconds seems like a good enough default.

srt32:
In the last week, the longest job took ~27 seconds.  `60` should probably be good.

screen shot 2015-02-16 at 12 50 02 pm

srt32:
hmm, actually, looks like the `LargeBuildJob`'s can take longer. Perhaps we should use `120` for now?
screen shot 2015-02-16 at 1 04 57 pm

teoljungberg:
I bumped the max timeout to 120s.
Ready for a final overview then we're good to merge

arunagw:
Looks good to merge.

croaky:
>  Perhaps we should use 120 for now?

If the purpose of this pull request is to kill long-running jobs in order to avoid them maxing out CPU and getting hung, do we want to set a value that is above normal job runs but below the outliers which are causing the problems? 120s might be too high for that? Maybe 90s?

Not something that should stop this from being merged, though, since we can set the `ENV` variable.

Past 7 days of `LargeBuildJob` response time:

screen shot 2015-02-17 at 10 43 18 am

srt32:
Do the jobs that hang and get killed report to New Relic? i.e. - is that graph all jobs or just the successful ones?

croaky:
@srt32 It must not. There's one that's an hour old right now and the image on this PR's Trello card shows three at 19m. 120s does seem solid.

teoljungberg:
120s seems like a good default in that case, we can always tweak the env variable to see how that works.

teoljungberg:
Thanks @srt32 @gylaz @croaky @arunagw for weighing in, Merging

salbertson:
@teoljungberg, will we know when jobs are long running and time out? If so, how?

teoljungberg:
@salbertson It raises a `TimeoutError` that will both be visible in the resque web ui and in sentry

PR #644, opened 2015-02-16
Track failed repo activations

arunagw:


spec/controllers/activations_controller_spec.rb
```diff
@@ -54,6 +54,30 @@
...
     end
+
+    it "tracks failed activation" do
+      membership = create(:membership)
+      repo = membership.repo
+      activator = double(
+        :repo_activator,
+        activate: false,
+        errors: []
+      ).as_null_object
```
teoljungberg: Since we're only actually calling `:activate` on the `RepoActivator`, there is no need to have it as a null object

spec/controllers/activations_controller_spec.rb
```diff
@@ -54,6 +54,30 @@
           with(repo: repo, github_token: token)
       end
     end
+
+    it "tracks failed activation" do
+      membership = create(:membership)
+      repo = membership.repo
+      activator = double(
+        :repo_activator,
```
teoljungberg: I know @gylaz have pointed out in the past that we should use `"RepoActivator"` instead of `:repo_activator` for doubles

teoljungberg:
Looks good to me

app/models/analytics.rb
```diff
@@ -10,6 +10,17 @@ def track_signed_in
     track(event: "Signed In")
   end
 
+  def track_repo_activation_failed(repo)
+    track(
+      event: "Repo Activation Failed",
+      properties: {
+        name: repo.full_github_name,
+        private: repo.private,
+        revenue: -repo.plan_price,
```
srt32: I'm not sure we're actually losing revenue when this happens. We are forgoing positive revenue but I think negative revenue would be confusing.

croaky: Yes, I think we do not want `revenue` in here. It could result in calculations we don't want. https://segment.com/docs/api/tracking/track/#special-properties

gylaz:
:+1:

PR #645, opened 2015-02-16
Scrub parameters before sending exception to Sentry

teoljungberg:
Instead send the parameters that might be interesting for the exception,
i.e `user_id`, `payload_data` etc.

### DIFF
--- a/app/jobs/accept_org_invitations_job.rb
+++ b/app/jobs/accept_org_invitations_job.rb
@@ -8,5 +8,7 @@ class AcceptOrgInvitationsJob
     github.accept_pending_invitations
   rescue Resque::TermException
     Resque.enqueue(self)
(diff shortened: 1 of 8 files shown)


gylaz:
Have we looked into what Raven's sanitizing processors are providing for us? http://raven.readthedocs.org/en/latest/config/#sanitizing-data

teoljungberg:
@gylaz neither of the options they provided did what we want. So it felt right doing a more "homegrown" solution

teoljungberg:
What we want is to remove the local parameters from the exception, where we send in things the like user's tokens. Which we don't want stored anywhere

gylaz:
:+1:

teoljungberg:
Thanks for the feedback @gylaz, merging

PR #646, opened 2015-02-16
Destroy memberships on `User` and `Repo` destroy

arunagw:
https://trello.com/c/kaZhpv02


spec/models/repo_spec.rb
```diff
@@ -13,6 +13,12 @@
     expect(subject).to validate_uniqueness_of(:github_id)
   end
 
+  it "destroys memberships if on destroy" do
```
teoljungberg: can't we use the should matchers here?
```ruby
should have_many(:memberships).dependent(:destroy)
```

srt32:
What would you think about adding foreign keys for these relationships?  https://github.com/thoughtbot/guides/blob/master/best-practices/README.md#rails

db/migrate/20150217090319_add_foreign_key_memberships.rb
```diff
@@ -0,0 +1,16 @@
+class AddForeignKeyMemberships < ActiveRecord::Migration
+  def up
+    ActiveRecord::Base.connection.execute <<-SQL
```
teoljungberg: Since this query takes _a long_ time to run locally, might it be a better fit as a one-off rails runner executed in prod?

srt32: How long is long?

srt32: Is this whole `up` command wrapped in a single transaction implicitly?

teoljungberg: @srt32 I timed the query on my local machine with a local copy of the database. Here is the result:
```ruby
>> t0 = Time.now; Membership.where('repo_id NOT IN (SELECT id FROM repos)').count; puts Time.now - t0
   (4274134.6ms)  SELECT COUNT(*) FROM "memberships" WHERE (repo_id NOT IN (SELECT id FROM repos))
4274.161292
=> nil
```

srt32: hmmm, maybe run `EXPLAIN` on it to see what's up?

gylaz: `ActiveRecord::Base.connection.execute` to just `execute`.

db/migrate/20150217090319_add_foreign_key_memberships.rb
```diff
@@ -0,0 +1,16 @@
...
+      DELETE FROM memberships
+      WHERE (repo_id NOT IN (SELECT id FROM repos))
+    SQL
```
gylaz: Should this be done for `WHERE (user_id NOT IN (SELECT id FROM users))`?

db/migrate/20150217090319_add_foreign_key_memberships.rb
srt32: Following up on my transaction question, I think we want a single transaction so that the deletions happen within the same transaction as the new FK's. Otherwise, I think it's possible that (especially if the query takes a long time) that new memberships with missing data could be created before the FK's get applied. Maybe I'm wrong since we don't programmatically delete `repos`?

db/migrate/20150217090319_add_foreign_key_memberships.rb
arunagw: We don't need to for users as fk migration was fine for users so there are no entries

arunagw:
I think we need to run this as a `rake` task as we don't know how much time it's going to take on production.

Should we split this into two PR's and first cleanup memberships and then add fk's ?

teoljungberg:
I think that's a good approach @arunagw

teoljungberg:
Or maybe we don't need to split the PR up? We can just make sure to run the rake task before deploying `master` to staging and then production

PR #647, opened 2015-02-17
Add log_level for production

arunagw:
This will fix warnings for production environment

https://trello.com/c/sq2tJ22Q

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -33,7 +33,7 @@ Houndapp::Application.configure do
   # config.force_ssl = true
 
   # See everything in the log (default is :info)
-  # config.log_level = :debug
+  config.log_level = :debug
 
   # Prepend all log lines with the following tags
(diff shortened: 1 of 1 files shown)


teoljungberg:
:+1:

config/environments/production.rb
```diff
@@ -33,7 +33,7 @@
   # config.force_ssl = true
 
   # See everything in the log (default is :info)
-  # config.log_level = :debug
+  config.log_level = :debug
```
gylaz: Should this be `:info` in production?

teoljungberg: Rails 5 will use `:debug` in all envs, so isn't it a good thing to have that now?

gylaz: If that will be the default then :+1: . `:debug` just felt heavy for production -- since there are a ton of logs as is.

PR #648, opened 2015-02-17
Do not run build if non PR hook

arunagw:
https://trello.com/c/0j5NoBwH

### DIFF
--- a/app/models/payload.rb
+++ b/app/models/payload.rb
@@ -45,7 +45,7 @@ class Payload
     repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
-  def pull_request_attached?
+  def pull_request?
     !pull_request.empty?
   end
 
--- a/app/models/payload.rb
+++ b/app/models/payload.rb
@@ -45,7 +45,7 @@ class Payload
     repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
-  def can_be_reviewed?
+  def pull_request_attached?
     !pull_request.empty?
   end
 
--- a/app/models/payload.rb
+++ b/app/models/payload.rb
@@ -45,6 +45,10 @@ class Payload
     repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
+  def can_be_reviewed?
+    !pull_request.empty?
+  end
+
   private
 
   def parse_data
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -20,7 +20,9 @@ class BuildRunner
   private
 
   def relevant_pull_request?
-    pull_request.opened? || pull_request.synchronize?
+    pull_request.can_be_reviewed? && (
+      pull_request.opened? || pull_request.synchronize?
+    )
   end
 
   def violations
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -20,7 +20,7 @@ class BuildRunner
   private
 
   def relevant_pull_request?
-    pull_request.can_be_reviewed? && (
+    pull_request.reviewable? && (
       pull_request.opened? || pull_request.synchronize?
     )
   end
--- a/spec/features/builds_spec.rb
+++ b/spec/features/builds_spec.rb
@@ -34,6 +34,25 @@ feature 'Builds' do
     end
   end
 
+  context "with non PR payload" do
+    scenario "does not run build" do
+      non_pr_payload = File.read(
+        "spec/support/fixtures/pull_request_opened_event.json"
+      )
+      parsed_payload = JSON.parse(non_pr_payload)
+      repo = create(
+        :repo,
+        :active,
+        github_id: parsed_payload["repository"]["id"],
+        full_github_name: parsed_payload["repository"]["full_name"]
+      )
+
+      page.driver.post builds_path, non_pr_payload
+
+      expect(repo.builds).to be_blank
+    end
+  end
+
   scenario 'a failed build' do
     create(:repo, :active, github_id: repo_id, full_github_name: repo_name)
     stub_github_requests_with_violations
--- a/spec/features/builds_spec.rb
+++ b/spec/features/builds_spec.rb
@@ -37,7 +37,7 @@ feature 'Builds' do
   context "with non PR payload" do
     scenario "does not run build" do
       non_pr_payload = File.read(
-        "spec/support/fixtures/pull_request_opened_event.json"
+        "spec/support/fixtures/push_event.json"
       )
       parsed_payload = JSON.parse(non_pr_payload)
       repo = create(
--- a/spec/features/builds_spec.rb
+++ b/spec/features/builds_spec.rb
@@ -49,7 +49,7 @@ feature 'Builds' do
 
       page.driver.post builds_path, non_pr_payload
 
-      expect(repo.builds).to be_blank
+      expect(repo.builds).to eq([])
     end
   end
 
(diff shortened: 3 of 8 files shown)


spec/jobs/buildable_spec.rb
```diff
@@ -33,6 +33,22 @@ class TestJob
 
       expect(Resque).to have_received(:enqueue).with(TestJob, "payload")
     end
+
+    it "will not run build runner if no PR" do
+      stub_const("Owner", "foo")
+      build_runner = double(:build_runner, run: nil)
```
teoljungberg: Prefer strings to symbols in mocks

teoljungberg:
1 minor comment, looks good to merge

gylaz:
This seems to be replicating the work in https://github.com/thoughtbot/hound/blob/master/app/services/build_runner.rb#L5

spec/jobs/buildable_spec.rb
```diff
@@ -33,6 +33,22 @@ class TestJob
 
       expect(Resque).to have_received(:enqueue).with(TestJob, "payload")
     end
+
+    it "will not run build runner if no PR" do
```
teoljungberg: We're using `it "does not run ...."`, so I think it's perferable to keep the grammar the same

arunagw:
@gylaz I initially thought that it's good to not even go into `BuildRunner` if payload is invalid but as that check already exists inside `BuildRunner` I will just write a feature spec to make sure that invalid hook will not raise anything.

spec/models/payload_spec.rb
```diff
@@ -1,8 +1,30 @@
 require "fast_spec_helper"
 require "attr_extras"
 require "app/models/payload"
+require "lib/github_api"
```
teoljungberg: Why does this need to be required?

spec/features/builds_spec.rb
```diff
@@ -34,6 +34,25 @@
...
+      repo = create(
+        :repo,
+        :active,
+        github_id: parsed_payload["repository"]["id"],
+        full_github_name: parsed_payload["repository"]["full_name"]
+      )
+
+      page.driver.post builds_path, non_pr_payload
+
+      expect(repo.builds).to be_blank
```
teoljungberg: isn't `empty?` better to check against?

spec/services/build_runner_spec.rb
```diff
@@ -232,7 +251,8 @@ def stubbed_pull_request
       :pull_request,
       pull_request_files: [double(:file)],
       config: double(:config),
-      opened?: true
+      opened?: true,
+      can_be_reviewed?: true
```
teoljungberg: Maybe add a trailing comma to avoid this kind of diff again?

spec/models/payload_spec.rb
arunagw: Because we are using `GithubApi::ORGANIZATION_TYPE` in `payload.rb`

app/models/payload.rb
```diff
@@ -45,6 +45,10 @@ def repository_owner_is_organization?
     repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
+  def can_be_reviewed?
+    !pull_request.empty?
+  end
```
gylaz: I'm curious if we've seen situations where `pull_request` is empty.

spec/services/build_runner_spec.rb
```diff
@@ -1,6 +1,25 @@
...
+  context "with non pr payload" do
+    it "does not run" do
+      repo = create(:repo, :active, github_id: 123)
+      payload = stubbed_payload(
+        github_repo_id: repo.github_id,
+        full_repo_name: repo.name
+      )
+      pull_request = stubbed_pull_request
+      allow(pull_request).
+        to receive_messages(can_be_reviewed?: false)
```
gylaz: Just `to receive(`?

spec/models/payload_spec.rb
```diff
@@ -1,8 +1,31 @@
...
 
 describe Payload do
+  describe "#can_be_reviewed?" do
+    context "with pull_request data" do
+      it "returns true" do
+        fixture_file = "spec/support/fixtures/pull_request_opened_event.json"
+        payload_json = File.read(fixture_file)
+        payload = Payload.new(payload_json)
+
+        expect(payload.can_be_reviewed?).to be_truthy
```
gylaz: Why not just explicitly `true` or false?

spec/models/payload_spec.rb
```diff
@@ -1,8 +1,31 @@
...
+      end
+    end
+
+    context "with no pull_request data" do
+      it "returns false" do
+        fixture_file = "spec/support/fixtures/push_event.json"
+        payload_json = File.read(fixture_file)
+        payload = Payload.new(payload_json)
+
+        expect(payload.can_be_reviewed?).to be_falsy
```
gylaz: Would it make sense to rename this method to `payload.reviewable?`, then this can be `expect(payload).to be_reviewable`.

app/models/payload.rb
```diff
@@ -45,6 +45,10 @@ def repository_owner_is_organization?
     repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
+  def can_be_reviewed?
```
gylaz: It seems odd that payload can be reviewed. Maybe `pull_request_reviewable?`?

app/services/build_runner.rb
```diff
@@ -20,7 +20,9 @@ def run
   private
 
   def relevant_pull_request?
-    pull_request.opened? || pull_request.synchronize?
+    pull_request.can_be_reviewed? && (
```
gylaz: I'm not sure that we need this check, since we only create a hook for PR events. We shouldn't be receiving any others.

spec/services/build_runner_spec.rb
```diff
@@ -1,6 +1,25 @@
 require 'spec_helper'
 
 describe BuildRunner, '#run' do
+  context "with non pr payload" do
+    it "does not run" do
```
gylaz: Maybe `does not create a build`?

app/models/payload.rb
arunagw: Yes. There is one error mentioned here https://trello.com/c/0j5NoBwH

gylaz:
This doesn't seem to be fixing anything that isn't already accounted for. We don't actually run builds for non-relevant PRs already.

srt32:
The intention of the card was to have a test case covering the reception of a GH hook that should not kick off a build that would have caught the issue (https://app.getsentry.com/hound-1/production/group/54752935/) that got raised when we deployed the `owners` update.

app/models/payload.rb
```diff
@@ -45,6 +45,10 @@ def repository_owner_is_organization?
     repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
+  def pull_request_attached?
```
teoljungberg: How about renaming it `pull_request?`?

gylaz:
To expand: `relevant_pull_request?` already did the check necessary for a valid PR. The issue was that we were assuming the `Payload` was a PR payload in `Buildable#perfrom`, in which case we haven't scheduled the job yet. I think what we may want to do here is not to even schedule a job if the Payload isn't PR.

gylaz:
Made a quick PR, to address this: https://github.com/thoughtbot/hound/pull/661

salbertson:
Thanks @arunagw, looks like @gylaz's covers the same thing.

Issue #627, opened 2015-02-12
Idea: Ignore `vendor/` directory by default?


gylaz:
In Rails, `vendor/` directory is usually third-party code. Should we just ignore this by default?

teoljungberg:
I'm all for it, but there is always going to be edge-cases. How about making it configurable and ignoring it default?

srt32:
:+1: to ignoring it by default with the ability to override the setting.

salbertson:
What would the override look like in the config?

teoljungberg:
@salbertson maybe something like:

```yaml
ignored_directories:
- vendor
# ...
```

salbertson:
@teoljungberg, we can ignore `vendor` if we add this to `config/style_guides/ruby.yml`.

```
AllCops:
  Exclude:
    - 'vendor/**/*'
    - db/schema.rb
```

I was asking about what the override would look like, but I realized users can simply add `Exclude:` to their own config.

salbertson:
@gylaz, sounds good to me. :+1:

teoljungberg:
@salbertson I didn't know you could do that, that's handy!

Then we have covered the case of ruby, but maybe we want to support ignoring `vendor` for node and JS applications aswell?

gylaz:
Ignoring `vendor/` via RuboCop only solves a small part of the issue -- since rarely ruby code ends up in `vendor/`. Ignoring Coffee, JS, and SCSS from `vendor/` would be much more useful.

teoljungberg:
Closed by 396d381355340e368756459c464a0327f522f5f8

PR #649, opened 2015-02-17
Ignore certain directories by default

teoljungberg:
- Ignore files under `vendor` per default.
- Users can override which directories they want ignored by configuring it in
  their `hound.yml`:

  ```yaml
  # ...
  # ignore `tmp` and `bin` instead of the default `vendor`
  ignored_paths:
    - tmp/**
    - bin/**
  # ...
  ```

https://trello.com/c/1Uhljes3
References #627

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -8,6 +8,7 @@ class RepoConfig
     "coffee_script" => "json",
     "scss" => "yaml",
   }
+  DEFAULT_IGNORED_DIRECTORIES = ["vendor"]
 
   pattr_initialize :commit
 
@@ -38,6 +39,11 @@ class RepoConfig
     else
       []
     end
+  end
+
+  def ignored_directories
+    remote_hound_config = load_file(HOUND_CONFIG, "yaml")
+    remote_hound_config["ignored_directories"] || DEFAULT_IGNORED_DIRECTORIES
   end
 
   private
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -42,8 +42,7 @@ class RepoConfig
   end
 
   def ignored_directories
-    remote_hound_config = load_file(HOUND_CONFIG, "yaml")
-    remote_hound_config["ignored_directories"] || DEFAULT_IGNORED_DIRECTORIES
+    hound_config["ignored_directories"] || DEFAULT_IGNORED_DIRECTORIES
   end
 
   private
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -8,7 +8,7 @@ class RepoConfig
     "coffee_script" => "json",
     "scss" => "yaml",
   }
-  DEFAULT_IGNORED_DIRECTORIES = ["vendor"]
+  DEFAULT_IGNORED_DIRECTORIES = ["vendor/**"]
 
   pattr_initialize :commit
 
@@ -41,8 +41,8 @@ class RepoConfig
     end
   end
 
-  def ignored_directories
-    hound_config["ignored_directories"] || DEFAULT_IGNORED_DIRECTORIES
+  def ignored_paths
+    hound_config["ignored_paths"] || DEFAULT_IGNORED_DIRECTORIES
   end
 
   private
--- a/app/models/style_guide/base.rb
+++ b/app/models/style_guide/base.rb
@@ -16,8 +16,8 @@ module StyleGuide
     private
 
     def directory_excluded?(file)
-      repo_config.ignored_directories.any? do |directory|
-        File.fnmatch?("#{directory}/*", file.filename)
+      repo_config.ignored_paths.any? do |pattern|
+        File.fnmatch?(pattern, file.filename)
       end
     end
 
--- a/app/models/style_guide/base.rb
+++ b/app/models/style_guide/base.rb
@@ -15,6 +15,12 @@ module StyleGuide
 
     private
 
+    def directory_not_excluded?(file)
+      !repo_config.ignored_directories.any? do |directory|
+        file.filename.start_with? directory
+      end
+    end
+
     def name
       self.class.name.demodulize.underscore
     end
--- a/app/models/style_guide/base.rb
+++ b/app/models/style_guide/base.rb
@@ -17,7 +17,7 @@ module StyleGuide
 
     def directory_excluded?(file)
       repo_config.ignored_directories.any? do |directory|
-        file.filename.start_with? directory
+        File.fnmatch?("#{directory}/*", file.filename)
       end
     end
 
--- a/app/models/style_guide/base.rb
+++ b/app/models/style_guide/base.rb
@@ -15,8 +15,8 @@ module StyleGuide
 
     private
 
-    def directory_not_excluded?(file)
-      !repo_config.ignored_directories.any? do |directory|
+    def directory_excluded?(file)
+      repo_config.ignored_directories.any? do |directory|
         file.filename.start_with? directory
       end
     end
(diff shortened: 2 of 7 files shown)


arunagw:
Looks fine to me. But need one more review before merging.

:+1:

app/models/repo_config.rb
```diff
@@ -40,6 +41,11 @@ def ignored_javascript_files
     end
   end
 
+  def ignored_directories
+    remote_hound_config = load_file(HOUND_CONFIG, "yaml")
```
gylaz: Any reason we are not using `hound_config` method?

app/models/style_guide/base.rb
```diff
@@ -17,6 +17,12 @@ def file_included?(_file)
 
     private
 
+    def directory_excluded?(file)
+      repo_config.ignored_directories.any? do |directory|
+        file.filename.start_with? directory
```
gylaz: Thoughts on using `File.fnmatch?` here?

app/models/style_guide/scss.rb
```diff
@@ -18,8 +18,7 @@ def violations_in_file(file)
             filename: file.filename,
             line: line,
             line_number: violation.location.line,
-            messages: [violation.description],
-            patch_position: line.patch_position,
+            messages: [violation.description], patch_position: line.patch_position,
```
gylaz: Split up?

spec/models/repo_config_spec.rb
```diff
@@ -306,6 +306,40 @@
       end
     end
 
+    describe "#ignored_directories" do
+      it "ignores `vendor` by default" do
+        hound_config = <<-EOS
+            java_script:
+              enabled: true
+              ignore_file: ".js_ignore"
+        EOS
```
gylaz: Indentation is too much

spec/models/repo_config_spec.rb
```diff
@@ -306,6 +306,40 @@
...
+
+        expect(repo_config.ignored_directories).to eq ["vendor"]
+      end
+
+      it "can be configured to allow `vendor` by the repository's config" do
+        hound_config = <<-EOS
+            java_script:
+              enabled: true
+              ignore_file: ".js_ignore"
+        EOS
```
gylaz: Indentation is too much

spec/models/repo_config_spec.rb
```diff
@@ -306,6 +306,40 @@
       end
     end
 
+    describe "#ignored_directories" do
+      it "ignores `vendor` by default" do
```
gylaz: How about `it "returns `vendor` by default" do`? It's nice to specify what this function does in the doc string.

spec/models/repo_config_spec.rb
```diff
@@ -306,6 +306,40 @@
...
+              enabled: true
+              ignore_file: ".js_ignore"
+        EOS
+        commit = stub_commit(hound_config: hound_config)
+        repo_config = RepoConfig.new(commit)
+
+        expect(repo_config.ignored_directories).to eq ["vendor"]
+      end
+
+      it "can be configured to allow `vendor` by the repository's config" do
```
gylaz: How about providing the condition in a `context`?
```rb
context "when configured by the repo's config" do
  it "returns configured directories" do
```

spec/models/repo_config_spec.rb
```diff
@@ -306,6 +306,40 @@
...
+              ignore_file: ".js_ignore"
+        EOS
+
+        repo_hound_config = <<-EOS
+            ignored_directories:
+              - tmp
+        EOS
+        commit = stub_commit(
+          hound_config: hound_config,
+          ".hound.yml" => repo_hound_config,
```
gylaz: I'm confused by this setup. These two files are the same. We should just modify `hound_config` variable to contain the ignored dirs.

spec/models/style_guide/scss_spec.rb
```diff
@@ -79,6 +79,40 @@
...
+
+      context "when a file is ignored" do
+        it "returns no violations" do
+          style_guide = build_style_guide
+          bad_content = ".a { .b { .c { background: #000; } } }"
+          file = build_file(bad_content)
+          config_double = double(
+            "SassConfig",
+            excluded_file?: true
+          )
```
gylaz: Fit on one line?

app/models/style_guide/base.rb
teoljungberg: `File.fnmatch?` doesn't match for a small subset of a string, like a directory that's in the beginning of a file path.
So `File.fnmatch?("foo', "foo/bar/baz/bax.js)"` won't work

app/models/style_guide/scss.rb
teoljungberg: That's probably a finger that shifted on `J`, thanks!

app/models/style_guide/base.rb
gylaz: But:
```rb
File.fnmatch?("foo/*", "foo/bar/baz/bax.js") # => true
# and 
File.fnmatch?("foo/**", "foo/bar/baz/bax.js") # => true
```
I'm only suggesting that we might want to favor the directory pattern matching, rather than strict directory names.

teoljungberg: I agree

app/models/style_guide/base.rb
```diff
@@ -17,6 +17,12 @@ def file_included?(_file)
 
     private
 
+    def directory_excluded?(file)
+      repo_config.ignored_directories.any? do |directory|
+        File.fnmatch?("#{directory}/*", file.filename)
```
gylaz: I guess my thought was to actually tell people to configure these directories like:
```
ignored_paths:
  - vendor/**
  - app/**/images
```
What do you think?

gylaz: Basically, the idea is to follow what RuboCop is doing.

teoljungberg: I like it a lot! More than hardly basing the ignoring around directories.

salbertson:
I don't think we should do this in code. If JSHint doesn't support ignoring directories, which it [looks like it does](https://groups.google.com/forum/#!msg/jshint/62N12QUhBCs/v9a1XC7I_j8J), then we should think about supporting a custom option in the JavaScript config. I believe we did something similar with `ShowCopName` in RuboCop before the option was officially supported.

gylaz:
We already have a way to configure ignored directories from JSHint: https://github.com/thoughtbot/hound/blob/master/app/models/style_guide/java_script.rb#L33

I think the original idea was to ignore directories like `tmp/`, and `vendor/` by default. I do like having one way to tell Hound to ignore directories, and not ignoring them on a per-linter basis. Do you see any issues with this @salbertson? Is there a situation where a user might want to ignore a directory for Ruby, but not for JS?

salbertson:
@gylaz, I'm not sure but I would rather avoid additional code and keep flexibility. I would also rather not add framework specific code, slippery slope.

salbertson:
@teoljungberg, closing this in favor of using configuration to ignore files and directories.

teoljungberg:
I'm with @gylaz on this, I prefer having this in code than configured per linter. What happends when a newly supported language's linter doesn't have an exclude feature?

salbertson:
@teoljungberg, we should add the support to the linter so it's configurable.

When do we stop adding directories if this is done in code? It feels too closely tied to Rails and our typical code bases. We plan on supporting Python and other languages which might not have `vendor` or other directories.

salbertson:
@gylaz @teoljungberg is there a library we currently use that doesn't allow users to exclude directories or files?

teoljungberg:
> When do we stop adding directories if this is done in code? It feels too closely tied to Rails and our typical code bases. We plan on supporting Python and other languages which might not have vendor or other directories.

That's why I made it so these settings can be configured. But if the libraries allows to ignore the directories needed - I'm all for it

salbertson:
Sounds good, thanks @teoljungberg.

PR #652, opened 2015-02-18
Add `HOST` environment variable

teoljungberg:
We're depending on `HOST` to be set in: `app/services/repo_activator.rb`
The default value is the standard url and port for `rails s`.


.sample.env
```diff
@@ -2,10 +2,11 @@ CHANGED_FILES_THRESHOLD=300
 ENABLE_HTTPS=no
 GITHUB_CLIENT_ID=github_client_id
 GITHUB_CLIENT_SECRET=github_client_secret
+HOST=0.0.0.0:3000
```
gylaz: Is this better than just `localhost`?

teoljungberg: No idea, I typed it out out of habbit

salbertson: `localhost:5000` for Foreman?

salbertson:
@teoljungberg, one comment, looks good. :+1:

PR #653, opened 2015-02-19
Blacklist public repos on demand

gylaz:
When we are having problems (like hitting our rate limit) we should have a way
to favor paying customers over free.

Setting `ENV["IGNORE_PUBLIC_REPOS"]` allows us to ignore builds for public
repos.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -2,7 +2,7 @@ class BuildRunner
   pattr_initialize :payload
 
   def run
-    if repo && relevant_pull_request?
+    if repo && relevant_pull_request? && !repo_blacklisted?
       track_subscribed_build_started
       create_pending_status
       repo.builds.create!(
@@ -90,5 +90,13 @@ class BuildRunner
 
   def github
     @github ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
+  end
+
+  def repo_blacklisted?
+    if ENV["IGNORE_PUBLIC_REPOS"] == "true"
+      !payload.private_repo?
+    else
+      false
+    end
   end
 end
--- a/app/models/payload.rb
+++ b/app/models/payload.rb
@@ -45,6 +45,10 @@ class Payload
     repository["owner"]["type"] == GithubApi::ORGANIZATION_TYPE
   end
 
+  def private_repo?
+    repository["private"]
+  end
+
   private
 
   def parse_data
--- a/spec/services/build_runner_spec.rb
+++ b/spec/services/build_runner_spec.rb
@@ -42,6 +42,7 @@ describe BuildRunner, '#run' do
     end
 
     it "comments a maximum number of times" do
+      allow(ENV).to receive(:[])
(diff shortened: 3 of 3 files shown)


app/services/build_runner.rb
```diff
@@ -91,4 +91,12 @@ def upsert_owner
   def github
     @github ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
   end
+
+  def repo_blacklisted?
+    if ENV["IGNORE_PUBLIC_REPOS"] == "true"
+      !payload.private_repo?
```
srt32: Would it be worth explicitly looking for a `subscription` as opposed to `private`? As of now `private` means there must be  a `subscription` so the current implementation is fine but in reality we want to prefer paid `repo`s, right?

salbertson:
@gylaz should we close this since our requests are somewhat under control now and we don't have long running jobs causing trouble?

gylaz:
@salbertson It's a small code change, and in case we have more fires we can push this button to make sure our paying customers are favored over free.

salbertson:
@gylaz, I'd say leave it out until we need it. I'll leave it up to you, follow your heart.

app/services/build_runner.rb
```diff
@@ -91,4 +91,12 @@ def upsert_owner
   def github
     @github ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
   end
+
+  def repo_blacklisted?
+    if ENV["IGNORE_PUBLIC_REPOS"] == "true"
```
teoljungberg: Can't we just follow the unix convention of checking the existance of the environment variable instead of it's contents?

gylaz: We can't. Because `ENV` variables come back as strings.
```rb
"false" == false # => false
```

app/services/build_runner.rb
gylaz: Yes, I think this makes sense from the correctness stand-point.

gylaz:
After talking IRL, we decided that it might make more sense to create a `free` queue, and put the free repo jobs there. In case of a fire, we can always blow away (or pause?) that queue, or scale down the workers, to prefer only the paying customers. Thoughts on this?

app/services/build_runner.rb
teoljungberg: That I know, I meant more that we set the environment variable when we need it and then remove it

teoljungberg:
I'm luke-warm to the blacklisting approach, I must prefer splitting up our queue into a higher prioritized one for paying customers and a lower prioritized one for public repositories.

croaky:
"free" queue sounds good.

app/services/build_runner.rb
gylaz: What advantage do you see to using `if ENV["IGNORE_PUBLIC_REPOS"]`? The advantage that I see to checking for `"true"` explicitly, is that we don't rely on someone removing `IGNORE_PUBLIC_REPOS`. Since there is no convention around this, or documentation, somewhere down the road someone can likely flip this to `"false"`.

salbertson:
:+1:

PR #654, opened 2015-02-19
Reduce the payload data for build jobs

gylaz:
We only need a subset of payload data when the jobs run. To reduce the amount of
memory a job takes up, pass a stripped down version of the payload.


salbertson:
:+1:

gylaz:
Merged as 937ad55f94e4c46beeab864c4ea44cabb82f3437.

PR #657, opened 2015-02-19
Remove responders gem

arunagw:
* We can set defaults in routes so we can reject other requests from routes
level
* `responders` is not a core rails part


config/routes.rb
```diff
@@ -9,16 +9,16 @@
 
   resource :account, only: [:show]
   resources :builds, only: [:create]
-  resource :credit_card, only: [:update]
+  resource :credit_card, only: [:update], defaults: { format: :json }
```
srt32: Would there be a non-namespace breaking way to wrap these json endpoints into a block with a single call to `defaults: { format: :json }`?

Koronen: [`Object#with_options`](http://apidock.com/rails/Object/with_options) maybe?

gylaz:
I thought responders where the recommended way with Rails. Did that change? pinging @sgrif

sgrif:
There's no official recommendation either way, we just moved `respond_with` into the responders gem because DHH doesn't like the feature and didn't want it in Rails.

salbertson:
Looks good. :+1:

config/routes.rb
```diff
@@ -9,16 +9,20 @@
 
   resource :account, only: [:show]
   resources :builds, only: [:create]
-  resource :credit_card, only: [:update]
 
   resources :repos, only: [:index] do
-    resource :activation, only: [:create]
-    resource :deactivation, only: [:create]
-    resource :subscription, only: [:create, :destroy]
+    with_options( defaults: { format: :json } ) do
```
teoljungberg: Remove the spaces around `(`

PR #658, opened 2015-02-19
Pass the parameters in the same order as the api client

teoljungberg:


arunagw:
We were sending wrong parameters from here https://github.com/thoughtbot/hound/blob/master/app/models/github_team.rb#L18

But now they are flipped should be good

gylaz:
:+1:

salbertson:
:+1:

salbertson:
Restarted build.

teoljungberg:
Closing in favor of #659

PR #659, opened 2015-02-19
Deactivation cleanup deux

teoljungberg:
- Re enable "deactivation cleanup"
- Pass the parameters to Octokit in the same order to avoid confusion
- Return `true` while deactivation the repo to avoid returning `nil` so we
don't break anything in the callback


arunagw:
Things are fine on staging on this branch.

PR #661, opened 2015-02-20
Ignore non-pr payloads

gylaz:
Don't schedule jobs for payloads that are not for a pull request
(e.g. commit push). No reason to waste the backround job round trip just to
do nothing.

### DIFF
--- a/spec/models/payload_spec.rb
+++ b/spec/models/payload_spec.rb
@@ -1,6 +1,7 @@
 require "fast_spec_helper"
 require "attr_extras"
 require "app/models/payload"
+require "lib/github_api"
 
 describe Payload do
   describe '#changed_files' do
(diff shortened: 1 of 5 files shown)


spec/models/payload_spec.rb
```diff
@@ -52,6 +53,28 @@
...
+      it "returns false" do
+        push_event = File.read("spec/support/fixtures/push_event.json")
+        payload = Payload.new(push_event)
+
+        expect(payload).not_to be_pull_request
+      end
+    end
+
+    context "when payload for pull request" do
+      it "returns false" do
```
thorncp: `false` -> `true`

thorncp:
Does this mean hound will no longer run when new commits are added to an open pull request?

croaky:
So this checks that it is a pull request, but we have other code that checks the `synchronize` event later, right? https://developer.github.com/v3/activity/events/types/#pullrequestevent

gylaz:
@thorncp No, that is considered a pull request event with an "synchronize" action. This filters our the push events to a branch, that's not been PR'd.

@croaky Yep. [`BuildRunner#run`](https://github.com/thoughtbot/hound/blob/master/app/services/build_runner.rb#L5) has a check for `opened` and `synchronize` type actions.

salbertson:
:+1:

PR #667, opened 2015-02-24
Add base Ruby style guide

salbertson:
This change uses a complete configuration file for Ruby, which includes what is merged by default and a few changes specific to Hound.


gylaz:
:+1:

PR #668, opened 2015-02-24
Exclude vendor when reviewing Ruby

salbertson:


gylaz:
:+1:

teoljungberg:
I haven't seen a ruby file in vendor in years, but we see plenty of `js` and `css` files. How can we handle those?

PR #669, opened 2015-02-24
Include links to full configuration files and reword documentation

salbertson:
This should make it very clear which style guide preferences are used by default and how users can override or disable the defaults.

screenshot 2015-02-23 21 26 52

### DIFF
--- a/app/views/pages/configuration.haml
+++ b/app/views/pages/configuration.haml
@@ -3,34 +3,30 @@
 %main.content-doc
   %h2 Configuration
 
-  %p
-    Hound uses the
-    #{link_to "thoughtbot style guide", "https://github.com/thoughtbot/guides/tree/master/style"}
-    by default, unless you configure it to use your own style guide.
-
-  %p
-    Configure Hound for Ruby, CoffeeScript, JavaScript, and SCSS
-    by following these instructions.
-
   %h3 Ruby
 
   %p
-    Hound uses #{link_to "RuboCop", "https://github.com/bbatsov/rubocop"}
-    internally so you can configure Hound by adding a
-    #{link_to "RuboCop config", "https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml"}
-    to your project and adding the following to
+    Hound uses
+    #{link_to "RuboCop", "https://github.com/bbatsov/rubocop", target: :blank}
+    with this
+    #{link_to "config", ruby_config_url, target: :blank}
+    by default.
+
+  %p
+    If you need to change the way Hound is configured, simply copy the
+    #{link_to "default config", ruby_config_url, target: :blank}
+    into your project, make changes and reference the file in your
     %em.code .hound.yml
-    in the root of your project.
 
     %code.code-block
       :preserve
         ruby:
-          config_file: config/.rubocop.yml
+          config_file: .ruby-style.yml
 
   %p
-    To disable Ruby style checking, add the following to
+    Add the following code to your
     %em.code .hound.yml
-    in the root of your project
+    to disable Ruby style checking.
 
     %code.code-block
       :preserve
@@ -40,22 +36,26 @@
   %h3 CoffeeScript
 
   %p
-    Hound uses #{link_to "CoffeeLint", "http://www.coffeelint.org/"}
-    internally so you can configure Hound by adding a
-    #{link_to "CoffeeLint config", "https://github.com/clutchski/coffeelint/blob/master/coffeelint.json"}
-    to your project and adding the following to
+    Hound uses
+    #{link_to "CoffeeLint", "http://www.coffeelint.org", target: :blank}
+    with this
+    #{link_to "config", coffeescript_config_url, target: :blank}.
+
+  %p
+    If you need to change the way Hound is configured, simply copy the
+    #{link_to "default config", coffeescript_config_url, target: :blank}.
+    into your project, make changes and reference the file in your
     %em.code .hound.yml
-    in the root of your project.
 
     %code.code-block
       :preserve
         coffee_script:
-          config_file: config/.coffeelint.json
+          config_file: .coffeescript-style.json
 
   %p
-    To disable CoffeeScript style checking, add the following to
+    Add the following code to your
     %em.code .hound.yml
-    in the root of your project
+    to disable CoffeeScript style checking.
 
     %code.code-block
       :preserve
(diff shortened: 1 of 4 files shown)


croaky:
Should link to `github.com/thoughtbot/hound`, not `gist.github.com/salbertson`?

salbertson:
@croaky, good idea, done.

app/views/pages/configuration.haml
```diff
@@ -3,34 +3,33 @@
...
+    Hound uses
+    #{link_to "RuboCop", "https://github.com/bbatsov/rubocop", target: :blank}
+    with this
+    #{link_to "configuration", ".ruby-style.yml", target: :blank}.
 
-  %h3 Ruby
+  %p
+    If you need to configure Hound differently, simply copy the default file
+    into your project and name it
+    %em.code .ruby-style.yml
```
croaky: This reads like the user needs to name the file this way, but they can name it whatever they want as long as they reference it correctly in `.hound.yml`?

app/views/pages/configuration.haml
```diff
@@ -113,24 +125,35 @@
...
-    in the root of your project
+    in the root of your project.
 
     %code.code-block
       :preserve
         scss:
           enabled: false
+
+  %p
+    Please tweet at us #{link_to "@houndci", "https://twitter.com/houndci", target: :blank} or email us at #{link_to "hound@thoughtbot.com", "mailto:hound@thoughtbot.com"} if you need a hand.
```
croaky: `mail_to`?

app/views/pages/configuration.haml
```diff
@@ -40,22 +39,30 @@
...
-    to your project and adding the following to
+    Hound uses
+    #{link_to "CoffeeLint", "http://www.coffeelint.org", target: :blank}
+    with this
+    #{link_to "configuration", ".coffeescript-style.json", target: :blank}.
+
+  %p
+    If you need to configure Hound differently, simply copy the default file
+    into your project and name it
+    %em.code .coffeescript-style.json
```
croaky: Should we go one step further with `curl` command instead of saying "copy" in prose?

app/views/pages/configuration.haml
```diff
@@ -40,22 +39,30 @@
...
+
+  %p
+    You also need to add a
     %em.code .hound.yml
-    in the root of your project.
+    to the root of your project and reference the config.
```
croaky: Same thought here: `curl` into `.hound.yml`?

app/views/pages/configuration.haml
salbertson: @croaky, how about this?

screenshot 2015-02-24 08 02 12

app/views/pages/configuration.haml
salbertson: Good idea, done.

app/views/pages/configuration.haml
salbertson: I think that would be great. What do you think about doing that in a follow up PR?

salbertson:
@croaky, what do you think about the changes?

app/views/pages/configuration.haml
```diff
@@ -3,34 +3,33 @@
...
-    by default, unless you configure it to use your own style guide.
+  %h3 Ruby
 
   %p
-    Configure Hound for Ruby, CoffeeScript, JavaScript, and SCSS
-    by following these instructions.
+    Hound uses
+    #{link_to "RuboCop", "https://github.com/bbatsov/rubocop", target: :blank}
+    with this
+    #{link_to "configuration", ".ruby-style.yml", target: :blank}
```
gylaz: I think we want this to be a link to Hound's full default config. I'm assuming that being `https://github.com/thoughtbot/hound/blob/master/config/style_guides/ruby.yml` once we make it contain the full configuration.

app/views/pages/configuration.haml
```diff
@@ -3,34 +3,33 @@
...
+    by default.
 
-  %h3 Ruby
+  %p
+    If you need to configure Hound differently, simply copy the default
+    configuration into your project.
```
gylaz: How about: `... into your project, and make desired changes.`

gylaz: Maybe we can provide a code block example for `curl`:
```
curl -O https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/ruby.yml
```

app/views/pages/configuration.haml
```diff
@@ -3,34 +3,33 @@
...
 
   %p
-    Hound uses #{link_to "RuboCop", "https://github.com/bbatsov/rubocop"}
-    internally so you can configure Hound by adding a
-    #{link_to "RuboCop config", "https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml"}
-    to your project and adding the following to
+    The file can be named anything, just make sure to reference it from
```
gylaz: I would almost would leave the `file can be named anything` out and make the users follow the examples. We eventually want to drop `.hound.yml` and if we setup a convention early (name the file `.ruby-style.yml` or `javascript-style.yml`), then it will only benefit us later.

app/views/pages/configuration.haml
```diff
@@ -65,45 +72,50 @@
...
-    Each line in the
+    Each line in
     %em.code .javascript_ignore
-    file can specify a pattern for file names to ignore.
+    should specify a pattern for files to ignore.
 
     %code.code-block
       :preserve
-        vendor/assets/javascripts/**.js
+        vendor/**/*
```
gylaz: Just `vendor/*` is enough.

gylaz:
@salbertson How about moving the config files you generated to `config/style_guides`, so those are the master blue-prints, then we can open them up via routes?

app/views/pages/configuration.haml
teoljungberg: I agree with ^

app/views/pages/configuration.haml
teoljungberg: I would prefer a link to the default configuration instead of a `curl` command

app/views/pages/configuration.haml
teoljungberg: I would skip this entirely, our target group knows this already

teoljungberg:
Why are the sample config files in public? I would prefer, as @gylaz said, to setup routes for the ones we have in `config/style_guides`

gylaz:
I'm cool with doing a follow up PR to address the discussions here.

app/views/pages/configuration.haml
salbertson: Good idea, done.

app/views/pages/configuration.haml
salbertson: I removed it, good call.

app/views/pages/configuration.haml
salbertson: I'm linking to our config now, so I removed this.

salbertson:
@gylaz @teoljungberg one last review?

app/views/pages/configuration.haml
```diff
@@ -3,34 +3,30 @@
...
+    #{link_to "config", "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/ruby.yml", target: :blank}
+    by default.
 
   %p
-    Hound uses #{link_to "RuboCop", "https://github.com/bbatsov/rubocop"}
-    internally so you can configure Hound by adding a
-    #{link_to "RuboCop config", "https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml"}
-    to your project and adding the following to
+    If you need to change the way Hound is configured, simply copy the
+    #{link_to "default config", "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/ruby.yml", target: :blank}
```
gylaz: Does it make sense to extract this into a view helper to avoid duplication and to make it easier to change in the future?

gylaz:
One comment about extracting the urls. Looks good to me.

app/views/pages/configuration.haml
salbertson: Works for me, done.

PR #671, opened 2015-02-24
Swallow erb exceptions when parsing `coffee.erb`

teoljungberg:
This happens when the file refers to variables or methods outside of
it's scope.

https://app.getsentry.com/hound-1/production/group/55884035/

### DIFF
--- a/app/models/style_guide/coffee_script.rb
+++ b/app/models/style_guide/coffee_script.rb
@@ -26,7 +26,11 @@ module StyleGuide
 
     def content_for_file(file)
       if erb? file
-        ERB.new(file.content).result
+        begin
+          ERB.new(file.content).result
+        rescue
+          ""
+        end
       else
         file.content
       end
--- a/spec/models/style_guide/coffee_script_spec.rb
+++ b/spec/models/style_guide/coffee_script_spec.rb
@@ -125,6 +125,24 @@ describe StyleGuide::CoffeeScript do
           ["Line exceeds maximum allowed length"]
         )
       end
+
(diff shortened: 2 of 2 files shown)


arunagw:
Looks fine to me.. Can be merged once travis is green.

app/models/style_guide/coffee_script.rb
```diff
@@ -26,7 +26,11 @@ def lint(content)
 
     def content_for_file(file)
       if erb? file
-        ERB.new(file.content).result
+        begin
+          ERB.new(file.content).result
+        rescue
```
gylaz: Is there a specific error we can rescue instead of just `StandardError`? Do we care?

salbertson: @teoljungberg, is there an exception in Sentry to led to this?

teoljungberg: @salbertson - if you look at the PR message the exception in linked there.

@gylaz - probably, any ideas?

gylaz: My thought is to just rescue `NoMethodError` and see if that is good enough. The thought here is, if there are other errors this code is raising, then we may want to fix them instead of silently ignoring them.

salbertson:
Thanks @teoljungberg, saw that after I commented, of course.

app/models/style_guide/coffee_script.rb
teoljungberg: I'll rescue from `NameError` which includes both undefined methods and variables

PR #672, opened 2015-02-24
Upgrade scss lint

arunagw:
* Locked to specific version to avoid upgrade problems
* Update configs from scss-lint default file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ gem "resque-sentry"
 gem "resque-timeout"
 gem "rubocop", "0.29.1"
 gem "sass-rails"
-gem "scss-lint", require: false
+gem "scss-lint", "0.34.0", require: false
 gem "sentry-raven"
 gem "stripe"
 gem "uglifier", ">= 1.0.3"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -276,7 +276,7 @@ GEM
     rufus-scheduler (3.0.9)
       tzinfo
     safe_yaml (1.0.4)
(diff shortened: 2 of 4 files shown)


gylaz:
Looks good. Does this fix the `bubbles?` problem?

salbertson:
@gylaz the build is failing, so I'm guessing no.

teoljungberg:
`bubbles?` is sadly still there, but it's nice to keep the gem up to date anyhow

Gemfile
```diff
@@ -28,7 +28,7 @@ gem "resque-sentry"
 gem "resque-timeout"
 gem "rubocop", "0.29.1"
 gem "sass-rails"
-gem "scss-lint", require: false
+gem "scss-lint", "0.34.0", require: false
```
gylaz: What is the reason to lock this explicitly? Shouldn't we just monitor `Gemfile.lock` changes in pull requests to make sure the version doesn't get bumped there?

teoljungberg: There was a breakage between `0.31.0` and `0.34.0` that broke the gem unless you added the new configurations. So to clarify that this gem might be risky to upgrade we added the lock in

croaky:
Looks good.

gylaz:
This should help us with the background jobs timing out. It looked like we were hitting an infinite loop situation which was fixed in `scss-lint` recently: https://github.com/causes/scss-lint/commit/d8be7c5125e68b963111a538c162b425cddf3e2e#diff-11767ea2aca3dca34f9d7d2cc03c73a8

PR #673, opened 2015-02-24
Read MAX_COMMENTS using ENV#fetch

Koronen:
Given a PR with violations in the diff, if `MAX_COMMENTS` is not set Hound will not report any violations. This is a kind of silent failure, which I think we should avoid.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -1,4 +1,6 @@
 class BuildRunner
+  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
+
   pattr_initialize :payload
 
   def run
@@ -28,7 +30,7 @@ class BuildRunner
   end
 
   def priority_violations
-    violations.take(ENV["MAX_COMMENTS"].to_i)
+    violations.take(MAX_COMMENTS)
   end
 
   def style_checker
--- a/spec/services/build_runner_spec.rb
+++ b/spec/services/build_runner_spec.rb
@@ -44,7 +44,7 @@ describe BuildRunner, '#run' do
     it "comments a maximum number of times" do
       allow(ENV).to receive(:[]).with("HOUND_GITHUB_TOKEN").
         and_return("something")
-      allow(ENV).to receive(:[]).with("MAX_COMMENTS").and_return("1")
+      stub_const("::BuildRunner::MAX_COMMENTS", 1)
       build_runner = make_build_runner
       stubbed_commenter
       violations = build_list(:violation, 2)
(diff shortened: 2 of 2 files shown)


app/services/build_runner.rb
```diff
@@ -28,7 +28,7 @@ def violations
   end
 
   def priority_violations
-    violations.take(ENV["MAX_COMMENTS"].to_i)
+    violations.take(ENV.fetch("MAX_COMMENTS").to_i)
```
Koronen: Should we add a default value here?
```rb
violations.take(ENV.fetch("MAX_COMMENTS", 10).to_i)
```

srt32: Thoughts on pulling this environment variable into a config var that is read and set at boot time?

Koronen: "config var"? Like a constant in `BuildRunner`?
```rb
class BuildRunner
  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS", 10).to_i
end
```

gylaz: @srt32 Do you mean something like `config/constants.rb`:
```
MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
RESQUE_RETRY = ENV.fetch("RESQUE_RETRY", 30).to_i
```

@Koronen I think either the default, or the approach like above would be good. If we don't default, we want to fail fast so the user knows to set that variable. Otherwise, the failure won't happen until some point later, when Hound tries to comment.

srt32: Yup! Something like `config/constants.rb` or use `Rails.config`.

Koronen: Good point @gylaz! I think I prefer setting a constant at boot time (and using `ENV.fetch` without a second argument), since it means we don't have to duplicate the default value (here and in `.sample.env`).

@srt32 would setting a constant in this class be okay for now? I'm not sure reworking the way we read constants (creating `config/constants.rb` or migrating to `Rails.config`) fits inside the scope of this PR.

srt32: :ok_hand:

Koronen: @srt32 @gylaz Updated!

Koronen:
No word from Travis on this PR? There's currently something wrong with my local development environment for Hound (tests failing for unknown reasons, e.g. #403), so I'd like Travis to check that these changes don't unexpectedly break the build.

gylaz:
Not sure what's up with Travis -- it didn't run this PR. Tests looked good on my machine. Thanks!

PR #674, opened 2015-02-24
Update JShintrb to support ES6 syntax

gylaz:
* Enables the usage of `"esnext": true` config option

https://trello.com/c/tUPpmibV


salbertson:
Yay, looks good to me. :+1:

Gemfile
```diff
@@ -14,7 +14,7 @@ gem "font-awesome-rails"
 gem "haml-rails"
 gem "high_voltage"
 gem "jquery-rails"
-gem "jshintrb"
+gem "jshintrb", github: "thoughtbot/jshintrb"
```
teoljungberg: Maybe add the `HEAD` commit?

croaky:
Awesome. This seems Intercom message-worthy, and updates to the homepage and config docs?

salbertson:
@croaky :+1:

Gemfile
gylaz: It's locked/specified in `Gemfile.lock`. What benefit do you see by using that here?

gylaz:
We might want to sit on this a bit before sending out messages. One issue that I can think of is that there are a few different file extensions people use with `ES6` syntax, like `.es6`, `.es` (besides the `.js`). @croaky @salbertson What do you think about giving this a week to bake before a public message?

croaky:
@gylaz Sounds good regarding holding off on public message.

PR #675, opened 2015-02-24
Ignore `vendor/` directory for JavaScript by default

gylaz:
This can still be overriden if the user provides their own `.jshintignore` file.


salbertson:
Related to https://github.com/thoughtbot/hound/pull/649

salbertson:
@gylaz, is there a reason not to use a default `.jshintignore` instead of doing this in code, the same way we use other default configs?

teoljungberg:
:+1:

gylaz:
@salbertson I can see that being useful for consistency. Good idea. Done in 3c39135.

app/models/repo_config.rb
```diff
@@ -33,11 +33,11 @@ def for(language)
   def ignored_javascript_files
     ignore_file_content = load_javascript_ignore
 
-    if ignore_file_content.present?
-      ignore_file_content.split("\n")
-    else
-      []
+    if ignore_file_content.blank?
+      ignore_file_content = File.read("config/style_guides/.jshintignore")
```
salbertson: Should we merge the default ignore file and the project ignore file?

salbertson:
Nice, looks good. I have one question. :+1:

app/models/repo_config.rb
gylaz: Then people would not be able to override default ignored directories.

I believe that in RuboCop if the user specifies `Exclude` it will override the default.

salbertson: Got it, makes sense.

PR #676, opened 2015-02-24
Update Octokit to v3.8.0

gylaz:
Allow us not to handle pagination ourselves for PR comments

https://github.com/octokit/octokit.rb/releases


salbertson:
:+1:

PR #677, opened 2015-02-25
Add a unique index on `subscriptions` `repo_id`

gylaz:
To prevent duplicate subscription records (and charges) for the same repo. Use a
unique partial index since we don't care for it to be unique if it was deleted.


salbertson:
Looks good to me. :+1:

PR #678, opened 2015-02-25
Add support for Go

salbertson:
https://github.com/thoughtbot/hound/pull/448

@gylaz, can we work together and see how difficult it will be to run Go on Heroku? Is shouldn't be too bad. We should work that out before thinking about merging.


salbertson:
I guess we need to figure out TravisCI setup too, build failed.

salbertson:
@kirs, have you run go or golint on Travis or Heroku?

kirs:
@salbertson wow, I'm glad to see Hound is going to finally merge my code :rocket: 

To run it on Travis we need to set up the $GOPATH variable (https://travis-ci.org/thoughtbot/hound/builds/52089646)
Can't suggest with Heroku as we don't use it.

spec/models/style_guide/go_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    style_guide = StyleGuide::Go.new(repo_config, "ownername")
+    style_guide.violations_in_file(build_file(content)).flat_map(&:messages)
+  end
+
+  def build_file(content)
+    line = double("Line", content: "doesntmatter", number: 1, patch_position: 2)
+    filename = Rails.root.join("test.go")
+    double("CommitFile", content: content, line_at: line, filename: filename)
+  end
+end
```
gylaz: Who about wrapping this code into `if ENV["GOPATH"]`? My thought is that not everyone who wants to work/contribute on Hound needs to have go installed.

gylaz:
There's a Heroku Go buildpack: https://github.com/kr/heroku-buildpack-go

Since we're looking into SOA stuff, does it make sense to work on this now?

croaky:
I don't think we should be adding more Ruby gems that wrap other language linters.

We should definitely support Go, but the approach should be the same as the SCSS IronWorker @arunagw and @teoljungberg are doing, except using the Go/binary IronWorker http://dev.iron.io/worker/languages/go/.

salbertson:
@croaky, yea, it doesn't make sense. I think Go support should follow whatever we end up doing with the server oriented approach.

salbertson:
@kirs, we've had a terrible time with Ruby wrappers. Our plan is to create open source workers that can be run anywhere, such as Iron.io, in the linter's native language.

@croaky, that sound accurate?

croaky:
Yup, that's the plan. Arun and Teo are making great progress this week with the Sass linter IronWorker. We hope that will define a clear API between workers and Hound. When it's been tested in production, we hope to open source that worker and provide guidance on creating workers for other languages (Go, PHP, Python, Java, JavaScript, Ruby, Shell, Scala, Clojure should all work well on IronWorker). We'll very likely take the lead on languages we use most often at thoughtbot (Ruby, JavaScript, SCSS) and need to lean on the community for contributions to other workers. We have quite a few Go, Scala, Clojure, and Haskell fans at the company, so I'm hoping this change opens up a flood of interest and ability to easily add new language support.

Early work seems promising but of course it could all blow up and plans could change. We'll do the best we can to comment on these threads with updates.

Apologies for not having it available long ago. We want to support everyone we can but have some foundational work to do first before we have the right balance of stability, maintainabilty, and team/community contribution.

salbertson:
@kirs, sorry for the fake out. This will be supported soon, but most likely written in Go and part of an external service. Thank you for all the work you put into this, it helped push supporting Go forward.

filib:
> We have quite a few Go, Scala, Clojure, and Haskell fans at the company, so I'm hoping this change opens up a flood of interest and ability to easily add new language support.

@croaky Any plans for that Haskell Hound? We'd love that over here at https://github.com/iconnect!

PR #679, opened 2015-02-25
README: Display correct ngrok URLs on GitHub

reshleman:
Github's markdown formatter was displaying the ngrok URLs incorrectly
because of the angle brackets. This commit surrounds the URLs with
backticks so they display as preformatted text.


gylaz:
Thanks @reshleman!

PR #680, opened 2015-02-25
Use ngrok url in .env.sample

dbackeus:
Make it conform with the README and avoid having a host which github webhooks can't reach.


teoljungberg:
Could you add in an update to the README asking the user to fill in the .env file aswell?

teoljungberg:
Looks great! I'll merge when you either rebase against master or squash the commits into 1

dbackeus:
@teoljungberg done

teoljungberg:
Thank you!

PR #681, opened 2015-02-25
Remove duplicated preferred methods in Ruby guide

reshleman:
Commit d86435b (#669) introduced two duplicate/conflicting lines in the
Ruby configuration file for preferred methods. The conflicts involved
`reduce`/`inject` and `find`/`detect`.

This commit removes the duplicates, preferring `inject` over `reduce`
and `detect` over `find`. Note that these conflict with [bbatsov]'s
style guide, which is linked to just above these settings in the config
file.

[bbatsov]: https://github.com/bbatsov/ruby-style-guide#map-find-select-reduce-size

(The [thoughtbot] style guide also prefers `reduce` over `detect`, but it's
unclear from the changes in d86435b if that's the intention is for Hound.
Let me know if I've got it backwards and I'll correct the PR.)

[thoughtbot]: https://github.com/thoughtbot/guides/blob/48f47748328c0b5f8575ed6f89f9a77c203b7063/style/ruby/README.md

### DIFF
--- a/config/style_guides/ruby.yml
+++ b/config/style_guides/ruby.yml
@@ -102,10 +102,8 @@ Style/CollectionMethods:
   PreferredMethods:
     collect: map
     collect!: map!
-    inject: reduce
-    detect: find
-    find_all: select
     find: detect
+    find_all: select
     reduce: inject
 Style/CommentAnnotation:
   Description: Checks formatting of special comments (TODO, FIXME, OPTIMIZE, HACK,


gylaz:
This seems correct in regards to what we had before. Originally, Hound deviated from the default RuboCop config, then thoughtbot's style guide forked off with config in `config/style_guides/thoughtbot`. Is there duplication in that file as well?

@salbertson What do you think about reverting to default RuboCop config for this rule? Or should we not do that because it'll be a sudden change on our existing customers (a la string quotes)?

salbertson:
@gylaz, RuboCop disables this Cop by default. Is that what you are suggesting?

I think we should stick closely to what we had before so we don't change up defaults on users.

```
CollectionMethods:
  Enabled: true
  PreferredMethods:
    find: detect
    reduce: inject
    collect: map
    find_all: select
```

@reshleman, should `reduce: inject` be added?

salbertson:
@reshleman thank you for catching this!

gylaz:
@salbertson Looks like `reduce: inject` is in there:
https://github.com/reshleman/hound/blob/re-duplicate-preferred-methods/config/style_guides/ruby.yml#L107

reshleman:
> thoughtbot's style guide forked off with config in `config/style_guides/thoughtbot`. Is there duplication in that file as well?

@gylaz There does not appear to be duplication in that file.

> should `reduce: inject` be added?

@salbertson It's in `style_guides/ruby.yml` in this PR -- and there are actually specs (passing) that check that `inject` is preferred over `reduce`, which I just noticed.

`style_guides/thoughtbot/ruby.yml` has `inject: reduce`

> thank you for catching this!

Sure thing. I was confused that Hound was flagging `reduce` in a PR, because I expected the defaults to match the thoughtbot style guide. I noticed the conflicting lines when trying to figure out what the default config was so I could override it.

gylaz:
Thanks @reshleman! Merged as fec939b9d5b4c03793cea763eb1f82a61ef7a854.

PR #682, opened 2015-02-25
Use the original jshintrb gem

gylaz:
It has been updated and the project owner seems active again.


salbertson:
@gylaz, is there anything we are going to lose that was in our fork?

salbertson:
One question, looks good to me. :+1:

gylaz:
No, nothing obvious.

PR #686, opened 2015-02-26
Disable PerceivedComplexity cop

salbertson:
We are testing these complexity cops out on Hound itself, I would rather wait until we have some feedback before enabling them for everyone by default.


gylaz:
Is this also enabled in thoughtbot config? Good to merge.

salbertson:
@gylaz, I think so, since it's on by default and the thoughtbot config doesn't override it. https://github.com/thoughtbot/hound/pull/687 should help with that issue.

PR #688, opened 2015-02-26
Deprecate ShowCopNames over DisplayCopNames

gylaz:
RuboCop supports `DisplayCopNames` option now. Having unsupported `ShowCopNames`
option in the RuboCop config seemed to cause issues for our users.

### DIFF
--- a/app/models/analytics.rb
+++ b/app/models/analytics.rb
@@ -62,6 +62,15 @@ class Analytics
     )
   end
 
+  def track_show_cop_names
+    track(
+      event: "Using ShowCopNames",
+      properties: {
+        owner: user
+      }
+    )
+  end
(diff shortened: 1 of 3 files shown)


croaky:
Is the effect of this that GitHub comments will include the cop name?

gylaz:
Yes, users can opt-in to see what cop is commenting for ease of configuring it.

salbertson:
Are we tracking the use so we can reach out to users?

:+1:

gylaz:
Yep. We can reach out to them, or if there's no use, then safely kill it.

gylaz:
Merged via d8268987de0214b081b601a2c6104c8501f50e8c.

PR #689, opened 2015-02-26
Whitelist Regexp for yaml loading

gylaz:
The RuboCop config uses a Ruby regex line in the config

Related error:

https://app.getsentry.com/hound-1/production/group/56103505/

### DIFF
--- a/spec/models/repo_config_spec.rb
+++ b/spec/models/repo_config_spec.rb
@@ -154,6 +154,27 @@ describe RepoConfig do
           "LineLength" => { "Max" => 90 },
         )
       end
+
+      context "with unsafe yaml" do
+        it "raises error" do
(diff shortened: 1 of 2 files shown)


spec/models/repo_config_spec.rb
```diff
@@ -154,6 +154,27 @@
...
+            StringLiterals: !ruby/object
+              foo:
+          EOS
+
+          expect { config.for("ruby") }.to raise_error Psych::DisallowedClass
+        end
+      end
+
+      context "with ruby regex in yaml" do
+        it "raises error" do
```
Koronen: `it "raises no error" do`?

gylaz: Yes! Thanks for catching that.

Koronen:
LGTM! :+1:

PR #690, opened 2015-02-26
Update repo owner on each build

gylaz:
If the repo has been transfered, this will be picked up during a build. Also,
this will populate our empty `repo.owner_id` fields.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -7,6 +7,7 @@ class BuildRunner
     if repo && relevant_pull_request?
       track_subscribed_build_started
       create_pending_status
+      upsert_owner
       repo.builds.create!(
         violations: violations,
         pull_request_number: payload.pull_request_number,
@@ -14,7 +15,6 @@ class BuildRunner
       )
       commenter.comment_on_violations(priority_violations)
       create_success_status
-      upsert_owner
       track_subscribed_build_completed
     end
   end
@@ -83,11 +83,12 @@ class BuildRunner
   end
 
   def upsert_owner
-    Owner.upsert(
+    owner = Owner.upsert(
       github_id: payload.repository_owner_id,
       name: payload.repository_owner_name,
       organization: payload.repository_owner_is_organization?
     )
+    repo.update(owner: owner)
   end
 
   def github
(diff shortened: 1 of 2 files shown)


thorncp:
LGTM

app/services/build_runner.rb
```diff
@@ -83,11 +83,12 @@ def create_success_status
   end
 
   def upsert_owner
-    Owner.upsert(
+    owner = Owner.upsert(
       github_id: payload.repository_owner_id,
       name: payload.repository_owner_name,
       organization: payload.repository_owner_is_organization?
     )
+    repo.update(owner: owner)
```
srt32: Should we be updating `repo` (and `owner`) before we go process the build (maybe on L8)? In that way, in the future when we pull config from the server, we'll have the `repo`'s `owner` before the `build` starts (so we can get the config).

salbertson:
@gylaz does this fix https://trello.com/c/9aE6gIdI/506-moving-a-repo-to-a-different-organization-or-owner-breaks-hound?

gylaz:
@salbertson No. I have left notes on that Trello card.

app/services/build_runner.rb
croaky: I don't think I understand the strategy of backfilling this data on each build. It feels a little dangerous to be doing this work in the critical path where we provide value to customers.

Don't we have this data in our database already? Can we backfill it in a rake task?

thorncp: This does seem like something we can backfill, since the only data we need from the GitHub API is the owner id, which is attainable through the public API without using the @houndci token.

gylaz: @croaky Regardless of the backfilling strategy, if the repo owner is ever changed (the repo was transferred), we want to update the owner -- that's the primary reason for this code. The added benefit of releasing it now, is it will begin backfilling the data. We will still need a rake task to finish backfilling, but this will alleviate our API rate limit hit.

@thorncp We will need to use the @houndci for private repos.

srt32: Here's a stab I took at the backfilling approach: https://github.com/thoughtbot/hound/compare/st-backfill-repo-owner-id-task.  We pulled back because of the concerns with the rate limit.

gylaz:
@salbertson @srt32 What do you think about merging this? It doesn't fix all the problems with a repo being moved from one org to another, but it will keep the owner updated if that ever happens.

salbertson:
@gylaz, looks good to me. :+1:

PR #691, opened 2015-02-27
Adds style_configs table, model, and associations

srt32:
Part of
https://trello.com/c/V0jY5LWs/452-migrate-existing-config-files-into-database

https://trello.com/c/0kAbrrqi/523-add-table-models-for-configs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ sudo: false
 cache: bundler
 rvm: 2.2.0
 
+addons:
+  postgresql: "9.3"
+
 before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
--- a/db/migrate/20150220174603_create_style_config.rb
+++ b/db/migrate/20150220174603_create_style_config.rb
@@ -1,0 +1,13 @@
+class CreateStyleConfig < ActiveRecord::Migration
+  def change
+    create_table :style_configs do |t|
+      t.boolean :enabled, default: true, null: false
+      t.string :language, null: false
+      t.text :rules, null: false
+      t.integer :owner_id, null: false
+    end
+
+    add_index :style_configs, [:owner_id, :language], unique: true
+    add_foreign_key :style_configs, :owners
+  end
+end
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -1,5 +1,6 @@
 class Owner < ActiveRecord::Base
   has_many :repos
+  has_many :style_configs, dependent: :destroy
 
   def self.upsert(github_id:, name:, organization:)
     owner = find_or_initialize_by(github_id: github_id)
--- a/app/models/style_config.rb
+++ b/app/models/style_config.rb
@@ -1,0 +1,3 @@
+class StyleConfig < ActiveRecord::Base
+  belongs_to :owner
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -68,6 +68,15 @@ ActiveRecord::Schema.define(version: 20150225001118) do
   add_index "repos", ["github_id"], name: "index_repos_on_github_id", using: :btree
   add_index "repos", ["owner_id"], name: "index_repos_on_owner_id", using: :btree
 
+  create_table "style_configs", force: :cascade do |t|
+    t.boolean "enabled",  default: true, null: false
+    t.string  "language",                null: false
+    t.text    "rules",                   null: false
+    t.integer "owner_id",                null: false
+  end
+
+  add_index "style_configs", ["owner_id", "language"], name: "index_style_configs_on_owner_id_and_language", unique: true, using: :btree
+
   create_table "subscriptions", force: :cascade do |t|
     t.datetime "created_at",                                                               null: false
     t.datetime "updated_at",                                                               null: false
@@ -108,4 +117,5 @@ ActiveRecord::Schema.define(version: 20150225001118) do
   add_foreign_key "memberships", "repos"
   add_foreign_key "memberships", "users"
   add_foreign_key "repos", "owners"
+  add_foreign_key "style_configs", "owners"
 end
--- a/spec/models/owner_spec.rb
+++ b/spec/models/owner_spec.rb
@@ -2,6 +2,7 @@ require "spec_helper"
 
 describe Owner do
   it { should have_many(:repos) }
+  it { should have_many(:style_configs).dependent(:destroy) }
 
   describe ".upsert" do
     context "when owner does not exist" do
--- a/spec/models/style_config_spec.rb
+++ b/spec/models/style_config_spec.rb
@@ -1,0 +1,5 @@
+require "spec_helper"
+
+describe StyleConfig do
+  it { should belong_to(:owner) }
+end


db/migrate/20150220174603_create_style_config.rb
```diff
@@ -0,0 +1,13 @@
+class CreateStyleConfig < ActiveRecord::Migration
+  def change
+    create_table :style_configs do |t|
+      t.boolean :enabled, default: true, null: false
```
srt32: Some open questions around the table structure:

1.  Should the concept of `enabled` be here at all? Should it be `repo` specific?
2.  How would we like to store the `language` type? This commit as is uses an `integer` so we could (once we have a use for this table) set up an `enum` on the `StyleConfig` model. Or just store strings.

gylaz: How about we just provide `enable` at the org level, since we still support per repo config files?

gylaz:
@srt32 Mind adding this to `.travis.yml` for Travis to be happy?
```
addons:
  postgresql: "9.3"
```

srt32:
Updated `.travis.yml` in 1792885

db/migrate/20150220174603_create_style_config.rb
srt32: Ah, interesting. Is there UI or a concept of an org disabling a language entirely? Maybe we just remove the concept of `enabled` from the server entirely?

gylaz: > Maybe we just remove the concept of enabled from the server entirely?

That's an interesting thought, and a good point to discuss in tomorrow's meeting (the one I'll miss).

croaky: Strings for the language seem simpler to me.

db/migrate/20150220174603_create_style_config.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.integer :language, null: false
+      t.json :rules, null: false, default: {}
+      t.integer :owner_id, null: false
+    end
+
+    add_index :style_configs, [:owner_id, :language], unique: true
```
croaky: Do we also want an index on owner_id by itself for lookup on the upcoming dashboard? Left hand side of Galen's new design.

spec/models/style_configs_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "spec_helper"
+
+describe StyleConfig do
+  it { should belong_to(:owner) }
+end
```
croaky: Duplicate file.

db/migrate/20150220174603_create_style_config.rb
gylaz: I was thinking about strings and `language` field. Food for thought, do we want to have configs tied to linters? For example, `JSHint` is moving away from doing style checking, to just JavaScript best practices checking. Then it's up to `JSCS` to handle strictly style checking. In this case, the `language` isn't the best fit. On the other hand, we can nest the rules ourselves for individual linter, say the db column would look something like:
```
rules
---------
["jshint": {...},
"jscs":  {,,,}]
```
Then there's ESLint which appears to do both, style checking and best practices checking.

spec/models/style_configs_spec.rb
gylaz: Yep, and pluralized to top it off.

db/migrate/20150220174603_create_style_config.rb
srt32: > For queries involving only x, the multicolumn index could be used, though it would be larger and hence slower than an index on x alone. 

From http://www.postgresql.org/docs/9.2/static/indexes-bitmap-scans.html. I think since `owner_id` is first in the list, I think we still are able to use the index on queries like `WHERE owner_id = 'xxx'`. I can run some `EXPLAIN`'s to confirm.

spec/models/style_configs_spec.rb
srt32: Yay for rebaseing ;)

db/migrate/20150220174603_create_style_config.rb
```diff
@@ -0,0 +1,13 @@
+class CreateStyleConfig < ActiveRecord::Migration
+  def change
+    create_table :style_configs do |t|
+      t.boolean :enabled, default: true, null: false
+      t.integer :language, null: false
+      t.json :rules, null: false, default: {}
```
salbertson: Why JSON and not text?

db/migrate/20150220174603_create_style_config.rb
srt32: We are able to use the first column of the compound index. So I think we're good. Sound right, @croaky?

```
irb(main):004:0> Owner.first.style_configs.explain
  Owner Load (0.4ms)  SELECT  "owners".* FROM "owners"  ORDER BY "owners"."id" ASC LIMIT 1
  StyleConfig Load (0.5ms)  SELECT "style_configs".* FROM "style_configs" WHERE "style_configs"."owner_id" = $1  [["owner_id", 1]]
=> EXPLAIN for: SELECT "style_configs".* FROM "style_configs" WHERE "style_configs"."owner_id" = $1 [["owner_id", 1]]
                                                QUERY PLAN
-----------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on style_configs  (cost=4.19..12.66 rows=5 width=45)
   Recheck Cond: (owner_id = 1)
   ->  Bitmap Index Scan on index_style_configs_on_owner_id_and_language  (cost=0.00..4.19 rows=5 width=0)
         Index Cond: (owner_id = 1)
(4 rows)
```

db/migrate/20150220174603_create_style_config.rb
srt32: The plan is store JSON. I think we could use the `text` type if we wanted to manually marshal / unmarshal the JSON in both directions. With the `json` type the data is stored as JSON so when we pull it out of the DB it's ready to use. I think having the data stored as JSON will also make debugging / data exploration easier. Thoughts, @salbertson?

srt32: Also, if we go with @gylaz's idea of storing multiple top-level keys we can index into the json using postgres (things also get better once jsonb lands). JSON should give us more flexibilty.

salbertson: Does JSON cover language configs we plan to support in the near future? I think I saw something about the PHP library using XML. I just don't want to be backing ourselves into a corner for no good reason.

croaky: Some existing languages we support have config files written in YAML. It seems like `text` is the simpler approach that solves our real/current, not potential/future needs.

db/migrate/20150220174603_create_style_config.rb
croaky: Yup, that should work.

croaky:
I'm into the "Change `json` column type to be `text`" idea but otherwise looks good to merge.

db/migrate/20150220174603_create_style_config.rb
srt32: Update to `text` in f8f09ce

db/migrate/20150220174603_create_style_config.rb
srt32: @gylaz intersting idea! Until we move away from language specific configs (https://github.com/thoughtbot/hound/tree/master/config/style_guides) it seems like maintaining concept of languages (over linters) still makes sense. I change the `language` column type to `string` in bc0c2f7

srt32:
@croaky @gylaz @salbertson ready for another quick :eyes:

.travis.yml
```diff
@@ -3,6 +3,9 @@ sudo: false
 cache: bundler
 rvm: 2.2.0
 
+addons:
+  postgresql: "9.3"
```
croaky: For my own curiosity, I was wondering if Travis supports Postgres 9.4 yet. Looks like they do (if experimental):

http://docs.travis-ci.com/user/build-environment-updates/2015-02-03/#PostgreSQL-9.4

I know Heroku supports 9.4. Looks like we're using 9.3, though, for production Hound.

Primary benefit I've read about for 9.4 is `jsonb` is supposed to be WICKED FAST compared to `json`.

croaky:
Good to merge.

srt32:
Thanks @croaky :bangbang: Squashed in 5a47ad5

db/migrate/20150220174603_create_style_config.rb
```diff
@@ -0,0 +1,13 @@
+class CreateStyleConfig < ActiveRecord::Migration
+  def change
+    create_table :style_configs do |t|
+      t.boolean :enabled, default: true, null: false
+      t.string :language, null: false
+      t.text :rules, null: false
```
gylaz: The configs that we pass to the linters is always a hash. I feel like `json` type was a better representation of a hash than text.

srt32:
Good point, @gylaz. If `json` ends up being a better choice we can move back that direction as the implementation gets fleshed out.

PR #694, opened 2015-02-27
Use activated instead of enabled

salbertson:
This change keep language consistent with code.


thorncp:
Looks good to me :+1:

P.S. I restarted the Travis build...

PR #700, opened 2015-03-02
Return unique list of repos from repos.json

srt32:
Currently, it is possible for there to be duplicate `memberships` tying a `user`
to `repos` and as a result, the user can sometimes see duplicate `repos` on
houndci.com. This update solves the UI problem. We'll have to dig a bit deeper
to figure out where the duplicates are coming from in the first place.

Part of
https://trello.com/c/GnvVhljg/532-repos-json-endpoint-returns-duplicate-repos

### DIFF
--- a/app/controllers/repos_controller.rb
+++ b/app/controllers/repos_controller.rb
@@ -11,7 +11,8 @@ class ReposController < ApplicationController
         repos = current_user.
           repos.
           order(active: :desc, full_github_name: :asc).
-          includes(:subscription)
+          includes(:subscription).
+          uniq
 
         render json: repos
       end
--- a/spec/controllers/repos_controller_spec.rb
+++ b/spec/controllers/repos_controller_spec.rb
@@ -27,4 +27,19 @@ describe ReposController do
       end
     end
   end
+
+  context "when current user has duplicate memberships" do
+    it "returns unique list of repos" do
+      user = create(:user)
+      repo = create(:repo)
+      repo.users << user
+      repo.users << user
+      stub_sign_in(user)
+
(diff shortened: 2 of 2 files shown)


app/controllers/repos_controller.rb
```diff
@@ -11,7 +11,8 @@ def index
         repos = current_user.
           repos.
           order(active: :desc, full_github_name: :asc).
-          includes(:subscription)
+          includes(:subscription).
```
srt32: This comment seems like something we want to ignore. Thoughts?

salbertson: This doesn't sound like it follows our style in this codebase. Want to open a separate PR updating [.ruby-style.yml](https://github.com/thoughtbot/hound/blob/master/.ruby-style.yml)?

salbertson:
Looks good to me. :+1:

app/controllers/repos_controller.rb
srt32: Cool, card here: https://trello.com/c/8plKSKKM/534-update-ruby-rules-to-ignore-operands-on-multiple-lines

gylaz:
Is there a related PR/card to make sure we don't get duplicate repos in the db? Would it be a better solution to solve this at the db/storage level, rather than adding a `uniq` call?

srt32:
@gylaz here's the card to figure out what's up at the persistence level: https://trello.com/c/rVCmKMc1/533-there-are-duplicate-memberships. I figured we could implement `uniq` (to not confuse users) until the underlying issue is sorted.

PR #703, opened 2015-03-03
Adds unique constraint to memberships

srt32:
There are duplicate `membership` records in the database:

```
SELECT distinct(user_id)
FROM memberships
GROUP BY user_id, repo_id
HAVING count(*) > 1;
```

Removing these records and then applying the unique key should prevent these
duplicates from happening in the future. If in routine usage, this unique
constraint is broken, we will now get exceptions (that we'll need to fix)
instead of bad data.

This PR removes the `membership` rake tasks because given the new DB constraints
it should not be possible to have bad data (that the tasks used to clean up).

https://trello.com/c/rVCmKMc1/533-there-are-duplicate-memberships


srt32:
I tried to create duplicate records by clicking around in my browser locally but I was not successful. Thoughts on taking the approach of applying the constraint in production and then investigating if the problem arises again?

db/migrate/20150303212157_add_unique_constraint_to_memberships.rb
```diff
@@ -0,0 +1,26 @@
+class AddUniqueConstraintToMemberships < ActiveRecord::Migration
+  def self.up
+    remove_duplicate_memberships
+
+    add_index :memberships, [:user_id, :repo_id], unique: true
+    remove_index :memberships, column: :user_id
+  end
```
gylaz: Does Rails run this in a transaction?

db/migrate/20150303212157_add_unique_constraint_to_memberships.rb
```diff
@@ -0,0 +1,26 @@
...
+    sql = <<-SQL
+      DELETE FROM memberships m
+      USING memberships m2
+      WHERE
+        m.user_id = m2.user_id AND
+        m.repo_id = m2.repo_id AND
+        m.id < m2.id
+    SQL
+
+    connection.execute(sql)
```
gylaz: Can just say `execute` without `connection`.

gylaz:
:+1:

db/migrate/20150303212157_add_unique_constraint_to_memberships.rb
srt32: I am not sure... I wrapped it in a transaction in 71fc67d

db/migrate/20150303212157_add_unique_constraint_to_memberships.rb
srt32: Nice! Updated in 71fc67d

app/controllers/repos_controller.rb
```diff
@@ -11,8 +11,7 @@ def index
         repos = current_user.
           repos.
           order(active: :desc, full_github_name: :asc).
-          includes(:subscription).
-          uniq
+          includes(:subscription)
```
gylaz: Seems like a bogus violation.

gylaz:
Ready to merge.

salbertson:
Looks good to me. :+1:

srt32:
I think the specs for repo_sync will fail after a rebase. I can take a look this weekend or next week.

arunagw:
@srt32 I have merged master in this branch to see if everything works and build is green. 

I think we can merge once everything is green

srt32:
@arunagw thanks for picking up this work. Can I help at all?

arunagw:
@srt32 build seems to be green here now. Is this looks fine to merge?

srt32:
:+1:

PR #704, opened 2015-03-04
Set user.refreshing_repos to true

srt32:
This update allows for the user's repo filtering to work after synced repos get
loaded up again.

repo_search

https://trello.com/c/6RlXlvgP/529-show-repos-after-searching-and-refreshing

### DIFF
--- a/spec/controllers/repo_syncs_controller_spec.rb
+++ b/spec/controllers/repo_syncs_controller_spec.rb
@@ -16,7 +16,7 @@ describe RepoSyncsController, "#create" do
   end
 
   context "user is not refreshing repos" do
-    it "enqueues repo sync job" do
+    it "sets user to refreshing repos true and enqueues repo sync job" do
       token = "usergithubtoken"
       user = create(:user)
       stub_sign_in(user, token)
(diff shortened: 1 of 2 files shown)


srt32:
We lost this functionality in https://github.com/thoughtbot/hound/commit/576954fdd36427baf3b898835b691069658f93ac

spec/controllers/repo_syncs_controller_spec.rb
```diff
@@ -24,6 +24,7 @@
 
       post :create
 
+      expect(user.reload.refreshing_repos).to be(true)
```
gylaz: How about `expect(user.reload).to be_refreshing_repos`?

gylaz:
:+1:

salbertson:
@srt32, if this is causing issues in production and the change is ready we should get it merged and deployed.

spec/controllers/repo_syncs_controller_spec.rb
srt32: Nice, updated in ce59f15

PR #706, opened 2015-03-05
Update documentation

salbertson:
Correct the first example, change `javascript` to `java_script`.

It might be a good idea to support both to avoid issues.


JoelQ:
:+1:

PR #707, opened 2015-03-05
Ignore return status of deactivation cleanup

gylaz:
If removing @houndci from the organization and repo fails for some reason, we
don't care, as we just want to make sure the hook is removed and the repo is
deactivated.


salbertson:
What do you think about adding a test since it's a regression?

Looks like it would fix the issue. :+1:

gylaz:
@salbertson Good idea. Added (and cleaned up) tests. Ready for another look.

PR #710, opened 2015-03-06
Introduce the concept of two GitHub API tokens

gylaz:
One token for dealing with private repos, one for public. This will keep the
non-paying customers separate from paying, and avoid chewing up our API rate
limit by the non-paying requests.

I created a @houndci-free to use that token for requests that can be handled
with a public token.


croaky:
Maybe name it @hound-oss?

.sample.env
```diff
@@ -3,7 +3,8 @@ ENABLE_HTTPS=no
 GITHUB_CLIENT_ID=github_client_id
 GITHUB_CLIENT_SECRET=github_client_secret
 HOST=<your-initials>-hound.ngrok.com
-HOUND_GITHUB_TOKEN=your_github_token
+PRIVATE_GITHUB_TOKEN=your_github_token
+PUBLIC_GITHUB_TOKEN=token_for_public_repos
 HOUND_GITHUB_USERNAME=your_github_username
 MAX_COMMENTS=10
 QUEUE=high,medium,low
```
croaky: `:sort`!

app/models/pull_request.rb
```diff
@@ -2,17 +2,17 @@ class PullRequest
...
-    @pull_request_files ||= api.
+    @pull_request_files ||= github.
       pull_request_files(full_repo_name, number).
       map { |file| build_commit_file(file) }
   end
 
   def comment_on_violation(violation)
-    api.add_pull_request_comment(
+    private_github.add_pull_request_comment(
       pull_request_number: number,
```
croaky: Why `private_github` here? Why not `github`?

app/models/pull_request.rb
```diff
@@ -43,8 +43,12 @@ def build_commit_file(file)
...
 
-  def api
-    @api ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
+  def github
+    @github ||= GithubApi.new(github_token)
+  end
+
+  def private_github
+    @private_github ||= GithubApi.new(ENV["PRIVATE_GITHUB_TOKEN"])
   end
```
croaky: This method seems unnecessary to me, duplicating `github`. Can it be deleted?

app/services/build_runner.rb
```diff
@@ -91,6 +91,6 @@ def upsert_owner
   end
 
   def github
-    @github ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
+    @github ||= GithubApi.new(ENV["PRIVATE_GITHUB_TOKEN"])
   end
```
croaky: Why not switch on public/private here, too?

This `GithubApi.new(public_or_private_token)` pattern is showing up in multiple places. Would it be simpler and reduce duplication to have `payload.github_token` or `payload.github_api_client`?

spec/fast_spec_helper.rb
```diff
@@ -11,7 +11,8 @@
 ENV["HOST"] = "test.host"
 ENV["SECRET_KEY_BASE"] = "test-key"
 ENV["HOUND_GITHUB_USERNAME"] = "houndci"
-ENV["HOUND_GITHUB_TOKEN"] = "houndgithubtoken"
+ENV["PRIVATE_GITHUB_TOKEN"] = "houndgithubtoken"
+ENV["PUBLIC_GITHUB_TOKEN"] = "public_github_token"
 ENV["ENABLE_HTTPS"] = "no"
 ENV["CHANGED_FILES_THRESHOLD"] = "300"
 ENV["MAX_COMMENTS"] = "10"
```
croaky: `:sort`!

croaky:
Makes sense to me. :+1:

app/models/pull_request.rb
gylaz: Originally, because I thought that only collaborators can make comments. Now that I think of it, anyone can make a comment on a public repo PR. Is that right?

app/models/pull_request.rb
gylaz: Yes, I think this can be removed now.

app/services/build_runner.rb
gylaz: I don't think status can be set by anyone on a public repo. I think, it has to be a collaborator with `push` permissions. But, we should verify this.

app/models/pull_request.rb
croaky: I think that's right, with the exception that owners can close a thread for comments.

app/services/build_runner.rb
croaky: Ah, right.

teoljungberg:
Looks interesting, @arunagw and I had this idea a few weeks back aswell :+1:

arunagw:
If we are going to this path for having multiple tokens I think we can keep same user.

As we know that personal token share same pool for rate limit so we cannot use that.

How about using authorized applications tokens. Hound can be authenticate against many applications and then I think each application token gets their own pool. If this is case then we can have a way to authenticate and generate authorized applications tokens for hound user only.

For now we can do this manually by doing an OAuth with a sample application and keep that token in ENV.

Generating multiple authorized applications tokens is also a hassle but I think it's less then creating more users.

We need to verify first that authorized applications pools are different.

salbertson:
Do we know what percentage of our requests are made for public vs private reviews?

gylaz:
Closing this in favor of #712.

PR #712, opened 2015-03-06
Use inidividual tokens for as much as possible

gylaz:
* Use user's token for PR fetching content, fetching comments, and setting status
* Use Hound's token for commenting and accepting org invites
* Save user's token if he doesn't have one upon viewing repos

Remaining changes?
- [x] Encrypt tokens
- [x] Figure out retry strategy when requests fail

### DIFF
--- a/app/controllers/repos_controller.rb
+++ b/app/controllers/repos_controller.rb
@@ -1,4 +1,6 @@
 class ReposController < ApplicationController
+  before_action :save_token
+
   def index
     respond_to do |format|
       format.html
@@ -15,6 +17,14 @@ class ReposController < ApplicationController
 
         render json: repos
       end
+    end
+  end
+
+  private
+
+  def save_token
+    if current_user.token.blank?
+      current_user.update!(token: session[:github_token])
     end
   end
 end
--- a/app/jobs/retryable.rb
+++ b/app/jobs/retryable.rb
@@ -5,5 +5,6 @@ module Retryable
 
   @retry_limit = 10
   @retry_delay = ENV.fetch("RESQUE_RETRY_DELAY", 30).to_i
+
   @fatal_exceptions = [Octokit::Unauthorized]
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,7 +39,24 @@ class User < ActiveRecord::Base
     repos.active.count > 0
   end
 
+  def token=(value)
+    encrypted_token = crypt.encrypt_and_sign(value)
+    write_attribute(:token, encrypted_token)
+  end
+
+  def token
+    encrypted_token = read_attribute(:token)
+    unless encrypted_token.nil?
+      crypt.decrypt_and_verify(encrypted_token)
+    end
+  end
+
   private
+
+  def crypt
+    secret_key_base = Rails.application.secrets.secret_key_base
+    ActiveSupport::MessageEncryptor.new(secret_key_base)
+  end
 
   def payment_gateway_customer
     @payment_gateway_customer ||= PaymentGatewayCustomer.new(self)
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -44,7 +44,16 @@ class BuildRunner
   end
 
   def pull_request
-    @pull_request ||= PullRequest.new(payload)
+    @pull_request ||= PullRequest.new(payload, token)
+  end
+
+  def token
+    @token ||= user_token || ENV["HOUND_GITHUB_TOKEN"]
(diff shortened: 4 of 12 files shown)


teoljungberg:
Should we also update the security document?

app/services/build_runner.rb
```diff
@@ -42,7 +42,17 @@ def commenter
...
+    @pull_request ||= PullRequest.new(payload, token)
+  end
+
+  def token
+    @token ||= user_token || ENV["HOUND_GITHUB_TOKEN"]
+  end
+
+  def user_token
+    user_with_token = repo.users.where.not(token: nil).sample
+    user_with_token && user_with_token.token
```
gylaz: Since tokens can be revoked and users can be removed from repos, without Hound knowing, what are some things we can do?
* Track the user that activated the repo -- won't work for existing active repos
* Get all the tokens and round-robin them until successful request
* Try the user token (just one like the implementation here), then fall back to Hound's token
* Something else

Thoughts?

app/jobs/retryable.rb
```diff
@@ -5,5 +5,7 @@ module Retryable
 
   @retry_limit = 10
   @retry_delay = ENV.fetch("RESQUE_RETRY_DELAY", 30).to_i
+
+  # to handle invalid tokens this needs to be removed
   @fatal_exceptions = [Octokit::Unauthorized]
```
gylaz: Retrying jobs seems to be currently broken. But once we fix it, we could remove this and then the `sample` logic should try a different user token. However, if there is just one user and its token is bad, then we'll get `Octokit::Unauthorized` 10 times before finally failing the job.

app/services/build_runner.rb
salbertson: I think trying the first token and retrying with Hound as the fallback is a good idea.

Should track that event to see how often it happens? Should we delete the user's token if it's invalid so we don't keep trying to reuse it?

app/jobs/retryable.rb
salbertson: What do you think about retrying with user tokens but if a user's token is invalid then deleting it so we don't try it again. If there are no users with present tokens then we fall back to Hound?

app/controllers/repos_controller.rb
```diff
@@ -18,4 +20,12 @@ def index
       end
     end
   end
+
+  private
+
+  def save_token
+    if current_user.token.blank?
+      current_user.update(token: session[:github_token])
```
salbertson: What do you think about doing some basic encryption so tokens are safe?

salbertson:
@gylaz, I created a checklist of remaining changes in the PR description. Do those look right?

app/controllers/repos_controller.rb
srt32: Should this line be `update!` to not swallow any potential errors?

db/migrate/20150306184045_add_token_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTokenToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :token, :string
```
srt32: Thoughts on having a `default: ""` so we don't have to check for `nil`s elsewhere?

salbertson:
@gylaz what's left besides updating security docs?

gylaz:
Nothing.

app/services/build_runner.rb
gylaz: Per discussion IRL: we don't expect there to be many expired tokens, and for now we'll treat it case by case.

db/migrate/20150306184045_add_token_to_users.rb
gylaz: I think it's better to see the `nil` and know it wasn't ever set, instead of guessing if we set this token on `""`.

app/services/build_runner.rb
salbertson: :+1:

app/models/user.rb
```diff
@@ -39,8 +39,25 @@ def has_active_repos?
     repos.active.count > 0
   end
 
+  def token=(value)
+    encrypted_token = crypt.encrypt_and_sign(value)
+    write_attribute(:token, encrypted_token)
```
teoljungberg: any reason why these violations where left un-resolved, @gylaz ?

gylaz: Yes. I chose to ignore them. thoughtbot doesn't have a guideline for this, and I disagree with default RuboCop config.

PR #713, opened 2015-03-09
Set GitHub status to failed when a config file is invalid

jakecraige:
https://trello.com/c/QI8TehB7/537-set-github-status-to-failed-when-any-config-files-are-invalid

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -9,6 +9,8 @@ class RepoConfig
     "scss" => "yaml",
   }
 
+  class ParserError < StandardError; end
+
   pattr_initialize :commit
 
   def enabled_for?(language)
@@ -106,13 +108,18 @@ class RepoConfig
 
   def parse_yaml(content)
     YAML.safe_load(content, [Regexp])
-  rescue Psych::SyntaxError
-    {}
+  rescue Psych::Exception => e
+    raise_repo_config_parser_error(e)
   end
 
   def parse_json(content)
     JSON.parse(content)
-  rescue JSON::ParserError
-    {}
+  rescue JSON::ParserError => e
+    raise_repo_config_parser_error(e)
+  end
+
+  def raise_repo_config_parser_error(e)
+    message = "#{e.class}: #{e.message}"
+    raise RepoConfig::ParserError.new(message)
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -6,6 +6,9 @@ en:
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
   success_status: "Hound has reviewed all the changes!"
+  config_error_status:
+    "Error parsing your config file. Click \"details\" for
+    assistance."
 
   account:
     bulk_pricing:
--- a/spec/models/repo_config_spec.rb
+++ b/spec/models/repo_config_spec.rb
@@ -234,6 +234,17 @@ describe RepoConfig do
         end
       end
 
+      context "with bad syntax" do
+        it "raises RepoConfig::ParserError error" do
+          config = config_for_file("config/rubocop.yml", <<-EOS.strip_heredoc)
+            StringLiterals: !ruby/object
+              ;foo:
+          EOS
+
+          expect { config.for("ruby") }.to raise_error(RepoConfig::ParserError)
+        end
+      end
+
       context "with unsafe yaml" do
         it "raises error" do
           config = config_for_file("config/rubocop.yml", <<-EOS.strip_heredoc)
@@ -241,7 +252,8 @@ describe RepoConfig do
               foo:
           EOS
 
-          expect { config.for("ruby") }.to raise_error Psych::DisallowedClass
+          expect { config.for("ruby") }.
+            to raise_error RepoConfig::ParserError, /Psych::DisallowedClass/
         end
       end
 
@@ -297,9 +309,8 @@ describe RepoConfig do
             }
           EOS
 
-          result = config.for("java_script")
-
-          expect(result).to eq({})
+          expect { config.for("java_script") }.
+            to raise_error(RepoConfig::ParserError)
         end
       end
     end
--- a/spec/services/build_runner_spec.rb
+++ b/spec/services/build_runner_spec.rb
@@ -140,6 +140,32 @@ describe BuildRunner, '#run' do
       )
       expect(repo.reload.owner).to eq Owner.first
     end
+
+    it "fails the GitHub status with invalid config" do
+      repo = create(:repo, :active, github_id: 123)
+      payload = stubbed_payload(
+        github_repo_id: repo.github_id,
+        full_repo_name: "test/repo",
+        head_sha: "headsha"
+      )
+      build_runner = BuildRunner.new(payload)
+      pull_request = stubbed_pull_request_with_file("random.js", "")
+      style_checker = stubbed_style_checker_with_invalid_javascript_config(
+        pull_request
+      )
+      allow(build_runner).to receive(:style_checker).and_return(style_checker)
+      allow(build_runner).to receive(:pull_request).and_return(pull_request)
+      github_api = stubbed_github_api
+
+      build_runner.run
+
+      expect(github_api).to have_received(:create_error_status).with(
+        "test/repo",
+        "headsha",
+        I18n.t(:config_error_status),
+        configuration_url
+      )
+    end
   end
 
   context 'without active repo' do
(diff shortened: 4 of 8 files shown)


config/locales/en.yml
```diff
@@ -6,6 +6,7 @@ en:
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
   success_status: "Hound has reviewed all the changes!"
+  config_error_status: "Hound has encountered an error parsing your config."
```
croaky: Can we link them to houndci.com/configuration to help them debug?

spec/services/build_runner_spec.rb
```diff
@@ -139,6 +139,33 @@
...
+        }
+      EOS
+      github_api = stubbed_github_api
+
+      build_runner.run
+
+      expect(github_api).to have_received(:create_error_status).with(
+        "test/repo",
+        "headsha",
+        "Hound has encountered an error parsing your config. JSON::ParserError: 399: unexpected token at ']\n}\n'"
```
croaky: Ouch, this error message is a little opaque. Is the 399 a line number?

spec/services/build_runner_spec.rb
```diff
@@ -239,14 +278,67 @@ def stubbed_pull_request
     pull_request
   end
 
+  def stubbed_pull_request_with_file(filename, file_content)
+      file         = double(:file,
```
croaky: Check out the style of some of the other hash arguments invocation in the project. This style should be fixed before merging.

https://github.com/thoughtbot/guides/tree/master/style#formatting

We also don't align tokens like the `=`.

```ruby
file = double(
  :file,
  filename: filename,
  content: file_content,
  removed?: false
)
```

spec/models/repo_config_spec.rb
```diff
@@ -218,9 +231,9 @@
             }
           EOS
 
-          result = config.for("java_script")
-
-          expect(result).to eq({})
+          expect {
```
croaky: I think these `{...}` errors are safe to ignore.

config/locales/en.yml
jakecraige: Added

spec/models/repo_config_spec.rb
```diff
@@ -155,6 +155,17 @@
         )
       end
 
+      context "with bad syntax" do
+        it "raises RepConfigParserError error" do
```
sgrif: Typo here

spec/models/repo_config_spec.rb
```diff
@@ -155,6 +155,17 @@
...
+          config = config_for_file("config/rubocop.yml", <<-EOS.strip_heredoc)
+            StringLiterals: !ruby/object
+              ;foo:
+          EOS
+
+          expect { config.for("ruby") }.to raise_error(RepoConfigParserError)
```
sgrif: Do you think it's worth testing the error message?

spec/services/build_runner_spec.rb
```diff
@@ -239,14 +264,99 @@ def stubbed_pull_request
     pull_request
   end
 
+  def stubbed_pull_request_with_file(filename, file_content)
+    file = double_file(filename, file_content)
+    pull_request = double_pull_request_with_files([file])
+    allow(PullRequest).to receive(:new).and_return(pull_request)
```
sgrif: Should we inject the pull request where it's needed, rather than stubbing `.new` (as per the style guide)

spec/services/build_runner_spec.rb
```diff
@@ -239,14 +264,99 @@ def stubbed_pull_request
...
+      repository_owner_name: "test"
+    )
+  end
+
+  def stubbed_style_checker_with_config_file(pull_request, filename, content)
+    config = config_for_file(filename, content)
+
+    style_checker = StyleChecker.new(pull_request)
+    allow(style_checker).to receive(:config).and_return(config)
+    allow(StyleChecker).to receive(:new).and_return(style_checker)
```
sgrif: Should we inject the style checker where it's needed, rather than stubbing .new (as per the style guide)

spec/services/build_runner_spec.rb
```diff
@@ -239,14 +264,99 @@ def stubbed_pull_request
...
+      hound_config: hound_config,
+      "#{file_path}" => content
+    )
+
+    RepoConfig.new(commit)
+  end
+
+  def configuration_url
+    Rails.application.routes.url_helpers.configuration_url(host: ENV["HOST"])
+  end
```
sgrif: Do you think if we changed the dependency graph of the objects being tested, we could get away with much less setup and just fake out a few direct collaborators?

app/models/repo_config.rb
```diff
@@ -1,3 +1,5 @@
+class RepoConfigParserError < StandardError; end
```
gylaz: How about putting this into its own file, or maybe an initializer?

config/locales/en.yml
```diff
@@ -6,6 +6,9 @@ en:
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
   success_status: "Hound has reviewed all the changes!"
+  config_error_status:
+    "Hound has encountered an error parsing your config. See
+    %{configuration_url} for assistance. %{error_message}"
```
gylaz: Any idea on how this looks in the PR build status? I'm curious if long sentences get cut off.

spec/models/repo_config_spec.rb
jakecraige: @sgrif In this specific case I don't think we need to because the typo just happens to be a specific type of error. It could be something else. [This line](https://github.com/thoughtbot/hound/pull/713/files#diff-51792d87c8babd27e80c21f70a1ba9c6R177) covers the error for a specific type

spec/services/build_runner_spec.rb
```diff
@@ -139,6 +139,31 @@
...
+
+    it "fails the GitHub status with invalid config" do
+      repo = create(:repo, :active, github_id: 123)
+      payload = stubbed_payload(
+        github_repo_id: repo.github_id,
+        full_repo_name: "test/repo",
+        head_sha: "headsha"
+      )
+      build_runner = BuildRunner.new(payload)
+
```
gylaz: Extra newline

spec/services/build_runner_spec.rb
```diff
@@ -239,14 +264,99 @@ def stubbed_pull_request
...
+      :file,
+      filename: filename,
+      content: file_content,
+      removed?: false
+    )
+  end
+
+  def double_pull_request_with_files(files)
+    double(
+      :pull_request,
```
gylaz: How about `"PullRequest"` for `instance_double` support?

config/locales/en.yml
jakecraige: @gylaz Turns out there's an undocumented 140 char limit. Thankfully their error message lets you know this. Updating to a shorter status

gylaz: Since it already says `hound` as part of the service name, how about just:
```
Error parsing your config file. See %{configuration_url} for assistance.
```

spec/services/build_runner_spec.rb
jakecraige: @sgrif Yeah I do. It took a decent amount of time to get the proper setup due to how deep the dependency graph goes

app/models/repo_config.rb
jakecraige: @gylaz How about moving it into the RepoConfig namespace? It then becomes `RepoConfig::ParserError` and is then container within the class.

spec/models/repo_config_spec.rb
jakecraige: Good catch, fixed

config/locales/en.yml
```diff
@@ -6,6 +6,9 @@ en:
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
   success_status: "Hound has reviewed all the changes!"
+  config_error_status:
+    "Error parsing your config file. Click \"details\" for
+    assistance."
```
gylaz: @frechg Thoughts about this copy?

spec/services/build_runner_spec.rb
jakecraige: I didn't like stubbing new too much myself. Updated to not use it now. Thanks

gylaz:
Looks good to me.

spec/services/build_runner_spec.rb
jakecraige: I've updated it accordingly, good catch.

spec/services/build_runner_spec.rb
jakecraige: Nice, I got rid of it

spec/services/build_runner_spec.rb
jakecraige: I've switched it now. Nice catch

gylaz:
Ready to merge.

salbertson:
@jakecraige, what needs to be done before this can be merged?

jakecraige:
@salbertson This one should be good

salbertson:
@jakecraige, great. Squash and more when you're ready.

PR #714, opened 2015-03-09
Gf account page cleanup

frechg:
* Update account layout to fit with the rest of the app UI
* Add bulk pricing message
* Show active paid repo usage

screen shot 2015-03-09 at 10 31 11 am

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -59,6 +59,10 @@ class Repo < ActiveRecord::Base
     ENV["EXEMPT_ORGS"] && ENV["EXEMPT_ORGS"].split(",").include?(organization)
   end
 
+  def total_violations
+    Violation.joins(:build).where(builds: { repo_id: self.id }).length
+  end
+
   private
 
   def organization
--- a/app/views/accounts/show.haml
+++ b/app/views/accounts/show.haml
@@ -3,47 +3,54 @@
 = render 'settings'
 
 %section.account-section
-  %h2.label-text
-    Your monthly cost
+  %aside.account-nav
+    %p.message
+  .account-content
+    .billing-breakdown
+      %h2.label-text
+        Your monthly cost
 
-    - if current_user.stripe_customer_id.present?
-      %span.section-action
-        %a{ 'ng-controller' => 'accountController', 'ng-click' => 'showStripeForm()' }
-          Update Credit Card
+        - if current_user.stripe_customer_id.present?
+          %span.section-action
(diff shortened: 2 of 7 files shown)


app/views/accounts/show.haml
```diff
@@ -3,47 +3,66 @@
...
-        - @account_page.repos.each do |repo|
-          %li.repo
-            %span.repo-name{ "ng-class" => "{private: #{repo.private}}" }
-              = repo.full_github_name
-    - else
-      %p.empty-state-message
-        You have no private repos billed to you
+    %article.active-private-repos
+      - if @account_page.repos.any?
+        %h3 Active Paid Repos
```
salbertson: What do you think about "Active Private Repos" instead?

app/models/repo.rb
```diff
@@ -59,6 +59,10 @@ def exempt?
     ENV["EXEMPT_ORGS"] && ENV["EXEMPT_ORGS"].split(",").include?(organization)
   end
 
+  def total_violations
+    Violation.joins(:build).where(builds: { repo_id: self.id }).length
```
salbertson: Redundant `self`?

salbertson: Should `#count` be used here instead of `#length`?

app/views/accounts/show.haml
```diff
@@ -3,47 +3,66 @@
 = render 'settings'
 
 %section.account-section
-  %h2.label-text
-    Your monthly cost
+  %aside.account-nav
+    %p.message
+      If you'd like to pay for Hound in bulk on a monthly or annual
+      basis we're happy to make that happen. Email
+      #{mail_to "hound@thoughtbot.com"}
```
salbertson: What do you think about moving the support email address to a translation so it's easy to change?

salbertson:
I like it; I really like the review and violation numbers for each repo, it hopefully helps show the value of Hound and that it's doing something worth paying for.

I have a couple questions, looks good. :+1:

spec/features/account_spec.rb
```diff
@@ -42,4 +42,20 @@
...
+  scenario "user sees paid repo usage" do
+    user = create(:user)
+    paid_repo = create(:repo, users: [user])
+    paid_repo.builds << create(:build, :failed)
+    paid_repo.builds << create(:build, :failed)
+    paid_repo.builds << create(:build)
+    create(:subscription, repo: paid_repo, user: user)
+
+    sign_in_as(user)
+
```
gylaz: This newline is unnecessary.

PR #717, opened 2015-03-10
Support RuboCop's inherit_from

jakecraige:
https://trello.com/c/hSQ6Xi7l/514-support-rubocop-s-inherit-from

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -23,7 +23,7 @@ class RepoConfig
       config_file_path = config_path_for(language)
 
       if config_file_path
-        load_file(config_file_path, FILE_TYPES.fetch(language))
+        load_config(config_file_path, FILE_TYPES.fetch(language))
       else
         {}
       end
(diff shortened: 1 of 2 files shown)


app/models/repo_config.rb
```diff
@@ -75,6 +75,17 @@ def config_path_for(language)
       hound_config[language]["config_file"]
   end
 
+  def load_config(config_file_path, file_type)
+    main_config = load_file(config_file_path, file_type)
+    inherit_from = main_config.fetch("inherit_from", [])
+
+    inherited_config = inherit_from.reduce({}) do |config, ancestor_file_path|
```
jakecraige: Style guide says the opposite "Prefer reduce over inject." [here](https://github.com/thoughtbot/guides/blob/master/style/ruby/README.md#ruby)

gylaz: Looks to be a bug in Hound. We should probably remove: https://github.com/thoughtbot/hound/blob/master/.ruby-style.yml#L93

app/models/repo_config.rb
```diff
@@ -75,6 +75,17 @@ def config_path_for(language)
...
+      config.merge(load_file(ancestor_file_path, file_type))
```
gylaz: What does this line return? Is it a path?

gylaz:
Nice work! :+1:

app/models/repo_config.rb
jakecraige: Seems like it's confusing. I've moved to load_file into a variable so clear up what it is in https://github.com/thoughtbot/hound/commit/87ac1382fa69683bad12938ab67d141116424c19

app/models/repo_config.rb
jakecraige: PR submitted https://github.com/thoughtbot/hound/pull/718

salbertson:
Awesome! @jakecraige, we've been wanting to do this for a while. We should get this merged and test it on staging. :+1:

PR #719, opened 2015-03-10
Retrieve account line items from Stripe instead of the database

jakecraige:
- Add subscriptions method to PaymentGatewayCustomer
- Account for stripe discounts
- Display raw subscription plan name on account line items

https://trello.com/c/lc8COE2s/508-pull-billing-summary-from-stripe-not-our-db


app/models/payment_gateway_customer.rb
```diff
@@ -49,6 +53,10 @@ class NoSubscription
     def retrieve(*_args)
       nil
     end
+
```
gabebw: Why isn't NoSubscription in its own file?

gabebw: Same for `NoRecord`

gylaz: Because they are implementation details of `PaymentGatewayCustomer`.

app/controllers/accounts_controller.rb
```diff
@@ -1,6 +1,9 @@
 class AccountsController < ApplicationController
   def show
-    @account_page = AccountPage.new(find_subscribed_repos)
+    @account_page = AccountPage.new(
+      find_subscribed_repos,
+      current_user.payment_gateway_customer
```
gylaz: What do you think about passing just the subscriptions here?

app/models/user.rb
```diff
@@ -39,12 +39,12 @@ def has_active_repos?
     repos.active.count > 0
   end
 
-  private
-
   def payment_gateway_customer
     @payment_gateway_customer ||= PaymentGatewayCustomer.new(self)
   end
```
gylaz: How about keeping this private, and just delegating `payment_subscriptions` to it?

app/presenters/monthly_line_item.rb
```diff
@@ -1,31 +1,42 @@
...
-    @repos = repos
+  def initialize(subscription)
+    @subscription = subscription
   end
 
   def title
-    case @repos.first.subscription_price
-    when 9
+    case @subscription.plan.name
+    when "Personal"
```
gylaz: Wonder if we can drop this now. We've got like 6 personal private legacy subs. We can just let the `else` handle it.

app/presenters/monthly_line_item.rb
```diff
@@ -1,31 +1,42 @@
...
       "Private Personal Repos"
-    when 12
+    when "Private"
       "Private Repos"
-    when 24
+    when "Organization"
```
gylaz: Same here.

app/presenters/monthly_line_item.rb
```diff
@@ -1,31 +1,42 @@
...
       "Private Org Repos"
+    else
+      @subscription.plan.name
     end
   end
```
gylaz: I'm thinking we don't need this method anymore, and just display `@subscription.plan.name` in the UI. @salbertson do you see anything worth keeping here?

app/controllers/accounts_controller.rb
jakecraige: :+1: Makes more sense since that's all I'm using anyways

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -78,4 +78,18 @@
...
+  describe "#subscriptions" do
+    context "when stripe_customer_id is present" do
+      it "retrieves subscriptions data" do
+        user = build_stubbed(:user, stripe_customer_id: stripe_customer_id)
+        stub_customer_find_request
+        payment_gateway_customer = PaymentGatewayCustomer.new(user)
+        subscriptions = payment_gateway_customer.subscriptions
+
+        expect(subscriptions.length).to eq 1
+        expect(subscriptions.first).to be_an_instance_of(Stripe::Subscription)
```
gylaz: Not a huge fan of checking type. How about just checking the fields that we care about?

app/models/user.rb
jakecraige: That seems good, I'll take care of it. I think I'm going to call it `payment_gateway_subscriptions` to keep the naming consistent with the customer one

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -145,4 +145,52 @@ def stub_failed_subscription_destroy_request
...
+    ).to_return(
+      status: 200,
+      body: merge_customer_subscriptions(
+        File.read("spec/support/fixtures/stripe_customer_find.json"),
+        subscriptions
+      )
+    )
+  end
+
+  def generate_subscriptions_response(subscriptions)
```
gylaz: Can we keep this close to the test, since it's only used there?

app/presenters/monthly_line_item.rb
jakecraige: Since we currently need the case statement anyways, might as well keep it around as long as they are. I don't see it hurting anything

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -145,4 +145,52 @@ def stub_failed_subscription_destroy_request
...
+    read_subscription_fixture("individual")
+  end
+
+  def private_subscription_response
+    read_subscription_fixture("private")
+  end
+
+  def org_subscription_response
+    read_subscription_fixture("org")
+  end
```
gylaz: Same for these methods too.

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -145,4 +145,52 @@ def stub_failed_subscription_destroy_request
...
     "https://api.stripe.com/v1/customers"
   end
+
+  def stub_customer_find_request_with_subscriptions(customer_id, subscriptions)
+    stub_request(:get, "#{stripe_base_url}/#{customer_id}").with(
+      headers: { "Authorization" => "Bearer #{ENV['STRIPE_API_KEY']}" }
+    ).to_return(
+      status: 200,
+      body: merge_customer_subscriptions(
+        File.read("spec/support/fixtures/stripe_customer_find.json"),
```
gylaz: Any reason this is an argument? Can we just move this into the method?

jakecraige: Good catch. I like that

jakecraige:
@gylaz Implemented your recommendations

spec/features/account_spec.rb
```diff
@@ -58,4 +67,52 @@
...
+
+  def private_subscription_response
+    read_subscription_fixture("private")
+  end
+
+  def org_subscription_response
+    read_subscription_fixture("org")
+  end
+
+  private
```
gylaz: > Avoid the private keyword in specs. 

[Source](https://github.com/thoughtbot/guides/blob/master/style/testing/README.md#testing)

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -78,4 +78,19 @@
...
     end
   end
+
+  describe "#subscriptions" do
+    context "when stripe_customer_id is present" do
+      it "retrieves subscriptions data" do
+        user = build_stubbed(:user, stripe_customer_id: stripe_customer_id)
+        stub_customer_find_request
+        payment_gateway_customer = PaymentGatewayCustomer.new(user)
+        subscription = payment_gateway_customer.subscriptions.first
```
gylaz: What is the execution step here? Can we separate with new lines?

gylaz:
A few minor comments. Looks good!

app/presenters/account_page.rb
```diff
@@ -3,32 +3,21 @@ class AccountPage
 
   attr_reader :repos
 
-  def initialize(repos)
+  def initialize(repos, subscriptions)
```
teoljungberg: How about using `pattr_initialize`?

app/presenters/monthly_line_item.rb
```diff
@@ -1,31 +1,42 @@
 class MonthlyLineItem
   include ActionView::Helpers::NumberHelper
 
-  def initialize(price, repos)
-    @price = price
-    @repos = repos
+  def initialize(subscription)
```
teoljungberg: How about using `pattr_initialize`?

teoljungberg:
There are a few missing trailing commas, but I didn't want to comment on each and every one of them

app/presenters/monthly_line_item.rb
jakecraige: I needed to use `vattr_initialize` since I need publicly accessibly properties, but nice catch!

app/presenters/account_page.rb
jakecraige: I needed to use `vattr_initialize` since I need publicly accessibly properties, but nice catch!

salbertson:
@jakecraige, does this take discounts into consideration? Some of our subscriptions in Stripe are are discounted using coupons. We should show the discounted price since that's what the subscribers are paying, and they might not know their subscription is discounted in some cases.

jakecraige:
@salbertson I don't think it does. It looks like there is a `discount` key in the json response from Stripe but I'm never using it. I'll have to check what that returns when there is a discount. I will either be able to get to it today or Friday.

app/presenters/monthly_line_item.rb
salbertson: @gylaz, yea, I think we can just display the plan name set up in Stripe.

salbertson:
@teoljungberg, do you think trailing comma preference should be added to the Hound style guide?

salbertson:
> It looks like there is a discount key in the json response from Stripe but I'm never using it.

@jakecraige, we should be taking the discount into consideration if the "total subscription price" doesn't. If we don't the price might be way more than what the subscriber is paying, which might freak them out, ha.

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -35,6 +35,20 @@ def stub_customer_find_request
     )
   end
 
+  def stub_customer_with_discount_find_request(customer_id = stripe_customer_id)
```
jakecraige: this should be ignored from https://github.com/thoughtbot/hound/commit/4a2b2ef7c93baee444ce76c860b00947201c95a4 ?

app/models/payment_gateway_customer.rb
```diff
@@ -56,4 +64,24 @@ def last4
       ""
     end
   end
+
+  class NullDiscount
```
srt32: Thoughts on calling this class `FullPrice` or something like it to better represent the domain concept?

spec/features/account_spec.rb
```diff
@@ -34,15 +43,39 @@
...
+    stub_customer_find_request_with_subscriptions(
+      user.stripe_customer_id,
+      generate_subscriptions_response([
+        discounted_amount_subscription_response,
+        discounted_percent_subscription_response,
+      ])
+    )
+
+    sign_in_as(user)
+
```
srt32: Maybe remove this line to group the execution steps?

spec/features/account_spec.rb
```diff
@@ -20,13 +21,21 @@
...
   scenario "user with multiple subscriptions views account page" do
-    user = create(:user)
+    user = create(:user, stripe_customer_id: "1234")
+    stub_customer_find_request_with_subscriptions(
+      user.stripe_customer_id,
+      generate_subscriptions_response([
+        individual_subscription_response,
+        private_subscription_response,
+        org_subscription_response,
+      ])
```
gylaz: Might be worth pulling this out into a var.

spec/features/account_spec.rb
gylaz: Is this newline necessary?

spec/features/account_spec.rb
```diff
@@ -34,15 +43,39 @@
...
+        discounted_percent_subscription_response,
+      ])
+    )
+
+    sign_in_as(user)
+
+    visit account_path
+
+    expect(page).to have_text("$700")
+
```
gylaz: Extra newline?

spec/features/account_spec.rb
```diff
@@ -34,15 +43,39 @@
...
+    expect(page).to have_text("Bulk - Yearly")
+    expect(page).to have_text("$250/mo")
+    expect(page).to have_text("$500")
+
```
gylaz: Extra newline?

spec/features/account_spec.rb
```diff
@@ -58,4 +91,59 @@
...
+      "object" => "list",
+      "total_count" => subscriptions.length,
+      "has_more" => false,
+      "url" => "/v1/customers/cus_2e3fqARc1uHtCv/subscriptions",
+      "data" => subscriptions,
+    }
+  end
+
+  def discounted_amount_subscription_response
+    # This subscription has a quanity of 2
```
gylaz: Why is this comment useful?

spec/support/helpers/stripe_api_helper.rb
gylaz: It should. Do you mind modifying that [these lines](https://github.com/thoughtbot/hound/blob/master/.ruby-style.yml#L511-L512) to?
```rb
Exclude: spec/`
```
Then delete these comments and push, to see if Hound will comment again?

app/models/payment_gateway_customer.rb
jakecraige: That makes sense. There's another called `NoSubscription` so I'm going to change this and `NullCoupon` to `NoDiscount` / `NoCoupon`.

spec/features/account_spec.rb
jakecraige: Sounds good

spec/features/account_spec.rb
jakecraige: I suppose it's not in this context. I'll move it to where it makes more sense to mention it. Thanks

spec/support/helpers/stripe_api_helper.rb
jakecraige: It didn't change @gylaz https://github.com/thoughtbot/hound/commit/f6e0b5568ee8072e2c910aeaae641d4b042ba368

teoljungberg:
Good idea! It should atleast be in thoughtbot's styleguide

---
Teo Ljungberg

salbertson:
@jakecraige any update on this?

srt32:
Is this PR still blocked on https://github.com/thoughtbot/hound/pull/720

jakecraige:
@srt32 Yeah I believe so. That one will need to be merged before this one

salbertson:
@jakecraige can we forward with this since https://github.com/thoughtbot/hound/pull/720 is merged?

PR #720, opened 2015-03-11
increment/decrement quantity instead of new subscription

jakecraige:
We increment/decrement existing subscriptions instead of creating a new one for each repo.

https://trello.com/c/XNL81IPQ/539-increment-subscription-quantity-instead-of-charging-for-each-repo-individually

### DIFF
--- a/app/models/payment_gateway_customer.rb
+++ b/app/models/payment_gateway_customer.rb
@@ -1,5 +1,10 @@
 class PaymentGatewayCustomer
-  pattr_initialize :user
+  attr_reader :user
+
+  def initialize(user, customer: nil)
+    @user = user
+    @customer = customer
+  end
 
   def email
     customer.email
@@ -19,6 +24,23 @@ class PaymentGatewayCustomer
     end
   end
 
+  def find_or_create_subscription(plan:, repo_id:)
+    subscriptions.detect { |subscription| subscription.plan == plan } ||
+      create_subscription(plan: plan, metadata: { repo_ids: [repo_id] })
+  end
+
+  def subscriptions
+    customer.subscriptions.map do |subscription|
+      PaymentGatewaySubscription.new(subscription)
+    end
+  end
+
+  def retrieve_subscription(stripe_subscription_id)
+    PaymentGatewaySubscription.new(
+      customer.subscriptions.retrieve(stripe_subscription_id)
+    )
+  end
+
   def update_card(card_token)
     customer.card = card_token
     customer.save
@@ -33,6 +55,13 @@ class PaymentGatewayCustomer
   end
 
   private
+
+  def create_subscription(options)
+    PaymentGatewaySubscription.new(
+      customer.subscriptions.create(options),
+      new_subscription: true,
+    )
+  end
 
   def default_card
     customer.cards.detect { |card| card.id == customer.default_card } ||
--- a/app/models/payment_gateway_subscription.rb
+++ b/app/models/payment_gateway_subscription.rb
@@ -1,0 +1,66 @@
+class PaymentGatewaySubscription
+  attr_reader :stripe_subscription
+
+  delegate :id, :metadata, :save, :delete, to: :stripe_subscription
+
+  def initialize(stripe_subscription, new_subscription: false)
+    @stripe_subscription = stripe_subscription
+    @new_subscription = new_subscription
+  end
+
+  def subscribe(repo_id)
+    append_repo_id_to_metadata(repo_id)
+    if existing_subscription?
+      increment_quantity
+    end
+  end
+
+  def unsubscribe(repo_id)
+    if stripe_subscription.quantity > 1
+      remove_repo_id_from_metadata(repo_id)
+      decrement_quantity
+    else
+      delete
+    end
+  end
+
+  def increment_quantity
+    stripe_subscription.quantity += 1
+    save
+  end
+
+  def decrement_quantity
+    stripe_subscription.quantity -= 1
+    save
+  end
+
+  def plan
+    stripe_subscription.plan.id
+  end
+
+  def existing_subscription?
+    !@new_subscription
+  end
+
+  private
+
+  def current_repo_ids
+    Array(metadata["repo_ids"] || metadata["repo_id"])
+  end
+
+  def append_repo_id_to_metadata(repo_id)
+    repo_ids = current_repo_ids + [repo_id]
+
+    if metadata["repo_id"]
+      metadata["repo_id"] = nil
+    end
+
+    metadata["repo_ids"] = repo_ids.map(&:to_s)
+  end
+
+  def remove_repo_id_from_metadata(repo_id)
+    metadata["repo_ids"] = current_repo_ids.reject do |id|
+      id.to_s == repo_id.to_s
+    end
+  end
+end
--- a/app/services/repo_subscriber.rb
+++ b/app/services/repo_subscriber.rb
@@ -14,10 +14,10 @@ class RepoSubscriber
   end
 
   def unsubscribe
-    stripe_subscription = payment_gateway_customer.subscriptions.retrieve(
-      repo.subscription.stripe_subscription_id
-    )
-    stripe_subscription.delete
+    payment_gateway_subscription = payment_gateway_customer.
+      retrieve_subscription(repo.subscription.stripe_subscription_id)
+
+    payment_gateway_subscription.unsubscribe(repo.id)
 
     repo.subscription.destroy
   rescue => error
@@ -28,15 +28,21 @@ class RepoSubscriber
   private
 
   def create_subscription
-    stripe_subscription = stripe_customer.subscriptions.create(
+    payment_gateway_subscription = customer.find_or_create_subscription(
       plan: repo.plan_type,
-      metadata: { repo_id: repo.id }
+      repo_id: repo.id,
     )
 
-    create_subscription_record(stripe_subscription.id)
+    payment_gateway_subscription.subscribe(repo.id)
+
+    repo.create_subscription!(
+      user_id: user.id,
+      stripe_subscription_id: payment_gateway_subscription.id,
+      price: repo.plan_price,
+    )
   rescue => error
     report_exception(error)
-    stripe_subscription.try(:delete)
+    payment_gateway_subscription.try(:delete)
     nil
   end
 
@@ -47,16 +53,18 @@ class RepoSubscriber
     )
   end
 
-  def stripe_customer
-    if user.stripe_customer_id.present?
-      payment_gateway_customer
-    else
-      create_stripe_customer
+  def customer
+    @customer ||= begin
+      if user.stripe_customer_id.present?
+        payment_gateway_customer
+      else
+        create_stripe_customer
+      end
     end
   end
 
   def payment_gateway_customer
-    @payment_gateway_customer ||= PaymentGatewayCustomer.new(user).customer
+    @payment_gateway_customer ||= PaymentGatewayCustomer.new(user)
   end
 
   def create_stripe_customer
@@ -68,7 +76,7 @@ class RepoSubscriber
 
     user.update(stripe_customer_id: stripe_customer.id)
 
-    stripe_customer
+    PaymentGatewayCustomer.new(user, customer: stripe_customer)
   end
 
   def create_subscription_record(stripe_subscription_id)
--- a/app/services/update_stripe_metadata.rb
+++ b/app/services/update_stripe_metadata.rb
@@ -11,7 +11,7 @@ class UpdateStripeMetadata
       )
 
       if stripe_subscription
-        stripe_subscription.metadata = { repo_id: repo.id }
+        stripe_subscription.metadata = { repo_ids: [repo.id.to_s] }
         stripe_subscription.save
       end
     end
--- a/spec/models/payment_gateway_customer_spec.rb
+++ b/spec/models/payment_gateway_customer_spec.rb
@@ -106,4 +106,18 @@ describe PaymentGatewayCustomer do
       expect(customer).to be payment_gateway_customer.customer
     end
   end
+
+  describe "#new with_customer" do
+    it "returns a new instance with customer set" do
+      user = build_stubbed(:user, stripe_customer_id: nil)
+      customer = :customer
+
+      payment_gateway_customer = PaymentGatewayCustomer.new(
+        user,
+        customer: customer,
+      )
+
+      expect(payment_gateway_customer.customer).to eq customer
+    end
+  end
 end
--- a/spec/models/payment_gateway_subscription_spec.rb
+++ b/spec/models/payment_gateway_subscription_spec.rb
@@ -1,0 +1,77 @@
+require "rails_helper"
+
+describe PaymentGatewaySubscription do
+  context ".subscribe" do
+    context "existing subscription" do
+      it "appends id to repo_ids metadata" do
+        stripe_subscription = MockStripeSubscription.new(repo_ids: [1])
+        subscription = PaymentGatewaySubscription.new(stripe_subscription)
+
+        expect(stripe_subscription.metadata["repo_ids"]).to eq ["1"]
+
+        subscription.subscribe(2)
+
+        expect(stripe_subscription.metadata["repo_ids"]).to eq ["1", "2"]
+      end
+
+      it "converts legacy format to new format" do
+        legacy_subscription = MockLegacyStripeSubscription.new(repo_id: 1)
+        subscription = PaymentGatewaySubscription.new(legacy_subscription)
+
+        expect(legacy_subscription.metadata["repo_id"]).to eq "1"
+
+        subscription.subscribe(2)
+
+        expect(legacy_subscription.metadata["repo_id"]).to be_nil
+        expect(legacy_subscription.metadata["repo_ids"]).to eq ["1", "2"]
+      end
+    end
+  end
+
+  context ".unsubscribe" do
+    it "removes repo_id from repo_ids" do
+      stripe_subscription = MockStripeSubscription.new(repo_ids: [1, 2])
+      allow(stripe_subscription).to receive(:delete)
+      allow(stripe_subscription).to receive(:save)
+      subscription = PaymentGatewaySubscription.new(stripe_subscription)
+
+      subscription.unsubscribe(2)
+
+      expect(stripe_subscription.metadata["repo_ids"]).to eq ["1"]
+      expect(stripe_subscription).not_to have_received(:delete)
+      expect(stripe_subscription).to have_received(:save)
+    end
+
+    it "doesn't blow up when unsubscribing from a legacy subscription" do
+      legacy_subscription = MockLegacyStripeSubscription.new(repo_id: 1)
+      allow(legacy_subscription).to receive(:delete)
+      allow(legacy_subscription).to receive(:save)
+      subscription = PaymentGatewaySubscription.new(legacy_subscription)
+
+      subscription.unsubscribe(1)
+
+      expect(legacy_subscription).to have_received(:delete)
+      expect(legacy_subscription).not_to have_received(:save)
+    end
+  end
+
+  class MockStripeSubscription
+    attr_accessor :quantity, :metadata
+
+    def initialize(repo_ids:)
+      @quantity = repo_ids.count
+      @metadata = { "repo_ids" => repo_ids.map(&:to_s) }
+    end
+
+    def save; end
+
+    def delete; end
+  end
+
+  class MockLegacyStripeSubscription < MockStripeSubscription
+    def initialize(repo_id:)
+      @quantity = 1
+      @metadata = { "repo_id" => repo_id.to_s }
+    end
+  end
+end
--- a/spec/services/repo_subscriber_spec.rb
+++ b/spec/services/repo_subscriber_spec.rb
@@ -3,34 +3,71 @@ require "rails_helper"
 describe RepoSubscriber do
   describe ".subscribe" do
     context "when Stripe customer exists" do
-      it "creates a new Stripe subscription and repo subscription" do
-        repo = create(:repo, private: true)
-        user =
-          create(:user, stripe_customer_id: stripe_customer_id, repos: [repo])
-        stub_customer_find_request
-        update_request = stub_customer_update_request
-        subscription_request = stub_subscription_create_request(
-          plan: repo.plan_type,
-          repo_id: repo.id,
-        )
+      context "when a subscription doesn't exist" do
+        it "creates a new Stripe subscription and repo subscription" do
+          repo = create(:repo, private: true)
+          user = create(
+            :user,
+            stripe_customer_id: stripe_customer_id,
+            repos: [repo],
+          )
+          stub_customer_find_request
+          update_request = stub_customer_update_request
+          subscription_request = stub_subscription_create_request(
+            plan: repo.plan_type,
+            repo_ids: [repo.id],
+          )
 
-        RepoSubscriber.subscribe(repo, user, "cardtoken")
+          RepoSubscriber.subscribe(repo, user, "cardtoken")
 
-        expect(subscription_request).to have_been_requested
-        expect(update_request).not_to have_been_requested
-        expect(repo.subscription.stripe_subscription_id).
-          to eq(stripe_subscription_id)
-        expect(repo.subscription_price).to(eq(Plan::PRICES[:private]))
+          expect(subscription_request).to have_been_requested
+          expect(update_request).not_to have_been_requested
+          expect(repo.subscription.stripe_subscription_id).
+            to eq(stripe_subscription_id)
+          expect(repo.subscription_price).to(eq(Plan::PRICES[:private]))
+        end
+      end
+
+      context "when a subscription exists" do
+        it "increments the Stripe subscription and updates repo subscription" do
+          repo = create(:repo, private: true)
+          user = create(
+            :user,
+            stripe_customer_id: stripe_customer_id,
+            repos: [repo],
+          )
+          stub_customer_find_request_with_subscriptions
+          repo_ids = [repo.id]
+          subscription_update_request = stub_subscription_update_request(
+            quantity: 2,
+            repo_ids: repo_ids,
+          )
+          subscription_create_request = stub_subscription_create_request(
+            plan: repo.plan_type,
+            repo_ids: repo_ids,
+          )
+
+          RepoSubscriber.subscribe(repo, user, "cardtoken")
+
+          expect(subscription_create_request).not_to have_been_requested
+          expect(subscription_update_request).to have_been_requested
+          expect(repo.subscription.stripe_subscription_id).
+            to eq(stripe_subscription_id)
+          expect(repo.subscription_price).to(eq(Plan::PRICES[:private]))
+        end
       end
     end
 
     context "when Stripe customer does not exist" do
       it "creates a new Stripe customer, subscription and repo subscription" do
         repo = create(:repo)
-        user = create(:user, repos: [repo], stripe_customer_id: "")
+        user = create(:user, repos: [repo], stripe_customer_id: "",)
         customer_request = stub_customer_create_request(user)
         subscription_request =
-          stub_subscription_create_request(plan: repo.plan_type)
+          stub_subscription_create_request(
+            plan: repo.plan_type,
+            repo_ids: [repo.id],
+        )
 
         RepoSubscriber.subscribe(repo, user, "cardtoken")
 
@@ -72,7 +109,10 @@ describe RepoSubscriber do
         repo = create(:repo)
         user = create(:user, repos: [repo])
         stub_customer_create_request(user)
-        stub_subscription_create_request(plan: repo.plan_type)
+        stub_subscription_create_request(
+          plan: repo.plan_type,
+          repo_ids: [repo.id],
+        )
         stripe_delete_request = stub_subscription_delete_request
         allow(repo).to receive(:create_subscription!).and_raise(StandardError)
 
@@ -101,38 +141,45 @@ describe RepoSubscriber do
   end
 
   describe ".unsubscribe" do
-    it "deletes Stripe subscription" do
-      user = create(:user, stripe_customer_id: stripe_customer_id)
-      subscription = create(
-        :subscription,
-        stripe_subscription_id: stripe_subscription_id,
-        user: user
-      )
-      user.repos << subscription.repo
-      stub_customer_find_request
-      stub_subscription_find_request(subscription)
-      stripe_delete_request = stub_subscription_delete_request
+    context "when the subscription quantity is 1" do
+      it "deletes Stripe subscription" do
+        subscription = subscription_with_user
+        stub_customer_find_request
+        stub_subscription_find_request(subscription)
+        stripe_delete_request = stub_subscription_delete_request
 
-      RepoSubscriber.unsubscribe(subscription.repo, user)
+        RepoSubscriber.unsubscribe(subscription.repo, subscription.user)
 
-      expect(stripe_delete_request).to have_been_requested
+        expect(stripe_delete_request).to have_been_requested
+      end
+
+      it "soft deletes the Repo subscription" do
+        subscription = subscription_with_user
+        stub_customer_find_request
+        stub_subscription_find_request(subscription)
+        stub_subscription_delete_request
+
+        RepoSubscriber.unsubscribe(subscription.repo, subscription.user)
+
+        expect(subscription.reload).to be_deleted
+      end
     end
 
-    it "soft deletes the Repo subscription" do
-      user = create(:user, stripe_customer_id: stripe_customer_id)
-      subscription = create(
-        :subscription,
-        stripe_subscription_id: stripe_subscription_id,
-        user: user
-      )
-      user.repos << subscription.repo
-      stub_customer_find_request
-      stub_subscription_find_request(subscription)
-      stub_subscription_delete_request
+    context "when the subscription quantity is greater 1" do
+      it "decrements the Stripe subscription quantity" do
+        subscription = subscription_with_user
+        stub_customer_find_request
+        stub_subscription_find_request(subscription, quantity: 2)
+        stripe_delete_request = stub_subscription_delete_request
+        subscription_update_request = stub_subscription_update_request(
+          quantity: 1
+        )
 
-      RepoSubscriber.unsubscribe(subscription.repo, user)
+        RepoSubscriber.unsubscribe(subscription.repo, subscription.user)
 
-      expect(subscription.reload).to be_deleted
+        expect(stripe_delete_request).not_to have_been_requested
+        expect(subscription_update_request).to have_been_requested
+      end
     end
 
     context "when Stripe unsubscription fails" do
@@ -159,5 +206,16 @@ describe RepoSubscriber do
         expect(Raven).to have_received(:capture_exception)
       end
     end
+  end
+
+  def subscription_with_user
+    user = create(:user, stripe_customer_id: stripe_customer_id)
+    subscription = create(
+      :subscription,
+      stripe_subscription_id: stripe_subscription_id,
+      user: user,
+    )
+    user.repos << subscription.repo
+    subscription
   end
 end
--- a/spec/support/helpers/stripe_api_helper.rb
+++ b/spec/support/helpers/stripe_api_helper.rb
@@ -35,6 +35,17 @@ module StripeApiHelper
     )
   end
 
+  def stub_customer_find_request_with_subscriptions
+    stub_request(:get, "#{stripe_base_url}/#{stripe_customer_id}").with(
+      headers: { "Authorization" => "Bearer #{ENV['STRIPE_API_KEY']}" }
+    ).to_return(
+      status: 200,
+      body: File.read(
+        "spec/support/fixtures/stripe_customer_find_with_subscriptions.json"
+      )
+    )
+  end
+
   def stub_customer_update_request(attrs = { card: "card-token" })
     stub_request(
       :post,
@@ -66,11 +77,11 @@ module StripeApiHelper
     )
   end
 
-  def stub_subscription_create_request(plan: "free", repo_id: nil)
-    body = { "plan" => plan }
-    if repo_id
-      body.merge!("metadata" => { "repo_id" => repo_id.to_s })
-    end
+  def stub_subscription_create_request(plan: "free", repo_ids: [])
+    body = {
+      "plan" => plan,
+      "metadata" => { "repo_ids" => repo_ids.map(&:to_s) }
+    }
     stub_request(
       :post,
       "#{stripe_base_url}/#{stripe_customer_id}/subscriptions"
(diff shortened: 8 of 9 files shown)


spec/support/helpers/stripe_api_helper.rb
```diff
@@ -35,6 +35,20 @@ def stub_customer_find_request
     )
   end
 
+  def stub_customer_find_request_with_subscriptions
```
jakecraige: I'm going to let this go since it's a testing stub method. Making it shorter doesn't help readability.

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -65,17 +79,32 @@ def stub_subscription_create_request(plan: "free", repo_id: nil)
     )
   end
 
-  def stub_subscription_find_request(subscription)
+  def stub_subscription_update_request(quantity: 1)
```
jakecraige: I'm going to let this go since it's a testing stub method. Making it shorter doesn't help readability.

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -65,19 +76,34 @@ def stub_subscription_create_request(plan: "free", repo_id: nil)
...
+      body: { "quantity" => quantity.to_s },
       headers: { "Authorization" => "Bearer #{ENV["STRIPE_API_KEY"]}" }
     ).to_return(
       status: 200,
-      body: File.read("spec/support/fixtures/stripe_subscription_find.json"),
+      body: File.read("spec/support/fixtures/stripe_subscription_update.json"),
     )
   end
 
+  def stub_subscription_find_request(subscription, quantity: 1)
```
jakecraige: I'm going to let this go since it's a testing stub method. Making it shorter doesn't help readability.

gylaz: Yep. Just an FYI we turned this setting on just for this repo to test drive it.

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -65,19 +76,34 @@ def stub_subscription_create_request(plan: "free", repo_id: nil)
...
+    body = JSON.parse(
+      File.read("spec/support/fixtures/stripe_subscription_find.json")
+    )
+    body["quantity"] = quantity
+    stub_request(
+      :get,
+      "#{stripe_base_url}/#{stripe_customer_id}/"\
+        "subscriptions/#{subscription.stripe_subscription_id}"
```
gylaz: How about capturing this into a var?

gylaz:
I like it! :+1:

app/services/repo_subscriber.rb
```diff
@@ -10,16 +10,19 @@ def self.unsubscribe(repo, user)
   end
 
   def subscribe
-    customer = if user.stripe_customer_id.present?
-      payment_gateway_customer
-    else
-      create_stripe_customer
+    stripe_subscription = customer.subscriptions.detect do |subscription|
+      subscription.plan.id == repo.plan_type
```
gabebw: Can this be `subscription.plan_id` instead?

gabebw: Oh, this is from Stripe.

app/services/repo_subscriber.rb
```diff
@@ -36,7 +39,13 @@ def unsubscribe
     stripe_subscription = payment_gateway_customer.subscriptions.retrieve(
       repo.subscription.stripe_subscription_id
     )
-    stripe_subscription.delete
+
+    if stripe_subscription.quantity > 1
```
gabebw: What if we moved this `if/else` around removing a subscription, and the `if/else` around adding a subscription, into the subscription itself? Then we could do `stripe_susbcription.increment_quantity` and `stripe_subscription.decrement_quantity` and have it just do the right thing.

Those if/else's feel like they could Tell more instead of Asking.

spec/services/repo_subscriber_spec.rb
```diff
@@ -3,24 +3,53 @@
...
-        subscription_request = stub_subscription_create_request(
-          plan: repo.plan_type,
-          repo_id: repo.id,
-        )
-
-        RepoSubscriber.subscribe(repo, user, "cardtoken")
+      context "when a subscription doesn't exist" do
+        it "creates a new Stripe subscription and repo subscription" do
+          repo = create(:repo, private: true)
+          user =
```
gabebw: how about:

```
user = create(
  :user,
  stripe_customer_id: stripe_customer_id,
  repos: [repo]
)
```

gabebw: That's what I see below, actually.

spec/support/helpers/stripe_api_helper.rb
jakecraige: sounds good

app/services/repo_subscriber.rb
jakecraige: Yeah. I've considered creating our own payment gateway subscription object so we can do stuff like that. Thoughts?

spec/services/repo_subscriber_spec.rb
jakecraige: sounds good to me, will do

app/services/repo_subscriber.rb
jakecraige: This relates to my previous comment about having a PaymentGateway subscriber object. Since we're currently dealing with a `Stripe::Subscription`, we can't add those methods. I'm going to implement that object because it will help clean up this part of the code.

app/services/repo_subscriber.rb
gabebw: I have never regretted wrapping third-party objects in my own custom object.

app/models/payment_gateway_customer.rb
```diff
@@ -1,6 +1,12 @@
 class PaymentGatewayCustomer
   pattr_initialize :user
 
+  def self.new_with_customer(user, customer)
+    payment_gateway_customer = new(user)
+    payment_gateway_customer.instance_variable_set(:@customer, customer)
```
gylaz: Why not make an setter for this?

app/models/payment_gateway_customer.rb
```diff
@@ -19,6 +25,32 @@ def customer
     end
   end
 
+  def find_or_create_subscription(plan:, repo_id:)
+    subscriptions.detect do |subscription|
+      subscription.plan == plan
+    end || create_subscription(plan: plan, metadata: { repo_id: repo_id })
```
gylaz: The `||` after end feels weird -- not a common style approach. How about this formatting?
```rb
subscriptions.detect { |subscription| subscription.plan == plan } ||
  create_subscription(plan: plan, metadata: { repo_id: repo_id })
```
or maybe extract the `detect` into it's own method.

app/models/payment_gateway_customer.rb
```diff
@@ -19,6 +25,32 @@ def customer
...
+    customer.subscriptions.map do |subscription|
+      PaymentGatewaySubscription.new(subscription)
+    end
+  end
+
+  def create_subscription(options)
+    subscription = PaymentGatewaySubscription.new(
+      customer.subscriptions.create(options)
+    )
+    subscription.new_subscription = true
```
gylaz: Any reason not to pass this to the `initialize`? We could default it to `false`.

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,45 @@
...
+  def plan
+    subscription.plan.id
+  end
+
+  def delete
+    subscription.delete
+  end
+
+  def new_subscription?
+    !!new_subscription
```
gylaz: Don't we expect `new_subscription` to be a boolean? Can we drop `!!` in that case?

app/services/repo_subscriber.rb
```diff
@@ -10,33 +10,31 @@ def self.unsubscribe(repo, user)
...
-    stripe_subscription.try(:delete)
+    payment_gateway_subscription.try(:delete)
     nil
   end
 
   def unsubscribe
-    stripe_subscription = payment_gateway_customer.subscriptions.retrieve(
-      repo.subscription.stripe_subscription_id
+    payment_gateway_subscription =
+      payment_gateway_customer.retrieve_subscription(
```
gylaz: How about leaving `payment_gateway_customer` on line 32, and breaking the line after the `.`?

spec/models/payment_gateway_customer_spec.rb
```diff
@@ -78,4 +78,17 @@
       expect(customer).to be payment_gateway_customer.customer
     end
   end
+
+  describe "#new_with_customer" do
+    it "returns a new instance with customer set" do
+      user = build_stubbed(:user, stripe_customer_id: nil)
+      customer = :customer
+      payment_gateway_customer = PaymentGatewayCustomer.new_with_customer(
```
gylaz: Newline between setup and exercise phases.

spec/services/repo_subscriber_spec.rb
```diff
@@ -85,38 +117,63 @@
...
+    context "when the subscription quantity is greater 1" do
+      it "decrements the Stripe subscription quantity" do
+        user = create(:user, stripe_customer_id: stripe_customer_id)
+        subscription = create(
+          :subscription,
+          stripe_subscription_id: stripe_subscription_id,
+          user: user
+        )
+        user.repos << subscription.repo
+        stub_customer_find_request
```
gylaz: Looks like 3 tests so far duplicate this much setup logic. Can we extract this to a function?

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -65,17 +76,31 @@ def stub_subscription_create_request(plan: "free", repo_id: nil)
...
+    )
+  end
+
+  def stub_subscription_find_request(subscription, quantity: 1)
+    body = JSON.parse(
+      File.read("spec/support/fixtures/stripe_subscription_find.json")
     )
+    body["quantity"] = quantity
+    request_url = "#{stripe_base_url}/#{stripe_customer_id}/"\
+                  "subscriptions/#{subscription.stripe_subscription_id}"
```
gylaz: Indent this line 2 spaces relative to the parent.

spec/support/helpers/stripe_api_helper.rb
gabebw: One possibility: make a spec helper so you can do `to_return(status: 200, body: fixture_file("stripe_subscription_update")`. But I agree, not a huge deal.

spec/support/helpers/stripe_api_helper.rb
jakecraige: :+1:

gabebw: Also, how about a space before that backslash?

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -65,17 +76,31 @@ def stub_subscription_create_request(plan: "free", repo_id: nil)
...
+
+    stub_request(:get, request_url).with(
+      headers: { "Authorization" => "Bearer #{ENV['STRIPE_API_KEY']}" }
```
gabebw: The ENV quotes can be double quotes too.

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,45 @@
+class PaymentGatewaySubscription
+  pattr_initialize :subscription
+
+  attr_accessor :new_subscription
+
+  def subscribe
+    if !new_subscription?
```
gabebw: This is only ever used in the negative. How about `if existing_subscription?` instead?

spec/services/repo_subscriber_spec.rb
jakecraige: Sounds good

app/services/repo_subscriber.rb
jakecraige: I was having trouble formatting that. That'll do :+1:

app/models/payment_gateway_subscription.rb
jakecraige: Yeah I do expect it to be a boolean, I was being safe. If we had types I could be sure of it, lol. They `!!` isn't necessary. I'll remove it

app/models/payment_gateway_customer.rb
jakecraige: I can change it to the format you recommended. No worries. Thanks

app/models/payment_gateway_customer.rb
jakecraige: That would be a little cleaner. Thanks

app/models/payment_gateway_customer.rb
jakecraige: Good catch. I like that much better

spec/models/payment_gateway_customer_spec.rb
jakecraige: Thanks

app/services/repo_subscriber.rb
```diff
@@ -10,33 +10,29 @@ def self.unsubscribe(repo, user)
...
     nil
   end
 
   def unsubscribe
-    stripe_subscription = payment_gateway_customer.subscriptions.retrieve(
-      repo.subscription.stripe_subscription_id
-    )
-    stripe_subscription.delete
+    payment_gateway_subscription = payment_gateway_customer.
+      retrieve_subscription(repo.subscription.stripe_subscription_id)
```
jakecraige: @gylaz Do you know to fix this one? I believe this is the change you recommended

gylaz: Hmm, I'm not really sure about this one. I says `operands` -- which would be what in this case?

The cop in question is `Style/MultilineOperationIndentation`, and its description `Checks indentation of binary operations that span more than one line`. I'm not seeing any binary operations here.

Ah, found [this](https://github.com/bbatsov/rubocop/issues/1633#issuecomment-73067109). So, let's change our setting to `indented` (since that's what thoughtbot does), also for thoughtbot config.

jakecraige: Sweet. I'll change it

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -48,11 +59,11 @@ def stub_customer_update_request(card_token = "cardtoken")
     )
   end
 
-  def stub_subscription_create_request(plan: "free", repo_id: nil)
-    body = { "plan" => plan }
-    if repo_id
-      body.merge!("metadata" => { "repo_id" => repo_id.to_s })
-    end
+  def stub_subscription_create_request(plan: "free", repo_ids: [])
```
jakecraige: I'm going to let this go since it's a testing stub method. Making it shorter doesn't help readability.

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -65,19 +76,39 @@ def stub_subscription_create_request(plan: "free", repo_id: nil)
     )
   end
 
-  def stub_subscription_find_request(subscription)
+  def stub_subscription_update_request(quantity: 1, repo_ids: nil)
```
jakecraige: I'm going to let this go since it's a testing stub method. Making it shorter doesn't help readability.

spec/support/helpers/stripe_api_helper.rb
gabebw: I like this setting. Might be nice to have a "skip spec/support" sub-setting, if that's possible.

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def append_repo_id_to_metadata(repo_id)
+    repo_ids = current_repo_ids.push(repo_id)
+
+    if metadata.repo_id
+      metadata.repo_id = nil
+    end
+
+    metadata.repo_ids = repo_ids.map(&:to_s)
+  rescue NoMethodError
```
gabebw: When might this occur?

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,70 @@
...
+    metadata.repo_ids = [repo_id.to_s]
+  end
+
+  def remove_repo_id_from_metadata(repo_id)
+    metadata.repo_ids = metadata.repo_ids.reject do |id|
+      id.to_s == repo_id.to_s
+    end
+  rescue NoMethodError
```
gabebw: When might this occur?

app/models/payment_gateway_subscription.rb
jakecraige: It's a `Stripe::StripeObject` and if you try to access a key that doesn't exist it throws the NoMethodError. I just checked out the source and there' s a way I should be able to set / access it without this. I'll give that a shot shortly

app/models/payment_gateway_subscription.rb
jakecraige: https://github.com/thoughtbot/hound/pull/720#discussion_r26342224

app/services/repo_subscriber.rb
graysonwright: There's a card for it: https://trello.com/c/8plKSKKM/534-update-ruby-rules-to-ignore-operands-on-multiple-lines

app/models/payment_gateway_subscription.rb
jakecraige: @gabebw Fixed now https://github.com/thoughtbot/hound/blob/jc-increment-subscription/app/models/payment_gateway_subscription.rb#L57-L71

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,72 @@
...
+  delegate :id, :metadata, :save, :delete, to: :subscription
+
+  def initialize(subscription, new_subscription: false)
+    @subscription = subscription
+    @new_subscription = new_subscription
+  end
+
+  def subscribe(repo_id)
+    append_repo_id_to_metadata(repo_id)
+    if !new_subscription?
```
teoljungberg: Why not use `unless`?

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,72 @@
...
+  def new_subscription?
+    @new_subscription
+  end
+
+  private
+
+  def current_repo_ids
+    repo_ids = metadata["repo_ids"] || []
+
+    if repo_ids.length == 0 && metadata["repo_id"]
```
teoljungberg: How about checking the size with `.zero?` instead?

app/models/payment_gateway_customer.rb
```diff
@@ -19,6 +29,30 @@ def customer
     end
   end
 
+  def find_or_create_subscription(plan:, repo_id:)
+    subscriptions.detect { |subscription| subscription.plan == plan } ||
+      create_subscription(plan: plan, metadata: { repo_ids: [repo_id] })
```
teoljungberg: How about extracting the creation of a subscription into a private method?

spec/models/payment_gateway_subscription_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def save; end
+
+    def delete; end
+  end
+
+  class MockLegacyStripeSubscription < MockStripeSubscription
+    def initialize(repo_id:)
+      @quantity = 1
+      @metadata = { "repo_id" => repo_id.to_s }
```
teoljungberg: Minor nitpick, ruby 2.2 allows us to write `{ "foo": 1 ]` instead of having the hash-rocket

spec/services/repo_subscriber_spec.rb
```diff
@@ -3,24 +3,58 @@
...
-        )
-
-        RepoSubscriber.subscribe(repo, user, "cardtoken")
+      context "when a subscription doesn't exist" do
+        it "creates a new Stripe subscription and repo subscription" do
+          repo = create(:repo, private: true)
+          user = create(
+            :user,
+            stripe_customer_id: stripe_customer_id,
+            repos: [repo]
```
teoljungberg: Missing a trailing `,`

spec/services/repo_subscriber_spec.rb
```diff
@@ -3,24 +3,58 @@
...
-        expect(repo.subscription.stripe_subscription_id).
-          to eq(stripe_subscription_id)
-        expect(repo.subscription_price).to(eq(Plan::PRICES[:private]))
+      context "when a subscription exists" do
+        it "increments the Stripe subscription and updates repo subscription" do
+          repo = create(:repo, private: true)
+          user = create(
+            :user,
+            stripe_customer_id: stripe_customer_id,
+            repos: [repo]
```
teoljungberg: Missing a trailing `,`

spec/services/repo_subscriber_spec.rb
```diff
@@ -3,24 +3,58 @@
...
+          )
+          stub_customer_find_request_with_subscriptions
+          repo_ids = [repo.id]
+          subscription_update_request = stub_subscription_update_request(
+            quantity: 2,
+            repo_ids: repo_ids
```
teoljungberg: Missing a trailing `,`

spec/services/repo_subscriber_spec.rb
```diff
@@ -30,7 +64,10 @@
         user = create(:user, repos: [repo], stripe_customer_id: "")
         customer_request = stub_customer_create_request(user)
         subscription_request =
-          stub_subscription_create_request(plan: repo.plan_type)
+          stub_subscription_create_request(
+            plan: repo.plan_type,
+            repo_ids: [repo.id]
```
teoljungberg: Missing a trailing `,`

teoljungberg:
There are some trailing commas missing in the param-declarations, but I wouldn't want to comment on all of them

spec/models/payment_gateway_subscription_spec.rb
jakecraige: Cool, I didn't know that. Unfortunately that breaks the tests. It turns the key into a symbol instead of a string and it has to be a string

app/models/payment_gateway_customer.rb
jakecraige: good catch. I was originally using it publicly but changed not to anymore. I'll change it

app/models/payment_gateway_subscription.rb
jakecraige: Yeah I can do that

app/models/payment_gateway_subscription.rb
jakecraige: I have a weird aversion to unless. In this case it's probably better to use it though. I'll switch it

salbertson:
@jakecraige, @gylaz brought up a potential issue on our call today. Check out https://trello.com/c/XNL81IPQ/539-increment-subscription-quantity-instead-of-charging-for-each-repo-individually.

spec/support/helpers/stripe_api_helper.rb
salbertson: Ha, what do you think about this preference in general? This is something we turned on for Hound to test it out.

spec/support/helpers/stripe_api_helper.rb
salbertson: @gabebw, good idea. https://github.com/thoughtbot/hound/pull/735

app/services/repo_subscriber.rb
salbertson: :+1:

jakecraige:
@salbertson @gylaz Is anything holding this up? I think what you referenced was related to updating the metadata? I believe I mentioned to @gylaz at some point that this is doing that. It converts `repo_id` to `repo_ids` when it encounters it.

app/models/payment_gateway_customer.rb
```diff
@@ -9,6 +14,10 @@ def card_last4
     default_card.last4
   end
 
+  def customer=(customer)
+    @customer = customer
+  end
```
gylaz: Do we need this method?

app/models/payment_gateway_customer.rb
```diff
@@ -19,6 +28,23 @@ def customer
     end
   end
 
+  def find_or_create_subscription(plan:, repo_id:)
+    subscriptions.detect { |subscription| subscription.plan == plan } ||
+      create_subscription(plan: plan, metadata: { repo_ids: [repo_id] },)
```
gylaz: We don't generally add a trailing comma, unless each argument is on its own line.

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,72 @@
...
+  def new_subscription?
+    @new_subscription
+  end
+
+  private
+
+  def current_repo_ids
+    repo_ids = metadata["repo_ids"] || []
+
+    if repo_ids.length.zero? && metadata["repo_id"]
```
gylaz: In Ruby, `size` is preferred of `length`.

gylaz: And since this is an array: `empty?`

app/models/payment_gateway_customer.rb
jakecraige: We can get around it, but you recommended it here: https://github.com/thoughtbot/hound/pull/720#discussion_r26323380

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,72 @@
...
+      repo_ids = [metadata["repo_id"]]
+    end
+
+    repo_ids
+  end
```
gylaz: Can this method be replaced with?
```rb
Array(metadata["repo_ids"] || metadata["repo_id"])
```

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def append_repo_id_to_metadata(repo_id)
+    repo_ids = current_repo_ids.push(repo_id)
```
gylaz: This is a bit misleading since `push` will modify `current_repo_ids`. I think `+` is more immutable-like?

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    if metadata["repo_id"]
+      metadata["repo_id"] = nil
+    end
+
+    metadata["repo_ids"] = repo_ids.map(&:to_s)
```
gylaz: Do we need this `map` call?

app/services/update_stripe_metadata.rb
```diff
@@ -11,7 +11,7 @@ def self.run
       )
 
       if stripe_subscription
-        stripe_subscription.metadata = { repo_id: repo.id }
+        stripe_subscription.metadata = { repo_ids: [repo.id.to_s] }
```
gylaz: Why is `to_s` necessary?

spec/models/payment_gateway_subscription_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+describe PaymentGatewaySubscription do
+  context ".subscribe" do
+    context "existing subscription" do
+      it "appends id to repo_ids metadata" do
+        stripe_subscription = MockStripeSubscription.new(repo_ids: [1])
+        subscription = PaymentGatewaySubscription.new(stripe_subscription)
+
+        expect(stripe_subscription.metadata["repo_ids"]).to eq ["1"]
+        subscription.subscribe(2)
```
gylaz: Separate execution step with newlines.

spec/models/payment_gateway_subscription_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        expect(stripe_subscription.metadata["repo_ids"]).to eq ["1", "2"]
+      end
+
+      it "converts legacy format to new format" do
+        legacy_subscription = MockLegacyStripeSubscription.new(repo_id: 1)
+        subscription = PaymentGatewaySubscription.new(legacy_subscription)
+
+        expect(legacy_subscription.metadata["repo_id"]).to eq "1"
+        subscription.subscribe(2)
```
gylaz: Separate execution step with newlines.

spec/models/payment_gateway_subscription_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        expect(legacy_subscription.metadata["repo_id"]).to be_falsy
```
gylaz: `to be nil`?

app/models/payment_gateway_subscription.rb
jakecraige: Yes it can. That's beautiful. Nice!

spec/support/helpers/stripe_api_helper.rb
```diff
@@ -65,19 +76,39 @@ def stub_subscription_create_request(plan: "free", repo_id: nil)
...
     )
   end
 
+  def stub_subscription_find_request(subscription, quantity: 1)
+    body = JSON.parse(
+      File.read("spec/support/fixtures/stripe_subscription_find.json")
+    )
+    body["quantity"] = quantity
+    request_url = "#{stripe_base_url}/#{stripe_customer_id}/"\
+                    "subscriptions/#{subscription.stripe_subscription_id}"
```
gylaz: Misindented.

app/models/payment_gateway_customer.rb
jakecraige: That makes sense. I'll take care of it

app/models/payment_gateway_subscription.rb
jakecraige: That's much nicer. Thanks!

app/services/update_stripe_metadata.rb
jakecraige: Stripe seems to return metadata as strings to matter what we send it.

Current code: https://github.com/thoughtbot/hound/blob/7164d4169fbb8e3be6219e714487ef6250c81931/app/services/update_stripe_metadata.rb#L14

Notice it's an integer.

Here's the relevant part of the curl response

```js
    "metadata": {
      "repo_id": "277648"
    }
```

I noticed this same phenomenon with arrays hence all metadata attributes being `to_s`'d

app/models/payment_gateway_subscription.rb
jakecraige: See https://github.com/thoughtbot/hound/pull/720#discussion_r26886745 for an explanation

app/models/payment_gateway_subscription.rb
jakecraige: Agreed. Changing to `repo_ids = current_repo_ids + [repo_id]`

spec/models/payment_gateway_subscription_spec.rb
jakecraige: that is what I'm testing :+1:  I'll change it out

spec/models/payment_gateway_subscription_spec.rb
jakecraige: Good catch.

app/models/payment_gateway_customer.rb
gylaz: In that particular case we can pass the customer into the initializer, right?

jakecraige: We actually do pass it to the initializer. I just kept this setter around when I should have removed it and set the instance variable in the initialize method. Fixing

salbertson:
@jakecraige are we still creating a new subscription record in our database each time a private repo is enabled? [It looks like we are](https://github.com/thoughtbot/hound/blob/jc-increment-subscription/app/services/repo_subscriber.rb#L20-L24), just making sure. This should address @gylaz's concern about having single subscription records for multiple repos in our database.

salbertson:
@jakecraige @gylaz it looks like there's a lot of code to support updating Stripe's metadata. Is it worth keeping repo IDs up to date on Stripe? I never use that data, does anyone else? I usually look up subscriptions based on `stripe_subscription_id`.

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def plan
+    subscription.plan.id
+  end
+
+  def existing_subscription?
+    !@new_subscription
+  end
+
+  def new_subscription?
```
salbertson: Is this used anywhere outside this class?

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,70 @@
...
+    subscription.quantity -= 1
+    save
+  end
+
+  def plan
+    subscription.plan.id
+  end
+
+  def existing_subscription?
+    !@new_subscription
```
salbertson: Can we tell whether or not it's an existing subscription besides passing an argument to `new`?

gylaz:
@salbertson I do find it useful to see what repo the subscription is for on Stripe's dashboard.

salbertson:
@gylaz, ok cool. Seems worth it then.

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,70 @@
...
+    if subscription.quantity > 1
+      remove_repo_id_from_metadata(repo_id)
+      decrement_quantity
+    else
+      delete
+    end
+  end
+
+  def increment_quantity
+    subscription.quantity += 1
```
srt32: Would we be able to use http://api.rubyonrails.org/classes/ActiveRecord/CounterCache/ClassMethods.html#method-i-increment_counter here and make the increment atomic?

gylaz: It's hard to actually tell, but I believe this is a Strip subscription. Maybe it would be a good idea to name the variable that.

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,70 @@
...
+    save
+  end
+
+  def decrement_quantity
+    subscription.quantity -= 1
+    save
+  end
+
+  def plan
```
srt32: Any reason not to call this method `plan_id1?

app/services/repo_subscriber.rb
```diff
@@ -10,33 +10,29 @@ def self.unsubscribe(repo, user)
...
       user_id: user.id,
-      stripe_subscription_id: stripe_subscription.id,
-      price: repo.plan_price
+      stripe_subscription_id: payment_gateway_subscription.id,
+      price: repo.plan_price,
     )
   rescue => error
     report_exception(error)
-    stripe_subscription.try(:delete)
+    payment_gateway_subscription.try(:delete)
```
srt32: This line is a refactoring but why do try to `delete` the `subscription` on error?

jakecraige: It seems like a safe assumption to make that if we got an uncaught exception here something went wrong and we should delete the subscription if one happened to be created

app/models/payment_gateway_subscription.rb
jakecraige: Looks like it's not. It should be removed.

app/models/payment_gateway_subscription.rb
jakecraige: Not really, we're expected to initialize this class with a valid `Stripe::Subscription`. It's been awhile but I think I tried to come up with a way to not have to do this with no avail. 

I believe it's because the `quantity` starts at one and there's no way to differentiate whether to keep it at 1 or increment it  when we call subscribe on this class.

app/models/payment_gateway_subscription.rb
jakecraige: Yeah we're dealing with a stripe subscription. Renaming it sounds good

app/models/payment_gateway_subscription.rb
jakecraige: The reason was because when you create a subscription you pass in `plan:` but it's really the `plan_id`, so it was to keep it consistent. See [here](https://github.com/thoughtbot/hound/blob/jc-increment-subscription/app/models/payment_gateway_customer.rb#L29) to [here](https://github.com/thoughtbot/hound/blob/jc-increment-subscription/app/models/payment_gateway_customer.rb#L53)

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def current_repo_ids
+    Array(metadata["repo_ids"] || metadata["repo_id"])
+  end
+
+  def append_repo_id_to_metadata(repo_id)
+    repo_ids = current_repo_ids + [repo_id]
+
+    if metadata["repo_id"]
+      metadata["repo_id"] = nil
```
srt32: Would it be better / okay to remove the key (`metadata.delete("repo_id")`) entirely instead of setting it to `nil`?

app/models/payment_gateway_subscription.rb
srt32: Did we want to remove this method?

app/models/payment_gateway_subscription.rb
srt32: @jakecraige were you thinking `stripe_subscription`?

app/models/payment_gateway_customer.rb
```diff
@@ -26,6 +48,13 @@ def update_card(card_token)
 
   private
 
+  def create_subscription(options)
+    PaymentGatewaySubscription.new(
+      customer.subscriptions.create(options),
```
srt32: Would `create!` be appropriate here since we're not checking the return value?

app/models/payment_gateway_subscription.rb
jakecraige: Depends on what you want in stripe. This specific block is for the "old style" storing of the repo id per subscription. This removes it so it's no longer in stripe. That's why we need to set it to nil rather than removing it

app/models/payment_gateway_customer.rb
jakecraige: It actually is being checked where it's used here, see https://github.com/thoughtbot/hound/pull/720/files#r26371515

srt32: I think I'm missing something but don't we use it here (https://github.com/thoughtbot/hound/pull/720/files#diff-3c6a907980d438c060c6d65ee6b1ab81R13) and we're not checking the return?

jakecraige: Ah you're right. That place I linked looked like it was within the `.detect` block. `create!` seems good here

salbertson:
@jakecraige, is this close to being ready to merge? What's left?

jakecraige:
@salbertson Sorry for taking forever to get back to you. I think my open PRs are mostly complete.. I think I should be able to get them finished at HoundConf 2015.

salbertson:
Sounds good, thanks @jakecraige!

app/models/payment_gateway_customer.rb
croaky: @jakecraige Should this be using `create!` now per your last comment?

app/models/payment_gateway_subscription.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def increment_quantity
+    stripe_subscription.quantity += 1
+    save
+  end
+
+  def decrement_quantity
+    stripe_subscription.quantity -= 1
+    save
```
croaky: is `stripe_subscription` an AR model? Could these lines use `increment(:quantity)`?

croaky:
Looks good to me.

app/models/payment_gateway_subscription.rb
jakecraige: No, it's a `Stripe::Subscription`

spec/models/payment_gateway_subscription_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      expect(stripe_subscription.metadata["repo_ids"]).to eq ["1"]
+      expect(stripe_subscription).not_to have_received(:delete)
+      expect(stripe_subscription).to have_received(:save)
+    end
+
+    it "doesn't blow up when unsubscribing from a legacy subscription" do
+      legacy_subscription = MockLegacyStripeSubscription.new(repo_id: 1)
+      legacy_subscription.stub(:delete)
+      legacy_subscription.stub(:save)
```
gylaz: For consistency, can these `stub`s become `allow(legacy_subscription).to receive(:delete)`?

spec/models/payment_gateway_subscription_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    attr_accessor :quantity, :metadata
+
+    def initialize(repo_ids:)
+      @quantity = repo_ids.count
+      @metadata = { "repo_ids" => repo_ids.map(&:to_s) }
+    end
+
+    def save; end
+
+    def delete; end
```
gylaz: Are these needed if we're stubbing them in the tests?

gylaz:
:shipit:

app/models/payment_gateway_customer.rb
jakecraige: @croaky Turns out this is a stripe object so there is no `create!` method. After some digging I determined that any stripe API requests [eventually raise an exception](https://github.com/stripe/stripe-ruby/blob/3106255495d5a078a64810796cb3f9c3053859c7/lib/stripe.rb#L316) if they fail so this `create` behaves like `create!` would.

spec/models/payment_gateway_subscription_spec.rb
jakecraige: Good catch. Fixing.

spec/models/payment_gateway_subscription_spec.rb
jakecraige: We're not stubbing them in the first 2 tests of this file so we do need them.

PR #722, opened 2015-03-12
Change javascript and coffeescript keys in hound.yml

jakecraige:
Kept backwards compatibility with old naming

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -1,11 +1,11 @@
 # Load and parse config files from GitHub repo
 class RepoConfig
   HOUND_CONFIG = ".hound.yml"
-  LANGUAGES = %w(ruby coffee_script java_script scss)
+  LANGUAGES = %w(ruby coffeescript javascript scss)
   FILE_TYPES = {
     "ruby" => "yaml",
-    "java_script" => "json",
-    "coffee_script" => "json",
+    "javascript" => "json",
+    "coffeescript" => "json",
     "scss" => "yaml",
   }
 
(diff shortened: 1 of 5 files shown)


app/models/repo_config.rb
```diff
@@ -103,4 +103,28 @@ def parse_json(content)
...
+    end
+
+    converted_config
+  end
+
+  def legacy_language_camelize(language)
+    case language.downcase
+    when "coffeescript"
+      "CoffeeScript"
+    when "javascript"
```
gylaz: Method name might be confusing since these aren't really legacy names.

gylaz:
One comment. :+1:

app/models/repo_config.rb
jakecraige: That makes sense. Fixed

teoljungberg:
How about renaming `app/models/style_guide/java_script.rb` aswell to follow the same convention?

salbertson:
This is great, it will help avoid issues. We should test this on staging to make sure all cases work. :+1:

salbertson:
@jakecraige, is there anything left before this can be merged?

jakecraige:
@salbertson I still should rename the file like @teoljungberg suggested. I can probably take care of that this afternoon

salbertson:
Sounds good. Squash and merge when you're ready.

jakecraige:
I actually changed my mind about renaming that file. JavaScript is the official spelling when it's capitalized. When people refer to it in lowercase it's normally "javascript". So while this change makes "javascript" acceptable. It seems to make sense that the class is still "JavaScript".

Thoughts @teoljungberg ?

teoljungberg:
I agree, let's keep the filenames, and the classes, unchanged
---
Teo Ljungberg

salbertson:
@jakecraige makes sense to me, merge when you're ready.

PR #723, opened 2015-03-12
Fix rules for indenting multiline assignments

graysonwright:
https://trello.com/c/8plKSKKM

See https://github.com/bbatsov/rubocop/issues/1633

### DIFF
--- a/.ruby-style.yml
+++ b/.ruby-style.yml
@@ -255,7 +255,7 @@ Style/MethodName:
 Style/MultilineOperationIndentation:
   Description: Checks indentation of binary operations that span more than one line.
   Enabled: true
-  EnforcedStyle: aligned
+  EnforcedStyle: indented
   SupportedStyles:
   - aligned
   - indented


jakecraige:
:+1:

gylaz:
Do you mind also adding in to thoughtbot's config?

gylaz:
@salbertson What are your thoughts on also making this change in default? It would mean drifting away from RuboCop's default, but this is really an ugly setting. I fear I might be getting too personally attached to the defaults.

salbertson:
@gylaz, I don't think we should be affecting existing users by updating defaults, unless we think it's important enough and we communicate the change to everyone.

salbertson:
@graysonwright, should this be added to the thoughtbot config too? Maybe in a different PR?

This looks good to me. :+1:

teoljungberg:
Our styleguide is not updated for this https://github.com/thoughtbot/guides/blob/master/style/README.md#formatting
We should open a PR to thoughtbot/guides before changing it in Hound, everyone at thoughtbot doens't follow the developments made in Hound but they watch Guides.
I personally don't agree with this styleguide and would like it to be discussed, if the majority is in favor of this - we keep it. If not, let's chage it back to the defaults

ping @thoughtbot/hound @jferris @graysonwright

gylaz:
@teoljungberg Just to clarify you prefer:
```rb
users = User.where(email: 'user@example.com').
             update(user_parameters)
```
This PR changes this to favor:
```rb
users = User.where(email: 'user@example.com').
  update(user_parameters)
```
And while it might not be explicitly in our guides, 99% of thoughtbot code follows the latter.

gylaz:
Actually, we do have a guideline for it:

> Indent continued lines two spaces.

teoljungberg:
Actually no, this rule enforces this style (sorry for a shitty example, typing from my phone)


users = Users.
               active.
               awesome.
               tall

---
Teo Ljungberg

gylaz:
@teoljungberg Can you clarify where you're getting your information from? This comment https://github.com/bbatsov/rubocop/issues/1633#issuecomment-73067109 seems to be inline with what I've said, and the general spirit of this PR.

teoljungberg:
@gylaz I pinged you in the repo where we have the problem

gylaz:
I don't know what's going on on that related PR. I've added unit tests for this setting in #772.

PR #724, opened 2015-03-13
Use display: none to hide loading spinners

graysonwright:
Turns out, the CPU load was caused by the many animated spinners that
are hidden on the page at any given time. Every repo has a spinner
that's only shown when it's being activated or deactivated.

With `visibility: hidden`, the browser was still animating these
spinners even when they weren't visible. You can see that in the Chrome
devtools by selecting a `.repo::before` pseudo-element, and you'll see a
spinning outline of where the component is hidden.

One problem with using `display: none` (the new approach) is that the
gear icon is not loaded until someone clicks on 'Activate' or
'Deactivate'. That shows up as a small delay, but I'm not sure how to
fix it.

### DIFF
--- a/app/assets/stylesheets/repos/_index.scss
+++ b/app/assets/stylesheets/repos/_index.scss
@@ -315,7 +315,6 @@ aside.org-nav {
 
     &:before {
       @include transition(all 0.25s linear);
-      @extend .fa-spin;
       color: $light-font-color;
       content: '\f013';
       display: inline-block;
@@ -438,6 +437,7 @@ aside.org-nav {
       &:before {
         opacity: 1;
         visibility: visible;
+        @extend .fa-spin;
       }
 
       .repo-name {


gylaz:
@frechg Any ideas on how to improve the visual delay?

gylaz:
@graysonwright Great find!

graysonwright:
Kudos to @wrightmhw for spotting it.

spinner

graysonwright:
Here's the delay - it only happens the first time someone clicks an activation/deactivation button

delay2

graysonwright:
The newest push simplifies everything. The gears have gone back to `visibility: hidden`, but the spinner class (`fa-spin`) is now only applied when the icon is visible.

graysonwright:
from slack:

> we're now only adding the `.fa-spin` class when it's processing
> so the gears still get loaded on the page as `visibility: hidden`, but they don't rotate until they're visible
> the opacity & transitions all work like they did originally

graysonwright:
@gylaz gave a :+1: in real life.

Issue #418, opened 2014-10-03
Use JSCS in addition to JSHint for JavaScript linting


mrjoelkemp:

Adam262:
Hey there. Does Hound plan to support JSCS? I noticed in JSHint docs that they plan to deprecate many options relating to code style.

thorncp:
It's pretty clear that JSHint alone isn't sufficient. There's been mentions of JSCS and ESLint (#738), but it's still undecided. Since they both have a dependency on Node, I don't think it's likely either of them will land until #727 is finalized and we can isolate the linters.

Quote from JSHint docs for http://jshint.com/docs/options/

> If you would like to enforce rules relating to code style, check out the JSCS project.

thorncp:
The new service architecture has landed in Hound, so we now have the capabilities of linting with JSCS. Is there still interest for it? Would people prefer ESLint (#738)?

codejet:
ESLint looks very good (traditional linting, style checking, JSX support...). Definitely interested in it (also having JSCS as an addition to JSHint if no ESLint).

eidge:
+1 I'd definitely be interested!

NewAlexandria:
Since [jsHint will stop checking for style](http://jshint.com/docs/options/#maxlen), something needs to be put in place.  I don't have more info yet to be any more opinionated than that.

teoljungberg:
We are working on adding ESLint and JSCS alongside of JSHint. Stay tuned!

meric426:
Great! +1 looking forward to JSCS

teoljungberg:
If you guys are interested in trying out JSCS, it is now in beta!

If you want to try it out, please head over to https://houndci.com/configuration#jscs and see how to enable it.
Please contact us at hound@thoughtbot.com if you have any questions or feedback.

Happy linting!

PR #727, opened 2015-03-16
Use SCSS workers

arunagw:
Flow: 

In this PR we are going to remove scss linting from hound and using a iron worker to generate violations and send back to hound and we are going to save those and do the rest of things.

1. For SCSS we are dispatching a job to iron worker with payload and expecting results back to hound on `/build_workers` url and then we review those violations and save them in DB.

2. For Ruby, JavaScript and CoffeeScript we are not dispatching any job to iron worker. We just use our `StyleGuide` to generate violations and post them back to `build_workers` URL so they can be saved.

To keep the flow same we have a class called `LegacyWorker` which basically used in ruby, javascript and coffeescript.

Also the BuildRunner is now small and only have a job to send calls to LangaugeWorker based on file type and every language worker know how to handle those files for checking violation either sending to iron worker or just fetch violations internally.


croaky:
This is the pull request that should get rid of the `bubbles?` non-deterministic test failures! WOOO HOOOO!

app/workers/language_worker/base.rb
```diff
@@ -0,0 +1,25 @@
+module LanguageWorker
+  class Base
+    pattr_initialize :build_worker, :pull_request_file,
+      :repo_config, :pull_request
```
teoljungberg: How about using
```ruby
pattr_initialize \
  :build_worker,
  :pull_request_file,
# ...
```

Instead?

teoljungberg:
@gylaz @salbertson @srt32 We would love feedback on this before continuing to work on this

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,34 @@
+class BuildWorkersController < ApplicationController
+  skip_before_action \
+    :authenticate,
+    :capture_campaign_params,
+    :verify_authenticity_token,
+
+  def update
```
teoljungberg: Bogus

spec/services/worker_dispatcher_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+      file_content: "",
+      head_commit: head_commit,
+      pull_request_files: [],
+      repository_owner_name: "some_org"
+    }
+
+    double("PullRequest", defaults.merge(options))
+  end
+
+  def stub_repo_config(options = {})
```
teoljungberg: Bogus

spec/services/worker_dispatcher_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+    repo_config = double(
+      "RepoConfig",
+      default_options.merge(options)
+    )
+    allow(RepoConfig).to receive(:new).and_return(repo_config)
+
+    repo_config
+  end
+
+  def stub_worker(name, options = {})
```
teoljungberg: Bogus

spec/features/builds_spec.rb
```diff
@@ -34,18 +34,8 @@
...
-    stub_pull_request_comments_request(repo_name, pr_number)
-    comment_request = stub_simple_comment_request
-    stub_status_requests(repo_name, pr_sha)
-
-    page.driver.post builds_path, payload: payload
-
-    expect(comment_request).to have_been_requested
-  end
+  # TODO not sure what needs to be tested here?
+  scenario "a failed build"
```
teoljungberg: @gylaz @salbertson does anyone know what this test should actually do?

.sample.env
```diff
@@ -10,3 +10,5 @@ QUEUE=high,medium,low
 RESQUE_ADMIN_PASSWORD=password
 RESQUE_JOB_TIMEOUT=120
 SECRET_KEY_BASE=secret_key_base
+SCSS_WORKER_URL=foo.iron.io
+BUILD_WORKERS_URL=localhost
```
srt32: should there be an example `port` on `BUILD_WORKERS_URL`?

Gemfile
```diff
@@ -6,10 +6,11 @@ gem "active_model_serializers", "0.8.3"
 gem "analytics-ruby", "~> 2.0.0", require: "segment/analytics"
 gem "angularjs-rails"
 gem "angular_rails_csrf"
-gem "attr_extras"
+gem "attr_extras", "~> 4.0"
```
srt32: Any specific reason to make this change?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    params[:file]
+  end
+
+  def find_build_worker
+    @find_build_worker ||= BuildWorker.find(params[:id])
+  end
+
+  def authenticate_with_token
+    authenticate_or_request_with_http_token do |token, _|
+      token == ENV.fetch("BUILD_WORKERS_TOKEN")
```
srt32: Could we move this token into a config variable of some kind that reads the ENV var? It would be nice to raise loudly during build time if we're missing this token as opposed to when we first receive a call on this controller.

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+  private
+
+  def violations_attrs
+    params[:violations]
+  end
+
+  def file
```
srt32: Thoughts on calling the method `file_name`? Would that be more accurate?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    else
+      error = "BuildWorker##{build_worker.id} has already been finished"
+
+      render json: { error: error }, status: 412
+    end
+  end
+
+  private
+
+  def violations_attrs
```
srt32: I'm curious why call this method `violations_attrs` and not just `violations`,

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  before_action :authenticate_with_token
+  skip_before_action \
+    :authenticate,
+    :capture_campaign_params,
+    :verify_authenticity_token,
+
+  def update
+    build_worker = find_build_worker
+
+    if not build_worker.completed?
```
srt32: Could we use `!` instead?

> Prefer ! over not

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      ReviewJob.perform_later(build_worker, file, violations_attrs)
+
+      render json: {}, status: 201
+    else
+      error = "BuildWorker##{build_worker.id} has already been finished"
+
+      render json: { error: error }, status: 412
```
srt32: Would `409` ([conflict](http://httpstatus.es/409)) be more appropriate?

db/migrate/20150226092224_create_build_workers.rb
```diff
@@ -0,0 +1,9 @@
+class CreateBuildWorkers < ActiveRecord::Migration
+  def change
+    create_table :build_workers do |t|
+      t.belongs_to :build, null: false
```
srt32: Could we add a foreign key here as well? https://robots.thoughtbot.com/referential-integrity-with-foreign-keys

app/models/build.rb
```diff
@@ -1,6 +1,7 @@
 class Build < ActiveRecord::Base
   belongs_to :repo
   has_many :violations, dependent: :destroy
+  has_many :build_workers, dependent: :destroy
```
srt32: Could we put this line above the `violations` one for :abc:  reasons?

> Order ActiveRecord associations alphabetically by association type, then attribute name.

doc/SECURITY.md
```diff
@@ -179,6 +179,7 @@ it makes a new HTTP request to GitHub's API to get
 the pull request's diff and file contents.
 Hound never fetches a complete version of your codebase.
 
+# FixMe
```
srt32: What do we need to do here?

app/models/pull_request.rb
```diff
@@ -39,8 +39,18 @@ def head_commit
...
+      PullRequestFile.new(
+        github_file.filename,
+        github_file.patch,
+        head_commit.file_content(github_file.filename)
+      )
+    end
+  end
+
+  def file_removed?(github_file)
+    github_file.status == "removed"
```
srt32: Thoughts on making `"removed"` a constant?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  private
+
+  def mark_build_worker_complete
+    build_worker.update!(completed_at: Time.now)
+  end
+
+  def track_subscribed_build_completed
+    if repo.subscription
```
srt32: Why do we limit this tracking to only those `repo`s with a `subscription`? Why not all of them?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
+class Reviewer
+  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
+
+  static_facade :run, :build_worker, :file, :violations_attrs
+
+  def run
+    commenter.comment_on_violations(priority_violations)
+    create_success_status
```
srt32: Should we rescue any failures here (`Octokit::NotFound` is already rescued in the api client)? 

It seems we should continue on with both
1) tracking the build and
2) marking the worker as completed

even if the Github API call breaks for some reason (heroku hiccup, github hiccup, etc). 

The downside here would be the user would be left with a pending status even though we're done (but I think this is already the case.... just now we'd have the data cleaned up on our end as well).  Thoughts?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+    track_subscribed_build_completed
+    mark_build_worker_complete
+  end
+
+  private
+
+  def mark_build_worker_complete
+    build_worker.update!(completed_at: Time.now)
```
srt32: Thoughts on extracting a method onto `BuildWorker` called `mark_as_completed!` that does this work but wraps up the AR interface?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+      I18n.t(:success_status)
+    )
+  end
+
+  def violations
+    violations_attrs.flat_map do |violation|
+      line = pull_request_file.line_at(violation[:line_number])
+
+      if line.changed?
+        build.violations.create!(
```
srt32: To please Hound we could extract this `build.violations.create!(...` into another private method called something like `create_violation!` ...

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+      build.commit_sha,
+      I18n.t(:success_status)
+    )
+  end
+
+  def violations
+    violations_attrs.flat_map do |violation|
+      line = pull_request_file.line_at(violation[:line_number])
+
+      if line.changed?
```
srt32: Thoughts on wrapping up this violation creation loop (outside the `flat_map` call) into a transaction so we're not committing on each `insert`? It should 1) speed things up a bit as well as 2) not leave partially reviewed builds.

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def pull_request
+    PullRequest.new(review_payload)
+  end
+
+  def build
+    build_worker.build
+  end
+
+  def pull_request_file
```
srt32: Should the returned value be memoized since `pull_request_file` is called from within the loop on L37?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+    PullRequestFile.new(file[:filename], file[:content], file[:patch])
+  end
+
+  def repo
+    build.repo
```
srt32: Not sure it matters but should we memoize this value as well since it gets called multiple times?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+      {
+        number: build.pull_request_number,
+        pull_request: { head: { sha: build.commit_sha } },
+        repository: { full_name: repo.full_github_name }
+      }.to_json
+    )
+  end
+
+  def github
+    @github ||= GithubApi.new(ENV["HOUND_GITHUB_TOKEN"])
```
srt32: And in a stunning reversal... I think we don't need to memoize this method since it gets called only one on L28. Perhaps I'm missing something?

app/services/worker_dispatcher.rb
```diff
@@ -0,0 +1,52 @@
...
+    worker_class_name.new(
+      new_build_worker,
+      file,
+      repo_config,
+      pull_request
+    )
+  end
+
+  def new_build_worker
+    build.build_workers.create
```
srt32: Should this call be `create!` since we're not checking the return value?

app/workers/language_worker/scss.rb
```diff
@@ -0,0 +1,53 @@
+module LanguageWorker
+  class Scss < Base
+    def run
+      Faraday.post do |request|
+        request.url = ENV.fetch("SCSS_WORKER_URL")
```
srt32: Thoughts on pulling `ENV.fetch("SCSS_WORKER_URL")` into a named method just to clean up this `run` method?

spec/controllers/build_workers_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
+require "rails_helper"
+
+describe BuildWorkersController do
+  describe "#update" do
+    context "auth" do
```
srt32: How about something like

```
context "with no token" do
  it "responds with 401" do
    # test here  
  end
end
```

spec/controllers/build_workers_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      it "unauthorized access without token" do
+        put :update, id: 1, format: :json
+
+        expect(response.status).to eq(401)
+      end
+    end
+
+    context "given an completed build" do
```
srt32: `s/an/a`

spec/features/builds_spec.rb
srt32: We could, at least, ensure that an error was reported as an exception...

spec/jobs/review_job_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "rails_helper"
+
+describe ReviewJob do
+  describe "perform" do
```
srt32: This describe should be `describe "#perform"` I believe.

spec/models/pull_request_file_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+        line = double("Line", number: 1)
+        patch = double("Patch", changed_lines: [line])
+        allow(Patch).to receive(:new).and_return(patch)
+
+        expect(pull_request_file.line_at(2)).to be_an UnchangedLine
+      end
+    end
+  end
+
+  def pull_request_file
```
srt32: Could we make this method private?

Gemfile
gylaz: Could we just `bundle update` and not lock it here?

app/controllers/build_workers_controller.rb
gylaz: Or `unless`.

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
+class BuildWorkersController < ApplicationController
+  before_action :authenticate_with_token
+  skip_before_action \
+    :authenticate,
+    :capture_campaign_params,
+    :verify_authenticity_token,
```
gylaz: How about wrapping this in parens, since this is just a method call (instead of breaking with `\`)?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def violations_attrs
+    params[:violations]
+  end
+
+  def file
+    params[:file]
+  end
+
+  def find_build_worker
```
gylaz: How about just calling this `build_worker` and using that, since it's already memoized?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def update
+    build_worker = find_build_worker
+
+    if not build_worker.completed?
+      ReviewJob.perform_later(build_worker, file, violations_attrs)
+
+      render json: {}, status: 201
+    else
+      error = "BuildWorker##{build_worker.id} has already been finished"
```
gylaz: `... has already finished.`?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    :capture_campaign_params,
+    :verify_authenticity_token,
+
+  def update
+    build_worker = find_build_worker
+
+    if not build_worker.completed?
+      ReviewJob.perform_later(build_worker, file, violations_attrs)
+
+      render json: {}, status: 201
```
gylaz: Can this be just `render status: 201`?

app/models/pull_request.rb
```diff
@@ -8,7 +8,7 @@ def comments
   def pull_request_files
     @pull_request_files ||= api.
       pull_request_files(full_repo_name, number).
-      map { |file| build_commit_file(file) }
+      map { |github_file| build_pull_request_file(github_file) }.compact
```
gylaz: I would be more functional to:
```ruby
  select { |github_file| !file_removed?(github_file) }.
  map { |github_file| build_pull_request_file(github_file) }
```

app/models/pull_request.rb
```diff
@@ -39,8 +39,18 @@ def head_commit
...
   private
 
-  def build_commit_file(file)
-    CommitFile.new(file, head_commit)
+  def build_pull_request_file(github_file)
+    unless file_removed?(github_file)
+      PullRequestFile.new(
+        github_file.filename,
+        github_file.patch,
+        head_commit.file_content(github_file.filename)
```
gylaz: We don't want to fetch content at this time -- that's the reason the fetching was within `CommitFile` itself. The reason we don't want to fetch, is that we don't know that we'll use the file -- it might not be a ruby, js, or scss file. Thus, we'll be fetching content for files that we don't need.

app/models/pull_request_file.rb
```diff
@@ -0,0 +1,21 @@
+require "app/models/patch"
+require "app/models/unchanged_line"
```
gylaz: Why do we need these `require`s here?

app/models/style_checker.rb
```diff
@@ -1,57 +0,0 @@
...
-      StyleGuide::Scss
-    else
-      StyleGuide::Unsupported
-    end
-  end
-
-  def config
-    @config ||= RepoConfig.new(pull_request.head_commit)
-  end
-end
```
gylaz: What did this become? I would make this more discoverable if we document this in the PR description.

app/models/style_guide/coffee_script.rb
```diff
@@ -5,12 +5,8 @@ class CoffeeScript < Base
 
     def violations_in_file(file)
       Coffeelint.lint(file.content, config).map do |violation|
-        line = file.line_at(violation["lineNumber"])
```
gylaz: Why don't we need lines anymore?

app/services/reviewer.rb
gylaz: ... but why the `!`?

app/services/reviewer.rb
gylaz: Since this is a move of existing code, I believe @croaky might be best able to answer this.

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+  def pull_request
+    PullRequest.new(review_payload)
+  end
+
+  def build
+    build_worker.build
+  end
+
+  def pull_request_file
+    PullRequestFile.new(file[:filename], file[:content], file[:patch])
```
gylaz: Can we benefit from kwargs here?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
...
+    build.repo
+  end
+
+  def review_payload
+    Payload.new(
+      {
+        number: build.pull_request_number,
+        pull_request: { head: { sha: build.commit_sha } },
+        repository: { full_name: repo.full_github_name }
+      }.to_json
```
gylaz: Thoughts on pulling this hash out into a var?

app/services/reviewer.rb
```diff
@@ -0,0 +1,87 @@
+class Reviewer
```
gylaz: What do you think about moving away from the "land of nouns" and calling this service something like `Review` or `RecordBuild`?

app/services/worker_dispatcher.rb
```diff
@@ -0,0 +1,52 @@
+class WorkerDispatcher
```
gylaz: Same here, no reason to pigeon-hole ourselves into nouns: `DispathcWorker`.

app/services/reviewer.rb
srt32: It seems nice to pull through the `!` so users know that this method call can raise.

app/services/reviewer.rb
srt32: Perhaps it's because the current implementation depends on ` user`?

app/services/reviewer.rb
gylaz: Nearly every method can raise. The original convention in Ruby was the denote mutational method that had their non-mutational counterparts. In this case, no counterpart exists.

Also, we have this in our guides:

> Avoid bang (!) method names. Prefer descriptive names.

The thought behind ^ is that `!` isn't descriptive enough to mark the method as "dangerous", and what exactly would that mean? It mutates? It raises? It's not very consistent.

app/services/worker_dispatcher.rb
```diff
@@ -0,0 +1,52 @@
...
+    pull_request.pull_request_files
+  end
+
+  def worker(file)
+    worker_class_name = worker_class_name(file)
+    worker_class_name.new(
+      new_build_worker,
+      file,
+      repo_config,
+      pull_request
```
gylaz: Use kwargs here?

app/services/reviewer.rb
srt32: :ok_hand:

app/workers/language_worker/scss.rb
```diff
@@ -0,0 +1,53 @@
+module LanguageWorker
```
gylaz: Why is the namespace `LanguageWorker`? These aren't workers. Maybe just `Language`?

app/workers/language_worker/scss.rb
gylaz: I'm assuming we can't runt without `SCSS_WORKER_URL` being setup. So, I'm thinking having a config initializer is a nice way to organize all these necessary env configs into one place.

srt32: :+1:

app/workers/language_worker/legacy_worker.rb
```diff
@@ -0,0 +1,44 @@
...
+  class LegacyWorker < Base
+    def run
+      Faraday.post do |request|
+        request.url = ENV.fetch("BUILD_WORKERS_URL")
+        request.token_auth(ENV.fetch("BUILD_WORKERS_TOKEN"))
+        request.body = hound_payload.to_json
+      end
+    end
+
+    attr_implement :style_guide_name
```
gylaz: I'm starting to like `attr_extras` less and less, because it adds another level of abstraction/dsl to our code.

app/services/worker_dispatcher.rb
gylaz: Looks like we only need the owner from `pull_request`. So maybe just pass that as opposed the whole obejct?

app/workers/language_worker/base.rb
```diff
@@ -0,0 +1,27 @@
+module LanguageWorker
+  class Base
+    pattr_initialize :build_worker,
+                     :pull_request_file,
+                     :repo_config,
+                     :pull_request
```
gylaz: We have a guideline against aligning tokens this way. How about:
```ruby
pattr_initialize(
  :build_worker,
  ...
  :pull_request,
)
```

app/workers/language_worker/coffee_script.rb
```diff
@@ -0,0 +1,9 @@
+module LanguageWorker
+  class CoffeeScript < LegacyWorker
+    private
+
+    def style_guide_name
+      StyleGuide::CoffeeScript
```
gylaz: Can't we just infer this from the current class name + `StyleGuide`?

gylaz:
@teoljungberg @arunagw I think there some great work in here. However, I think this PR is just too big. It's too risky to release this many changes. Is there a way to just isolate the changes necessary to handle SCSS worker differently? Another idea, (credit goes to @srt32) is to run the new worker silently alongside the old way -- that way we can compare the outputs, have the new worker running and doing work, but not actually commenting.

Gemfile
arunagw: Yes bundle update will lock things but I am using something which will not available in lower version.. I think that's why we lock things in Gemfile.

app/workers/language_worker/coffee_script.rb
arunagw: Yes we totally can. I wanted to be specific here as these files anyway going to change soon once we start migrating other workers. 

But will fix these

app/models/style_guide/coffee_script.rb
arunagw: Because we calculate this back in reviewer

app/controllers/build_workers_controller.rb
arunagw: This hash contain content as well

doc/SECURITY.md
arunagw: We need to add some document that we are not using StyleChecker anymore and what is new flow in code

app/models/pull_request.rb
arunagw: Copied from older code base.. but sure we can move this in a constant

app/controllers/build_workers_controller.rb
teoljungberg: Good idea @srt32

app/controllers/build_workers_controller.rb
teoljungberg: It's just not the file, it's the `name`, `content`, and `patch`

app/controllers/build_workers_controller.rb
teoljungberg: I'm not a big fan of `unless + else` statement, I'll extract `incomplete?` out.

app/services/reviewer.rb
teoljungberg: This is being cached internally by rails since it's fetched through the association

app/services/worker_dispatcher.rb
teoljungberg: Don't we? Since we pass it in on L23?

app/workers/language_worker/scss.rb
teoljungberg: Good idea, fixed in 8aba20a

spec/models/pull_request_file_spec.rb
teoljungberg: Our styleguide says to keep helper methods public in the specs
> Avoid the private keyword in specs.

app/workers/language_worker/legacy_worker.rb
teoljungberg: I really like `attr_implement`, it's a very common pattern we use. Why is it a downside having those things more DRY?

app/models/pull_request.rb
teoljungberg: I totally agree, I would use `reject` instead of `select` to have it read a bit nicer. And skip the `!`

app/models/pull_request_file.rb
teoljungberg: This is the purpose of why #734, maybe that conversation should be in there?

teoljungberg: This is the purpose of why #734, maybe that conversation should be in there?

db/migrate/20150226092224_create_build_workers.rb
arunagw: added here 55d95481926a04d043b25e931084aef0967dd21a

app/models/pull_request.rb
arunagw: Yes. this seems to be a problem here. Need to look another solution. May be bring back `CommitFile` and keep `PullRequestFile` for lines

app/models/style_checker.rb
arunagw: I have added a better description into PR.

app/controllers/build_workers_controller.rb
arunagw: The problem is that we can't move this in initizlizer because our current setup fails on travis and we got failures because of that.

We have created a card in trello to handle that in complete code base from master. I think it's better if we remember setting these variables on heroku for now

app/models/pull_request.rb
```diff
@@ -39,8 +43,16 @@ def head_commit
 
   private
 
-  def build_commit_file(file)
-    CommitFile.new(file, head_commit)
+  def build_pull_request_file(github_file)
+    PullRequestFile.new(
+      github_file.filename,
+      github_file.patch,
+      head_commit.file_content(github_file.filename)
```
arunagw: Need to fix this to make sure we are fetching content for every file

srt32:
Build failed on bubbles. I restarted it.

.sample.env
```diff
@@ -10,3 +10,7 @@ QUEUE=high,medium,low
 RESQUE_ADMIN_PASSWORD=password
 RESQUE_JOB_TIMEOUT=120
 SECRET_KEY_BASE=secret_key_base
+SCSS_WORKER_URL=foo.iron.io
+BUILD_WORKERS_URL=<your-initials>-hound.ngrok.com
+BUILD_WORKERS_TOKEN=token
+IRON_WORKER_DISABLED=1
```
gylaz: This config var doesn't seem to be used anywhere.

.sample.env
```diff
@@ -10,3 +10,7 @@ QUEUE=high,medium,low
 RESQUE_ADMIN_PASSWORD=password
 RESQUE_JOB_TIMEOUT=120
 SECRET_KEY_BASE=secret_key_base
+SCSS_WORKER_URL=foo.iron.io
+BUILD_WORKERS_URL=<your-initials>-hound.ngrok.com
```
gylaz: Do we need this, since we already have `HOST`?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,40 @@
+class BuildWorkersController < ApplicationController
+  before_action :authenticate_with_token
+  skip_before_action(
+    :authenticate,
+    :capture_campaign_params,
+    :verify_authenticity_token,
```
gylaz: Where is this method coming from?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  )
+
+  def update
+    if build_worker.incomplete?
```
gylaz: `incomplete` seems like the wrong word here. I think we at trying to denote that the worker has not finished yet. How about some alternatives: `unless build_worker.finished?`, `if build_worker.dispatched?`, `if build_worker.running?`?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      ReviewJob.perform_later(build_worker, file, violations)
```
gylaz: How about passing just the build worker id, instead of having to serialize the whole object?

app/controllers/build_workers_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      render json: {}, status: 201
+    else
+      error = "BuildWorker##{build_worker.id} has already been finished"
+
+      render json: { error: error }, status: 409
+    end
```
gylaz: Who is going to be consuming the `else` case? It seems that we can just not handle the bad path situation, since we don't expect it to happen.

app/models/pull_request.rb
```diff
@@ -8,7 +10,9 @@ def comments
   def pull_request_files
     @pull_request_files ||= api.
       pull_request_files(full_repo_name, number).
-      map { |file| build_commit_file(file) }
+      reject { |github_file| file_removed?(github_file) }.
+      map { |github_file| build_pull_request_file(github_file) }.
+      compact
```
gylaz: I think we can remove this `compact` call now.

app/models/pull_request.rb
gylaz: I think @arunagw means "not fetching" :smile:.

app/workers/language_worker/legacy_worker.rb
gylaz: Because it adds a burden of learning another DSL. `attr_extras` isn't common enough, and `attr_implement` (or `static_facade`) isn't self-documenting, which causes a developer no familiar with `attr_extras` to go look up this method. Does it save us that much over plain Ruby?

app/models/pull_request.rb
srt32: @arunagw @gylaz could you elaborate on what needs to happen here?

spec/services/build_runner_spec.rb
```diff
@@ -141,57 +96,7 @@
...
-      pull_request = stubbed_pull_request
-      allow(pull_request).
-        to receive_messages(opened?: false, synchronize?: false)
-      allow(Commenter).to receive(:new)
-
-      build_runner.run
-
-      expect(Commenter).not_to have_received(:new)
-    end
+    it "fails the GitHub status with invalid config"
```
srt32: I think we lost a spec here

app/services/review.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  private
+
+  def mark_build_worker_complete
+    build_worker.update!(completed_at: Time.now)
+  end
+
+  def track_subscribed_build_completed
+    if repo.subscription
```
srt32: Sorry if I missed this discussion from earlier but why don't we track all `builds`?

app/services/review.rb
```diff
@@ -0,0 +1,102 @@
...
+          patch_position: line.patch_position,
+          line_number: violation[:line_number],
+          messages: violation[:messages],
+        )
+      end
+    end
+  end
+
+  def create_violation(attributes)
+    Violation.transaction do
```
srt32: I think this `transaction` needs to be moved out of this method and into `def violations` on L36 so that it's outside the loop,

app/models/pull_request.rb
gylaz: As is, this will fetch content for every file in the PR that is not `removed`. This would include html, haml, erb, etc files that we don't support. Also, since not everyone has JS/Coffee/SCSS enabled, then we might not need those files either. The other reason, is that some files/directories are ignored on per linter basis.

All this boils down to us not fetching what we don't need ahead of time, and only fetching the content when we have actually decided that the file should be checked -- that happens in `StyleChecker` and each respective `StyleGuide` class.

app/services/review.rb
srt32: Updated in d70dade

app/workers/language/scss.rb
```diff
@@ -0,0 +1,53 @@
+module Language
+  class Scss < Base
+    def run
+      Faraday.post do |request|
```
srt32: Does this `POST` not also need `request.token_auth(ENV.fetch("BUILD_WORKERS_TOKEN"))`?

srt32: Nevermind, this `POST` is going out to Iron.

db/migrate/20150226092224_create_build_workers.rb
srt32: :smiley:

app/controllers/build_workers_controller.rb
thorncp: `==` is vulnerable to timing attacks. We probably want to ensure the given token is the same length and use a secure comparison algorithm. E.g.

```ruby
worker_token = ENV.fetch("BUILD_WORKERS_TOKEN")
token.size == worker_token.size && ActiveSupport::SecurityUtils.secure_compare(token, worker_token)
```

app/services/review.rb
croaky: We were having an issue not getting "Reviewed Repo" events to record correctly.

https://github.com/thoughtbot/hound/commit/dc71071d8dce25d241e6e3421c26ed27e3ad7573

There's still something wrong and we don't know what it is:

https://mixpanel.com/report/468495/funnels/#view/1003465/from_date:-30,to_date:0

We haven't been getting "Build Completed" event at all.

spec/factories.rb
```diff
@@ -10,6 +10,14 @@
     end
   end
 
+  factory :build_worker do
+    build
+
+    trait :completed do
+      completed_at Time.now
```
croaky: `Time.zone.now`?

app/controllers/build_workers_controller.rb
teoljungberg: The convention for activejob is to use globalid - https://github.com/rails/globalid, which handles the serialization for us safely

app/controllers/build_workers_controller.rb
teoljungberg: Actually no-one, let's kill it

app/models/pull_request.rb
teoljungberg: Good call

app/controllers/build_workers_controller.rb
teoljungberg: Rails internals, before each `POST` and `PUT/PATCH` request Rails verifies the authenticity of the requester http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html
We can probably leverage this `skip_before_action :verify_authenticity_token, if: :json_request?` for clarity

spec/factories.rb
teoljungberg: Thanks

app/services/review.rb
```diff
@@ -0,0 +1,102 @@
...
+            messages: violation[:messages],
+          )
+        end
+      end
+    end
+  end
+
+  def create_violation(attributes)
+    build.violations.create!(
+      filename: attributes[:filename],
```
thorncp: Seems like `attributes` can be passed to `create!` as-is.

.sample.env
thorncp: There's two checks for `IRON_WORKER_ENABLED` though. Should this be that?

app/controllers/build_workers_controller.rb
gylaz: So, activejob uses `globalid` behind the scenes?

thorncp: Seems so. Neat! http://guides.rubyonrails.org/active_job_basics.html#globalid

thorncp:
Tried to knock out several minor things in 3725eac0fdc2085e8c958fe73b61aa032aede89e

app/services/dispatch_workers.rb
```diff
@@ -0,0 +1,60 @@
+class DispatchWorkers
```
salbertson: This name is a little misleading, not all languages use workers/SOA to review code. Is there a better name for this?

app/models/build_worker.rb
```diff
@@ -0,0 +1,13 @@
+class BuildWorker < ActiveRecord::Base
```
salbertson: What does this class have to do with the worker? Should we name it something differently, something that doesn't include "worker"? What about `PendingReview`?

app/services/dispatch_workers.rb
gylaz: This dispatches a worker per file. Where is the confusion coming from?

app/models/build_worker.rb
gylaz: Since this is generated for every file, how about calling something like `FileReview`? I think pending is a bit of a bad match, because these aren't always pending.

app/services/dispatch_workers.rb
salbertson: I'm saying including "worker" when naming this class or the language modules doesn't make sense to me. What does this code have to do with workers? Why are we naming this class and our language modules as "workers"?

When I think of "worker" I'm thinking background process or something like IronWorker, not a class that dispatches requests.

app/workers/language/legacy_worker.rb
```diff
@@ -0,0 +1,44 @@
+module Language
+  class LegacyWorker < Base
```
salbertson: Should we name this something other than "legacy"? That prefix doesn't help describe what this class does.

app/workers/language/legacy_worker.rb
```diff
@@ -0,0 +1,44 @@
...
+    def run
+      Faraday.post do |request|
```
salbertson: So, all languages make requests to the `BuildWorkersController`? That seems unnecessary for languages that can be reviewed in process. Are we trying to keep this consistent with what SCSS reviews are doing or what?

app/services/dispatch_workers.rb
gylaz: Got it.

app/workers/language/legacy_worker.rb
teoljungberg: I agree, any suggestions? `LocalWorker`?

app/workers/language/legacy_worker.rb
teoljungberg: We were trying to keep the "old" languages consistent with how SCSS works with iron.io
Do you think we should avoid this and keep it in-process?

app/models/build_worker.rb
teoljungberg: Not having "worker" in the class-name sounds great, I want to name it something along the lines of `File`. Do we have any other contenders thatn `FileReview`?

app/services/dispatch_workers.rb
teoljungberg: Good point @salbertson, do you have any name suggestions?
I think we got a bit tainted/influenced by looking into IronWorker, which is why we have the word "worker" spread across the files.
To keep it on a higher level, maybe if we name it something like `DispatchLinters`? And then rename the language specific linters as `RubyLinter` unt so weiter?

app/services/review.rb
```diff
@@ -0,0 +1,106 @@
+class Review
+  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
+
+  static_facade :run, :build_worker, :file, :violations_attributes
+
+  def run
+    commenter.comment_on_violations(priority_violations)
+    create_success_status
```
thorncp: Do we want to do this here? If we're sending payloads to Iron for each file, we probably want to wait until all of those individual reviews have happened before setting the success status.

app/services/review.rb
```diff
@@ -0,0 +1,106 @@
...
+    track_subscribed_build_completed
```
thorncp: Same as above comment: does it make sense to mark the build as completed for every file review that Iron responds back with?

app/services/review.rb
```diff
@@ -0,0 +1,106 @@
+class Review
+  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
+
+  static_facade :run, :build_worker, :file, :violations_attributes
+
+  def run
+    commenter.comment_on_violations(priority_violations)
```
thorncp: It may make sense to not send any comments until all the file reviews have finished. If we do it per file, the max number of comments will only apply per file, so we could end up submitting quite a lot of comments to a PR if it has many files with many violations.

spec/workers/language/scss_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+            config: {
+              custom: "custom",
+              default: default_config_file
+            },
+            file: {
+              name: "test.scss",
+              content: "some content",
+              patch_body: ""
+            },
+            hound_url: "#{ENV.fetch("BUILD_WORKER_CALLBACK_URL")}/#{build_worker.id}",
```
teoljungberg: I think the later is something we don't want to do, thoughts @thoughtbot/hound

app/services/review.rb
teoljungberg: To keep up with the current behaviour, I think should only comment once
all of the reviews are finished. We can revise that idea as soon as we
feel comforatble with this shipped

app/services/review.rb
teoljungberg: @thorncp this should be fixed in 86d6526, aswell as the other comments you made

app/services/review.rb
teoljungberg: This might return `nil` if it's a public repository, will that change anything for the background job?

app/models/build_worker.rb
teoljungberg: I haven't changed this since it's added seperately in #765
/cc @gylaz @salbertson @thorncp

teoljungberg:
@thorncp sounds like a good idea, mind taking over that work in this PR? (I'm off today)

app/models/style_checker.rb
srt32: Seems like this class became `DispatchWorkers`.

app/services/review.rb
```diff
@@ -0,0 +1,113 @@
+class Review
+  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
+
+  static_facade :run, :build_worker, :file, :violations_attributes
+
+  def run
+    mark_build_worker_complete
+
+    if all_workers_completed?
+      commenter.comment_on_violations(priority_violations)
```
thorncp: `priority_violations` is indirectly creating the violations records, but is only called when the last worker reports, so we're throwing away the violations reporting by the first workers to report back.

blackjid:
Why this is specific to SCSS and coffeescript, javascript and ruby are treated differently?

gylaz:
@blackjid Eventually all the languages will run this way. SCSS is just the first.

blackjid:
cool..it will be tightly coupled to iron.io or is just and example?

gylaz:
It's more of a spike. We are now considering using Resque, but Iron.io is one of the alternatives.

blackjid:
This is a lazy question (sorry), I should dig in the code, but isn't hound already using resque??? I've setup hound for my company using docker, and I did have to setup a resque worker.

gylaz:
Yes, Hound is using Resque already. That's why we prefer sticking to Resque as we consider more of an SOA approach.

blackjid:
I like that!!, Now that this issue is closed how do you see issues like https://github.com/thoughtbot/hound/issues/738 and https://github.com/thoughtbot/hound/issues/418 regards using JSCS or ESLint are going to make progress?? I thought they where stuck waiting this issue to be solved...

gylaz:
@blackjid I suggest watching the progress of https://github.com/thoughtbot/hound-scss. Once we hook that up, we'll be able to make further progress on other linters.

teoljungberg:
@blackjid We still have a long way to go before we're migrating over the Javascript linter and considering replacing JSHint. But it's in the backlog

blackjid:
thanks @gylaz, do you mean this one?? https://github.com/thoughtbot/hound-scss-worker

@gylaz @teoljungberg but if I would like to implement other linters in my own fork, would you recommend to wait hound-scss-worker?? or do you think it's posible with the current implementation of runners and workers?

gylaz:
@blackjid No. I meant the link I posted.

blackjid:
oh, maybe is private, because I get a 404...

PR #728, opened 2015-03-16
Convert inherit_from value to array

gylaz:
The value can be a single entry:

    inherit_from: .rubocop_todo.yml

which will be parsed as just a string.

This change ensures we are always dealing with an array.

### DIFF
--- a/spec/models/repo_config_spec.rb
+++ b/spec/models/repo_config_spec.rb
@@ -155,49 +155,83 @@ describe RepoConfig do
         )
       end
 
-      it "supports inherit_from" do
-        hound_config = <<-EOS.strip_heredoc
-          ruby:
-            enabled: true
-            config_file: .rubocop.yml
-        EOS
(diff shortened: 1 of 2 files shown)


jakecraige:
Could you add a test for this?

gylaz:
@jakecraige Done.

spec/models/repo_config_spec.rb
```diff
@@ -155,49 +155,83 @@
...
+            ".rubocop.yml" => rubocop,
+            "config/base.yml" => base,
+          )
+          config = RepoConfig.new(commit)
+
+          result = config.for("ruby")
+
+          expect(result).to eq(
+            "LineLength" => { "Max" => 40 },
+            "Style/Encoding" => { "Enabled" => true }
```
jakecraige: missing trailing comma?

jakecraige:
Other than that one comment. LGTM

PR #729, opened 2015-03-17
Disable SCSS property count linter

tysongach:
https://github.com/causes/scss-lint/blob/master/lib/scss_lint/linter/README.md#propertycount

### DIFF
--- a/config/style_guides/thoughtbot/scss.yml
+++ b/config/style_guides/thoughtbot/scss.yml
@@ -69,9 +69,7 @@ linters:
   PlaceholderInExtend:
     enabled: false
   PropertyCount:
-    enabled: true
-    include_nested: false
-    max_properties: 10
+    enabled: false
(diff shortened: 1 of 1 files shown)


gylaz:
LGTM, but I'm curious what's the story/rational behind this?

tysongach:
I don’t think property count is a big enough issue/code smell to warrant a comment from Hound.

joshuaogle:
:+1: I agree. It could theoretically be a smell, but the vast majority of the time it would be flagging on legitimate uses.

tysongach:
@gylaz Do we know why builds are failing here and on `master`?

teoljungberg:
@tysongach it's a random build-failure we're having. We're working on getting it fixed

tysongach:
@teoljungberg Cool. Good to go ahead an merge this? I was failing at one point…

PR #730, opened 2015-03-18
Use rails_helper

arunagw:
As latest rspec uses rails_helper and spec_helper.

* Rename fast_spec_helper to spec_helper
* Rename spec_helper to rails_helper


teoljungberg:
LGTM

PR #734, opened 2015-03-20
Remove unused require

arunagw:
Also move requires to right places

### DIFF
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -1,3 +1,6 @@
+require "attr_extras"
+require "octokit"
+
 class Commit
   pattr_initialize :repo_name, :sha, :github
   attr_reader :repo_name, :sha
--- a/spec/jobs/buildable_spec.rb
+++ b/spec/jobs/buildable_spec.rb
@@ -1,8 +1,4 @@
-require "spec_helper"
-require "app/jobs/buildable"
-require "app/models/payload"
-require "app/services/build_runner"
-require "raven"
+require "rails_helper"
 
 describe Buildable do
   class TestJob < ActiveJob::Base
--- a/spec/lib/github_api_spec.rb
+++ b/spec/lib/github_api_spec.rb
@@ -1,8 +1,5 @@
 require "spec_helper"
-require "attr_extras"
 require "lib/github_api"
-require "json"
-require "app/models/github_user"
 
 describe GithubApi do
   describe "#repos" do
--- a/app/models/commit_file.rb
+++ b/app/models/commit_file.rb
@@ -1,3 +1,6 @@
+require "app/models/patch"
+require "app/models/unchanged_line"
+
 class CommitFile
   pattr_initialize :file, :commit
 
--- a/app/models/commit_file.rb
+++ b/app/models/commit_file.rb
@@ -1,3 +1,4 @@
+require "attr_extras"
 require "app/models/patch"
 require "app/models/unchanged_line"
 
--- a/app/models/default_config_file.rb
+++ b/app/models/default_config_file.rb
@@ -1,3 +1,5 @@
+require "attr_extras"
+
 class DefaultConfigFile
   CONFIG_DIR = "config/style_guides"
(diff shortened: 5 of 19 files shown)


teoljungberg:
Looks great, thank you! LGTM

spec/jobs/buildable_spec.rb
```diff
@@ -1,8 +1,4 @@
-require "spec_helper"
-require "app/jobs/buildable"
-require "app/models/payload"
-require "app/services/build_runner"
-require "raven"
+require "rails_helper"
```
gylaz: This will make the spec slower. Is there a reason for this change?

spec/lib/github_api_spec.rb
```diff
@@ -1,8 +1,5 @@
 require "spec_helper"
-require "attr_extras"
 require "lib/github_api"
-require "json"
-require "app/models/github_user"
```
gylaz: Does this spec run stand alone?

app/models/commit.rb
```diff
@@ -1,3 +1,6 @@
+require "attr_extras"
+require "octokit"
```
gylaz: What's the reasoning behind adding these to the files as opposed to the spec file? These requires aren't necessary for the code to run under Rails, right? What prompted us to move them here?

salbertson: I'd also to like to know the reason for moving these requires. @arunagw?

spec/jobs/buildable_spec.rb
salbertson: @arunagw, Is there something in this test that needs the entire Rails environment?

spec/lib/github_api_spec.rb
salbertson: @gylaz, what do you mean?

gylaz: I was curious if these requires were necessary for the test to run in isolation. I've tested this myself since, and it looks like it works.

spec/jobs/buildable_spec.rb
arunagw: Yes we need to have ActiveJob here.

app/models/commit.rb
arunagw: I think we should require where they belongs to. if we are going to spec_helper here which is fast helper then we need to have these requires inside production code because i think test should just load file and helper not other things

But in rails console we load everything so it's not a problem.

gylaz: I'm not sure I understood, but the proposal isn't to require `spec_helper` here. Rails requires/loads everything on boot, so these requires aren't actually necessary. Thus, we should move them to the spec instead.

teoljungberg: The idea, and the convention, is that you require the things you need your specific file. That's why we moved everything from the `spec_helper` to the classes where they are used. It makes unittesting easier and cleaner.

This is a less common pattern in rails applications, since rails requires everything at boot, but it's commonly found in gems.

gylaz: I agree with the general idea here -- this is very common in other-languages (and like you mentioned non-Rails Ruby libs). But, I wonder if this is breaking the Rails contention too much, and going against the grain? Thoughts?

teoljungberg: I actually prefer doing this for PORO objects, I've done so for several applications in the past and I think it helps in emphesising that it's a PORO object and not related to Rails at all

---
Teo Ljungberg

gylaz: One situation I can think of is, some can at a later time remove these from this file as "unnecessary". The spec suite will still run fine, but the individual spec files will fail. It could cause us to get into this loop of removal/addition of these `require`s.

teoljungberg: Good point, but we can avoid that case by updating the contribution guidelines to enforce that we want the dependencies of a PORO object to be in that file. 

I agree that this goes against a lot of what we're used to, but I believe that this will be an interesting experiment
---
Teo Ljungberg

arunagw:
I think this can be think later and it's outdated anyway. Closing this for now

PR #735, opened 2015-03-20
Exclude spec directory when checking for long methods

salbertson:
@gylaz, this should work, right?


.ruby-style.yml
```diff
@@ -508,6 +508,8 @@ Metrics/MethodLength:
   Enabled: true
   CountComments: true
   Max: 10
+  Exclude:
+  - "spec/**/*"
```
gylaz: This can be just written as `Exclude: spec/`

gylaz:
:+1:

PR #740, opened 2015-03-27
Extract `ENV.fetch` into initializer

teoljungberg:


gylaz:
Why was this closed? Seemed like a good idea.

teoljungberg:
There was a massive test-breakage doing this, I and I wasn't feeling up for it today. If someone else is willing to continue, please do

Issue #331, opened 2014-07-30
[Feature] Run hound on templates


rupurt:
Howdy,

Is it possible to run hound on templates e.g. `.erb`, `.haml`...

gylaz:
Hound now supports reviewing HAML (in beta).
To enable it add this to your `.hound.yml`:
```yaml
haml:
  enabled: true
```

ckundo:
:tada:

salbertson:
Annnd we will soon support ERBLint too! https://github.com/houndci/hound/pull/1641

salbertson:
Hound now reviews ERB files. https://github.com/houndci/hound/issues/1599#issuecomment-440537267 💥

PR #741, opened 2015-03-27
Add linter for haml templates via haml-lint gem.

ckundo:
- Adds haml-lint integration
- Invokes the `HamlLint::Runner` class, which is used by the CLI and accepts command line options.
- Uses haml-lint's default linter configuration
- Uses a fixture in test instead of stubbing, since the contents get parsed at a low level in haml-lint and it'd be difficult to stub

[#331]


spec/support/fixtures/div.haml
```diff
@@ -0,0 +1 @@
+%div.some-class This is content
```
ckundo: I feel a little uncomfortable with the indirection of using a fixture, and it deviates from the other specs, but whereas the other linter runners accept contents of a file, [`HamlLint::Runner`](https://github.com/causes/haml-lint/blob/master/lib/haml_lint/runner.rb) accepts an options hash with file names, and parses the file contents later. Open to alternatives and suggestions.

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,50 @@
...
+    def build_runner
+      HamlLint::Runner.new
+    end
+
+    def options
+      HamlLint::Options.new.parse(options_args)
+    end
+
+    def options_args
+      ["--config", default_config_file, file.filename]
```
ckundo: Does it feel weird to be emulating command line arguments here? It feels a little funny to me, but looking for feedback.

ckundo:
I have not added Hound specific config for haml-lint yet, curious if there any any rules we want to exclude.

rupurt:
You're a gun @ckundo!

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,49 @@
+module StyleGuide
+  class Haml < Base
+    DEFAULT_CONFIG_FILENAME = "haml.yml"
+
+    def violations_in_file(file)
+      require "haml_lint"
```
gylaz: Does this not get required by bundler?

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,49 @@
...
+      require "haml_lint/options"
```
gylaz: Is this necessary?

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      @file = file
```
gylaz: Can we pass the filename to `options` instead of mutating instance state?

app/models/style_guide/haml.rb
gylaz: Are any alternatives? If `HamlLint::Options.new.parse` expects an array like this, then this seems fine.

spec/models/style_guide/haml_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "spec_helper"
+
+describe StyleGuide::Haml do
+  describe "#violations_in_file" do
+    context "with default configuration" do
+      describe "for implicit %div violation" do
```
gylaz: Should this be `context`?

spec/models/style_guide/haml_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+          )
+        end
+      end
+    end
+  end
+
+  private
+
+  def violations
+    style_guide = build_style_guide(nil)
```
gylaz: `(nil)` can be dropped.

spec/support/fixtures/div.haml
gylaz: The file isn't present on the filesystem. All the other linters deal with file content directly. I might have missed that in this PR, but I didn't see where file's content is being given to the linter.

teoljungberg:
I would love to support this, can it wait until #727 is released?

gylaz:
I think the biggest issue here is that the way this PR stands it expects the file to exist on the filesystem, which is not the case in Hound -- the file content is fetched from GitHub and lives in memory. Is there a way to pass the *file content* to the linter, as opposed to just the filepath?

thorncp:
From a quick look through of haml-lint's code, it looks like it supports passing in the content directly, based on the parameter names in this [internal object](https://github.com/brigade/haml-lint/blob/master/lib/haml_lint/parser.rb#L14).

ckundo:
I’ll revise the PR to make the linter accept file contents directly.

ckundo:
@teoljungberg sure, I can hold off until #727 is merged.

srt32:
@ckundo while it may add some complexity to wrapping up the worker PR it'd be great to get this `haml-lint` work merged once the file content issue is worked out. What can we do to help?

ckundo:
@srt32 I updated the haml class to accept file contents. I think it needs more testing around accepting different configuration options, but welcome more feedback in the mean time.

app/models/style_checker.rb
```diff
@@ -44,6 +44,8 @@ def style_guide_class(filename)
       StyleGuide::CoffeeScript
     when /.+\.js\z/
       StyleGuide::JavaScript
+    when /.+\.haml\z/
+      StyleGuide::Haml
```
ckundo: I'm not sure if this should be here.

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,68 @@
...
+      @config ||= HamlLint::ConfigurationLoader.load_file(default_config_file)
+    end
+
+    def options
+      {}
+    end
+
+    def run_linters
+      results = []
+      linters.each do |linter|
```
srt32: Could we refactor this method to use `reduce`? Something like:

```
linters.reduce([]) do |results, linter|
  results += linter.run(parser)
end
```

app/models/style_checker.rb
srt32: What makes you say that? Seems pretty resonable!

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,50 @@
+module StyleGuide
+  class Haml < Base
+    DEFAULT_CONFIG_FILENAME = "haml.yml"
+
+    def violations_in_file(file)
```
srt32: We could extract L11-19 into a private method called something like `violations_in_line`. Thoughts?

app/models/style_checker.rb
ckundo: :+1:

app/models/style_guide/haml.rb
ckundo: done!

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def parser
+      @parser ||= HamlLint::Parser.new(file.content, {})
+    end
+
+    def config
+      @config ||= HamlLint::ConfigurationLoader.load_file(default_config_file)
+    end
+
+    def options
```
ckundo: This is probably superfluous

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,65 @@
...
+    def parser
+      @parser ||= HamlLint::Parser.new(file.content, {})
+    end
+
+    def config
+      @config ||= HamlLint::ConfigurationLoader.load_file(default_config_file)
+    end
+
+    def run_linters
+      linters.inject([]) do |results, linter|
```
gylaz: Strange, this is wrong (I know Hound left a comment). We prefer `reduce` now.

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,65 @@
...
+    def parser
+      @parser ||= HamlLint::Parser.new(file.content, {})
+    end
+
+    def config
+      @config ||= HamlLint::ConfigurationLoader.load_file(default_config_file)
+    end
+
+    def run_linters
+      linters.reduce([]) do |results, linter|
```
teoljungberg: This shouldn't be in our styleguide, right? /cc @gylaz

gylaz: Yep. Hound comment is wrong, but not sure why it's commenting. Here's the config line: https://github.com/thoughtbot/hound/blob/master/.ruby-style.yml#L93

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    def custom_config
+      @options ||= HamlLint::Configuration.new(repo_config_for_haml)
+    end
+
+    def repo_config_for_haml
+      repo_config_for_haml = repo_config.for(name)
+
+      if repo_config_for_haml.is_a? Hash
```
ckundo: `repo_config` returns `ActiveRecord::Base` when there is no custom configuration (instead of an empty hash which seems more expected). I can take a stab at refactoring that method, but for the moment this type checking works.

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    def custom_config
+      @options ||= HamlLint::Configuration.new(repo_config_for_haml)
+    end
+
+    def repo_config_for_haml
+      options = repo_config.for(name)
+
+      if options.is_a? Hash
```
ckundo: `repo_config` returns `ActiveRecord::Base` when there is no custom configuration (instead of an empty hash which seems more expected). I can take a stab at refactoring that method, but for the moment this type checking works.

ckundo: ah wow, repo_config is stubbed in test. nevermind!

ckundo:
Do we want to add/exclude any linters in the default config file?

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,73 @@
+require "haml_lint"
```
gylaz: Do we need this require? Shouldn't it be already required by Bundler?

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    def options
+      {}
+    end
+
+    def run_linters
+      linters.reduce([]) do |results, linter|
+        linter.run(parser)
+        results += linter.lints
```
gylaz: The last thing in the block will automatically be assigned to `results` in the block, so it's true just `+` will do.

gylaz:
> Do we want to add/exclude any linters in the default config file?

@ckundo I'm not sure what you mean by this. By "linters" do you mean configuration rules?

ckundo:
@gylaz yes, the default configuration I mean

gylaz:
@ckundo We want to go with defaults for everyone, but tweak it (as we find inconsistencies) for thoughtbot's settings. I don't think we know enough to do anything at this time.

app/models/style_checker.rb
srt32: On second thought, could we add a spec to https://github.com/thoughtbot/hound/blob/cc-haml-lint/spec/models/style_checker_spec.rb to support this change?

salbertson:
@ckundo, any update on this?

ckundo:
@salbertson I think the only remaining is to add a spec. I'll submit for re-review this afternoon.

app/models/style_guide/haml.rb
```diff
@@ -0,0 +1,71 @@
...
+      linters.reduce([]) do |results, linter|
+        linter.run(parser)
+        results + linter.lints
+      end
+    end
+
+    def linters
+      included_linters = HamlLint::LinterRegistry.linters
+
+      @linters = included_linters.map do |linter_class|
```
salbertson: Is `@linters` used anywhere?

salbertson:
@ckundo, looks good to me. Would you like to include updates to the [documentation](https://houndci.com/configuration) in this PR or a new one?

app/models/style_guide/haml.rb
ckundo: I think that was meant to be memoization. Will fix.

ckundo:
Thanks @salbertson. I think including the docs in this PR makes sense.

ckundo:
@salbertson updated the docs.

salbertson:
:+1:

PR #743, opened 2015-04-01
Include violation count in GitHub status message

thorncp:
* Treats multiple messages on a line as separate violations.
* The status message will be overwritten when new commits are pushed to
  the PR.
* Updates feature spec stubbing to not care what message is passed.

https://trello.com/c/oCqXJB1V

yep_by_thorncp_ _pull_request__22_ _thornco_dumb-ruby
<hr>
yep_by_thorncp_ _pull_request__22_ _thornco_dumb-ruby
<hr>
yep_by_thorncp_ _pull_request__23_ _thornco_dumb-ruby

### DIFF
--- a/app/models/violation.rb
+++ b/app/models/violation.rb
@@ -9,6 +9,8 @@ class Violation < ActiveRecord::Base
 
   attr_writer :line
 
+  delegate :count, to: :messages, prefix: true
+
   def add_messages(new_messages)
     self[:messages].concat(new_messages)
   end
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -89,7 +89,7 @@ class BuildRunner
     github.create_success_status(
       payload.full_repo_name,
       payload.head_sha,
-      I18n.t(:success_status)
+      I18n.t(:success_status, count: violation_count)
     )
   end
 
@@ -117,5 +117,9 @@ class BuildRunner
 
   def configuration_url
     Rails.application.routes.url_helpers.configuration_url(host: ENV["HOST"])
+  end
+
+  def violation_count
+    violations.sum(&:messages_count)
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -5,7 +5,10 @@ en:
   active_repos: "Active repos"
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
-  success_status: "Hound has reviewed all the changes!"
+  success_status:
+    zero: "No violations found. Woof!"
+    one: "1 violation found."
+    other: "%{count} violations found."
   config_error_status:
     "Error parsing your config file. Click \"details\" for
     assistance."
--- a/spec/support/helpers/github_api_helper.rb
+++ b/spec/support/helpers/github_api_helper.rb
@@ -431,7 +431,7 @@ module GithubApiHelper
       repo_name,
       sha,
       "success",
-      "Hound has reviewed all the changes!"
+      anything
     )
   end
 
--- a/spec/models/violation_spec.rb
+++ b/spec/models/violation_spec.rb
@@ -41,4 +41,12 @@ describe Violation do
       expect(changed).to eq false
     end
   end
+
+  describe "#messages_count" do
+    it "returns the number of violation messages" do
+      violation = build(:violation, messages: ["foo", "bar"])
+
+      expect(violation.messages_count).to eq 2
+    end
+  end
 end
--- a/spec/services/build_runner_spec.rb
+++ b/spec/services/build_runner_spec.rb
@@ -96,7 +96,11 @@ describe BuildRunner, '#run' do
       )
       build_runner = BuildRunner.new(payload)
       stubbed_pull_request
-      stubbed_style_checker_with_violations
+      violations = [
+        build(:violation),
+        build(:violation, messages: ["wrong", "bad"]),
+      ]
+      stubbed_style_checker(violations: violations)
       stubbed_commenter
       github_api = stubbed_github_api
 
@@ -110,7 +114,7 @@ describe BuildRunner, '#run' do
       expect(github_api).to have_received(:create_success_status).with(
         "test/repo",
         "headsha",
-        "Hound has reviewed all the changes!"
+        "3 violations found."
       )
     end
 


config/locales/en.yml
```diff
@@ -5,7 +5,9 @@ en:
   active_repos: "Active repos"
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
-  success_status: "Hound has reviewed all the changes!"
+  success_status:
+    "Hound has reviewed all the changes and found %{violation_count}
```
teoljungberg: How about something shorter `NUM violations found`?

app/services/build_runner.rb
```diff
@@ -109,4 +109,8 @@ def github
   def configuration_url
     Rails.application.routes.url_helpers.configuration_url(host: ENV["HOST"])
   end
+
+  def violation_count
+    violations.map { |v| v.messages.size }.sum
```
srt32: Thoughts on extracting a method to `Violation` to handle the summing so that `messages` doesn't leak into `BuildRunner`?

croaky:
Can we make the status a failure if there are violations? "All is well" doesn't really fit and we've had confused users in customer support.

teoljungberg:
> Can we make the status a failure if there are violations? "All is
> well" doesn't really fit and we've had confused users in customer
> support.
Agreed, I've been thinking about adding this aswell

srt32:
It seems the change to failure (setting a failed CI status) is a larger change than just the amount of violations. Perhaps we separate out the two features into their PR's?

config/locales/en.yml
thorncp: Sounds good. Done in 7eba966

app/services/build_runner.rb
thorncp: Makes sense to me. Done in 6898a85. I called it `Violation#count` but I don't really like it. It sounds weird to ask a violation (singular) how many things it has. Any suggestions that are more clear?

teoljungberg: `Violation#total` ?

thorncp:
Failing the build sounds like a cool idea, but agree that it should be a different PR. Created a card for it https://trello.com/c/MtnDjmfF

config/locales/en.yml
```diff
@@ -5,7 +5,7 @@ en:
   active_repos: "Active repos"
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
-  success_status: "Hound has reviewed all the changes!"
+  success_status: "%{violation_count} violations found."
```
gylaz: Can we embed ERB here for this? `pluralize(violation_count, "violation")`?

gylaz:
Nice work! Now, hopefully, people will actually read the status message :smile:.

app/models/violation.rb
```diff
@@ -21,6 +21,10 @@ def on_changed_line?
     line.changed?
   end
 
+  def count
```
srt32: I think `#count` be could be confusing with `.count`. How about `messages_count`? More verbose but maybe less confusing?

thorncp: Done in c8c0c24

config/locales/en.yml
thorncp: How about using the approach from the [Rails docs](http://guides.rubyonrails.org/i18n.html#pluralization)? Done in 04e19c50f057df431fc6b0bd4193526d961084eb

config/locales/en.yml
```diff
@@ -5,7 +5,9 @@ en:
   active_repos: "Active repos"
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
-  success_status: "Hound has reviewed all the changes!"
+  success_status:
+    one: "1 violation found."
+    other: "%{count} violations found."
```
gylaz: Sweet! Is there a way to do `zero: No violations are found. Woof!` or something cheezy like that?

srt32: :+1:

thorncp: Quick test says yes! Done in c409abbba07c86748d2da0e7a181cbe94570169c

```
         expected: ("test/repo", "headsha", "3 violations found.")
              got: ("test/repo", "headsha", "No violations found. Woof!")
```

app/models/violation.rb
```diff
@@ -21,6 +21,10 @@ def on_changed_line?
     line.changed?
   end
 
+  def messages_count
+    messages.size
```
teoljungberg: How about `delegate`-ing `messages_count` to `messages.count`?

thorncp: Sounds good. Done in e97b748f356b327ed5d92bcbb71037a871cbbe4f

app/models/violation.rb
```diff
@@ -9,6 +9,8 @@ class Violation < ActiveRecord::Base
 
   attr_writer :line
 
+  delegate :count, to: :messages, prefix: true
```
teoljungberg: How about adding a testcase for the delegation?

teoljungberg:
Other than the comments laid and addressed, this looks great!

app/models/violation.rb
thorncp: There's already a single test for it. Do you think more are needed?

https://github.com/thoughtbot/hound/blob/ct-github-status-count/spec/models/violation_spec.rb#L45-L51

teoljungberg: Ah, I missed that. What I would do is to test the delegation itself, and
not the actual implementation of the method

salbertson:
@thorncp will it be confusing if there are a lot of violations and this number doesn't match the number of comments?

app/models/violation.rb
thorncp: Wouldn't that be testing an implementation detail instead of the feature?

thorncp:
@salbertson: that's definitely a possibility. Do you think the message should reflect the total number versus the number commented? Something like `Hound reported 10 of 17 violations`?

srt32:
@thorncp or `Hound commented on 10 of 17 violations`?

gylaz:
So if a user fixes a few comments, the next status would say `Hound commented on 10 of 13 violations`? Seems confusing.

How about we leave as is, and then revisit when we get some feedback?

srt32:
:ok_hand:

thorncp:
Sounds good to me.

spec/support/helpers/github_api_helper.rb
```diff
@@ -425,13 +425,13 @@ def stub_status_requests(repo_name, sha)
...
       "pending",
-      "Hound is busy reviewing changes..."
+      anything
     )
     stub_status_request(
       repo_name,
       sha,
       "success",
-      "Hound has reviewed all the changes!"
+      anything
```
srt32: Maybe I'm missing something but where is `anything` defined?

spec/support/helpers/github_api_helper.rb
```diff
@@ -425,13 +425,13 @@ def stub_status_requests(repo_name, sha)
       repo_name,
       sha,
       "pending",
-      "Hound is busy reviewing changes..."
+      anything
```
srt32: Why do we need to change this stub? If it's `pending` isn't the message always the same ( "Hound is busy reviewing changes...")? The other change below makes sense though since the amount of violations will be variable.

srt32:
A couple questions around the stub changes but after those questions LGTM

spec/support/helpers/github_api_helper.rb
gylaz: @srt32 http://www.rubydoc.info/gems/rspec-mocks/RSpec/Mocks/ArgumentMatchers#anything-instance_method

gylaz:
Have the same question as @srt32 regarding the pending stub, other than that -- looks ready to me!

spec/support/helpers/github_api_helper.rb
srt32: TIL!

spec/support/helpers/github_api_helper.rb
thorncp: I'm not sure. Maybe I was thinking we'd tweak all the text? Anyway, updated in a8f961e76bf918fa1288dfafc1eae984ac0fc36c

PR #744, opened 2015-04-02
Use absolute file paths to fix excluding

gylaz:
We call `make_excludes_absolute` on the configuration
to satisfy RuboCop's `file_to_exclude?` since it calls `File.expand_path`
on the filename that we pass it.
Thus, all of our `Exclude` paths for individual cops get expanded to absolute as
well. We need to call `File.expand_path` on the filename in order for the two
paths to match.

This is kind of silly, since GitHub does not give us absolute paths, so we
should be fine with just the relative paths (relative to project root),
because they would match what the user puts in their config.

Basically, all of this (plus `make_excludes_absolute`) is because of
`file_to_exclude?`.

### DIFF
--- a/app/models/style_guide/ruby.rb
+++ b/app/models/style_guide/ruby.rb
@@ -28,7 +28,8 @@ module StyleGuide
     end
 
     def parsed_source(file)
-      RuboCop::ProcessedSource.new(file.content, file.filename)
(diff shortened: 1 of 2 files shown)


teoljungberg:
Looks great, merge at your own discretion

salbertson:
:+1:

PR #745, opened 2015-04-03
Update security document

salbertson:
We are now storing GitHub tokens in Postgres, so this update includes that.

### DIFF
--- a/doc/SECURITY.md
+++ b/doc/SECURITY.md
@@ -38,7 +38,8 @@ What happens when you authenticate your GitHub account
 ------------------------------------------------------
 
 Hound uses the [OmniAuth GitHub] Ruby gem to
-authenticate your GitHub account using [GitHub's OAuth2 flow][gh-oauth].
+authenticate your GitHub account using [GitHub's OAuth2 flow][gh-oauth]
+and provide Hound with a GitHub token.
 
 [OmniAuth GitHub]: https://github.com/intridea/omniauth-github
 [gh-oauth]: https://developer.github.com/v3/oauth/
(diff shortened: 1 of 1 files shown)


doc/SECURITY.md
```diff
@@ -46,9 +47,9 @@ authenticate your GitHub account using [GitHub's OAuth2 flow][gh-oauth].
 Using OAuth2 means we do not access your GitHub password
 and that you can revoke our access at any time.
 
-We need this token in order to refresh your GitHub repositories with Hound,
-which we do once, immediately after you authenticate your GitHub account.
-Later, you can manually refresh your GitHub repositories with Hound at any time.
+Your GitHub token in needed in order to fetch file content, comments, repo
```
srt32: `Your GitHub token in` => `Your GitHub token is`

doc/SECURITY.md
```diff
@@ -46,9 +47,9 @@ authenticate your GitHub account using [GitHub's OAuth2 flow][gh-oauth].
...
+information and update Pull Request status. This token is stored in our
+Postgres database on Heroku and is base64 encoded using a protected key.
```
gylaz: Link to (PostgreSQL) http://www.postgresql.org/?

doc/SECURITY.md
```diff
@@ -66,8 +67,7 @@ We pass your GitHub token to our [Ruby on Rails] app
 which runs on [Heroku].
 
 We temporarily store your GitHub token in the Redis database
```
gylaz: Should we add `Additionally,` to the beginning of the sentence, to not make it sound like that's the only place we store the token?

salbertson: I don't think so.

doc/SECURITY.md
salbertson: Thanks, done.

PR #746, opened 2015-04-03
Implement A/B testing on landing page

srt32:
* Configures https://github.com/splitrb/split and adds a dashboard at `/split`.
  The dashboard is behind basic auth.
* Runs an A/B test on the auth button text on the landing page. The tests are
  defined in `config/experiments.yml`.
* Upate auth button analytics tracking to pass through the `variation`.

**Deployment**
Setting the `SPLIT_ADMIN_PASSWORD` environment variable is required.

https://trello.com/c/qmurdq4Z


config/experiments.yml
```diff
@@ -0,0 +1,5 @@
+auth_button:
+  alternatives:
+    - "Sign In with GitHub"
+    - "Get Started"
```
srt32: What text would we like to test?

spec/requests/split_route_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  context "with auth" do
+    it "allows access" do
+      credentials = ActionController::HttpAuthentication::Basic.
+        encode_credentials(
+          "admin",
+          ENV["SPLIT_ADMIN_PASSWORD"]
+        )
+      get "/split", nil, { "HTTP_AUTHORIZATION" =>  credentials }
```
srt32: I enjoy seeing the curly braces for readability.

.sample.env
```diff
@@ -7,6 +7,8 @@ HOUND_GITHUB_TOKEN=your_github_token
 HOUND_GITHUB_USERNAME=your_github_username
 MAX_COMMENTS=10
 QUEUE=high,medium,low
+REDIS_URL=redis://localhost:6379
```
gylaz: Is this required? Redis gem should have sensible defaults.

app/views/application/_auth_button.haml
```diff
@@ -1,8 +1,12 @@
...
+  %span= ab_test("auth_button")
 
 - content_for :analytics do
   :javascript
     var link = document.getElementsByClassName("auth")[0];
-    window.analytics.trackLink(link, "Clicked Sign In");
+    window.analytics.trackLink(
+      link,
+      "Clicked Sign In",
+      { variation: #{ ab_test("auth_button") } }
```
gylaz: No spaces in `{}` when interpolating?

config/initializers/split.rb
```diff
@@ -0,0 +1,9 @@
+Split.configure do |config|
+  config.experiments = YAML.load_file "config/experiments.yml"
+end
+
+Split.redis = ENV.fetch("REDIS_URL")
```
gylaz: I don't think we need this. It should just pick up defaults.

.sample.env
```diff
@@ -7,6 +7,8 @@ HOUND_GITHUB_TOKEN=your_github_token
...
 RESQUE_ADMIN_PASSWORD=password
 RESQUE_JOB_TIMEOUT=120
 SECRET_KEY_BASE=secret_key_base
+SPLIT_ADMIN_PASSWORD="split_password"
```
gylaz: How about dropping the quotes here for consistency?

.sample.env
srt32: We have to set this variable on staging / production so it seems reasonable to require it here.

config/initializers/split.rb
srt32: You are correct but the implementation (https://github.com/splitrb/split/blob/ee1f6208fcf4506f02207d829eb2911b6898d3e5/lib/split.rb#L56) defaults to localhost so we could deploy the app without the variable and not know until something breaks later. With the setting here, the app breaks on boot if the variable is missing.

app/views/application/_auth_button.haml
srt32: Nice! Updated in 4a63e4f

.sample.env
srt32: Good call! Done in 43c7939

srt32:
@gylaz left a couple responses. Curious what you think.

gylaz:
Looks good to me.

spec/requests/split_route_spec.rb
thorncp: What about the unnecessary spacing? Two spaces after the hash rocket.

srt32: Wah! Missed that. Updated in 3f0a4e4

config/experiments.yml
srt32: Ping @frechg, did you have any ideas?

srt32:
The build is failing but will be fixed when https://github.com/thoughtbot/hound/commit/081fcb642ccd291e58027e5d4b6d19d8522ebf2e lands.

PR #748, opened 2015-04-04
Lint `coffee.erb`

teoljungberg:
To avoid eval-ing `erb` tags, due to security reasons, we could instead
replace it the tag with a dummy value. I picked an emtpy string since
that seemed the easier way of solving it.

Thoughts?

### DIFF
--- a/app/models/style_checker.rb
+++ b/app/models/style_checker.rb
@@ -40,7 +40,7 @@ class StyleChecker
     case filename
     when /.+\.rb\z/
       StyleGuide::Ruby
-    when /.+\.coffee(\.js)?\z/
+    when /.+\.coffee(\.js)?(\.erb)?\z/
       StyleGuide::CoffeeScript
     when /.+\.js\z/
       StyleGuide::JavaScript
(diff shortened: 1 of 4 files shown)


teoljungberg:
Should we wait with merging this unil #727 is merged?

gylaz:
Looks good to me. I say merge it.

teoljungberg:
@thorncp do you see anything suspicous about this approach?

thorncp:
The only thing I can think of is the possibility of substituting `123` leaving the content in a state that causes the linter to report a false violation. Does that seem likely? I can't think of a situation like that offhand.

teoljungberg:
@thorncp good point, would it suffice to just test to run against https://github.com/thoughtbot/hound/blob/master/app/assets/javascripts/directives/repo_list.js.coffee.erb and see if any problems are caused?

thorncp:
I'm not sure how likely it is to be an issue. It may just make sense to ship it and see what happens. I think at worst Hound comments on something that isn't an issue, which it already does occasionally...

teoljungberg:
I'll ship this in that case, if there is a problem we'll get it sorted out when it occurs.

PR #747, opened 2015-04-03
Improve security documentation

gylaz:
* Correct a few points that weren't quite correct.
* Update for SCSS-Lint

PR #749, opened 2015-04-04
Run migrations on travis

teoljungberg:


teoljungberg:
The test-failure seems to be fixed by, https://github.com/thoughtbot/hound/commit/081fcb642ccd291e58027e5d4b6d19d8522ebf2e so this PR is waiting for #747 to be merged

gylaz:
We don't actually need to run migrations on Travis, as it re-creates the db and loads the schema on each turn.

teoljungberg:
Oh, in that case I'll close this

PR #752, opened 2015-04-10
Wrap repo sync inserts in a transaction

srt32:
This wraps the loop in `RepoSynchronization#start` in a transaction. This way
all the `INSERT`s will happen together or not at all, which should lead to more
consistent repo syncs.

### DIFF
--- a/app/services/repo_synchronization.rb
+++ b/app/services/repo_synchronization.rb
@@ -4,10 +4,13 @@ class RepoSynchronization
 
   def start
     user.repos.clear
+    repos = api.repos
 
-    api.repos.each do |resource|
-      attributes = repo_attributes(resource.to_hash)
-      user.repos << Repo.find_or_create_with(attributes)
+    Repo.transaction do
+      repos.each do |resource|
+        attributes = repo_attributes(resource.to_hash)
+        user.repos << Repo.find_or_create_with(attributes)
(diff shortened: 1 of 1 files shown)


app/services/repo_synchronization.rb
```diff
@@ -5,9 +5,11 @@ class RepoSynchronization
   def start
     user.repos.clear
 
-    api.repos.each do |resource|
-      attributes = repo_attributes(resource.to_hash)
-      user.repos << Repo.find_or_create_with(attributes)
+    Repo.transaction do
+      api.repos.each do |resource|
```
gylaz: `api.repos` is an expensive call, as it fetches all of the repos for a user from GitHub (some users have up to a thousand of repos). Does it make sense to move it out of the `transaction` block? It might not affect anything, and this comment can be ignored.

gylaz:
One comment. :+1:

app/services/repo_synchronization.rb
srt32: Sounds good, done in 1392999

srt32:
@gylaz thanks for the feedback. Mind taking one more look?

gylaz:
Ready to merge.

srt32:
Thanks @gylaz!

PR #753, opened 2015-04-10
Set StyleGuide::CoffeeScript#name

srt32:
This PR is a follow up to
https://github.com/thoughtbot/hound/commit/6cec20ad0cb47ce807e631ede531efd3f39ba067
and fixes an issue where the new `coffeescript` would not be picked up.


thorncp:
LGTM :swimmer:

TIL about Rails' `String#strip_heredoc` :+1:

srt32:
Thanks @thorncp

PR #755, opened 2015-04-14
Update dotenv-rails

teoljungberg:
To quell warning:
```
Auto-loading of `.env.test` will be removed in 1.0. See
"https://github.com/bkeepers/dotenv-deployment" if you would like to
continue using this feature.
```

`dotenv-deployment` is deprecated, so updating `dotenv-rails` quells the warning


gylaz:
`dotenv-deployment` is deprecated. I don't think it's worth adding another, deprecated gem in this case. Does updating `dotenv-rails` get rid of the warning? (You'll also have to upgrade `foreman` first).

teoljungberg:
That worked perfectly @gylaz fixed in a69c53a

gylaz:
Great! Mind updating the commit (and the PR) message/description? Then ready to merge.

thorncp:
LGTM!

PR #756, opened 2015-04-14
Automatically retry failed jobs

bernerdschaefer:
This changeset removes resque-retry and resque-timeout,
which are incompatible with resque under ActiveJob.
They are replaced with a new ApplicationJob class
and a rewrite of the Retryable module.

A new scheduler task is added to the Procfile,
which runs an instance of resque-scheduler and
executes delayed jobs.

Exceptions in jobs are no longer explicitly
caught and sent via `Raven.capture_exception`,
and instead fall through to the default
Resque::Failure::Sentry handler.


Gemfile
```diff
@@ -8,6 +8,7 @@ gem "angularjs-rails"
 gem "angular_rails_csrf"
 gem "attr_extras"
 gem "bourbon"
+gem "byebug", require: false
```
croaky: Should this be dev/test only?

Procfile
```diff
@@ -1,2 +1,3 @@
 web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
 resque: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 bundle exec rake resque:work
+scheduler: bundle exec rake resque:scheduler
```
croaky: Not a big deal, but can we name this process something different? Could be confused in logs with Heroku Scheduler if we have any or add any Heroku Scheduler tasks.

config/initializers/resque.rb
```diff
@@ -12,17 +13,19 @@
...
+Resque::Failure.backend = Resque::Failure::Multiple
 
-Resque::Plugins::Timeout.timeout = (ENV["RESQUE_JOB_TIMEOUT"] || 120).to_i
+Resque::Failure::Sentry.logger = "resque"
 
 Resque::Server.use(Rack::Auth::Basic) do |user, password|
   password == ENV['RESQUE_ADMIN_PASSWORD']
 end
+
+ApplicationJob.timeout = (ENV["RESQUE_JOB_TIMEOUT"] || 120).to_i
```
croaky: `ENV.fetch` for consistency?

lib/ext/active_job/base.rb
```diff
@@ -0,0 +1,25 @@
+# Backport of https://github.com/rails/rails/pull/18260.
```
croaky: Oof! Any idea when it will be in a release? Looks like this was back in December, I would think we wouldn't need the backport at this point. Are we behind on Rails versions?

spec/jobs/accept_org_invitations_job_spec.rb
```diff
@@ -22,22 +22,10 @@
     allow(GithubApi).to receive(:new).and_raise(Resque::TermException.new(1))
     allow(AcceptOrgInvitationsJob.queue_adapter).to receive(:enqueue)
 
-    job = AcceptOrgInvitationsJob.perform_now
+    job = AcceptOrgInvitationsJob.new
+    job.perform_now
 
     expect(AcceptOrgInvitationsJob.queue_adapter).
       to have_received(:enqueue).with(job)
```
croaky: Any benefit to checking `enqueued_jobs` or `performed_jobs`?

https://github.com/rails/rails/blob/master/activejob/lib/active_job/test_helper.rb#L282-L288

Gemfile
gylaz: We already have `byebug` in dev/test. Maybe this was for debugging?

app/jobs/application_job.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  cattr_accessor :timeout
+
+  rescue_from(Octokit::Unauthorized) do |exception|
+    raise exception
+  end
+
+  around_perform do |_, block|
+    Timeout::timeout(ApplicationJob.timeout, &block)
+  end
```
gylaz: In which order do the multiple `around_permorm` blocks fire? As defined in the file?

lib/ext/active_job/base.rb
bernerdschaefer: From the PR:

> > Also, is there any chance of this making it onto 4-2-stable? I'm pretty sure it's completely non-breaking.
>
> I don't think it's a good idea to apply to stable. It's a feature upgrade.

There are no released versions of rails which include this change.

Gemfile
bernerdschaefer: Thanks. Removed in 444939e.

Procfile
bernerdschaefer: Renamed to resque_scheduler.

config/initializers/resque.rb
bernerdschaefer: Done 444939e.

app/jobs/application_job.rb
bernerdschaefer: It would appear they run in the order they are defined.

lib/ext/active_job/base.rb
bernerdschaefer: @croaky This will be in Rails 5.

app/jobs/application_job.rb
gylaz: Does it make sense to combine them? Would that even work?

config/initializers/resque.rb
```diff
@@ -12,17 +13,19 @@
...
-Resque::Plugins::Timeout.timeout = (ENV["RESQUE_JOB_TIMEOUT"] || 120).to_i
+Resque::Failure::Sentry.logger = "resque"
 
 Resque::Server.use(Rack::Auth::Basic) do |user, password|
   password == ENV['RESQUE_ADMIN_PASSWORD']
 end
+
+ApplicationJob.timeout = ENV.fetch("RESQUE_JOB_TIMEOUT", 30).to_i
+Retryable.retry_delay = ENV.fetch("RESQUE_RETRY_DELAY", 30).to_i
+Retryable.retry_attempts = ENV.fetch("RESQUE_RETRY_ATTEMPTS", 10).to_i
```
gylaz: Does it make sense to change these prefixes to just `JOB_`?

config/initializers/resque.rb
```diff
@@ -12,17 +13,19 @@
...
+Resque::Failure.backend = Resque::Failure::Multiple
 
-Resque::Plugins::Timeout.timeout = (ENV["RESQUE_JOB_TIMEOUT"] || 120).to_i
+Resque::Failure::Sentry.logger = "resque"
 
 Resque::Server.use(Rack::Auth::Basic) do |user, password|
   password == ENV['RESQUE_ADMIN_PASSWORD']
 end
+
+ApplicationJob.timeout = ENV.fetch("RESQUE_JOB_TIMEOUT", 30).to_i
```
gylaz: `30` is pretty low. We do have long running jobs if there are a lot of files to fetch and review. It was `120` previously, should we keep that number?

spec/jobs/application_job_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    expect(job).to have_received(:retry_job)
+  end
+
+  it "does not retry on octokit authorization exception" do
+    job = ApplicationJob.new
+    allow(job).to receive(:perform).and_raise(Octokit::Unauthorized)
+    allow(job).to receive(:retry_job)
+
+    expect { job.perform_now }.to raise_error(Octokit::Unauthorized)
```
gylaz: Newline after this to denote the execution phase?

spec/jobs/accept_org_invitations_job_spec.rb
bernerdschaefer: Maybe as a followup PR? I tried not to muck around with the existing specs too much.

config/initializers/resque.rb
bernerdschaefer: Must have been a bad copy/paste. Addressed in b8a4fac.

spec/jobs/application_job_spec.rb
bernerdschaefer: Done in b8a4fac.

app/jobs/application_job.rb
bernerdschaefer: It would work, I'm not sure what advantage it gives. As it is, each around block provides a single feature. If its the ordering that you're worried about, I think I'm okay with these running in whatever order activesupport chooses.

gylaz: Sounds good on keeping them as is. My concern was that there are multiple `around_perform` blocks, potentially causing a head scratch why there are two of them.

lib/ext/active_job/base.rb
croaky: Word. Maybe a Trello card to remind us to come back and remove this during Rails 5 upgrade? And/or comment about "Rails 5" for greppability later?

spec/jobs/retryable_spec.rb
```diff
@@ -0,0 +1,73 @@
+require "spec_helper"
```
thorncp: Noticed this file isn't syntax highlighted on GitHub, then saw

```
% file spec/jobs/retryable_spec.rb
spec/jobs/retryable_spec.rb: ASCII C++ program text
```

What's up with that?

For reference

```
% file spec/jobs/small_build_job_spec.rb
spec/jobs/small_build_job_spec.rb: ASCII text
```

lib/ext/active_job/base.rb
bernerdschaefer: Done in 4b5370f9e22cfa21b94833b4a66fc216213055f6.

spec/jobs/retryable_spec.rb
bernerdschaefer: No idea what's up with this. The `file` check is likely unrelated, since almost every ruby file in the project shows up like that:

```
> find . -name "*.rb" | xargs file | grep -c "C++"
128
```

gylaz:
:+1:

PR #758, opened 2015-04-17
Refactor display_onboarding? query

srt32:
There was an N+1 query firing whern we checked if a `user` has had a previous
`build` (used to determine if we should display onboarding copy).

This change refactors the check to a single query. It may make sense later to
move this query elsewhere entirely.

Fixes
https://trello.com/c/axeF15jZ/566-investigate-n-1-query-on-builds-in-web-process

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,6 +5,9 @@ module ApplicationHelper
   end
 
   def display_onboarding?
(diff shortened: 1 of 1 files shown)


thorncp:
I do think this is beyond the concern of a helper, but it LGTM to fix the production issue.

srt32:
Thanks, @thorncp!

PR #762, opened 2015-04-21
Return an existing subscription during activation

gylaz:
We've recently encountered a situation where we believe the customer had
a subscription, but was attempting to activate again -- this can also be
caused by duplicate, simultaneous requests for the same user.

There is not reason to not return the existing subscription, if it
exists and avoid charging the user again via Stripe.

### DIFF
--- a/app/services/repo_subscriber.rb
+++ b/app/services/repo_subscriber.rb
@@ -10,26 +10,7 @@ class RepoSubscriber
   end
 
   def subscribe
-    customer = if user.stripe_customer_id.present?
-      payment_gateway_customer
-    else
-      create_stripe_customer
-    end
-
-    stripe_subscription = customer.subscriptions.create(
-      plan: repo.plan_type,
-      metadata: { repo_id: repo.id }
-    )
-
-    repo.create_subscription!(
-      user_id: user.id,
-      stripe_subscription_id: stripe_subscription.id,
-      price: repo.plan_price
-    )
-  rescue => error
-    report_exception(error)
-    stripe_subscription.try(:delete)
-    nil
+    repo.subscription || create_subscription
   end
 
   def unsubscribe
@@ -46,11 +27,32 @@ class RepoSubscriber
 
   private
 
+  def create_subscription
+    stripe_subscription = stripe_customer.subscriptions.create(
+      plan: repo.plan_type,
+      metadata: { repo_id: repo.id }
+    )
+
+    create_subscription_record(stripe_subscription.id)
+  rescue => error
+    report_exception(error)
+    stripe_subscription.try(:delete)
+    nil
+  end
+
   def report_exception(error)
     Raven.capture_exception(
       error,
       extra: { user_id: user.id, repo_id: repo.id }
     )
+  end
+
+  def stripe_customer
+    if user.stripe_customer_id.present?
+      payment_gateway_customer
+    else
+      create_stripe_customer
+    end
   end
 
   def payment_gateway_customer
@@ -67,5 +69,13 @@ class RepoSubscriber
     user.update(stripe_customer_id: stripe_customer.id)
 
     stripe_customer
+  end
+
+  def create_subscription_record(stripe_subscription_id)
+    repo.create_subscription!(
+      user_id: user.id,
+      stripe_subscription_id: stripe_subscription_id,
+      price: repo.plan_price
+    )
   end
 end
--- a/spec/services/repo_subscriber_spec.rb
+++ b/spec/services/repo_subscriber_spec.rb
@@ -82,6 +82,22 @@ describe RepoSubscriber do
         expect(stripe_delete_request).to have_been_requested
       end
     end
+
+    context "when repo already has a subscription" do
+      context "and it is not marked as deleted" do
+        it "returns the existing subscription" do
+          subscription = create(:subscription)
+
+          result = RepoSubscriber.subscribe(
+            subscription.repo,
+            subscription.user,
+            "cardtoken"
+          )
+
+          expect(result).to eq subscription
+        end
+      end
+    end
   end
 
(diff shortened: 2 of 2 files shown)


spec/services/repo_subscriber_spec.rb
```diff
@@ -82,6 +82,20 @@
         expect(stripe_delete_request).to have_been_requested
       end
     end
+
+    context "when repo already has a subscription" do
```
srt32: Worth adding a test around the behavior when a repo has a subscription that has been deleted?

app/services/repo_subscriber.rb
```diff
@@ -10,6 +10,24 @@ def self.unsubscribe(repo, user)
...
+
+    repo.subscription.destroy
+  rescue => error
+    report_exception(error)
+    nil
+  end
+
+  private
+
+  def create_subscription
```
thorncp: I guess you *technically* changed this method, but the violation isn't really caused by this PR.

app/services/repo_subscriber.rb
```diff
@@ -10,6 +10,24 @@ def self.unsubscribe(repo, user)
   end
 
   def subscribe
+    repo.subscription || create_subscription
```
srt32: Would we want the status code returned to the user (would require an update elsewhere) to be different?

srt32:
A couple questions but LGTM

spec/services/repo_subscriber_spec.rb
gylaz: Why not!?

app/services/repo_subscriber.rb
gylaz: Caught me!

app/services/repo_subscriber.rb
gylaz: Status code of Stripe?

thorncp:
Should we update any of the existing subscription's attributes when this happens? Does it seem likely that the price could change between an existing subscription being deactivated then reactivated again? Is that handled somewhere else?

gylaz:
@thorncp Historically, we've been grandfathering people into better prices.

gylaz:
@srt32 @thorncp Ready for RR.

spec/services/repo_subscriber_spec.rb
```diff
@@ -82,6 +82,47 @@
...
+        end
+      end
+
+      context "and it is marked as deleted" do
+        it "returns the existing subscription" do
+          repo = create(:repo, private: true)
+          subscription = create(:subscription, deleted_at: Time.now, repo: repo)
+          stub_customer_find_request
+          customer_request = stub_customer_create_request(subscription.user)
+          subscription_request =
```
thorncp: Looks like L107-108 can be deleted?

thorncp:
LGTM when Hound is happy :+1: :clock7:

spec/services/repo_subscriber_spec.rb
```diff
@@ -82,6 +82,44 @@
...
+            repo.reload,
+            subscription.user,
+            "cardtoken"
+          )
+
+          expect(customer_request).to have_been_requested
+          expect(subscription_request).to have_been_requested
+          expect(result).not_to be_deleted
+        end
+      end
```
gylaz: Now that I added this test, I'm not sure it brings any value. A deleted subscription is a responsibility of `paranoia` gem, and this feels like making sure that it's behaving properly -- which is a heavy test at this layer, or arguably isn't even our concern.

The only way to break this test would be to do `Subscription.unscoped.find_by(repo_id: repo.id)` instead of `repo.subscription`, and that likely would be a flag in itself.

Looking for thoughts on this one?

srt32: My thought was that I didn't know what was supposed to happen if a user 1) activates a repo and subscribes, then 2) unsubscribes (so there is a deleted subscription) and then 3) reactivates and subscribes again. What happens? Perhaps that logic belongs elsewhere as you said.

spec/services/repo_subscriber_spec.rb
```diff
@@ -82,6 +82,44 @@
...
+            subscription.user,
+            "cardtoken"
+          )
+
+          expect(result).to eq subscription
+        end
+      end
+
+      context "and it is marked as deleted" do
+        it "returns the existing subscription" do
```
thorncp: Does this describe what's being tested? It looks like the code *isn't* returning the existing subscription, and is instead requesting a new one from Stripe.

gylaz: Yes. Good catch.

spec/services/repo_subscriber_spec.rb
thorncp: I think it makes sense for the existing, deleted subscriptions to remain untouched. They have pricing info that has potential to change, so it would make sense to preserve that information by not reusing the record.

As for this spec, I think I agree with @gylaz. It's basically testing that `paranoia` is doing its job. I think it's ok to :scissors: it.

gylaz: @srt32 This code doesn't cover it. I think the de-activation, then re-activation is a whole topic on its own.

srt32:
The code looks good as is but thinking about it some more would it make sense to put this logic into the controller level (https://github.com/thoughtbot/hound/blob/c0a8e290f2e8f04c74dd60d64b7458951953035b/app/controllers/subscriptions_controller.rb#L44) since we're dealing with what we think is a duplicate request problem as opposed to a business logic problem?

gylaz:
> would it make sense to put this logic into the controller level 

This is definitely up for debate.  I can see that calling `RepoSubscriber.subscribe(...)` repeatedly with the same data, should return the subscription that was created -- thus business logic.

PR #763, opened 2015-04-23
Update capybara-webkit

jferris:
In some situations, JavaScript tests were displaying native alert
dialogs and crashing. This is now fixed in capybara-webkit.

Also blocks remote URLs to prevent warnings and slow tests.

Fixes #403.


srt32:
:+1:

PR #764, opened 2015-04-25
Only return current pull request files

thorncp:
`PullRequest#pull_request_files` now only returns the *current* files of a pull request. Previously it would return files that had been removed, and left the responsibility of not operating on them up to other parts of the application.

This is the first step in an attempt to split up #727 into smaller chunks that can be reviewed more thoroughly. I started with this small change so I could get people's thoughts on this approach to try to determine if splitting it up is worth pursuing.

### DIFF
--- a/app/models/pull_request.rb
+++ b/app/models/pull_request.rb
@@ -1,14 +1,14 @@
 class PullRequest
   pattr_initialize :payload, :token
 
+  FILE_REMOVED_STATUS = "removed"
+
   def comments
     @comments ||= user_github.pull_request_comments(full_repo_name, number)
   end
 
   def pull_request_files
-    @pull_request_files ||= user_github.
-      pull_request_files(full_repo_name, number).
-      map { |file| build_commit_file(file) }
+    @pull_request_files ||= changed_pull_request_files
   end
 
   def comment_on_violation(violation)
@@ -41,6 +41,13 @@ class PullRequest
 
   def build_commit_file(file)
     CommitFile.new(file, head_commit)
+  end
+
+  def changed_pull_request_files
+    user_github.
+      pull_request_files(full_repo_name, number).
+      reject { |file| file.status == FILE_REMOVED_STATUS }.
+      map { |file| build_commit_file(file) }
   end
 
   def user_github
--- a/spec/models/commit_file_spec.rb
+++ b/spec/models/commit_file_spec.rb
@@ -6,24 +6,6 @@ require "app/models/patch"
 require "app/models/unchanged_line"
 
 describe CommitFile do
-  describe "#removed?" do
-    context "when status is removed" do
-      it "returns true" do
-        commit_file = commit_file(status: "removed")
-
-        expect(commit_file).to be_removed
-      end
-    end
-
-    context "when status is added" do
-      it "returns false" do
-        commit_file = commit_file(status: "added")
-
-        expect(commit_file).not_to be_removed
-      end
-    end
-  end
-
   describe "#line_at" do
     context "with a changed line" do
       it "returns a line at the given line number" do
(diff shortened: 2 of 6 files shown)


croaky:
Nice, do this reduce the number of our requests to GitHub?

app/models/pull_request.rb
```diff
@@ -1,14 +1,15 @@
 class PullRequest
   pattr_initialize :payload, :token
 
+  FILE_ADDED_STATUS = "added"
```
croaky: Do we need this constant yet?

spec/models/commit_file_spec.rb
```diff
@@ -47,20 +29,10 @@
   end
 
   describe "#content" do
-    context "when file is removed" do
-      it "returns nil" do
-        commit_file = commit_file(status: "removed")
+    it "returns content string" do
+      commit_file = commit_file(status: "modified")
```
croaky: Is `status: "modified"` important in this spec?

spec/models/pull_request_spec.rb
```diff
@@ -80,6 +80,27 @@
...
+  describe "#pull_request_files" do
+    it "does not include removed files" do
+      added_file = double(
+        filename: "foo.rb",
+        status: PullRequest::FILE_ADDED_STATUS,
+      )
+      removed_file = double(
+        filename: "bar.rb",
+        status: PullRequest::FILE_REMOVED_STATUS,
+      )
```
croaky: Should we include a modified file in the setup data?

croaky:
This looks good to me. I agree these smaller pull requests are easier to manage.

app/models/pull_request.rb
thorncp: I added it because the spec uses it. Do you think it's overkill to have a constant if it's not used in production code?

spec/models/commit_file_spec.rb
thorncp: I don't think so, but I wanted to keep the changes here focused on just removing "removed" support. This spec has another issue too, where the `"some content"` expectation is very magical, and I don't think I want to include a refactoring for it here.

spec/models/pull_request_spec.rb
thorncp: Sounds good to me. Done in e7d1532dd45b201f8822596acf2600bbab05bf06

app/models/pull_request.rb
thorncp: Removed in e7d1532dd45b201f8822596acf2600bbab05bf06

thorncp:
It won't reduce the number of GitHub requests. It just centralizes the responsibility of checking if the file is relevant.

teoljungberg:
Looks great @thorncp

app/models/pull_request.rb
```diff
@@ -43,6 +43,13 @@ def build_commit_file(file)
     CommitFile.new(file, head_commit)
   end
 
+  def current_pull_request_files
```
gylaz: We've used the concept of "changed" files elsewhere (for modified and added files). Might that be a more descriptive prefix instead of "current"?

gylaz:
I had one question. LGTM.

app/models/pull_request.rb
teoljungberg: I agree, fixed in 45ce6d2

gylaz:
Looks good to merge.

teoljungberg:
@thorncp You're current committing as @houndci

thorncp:
@teoljungberg I amended the commit author to be `Hound Team <hound@thoughtbot.com>" in an attempt to capture everyone that put effort in it it since it stems from #727. Didn't anticipate it attaching it to the Hound's GitHub user though...

thorncp:
Updated author in 4cfdb03bc7e7779ffab320ae0b8c8ed5b2ac2443

PR #765, opened 2015-04-25
Introduce FileReview model

thorncp:
This model will represent the progress of a file review, particularly those done by an external system, e.g. IronWorker.

This is another extraction of a small chunk of functionality from #727. It went under the name `BuildWorker` in that PR.

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -1,5 +1,6 @@
 class Build < ActiveRecord::Base
   belongs_to :repo
+  has_many :file_reviews, dependent: :destroy
   has_many :violations, dependent: :destroy
 
   before_create :generate_uuid
--- a/app/models/file_review.rb
+++ b/app/models/file_review.rb
@@ -1,0 +1,13 @@
+class FileReview < ActiveRecord::Base
+  belongs_to :build
+
+  validates :build, presence: :true
+
+  def completed?
+    completed_at?
+  end
+
+  def running?
+    !completed?
+  end
+end
--- a/spec/models/build_spec.rb
+++ b/spec/models/build_spec.rb
@@ -1,9 +1,15 @@
 require "rails_helper"
 
 describe Build do
-  it { should belong_to :repo }
-  it { should validate_presence_of :repo }
-  it { should have_many(:violations).dependent(:destroy) }
+  describe "associations" do
+    it { should belong_to :repo }
+    it { should have_many(:file_reviews).dependent(:destroy) }
+    it { should have_many(:violations).dependent(:destroy) }
+  end
+
+  describe "validations" do
+    it { should validate_presence_of :repo }
+  end
 end
 
 describe Build, '#status' do
--- a/spec/models/file_review_spec.rb
+++ b/spec/models/file_review_spec.rb
@@ -1,0 +1,39 @@
+require "rails_helper"
+
+describe FileReview do
+  describe "associations" do
+    it { should belong_to :build }
+  end
+
+  describe "validations" do
+    it { should validate_presence_of :build }
+  end
+
+  describe "#completed?" do
+    it "returns true when completed_at is set" do
+      file_review = FileReview.new(completed_at: Time.zone.now)
+
+      expect(file_review).to be_completed
+    end
+
+    it "returns false when completed_at is nil" do
+      file_review = FileReview.new
+
+      expect(file_review).not_to be_completed
+    end
+  end
+
+  describe "#running?" do
+    it "returns true when complete_at is set" do
+      file_review = FileReview.new(completed_at: Time.zone.now)
+
+      expect(file_review).not_to be_running
+    end
+
+    it "returns false when completed_at is nil" do
+      file_review = FileReview.new(completed_at: nil)
+
+      expect(file_review).to be_running
+    end
+  end
+end
--- a/db/migrate/20150425204143_create_file_reviews.rb
+++ b/db/migrate/20150425204143_create_file_reviews.rb
@@ -1,0 +1,10 @@
+class CreateFileReviews < ActiveRecord::Migration
+  def change
+    create_table :file_reviews do |t|
+      t.belongs_to :build, index: true, foreign_key: true, null: false
+      t.datetime :completed_at
+
+      t.timestamps null: false
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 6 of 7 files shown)


app/models/build.rb
```diff
@@ -1,6 +1,7 @@
 class Build < ActiveRecord::Base
   belongs_to :repo
   has_many :violations, dependent: :destroy
+  has_many :file_reviews, dependent: :destroy
```
croaky: > Order ActiveRecord associations alphabetically by association type, then attribute name. [Example](https://github.com/thoughtbot/guides/blob/master/style/rails/sample.rb#L2-L4)

spec/models/file_review_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    it "returns false where completed_at is nil" do
+      file_review = FileReview.new
+
+      expect(file_review.completed?).to eq false
+    end
+  end
+
+  describe "#running?" do
+    it "returns true where complete_at is nil" do
+      file_review = FileReview.new(completed_at: Time.now)
```
croaky: `Time.zone.now` in this spec?

spec/models/file_review_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      expect(file_review.running?).to eq false
+    end
+
+    it "returns false where completed_at is set" do
+      file_review = FileReview.new(completed_at: nil)
+
+      expect(file_review.running?).to eq true
```
croaky: > Avoid checking boolean equality directly. Instead, write predicate methods and use appropriate matchers. [Example](https://github.com/thoughtbot/guides/blob/master/style/testing/predicate_tests_spec.rb).

```ruby
expect(file_review).to be_running
```

Same for others in this spec?

spec/models/file_review_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  describe "#running?" do
+    it "returns true where complete_at is nil" do
+      file_review = FileReview.new(completed_at: Time.now)
+
+      expect(file_review.running?).to eq false
+    end
+
+    it "returns false where completed_at is set" do
```
croaky: maybe "is false", "is true" in the spec descriptions? `#running?` feels like we're talking about the `FileReview`'s state.

Maybe also "when" instead of "where" since we're talking about time.

"Describe FileReview#running? It is false when completed_at is set"

croaky:
Looks good to me.

spec/models/file_review_spec.rb
thorncp: Reads much better. Done in 654d59a8b6ac00679d256ea0de8e50729437dee8

spec/models/file_review_spec.rb
thorncp: Sure! 654d59a8b6ac00679d256ea0de8e50729437dee8

app/models/build.rb
thorncp: Done in 654d59a8b6ac00679d256ea0de8e50729437dee8. This would be a cool if Hound could bark at things like this.

spec/models/file_review_spec.rb
thorncp: Tweaked wording in 654d59a8b6ac00679d256ea0de8e50729437dee8. Also noticed that the original descriptions were the opposites of their spec implementations!

teoljungberg:
First of all, thank you @thorncp for this! I've been meaning to start extracting the work done in #727 into smaller and more managable PRs. I think I have time to continue your work next week, could you create a trello card with checklists for the different parts, and check those you've done, so we don't overlap any work?

We discussed having a `language` field on the model for easier debugging, do you think we should add that here aswell?

app/models/build.rb
```diff
@@ -1,5 +1,6 @@
 class Build < ActiveRecord::Base
   belongs_to :repo
+  has_many :file_reviews, dependent: :destroy
```
teoljungberg: How about seperating the different associations with a newline?

spec/models/build_spec.rb
```diff
@@ -1,9 +1,15 @@
 require "rails_helper"
 
 describe Build do
-  it { should belong_to :repo }
-  it { should validate_presence_of :repo }
-  it { should have_many(:violations).dependent(:destroy) }
+  describe "associations" do
+    it { should belong_to :repo }
+    it { should have_many(:violations).dependent(:destroy) }
```
teoljungberg: Should we sort these tests to reflect the order of the associations in the model?

spec/models/file_review_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    it { should belong_to :build }
+  end
+
+  describe "validations" do
+    it { should validate_presence_of :build }
+  end
+
+  describe "#completed?" do
+    it "returns true when completed_at is set" do
+      file_review = FileReview.new(completed_at: Time.zone.now)
```
teoljungberg: Have you considered using factory girl in the tests?

teoljungberg:
Besides a few remarks, looks great!

app/models/file_review.rb
```diff
@@ -0,0 +1,13 @@
+class FileReview < ActiveRecord::Base
+  belongs_to :build
+
+  validates :build, presence: :true
+
+  def completed?
+    completed_at?
+  end
+
+  def running?
```
teoljungberg: What about renaming this to `in_progress?`

spec/models/file_review_spec.rb
gylaz: Looks like this test only depends on the presence of `completed_at`, thus just the object is enough.
As far as the speed of tests: `double > Ruby object > build > build_stubbed > create`

gylaz:
:+1:

spec/models/file_review_spec.rb
teoljungberg: I prefer never stubbing the object under test, a double feels unnecessary

---
Teo Ljungberg

gylaz: My point wasn't for `double` here -- as it obviously doesn't make sense to test what you've just stubbed. I simply provided a general speed hierarchy.

app/models/build.rb
srt32: Do we get `cascade: delete` with the FK? If so, can we remove `dependent: :destroy`?

spec/models/file_review_spec.rb
srt32: "speed hierarchy" - The Lazarev Doctrine.

srt32:
LGTM

PR #766, opened 2015-04-27
Update SCSSLint

teoljungberg:
- Update SCSS config
- Due to a change in `0.35.0`, we have to support create an actual file
  for each file we want to lint. This is being handled by creating a
  tempfile and then unlinking it after each run

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ gem "resque-scheduler"
 gem "resque-sentry"
 gem "rubocop", "0.29.1"
 gem "sass-rails"
-gem "scss-lint", "0.34.0", require: false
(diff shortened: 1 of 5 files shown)


teoljungberg:
We're currently discussing in brigade/scss-lint#435 if we can
re-implement support for giving `SCSSLint::Runner#run` file contents
aswell as proper files

gylaz:
Looks good.

teoljungberg:
I'm going to let this sit until we see what happeneds with
https://github.com/brigade/scss-lint/issues/435

thorncp:
Closing this since SCSS is now a worker. Created https://github.com/thoughtbot/hound-scss/issues/10 in its stead.

PR #768, opened 2015-05-01
Handle single (non-array) SCSS excludes

gylaz:
Users can provide excludes for SCSS linters as an array:

    exclude:
      - lib/**
      - spec/assets/**

or as a string:

    exclude: spec/assets/**

We need to run the config hash through SCSSLint's methods to normalize
all of the options. In order to do that, we need to write to a file and
provide a path to that file.

Related error: https://app.getsentry.com/hound-1/production/group/63380345/

### DIFF
--- a/app/models/style_guide/scss.rb
+++ b/app/models/style_guide/scss.rb
@@ -32,24 +32,31 @@ module StyleGuide
     end
 
     def config
-      SCSSLint::Config.new(custom_options || default_options)
+      @config ||= SCSSLint::Config.load(
+        custom_config_file.path,
+        merge_with_default: false
+      )
     end
 
-    def custom_options
-      if options = repo_config.for(name)
-        merge(default_options, options)
+    def custom_config_file
+      merged_config = SCSSLint::Config.send(
+        :smart_merge,
+        default_options,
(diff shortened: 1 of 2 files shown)


app/models/style_guide/scss.rb
```diff
@@ -32,24 +32,31 @@ def build_runner
...
+        merge_with_default: false
+      )
     end
 
-    def custom_options
-      if options = repo_config.for(name)
-        merge(default_options, options)
+    def custom_config_file
+      merged_config = SCSSLint::Config.send(
+        :smart_merge,
```
thorncp: `smart_merge` is private?

gylaz: Yep, the joys of relying on third-party.

spec/models/style_guide/scss_spec.rb
```diff
@@ -64,6 +64,23 @@
           expect(violations_in(content, config)).to eq []
         end
       end
+
+      context "when exclude is provides as string" do
```
srt32: s/provides/provided

teoljungberg:
Looks good, merge away

PR #769, opened 2015-05-01
Extract build finalization to job

thorncp:
`BuildReportJob` should stand in place of the GitHub interactions in
`Review` in #727 with the plan of using it in the build workers
controller. Also change `Commenter` to accept a `Payload` and create its
own `PullRequest` instance.

Up next will be implementing the `BuildWorkers` controller
that will check for build completion, persist violations, and enqueue
this job.

https://trello.com/c/CdExHIVN

### DIFF
--- a/app/services/build_report.rb
+++ b/app/services/build_report.rb
@@ -1,0 +1,59 @@
+require "attr_extras"
+
+class BuildReport
+  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
+
+  def self.run(pull_request, build)
+    new(pull_request, build).run
+  end
+
+  pattr_initialize :pull_request, :build
+
+  def run
+    commenter.comment_on_violations(priority_violations)
+    create_success_status
+    track_subscribed_build_completed
+  end
+
+  private
+
+  def commenter
+    Commenter.new(pull_request)
+  end
+
+  def token
+    ENV.fetch("HOUND_GITHUB_TOKEN")
+  end
+
+  def violations
+    build.violations
+  end
+
+  def priority_violations
+    violations.take(MAX_COMMENTS)
+  end
+
+  def create_success_status
+    github.create_success_status(
+      pull_request.head_commit.repo_name,
+      pull_request.head_commit.sha,
+      I18n.t(:success_status, count: violation_count),
+    )
+  end
+
+  def violation_count
+    violations.map(&:messages_count).sum
+  end
+
+  def github
+    @github ||= GithubApi.new(token)
+  end
+
+  def track_subscribed_build_completed
+    if build.repo.subscription
+      user = build.repo.subscription.user
+      analytics = Analytics.new(user)
+      analytics.track_build_completed(build.repo)
+    end
+  end
+end
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -1,8 +1,6 @@
 class BuildRunner
   class ExpiredToken < StandardError; end
 
-  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
-
   pattr_initialize :payload
 
   def run
@@ -26,10 +24,8 @@ class BuildRunner
     track_subscribed_build_started
     create_pending_status
     upsert_owner
-    create_build
-    commenter.comment_on_violations(priority_violations)
-    create_success_status
-    track_subscribed_build_completed
+    build = create_build
+    BuildReport.run(pull_request, build)
   end
 
   def relevant_pull_request?
@@ -40,16 +36,8 @@ class BuildRunner
     @violations ||= style_checker.violations
   end
 
-  def priority_violations
-    violations.take(MAX_COMMENTS)
-  end
-
   def style_checker
     StyleChecker.new(pull_request)
-  end
-
-  def commenter
-    Commenter.new(pull_request)
   end
 
   def create_build
@@ -78,8 +66,10 @@ class BuildRunner
   end
 
   def repo
-    @repo ||= Repo.active.
-      find_and_update(payload.github_repo_id, payload.full_repo_name)
+    @repo ||= Repo.active.find_and_update(
+      payload.github_repo_id,
+      payload.full_repo_name,
+    )
   end
 
   def reset_token
@@ -97,27 +87,11 @@ class BuildRunner
     end
   end
 
-  def track_subscribed_build_completed
-    if repo.subscription
-      user = repo.subscription.user
-      analytics = Analytics.new(user)
-      analytics.track_build_completed(repo)
-    end
-  end
-
   def create_pending_status
     github.create_pending_status(
       payload.full_repo_name,
       payload.head_sha,
       I18n.t(:pending_status)
-    )
-  end
-
-  def create_success_status
-    github.create_success_status(
-      payload.full_repo_name,
-      payload.head_sha,
-      I18n.t(:success_status, count: violation_count)
     )
   end
 
@@ -145,9 +119,5 @@ class BuildRunner
 
   def configuration_url
     Rails.application.routes.url_helpers.configuration_url(host: ENV["HOST"])
-  end
-
-  def violation_count
-    violations.sum(&:messages_count)
   end
 end
--- a/spec/services/build_report_spec.rb
+++ b/spec/services/build_report_spec.rb
@@ -1,0 +1,97 @@
+require "rails_helper"
+
+describe BuildReport do
+  describe ".run" do
+    context "when build has violations" do
+      it "comments a maximum number of times" do
+        stub_const("::BuildReport::MAX_COMMENTS", 1)
+        commenter = stubbed_commenter(comment_on_violations: true)
+        build = create(:build, violations: build_list(:violation, 2))
+        stubbed_github_api
+        pull_request = stubbed_pull_request
+
+        BuildReport.run(pull_request, build)
+
+        expect(commenter).to have_received(:comment_on_violations).
+          with(build.violations.take(BuildReport::MAX_COMMENTS))
+      end
+
+      it "creates GitHub statuses" do
+        repo = create(:repo, full_github_name: "test/repo")
+        build = create(
+          :build,
+          commit_sha: "headsha",
+          repo: repo,
+          violations: [
+            build(:violation, messages: ["wrong", "bad"]),
+          ],
+        )
+        stubbed_commenter
+        github_api = stubbed_github_api
+        pull_request = stubbed_pull_request
+
+        BuildReport.run(pull_request, build)
+
+        expect(github_api).to have_received(:create_success_status).with(
+          "test/repo",
+          "headsha",
+          "2 violations found."
+        )
+      end
+    end
+
+    context "with subscribed private repo and opened pull request" do
+      it "tracks build events" do
+        repo = create(:repo, :active, github_id: 123, private: true)
+        build = create(:build, repo: repo)
+        create(:subscription, repo: repo)
+        stubbed_commenter
+        stubbed_github_api
+        pull_request = stubbed_pull_request
+
+        BuildReport.run(pull_request, build)
+
+        expect(analytics).to have_tracked("Build Completed").
+          for_user(repo.subscription.user).
+          with(properties: { name: repo.full_github_name, private: true })
+      end
+    end
+
+    def stubbed_commenter(options = {})
+      commenter = double(:commenter, options).as_null_object
+      allow(Commenter).to receive(:new).and_return(commenter)
+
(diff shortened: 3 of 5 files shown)


app/jobs/build_report_job.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def track_subscribed_build_completed
+    if build.repo.subscription
+      user = build.repo.subscription.user
+      analytics = Analytics.new(user)
+      analytics.track_build_completed(build.repo)
+    end
+  end
+end
```
gylaz: This job is pretty hefty on logic -- thoughts on pulling this out into a model/service?

app/services/build_runner.rb
```diff
@@ -8,14 +6,14 @@ def run
...
-        violations: violations,
+      build = repo.builds.create!(
         pull_request_number: payload.pull_request_number,
         commit_sha: payload.head_sha,
       )
-      commenter.comment_on_violations(priority_violations)
-      create_success_status
-      track_subscribed_build_completed
+
+      build.update!(violations: find_violations)
```
gylaz: Can this be combined with the `create!` call?

app/services/build_runner.rb
```diff
@@ -27,22 +25,14 @@ def relevant_pull_request?
     pull_request.opened? || pull_request.synchronize?
   end
 
-  def violations
+  def find_violations
```
gylaz: Any particular reason for this rename? It doesn't really reflect the nature of `find`.

app/jobs/build_report_job.rb
```diff
@@ -0,0 +1,70 @@
...
+      "pull_request" => {
+        "head" => {
+          "sha" => build.commit_sha,
+        },
+      },
+      "repository" => {
+        "full_name" => build.repo.full_github_name,
+      },
+    )
+  end
```
gylaz: If we ever need to pull more information from the original payload data, we'd have to modify it in two places. How about either passing `payload` to this job, or storing the payload on the build?

app/services/build_runner.rb
```diff
@@ -8,14 +6,14 @@ def run
...
         pull_request_number: payload.pull_request_number,
         commit_sha: payload.head_sha,
       )
-      commenter.comment_on_violations(priority_violations)
-      create_success_status
-      track_subscribed_build_completed
+
+      build.update!(violations: find_violations)
+
+      BuildReportJob.perform_later(build)
```
gylaz: `BuildRunner` already runs as a job. Is there a reason for spawning another job here?

app/jobs/build_report_job.rb
teoljungberg: It also follows our current style, having a job that just calls a service

app/services/build_runner.rb
teoljungberg: Fixed and pushed

app/services/build_runner.rb
teoljungberg: I agree, I renamed it and pushed it

app/services/build_runner.rb
teoljungberg: It can, done and pushed

app/jobs/build_report_job.rb
teoljungberg: Done and pushed

app/jobs/build_report_job.rb
teoljungberg: I'm unsure what's the best way of solving this. It kinda feels like we should pass the build (that has most of this information to the `Commenter`

thorncp: I like the idea of storing it on the build. When the workers are introduced, we won't have to pass the payload (or a subset of it) back and forth between Hound an Iron, since we can just pull it from the build. That should probably be a different PR though.

app/services/build_runner.rb
thorncp: I think this was done with the workers in mind, where the build worker controller would kick off the job when it gets a payload from Iron. I like it better as a service, and we can another job for the build worker case.

app/services/build_runner.rb
thorncp: I'm having a hard to remembering what our thought process was here. I think we were trying to indicate the severity of this method call, in that it goes through many levels of abstraction and does quite a lot of work. I don't think `find_violations` captures that, but I do think `violations` sort of trivializes what's happening.

app/services/commenter.rb
```diff
@@ -15,4 +15,8 @@ def comment_on_violations(violations)
   def commenting_policy
     @commenting_policy ||= CommentingPolicy.new(pull_request)
   end
+
+  def token
+    ENV["HOUND_GITHUB_TOKEN"]
+  end
```
gylaz: We don't need this anymore, right?

spec/services/build_report_spec.rb
```diff
@@ -0,0 +1,111 @@
+require "rails_helper"
+
+describe BuildReport, ".run" do
```
gylaz: We have a guideline against doing this:

> Use a single, top-level describe ClassName block.

spec/services/commenter_spec.rb
```diff
@@ -9,6 +9,7 @@
         it "comments on the violation" do
           pull_request = double(:pull_request, comment_on_violation: nil)
           violation = double(:violation)
+          payload = Payload.new({})
```
gylaz: Left over?

gylaz:
This is looking great! Couple of small things, then good to merge.

teoljungberg:
I rebased against master to get the changes made in #774 and rewrote the commit message. Thoughts @gylaz and @thorncp ?

app/services/build_report.rb
```diff
@@ -0,0 +1,55 @@
+require "attr_extras"
+
+class BuildReport
+  MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
+
+  static_facade :run, :pull_request, :build
```
gylaz: Ugh, I really hate this syntax. It's not straight forward to understand that `pull_request` and `build` are params of `run`. What do you think about using plain Ruby for `self.run` definition?

app/services/build_report.rb
```diff
@@ -0,0 +1,55 @@
...
+  def run
+    commenter.comment_on_violations(priority_violations)
+    create_success_status
+    track_subscribed_build_completed
+  end
+
+  private
+
+  def commenter
+    Commenter.new(pull_request)
```
gylaz: What is `pull_request` defined? This feels like too much magic.

app/services/build_report.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def token
+    ENV["HOUND_GITHUB_TOKEN"]
```
gylaz: `fetch`?

app/services/build_runner.rb
```diff
@@ -26,10 +24,8 @@ def review_pull_request
     track_subscribed_build_started
     create_pending_status
     upsert_owner
-    create_build
-    commenter.comment_on_violations(priority_violations)
-    create_success_status
-    track_subscribed_build_completed
+    build = create_build
+    BuildReport.run(pull_request, build)
```
gylaz: Thoughts on calling `create_build` inline?

app/services/build_report.rb
teoljungberg: It get's passed into the initializer from `static_facade`

app/services/build_report.rb
teoljungberg: I actually prefer it, but since we don't have it incorporated in the project I'm fine leaving it out

app/services/build_report.rb
teoljungberg: Good idea

spec/services/build_report_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        build = create(:build, violations: build_list(:violation, 2))
+        commenter = stubbed_commenter
+        stubbed_github_api
+        pull_request = stubbed_pull_request
+
+        BuildReport.run(pull_request, build)
+
+        expect(commenter).to have_received(:comment_on_violations).
+          with(build.violations.take(BuildReport::MAX_COMMENTS))
+      end
```
gylaz: Does this test duplicate what test on line 6 does? Can we remove that test?

spec/services/build_report_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        BuildReport.run(pull_request, build)
+
+        expect(commenter).to have_received(:comment_on_violations).
+          with(build.violations)
+      end
+
+      it "comments a maximum number of times" do
+        stub_const("::BuildReport::MAX_COMMENTS", 1)
+        commenter = double("Commenter", comment_on_violations: true)
```
gylaz: What do you think about modifying the `stubbed_commenter` function to take the changing variables? Similar to how we do that here: https://github.com/thoughtbot/hound/blob/master/spec/models/pull_request_spec.rb#L101-L108

spec/services/build_report_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        allow(Commenter).to receive(:new).and_return(commenter)
+        build = create(:build, violations: build_list(:violation, 2))
+        commenter = stubbed_commenter
```
gylaz: This doesn't look right.

spec/services/build_report_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      it "creates GitHub statuses" do
+        repo = create(:repo, full_github_name: "test/repo")
+        build = create(
+          :build,
+          commit_sha: "headsha",
+          repo: repo,
+          violations: [
+            build(:violation),
+            build(:violation, messages: ["wrong", "bad"]),
+          ],
```
gylaz: Is one violations enough to test that the status is created? Do we need any violations?

spec/services/build_runner_spec.rb
```diff
@@ -29,34 +29,14 @@
...
       stubbed_github_api
+      pull_request = stubbed_pull_request
+      stubbed_style_checker_with_violations
+      allow(BuildReport).to receive(:run)
 
       build_runner.run
 
-      expect(commenter).to have_received(:comment_on_violations).
-        with(violations.take(BuildRunner::MAX_COMMENTS))
+      expect(BuildReport).to have_received(:run).with(pull_request, Build.last)
```
gylaz: How about updating the doc string of this test?

app/services/build_report.rb
gylaz: So, `static_facade` does two things? Define a `self.run` method, as well as create private `attr_reader`s?

spec/services/build_report_spec.rb
teoljungberg: :+1:

spec/services/build_report_spec.rb
teoljungberg: How so?

spec/services/build_report_spec.rb
teoljungberg: Good catch

app/services/build_report.rb
teoljungberg: That's correct

app/services/build_runner.rb
teoljungberg: I would keep this as is - I don't think it's going to be clearer of inlining `create_build`

spec/services/build_report_spec.rb
gylaz: There's a `commenter` being created on line 20.

teoljungberg:
@gylaz all comments are fixed and pushed. Ready for a last pass

spec/services/build_runner_spec.rb
teoljungberg: Done in 9c52af4

spec/services/build_report_spec.rb
teoljungberg: Done in 7a2861e

spec/services/build_report_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    context "with subscribed private repo and opened pull request" do
+      it "tracks build events" do
+        repo = create(:repo, :active, github_id: 123, private: true)
+        build = create(:build, repo: repo)
+        create(:subscription, repo: repo)
+        stubbed_commenter
+        stubbed_pull_request
+        stubbed_github_api
+        pull_request = stubbed_pull_request
```
gylaz: Line 49 duplicates this.

gylaz:
One more note, then good to merge.

teoljungberg:
Thanks for reviewing @gylaz

PR #773, opened 2015-05-07
Consistent style of one GithubApiHelper method

JuanitoFatas:
Spotted while reading the code.


gylaz:
This is more lines of code, and it adds no value.

JuanitoFatas:
Got it. Thanks @gylaz!

PR #774, opened 2015-05-07
Remove expired token when running a build

gylaz:


app/services/build_runner.rb
```diff
@@ -1,28 +1,37 @@
...
-        commit_sha: payload.head_sha,
-      )
-      commenter.comment_on_violations(priority_violations)
-      create_success_status
-      track_subscribed_build_completed
+      review_pull_request
     end
   rescue RepoConfig::ParserError
     create_config_error_status
+  rescue Octokit::Unauthorized
```
thorncp: Rescuing config errors and authorization errors in the same place feels wrong. However, since this class is already up for a refactoring (with the changes for workers), *maybe* it's ok to leave this so we can deploy the fix, then include this in the upcoming tweaks.

teoljungberg: Agreed ^

app/services/build_runner.rb
```diff
@@ -1,28 +1,37 @@
 class BuildRunner
+  class ExpiredToken < StandardError; end
```
teoljungberg: How about using

```ruby
ExpiredToken = Class.new(StandardError)
```
instead?

app/services/build_runner.rb
```diff
@@ -1,28 +1,37 @@
...
+
   MAX_COMMENTS = ENV.fetch("MAX_COMMENTS").to_i
 
   pattr_initialize :payload
 
   def run
     if repo && relevant_pull_request?
-      track_subscribed_build_started
```
teoljungberg: Nice to finally extract this out, :+1:

app/services/build_runner.rb
```diff
@@ -1,28 +1,37 @@
...
-      commenter.comment_on_violations(priority_violations)
-      create_success_status
-      track_subscribed_build_completed
+      review_pull_request
     end
   rescue RepoConfig::ParserError
     create_config_error_status
+  rescue Octokit::Unauthorized
+    if users_with_token.any?
+      reset_token
```
teoljungberg: Should this be called `reset_tokens` since it's for multiple users?

app/services/build_runner.rb
```diff
@@ -61,6 +82,11 @@ def repo
       find_and_update(payload.github_repo_id, payload.full_repo_name)
   end
 
+  def reset_token
```
teoljungberg: Should this be called `reset_tokens` since it's for multiple users?

teoljungberg:
I like the overall concept, made a few small remarks on naming

app/services/build_runner.rb
gylaz: Is there any particular benefit to this? The way I have is how I've seen this generally done. I wouldn't break away from the norm, unless there's a benefit.

app/services/build_runner.rb
gylaz: It only resets the current token -- one token at a time (per job try).

gylaz:
@teoljungberg @thorncp @salbertson Ready for another look, especially the `reset_token` method.

spec/services/build_runner_spec.rb
```diff
@@ -225,6 +225,23 @@
     end
   end
 
+  context "with expired token" do
+    it "removes the token and falls back to Hound's token" do
```
thorncp: How does this test handle the "falls back to Hound's token" part?

gylaz: Good point. It doesn't. Should probably be renamed to `it "removes expired token" do`.

app/services/build_runner.rb
teoljungberg: It's simply a stylistic thing, if we're using the style you wrote in other places in hound; let's not change this

app/services/build_runner.rb
```diff
@@ -61,6 +82,14 @@ def repo
...
+    token_user = repo.users.detect { |user| user.token == token }
+    token_user[:token] = nil
```
teoljungberg: How about using update_attribute instead of #[] + #save ?

spec/services/build_runner_spec.rb
```diff
@@ -225,6 +225,23 @@
     end
   end
 
+  context "with expired token" do
+    it "removes the expired token" do
+      repo = create(:repo, :active)
+      user = create(:user, token: "123abc")
```
teoljungberg: How about using "expiredtoken" to clarify it's intent instead of "123abc"?

teoljungberg:
A few remarks, looks good to squash and merge

app/services/build_runner.rb
gylaz: `update_attribute` won't work because it hits the setter, but `update_column` does the trick.

spec/services/build_runner_spec.rb
gylaz: Good idea.

PR #779, opened 2015-05-12
Add missing EXEMPT_ORGS env variable

JensDebergh:
Adds missing EXEMPT_ORGS variable responsible for exempting users or organisations from paying for repo reviews.

### DIFF
--- a/.sample.env
+++ b/.sample.env
@@ -1,5 +1,6 @@
 CHANGED_FILES_THRESHOLD=300
 ENABLE_HTTPS=no
+EXEMPT_ORGS=organisation_name,user_name
 GITHUB_CLIENT_ID=github_client_id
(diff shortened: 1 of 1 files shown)


.sample.env
```diff
@@ -1,5 +1,6 @@
 CHANGED_FILES_THRESHOLD=300
 ENABLE_HTTPS=no
+EXEMPT_ORGS=organisation_name, user_name
```
JuanitoFatas: I think there is no space before and after the comma. https://github.com/thoughtbot/hound/blob/master/app/models/repo.rb#L59

JensDebergh: You are right! Thanks for the heads up. I've pushed the change.

salbertson:
@JensDebergh, merged. Thanks!

PR #780, opened 2015-05-13
Return correct owner ID and login from #build_data

salbertson:


app/models/payload.rb
```diff
@@ -63,8 +63,8 @@ def build_data
         "id" => github_repo_id,
         "full_name" => full_repo_name,
         "owner" => {
-          "id" => repository_owner_name,
-          "login" => repository_owner_id,
+          "id" => repository_owner_id,
+          "login" => repository_owner_name,
```
gylaz: Haha, what? How did that happen?

gylaz:
:+1:

app/models/payload.rb
salbertson: https://github.com/thoughtbot/hound/commit/937ad55f94e4c46beeab864c4ea44cabb82f3437#diff-ceee0f960ebd4118609a893634b410e0

PR #781, opened 2015-05-15
Log HTTP requests to GitHub

thorncp:
* Log the URL of all HTTP requests made via Octokit. Log lines are
  tagged with `[BIG_BROTHER]` for easy searching.
* Add hard dependency on Faraday since we have library code relying on
  it.

https://github.com/octokit/octokit.rb/tree/f5f9f2fab804cd2231bd8a33c3a3234504782243#debugging

<hr />

Example from refreshing repo list locally:

```text
[ActiveJob] [RepoSynchronizationJob] [b2582ad4-6655-4134-87a9-751f78bbca36] [BIG_BROTHER] https://api.github.com/user/repos?per_page=100
[ActiveJob] [RepoSynchronizationJob] [b2582ad4-6655-4134-87a9-751f78bbca36] [BIG_BROTHER] https://api.github.com/user/orgs
[ActiveJob] [RepoSynchronizationJob] [b2582ad4-6655-4134-87a9-751f78bbca36] [BIG_BROTHER] https://api.github.com/orgs/thoughtbot/repos?per_page=100
```


lib/big_brother.rb
```diff
@@ -0,0 +1,7 @@
+require "faraday"
+
+class BigBrother < Faraday::Response::Middleware
```
gylaz: Ha, love this name.

gylaz:
Would it be helpful, or possible, to log the user that is making the request? So, we can judge houndci's token usage vs personal tokens.

thorncp:
We have access to the request headers, so we could do a simple comparison of the token to Hound's token from ENV and include `[hound]` or `[user]` (or whatever) in the log entry with it.

thorncp:
Added those flags in 93309219d9cd5b625c379db72d770a8eec266f7d

gylaz:
:+1:

PR #782, opened 2015-05-18
Adjust the max SelectorDepth in scss lint to 4

christoomey:
Our style guide currently includes [the following with regard to selector depth][]:

> Avoid nesting more than 4 selectors deep.
> Don't nest more than 6 selectors deep.

SCSSlint has two seemingly applicable settings, [NestingDepth][] and [SelectorDepth][]. The former is set to 4 in our hound configuration, but the latter is set to 2. My thinking is that this setting (`SelectorDepth`) should also be 4. This is well out of my area of expertise, but I have run into it enough on PRs and would rather not just ignore.

/cc @thoughtbot/design 

[the following with regard to selector depth]: https://github.com/thoughtbot/guides/tree/c3394adfa486ce37bf205a85e8dc1c5dffd6c6da/style/sass#selectors
[NestingDepth]: https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#nestingdepth
[SelectorDepth]: https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#selectordepth

### DIFF
--- a/config/style_guides/thoughtbot/scss.yml
+++ b/config/style_guides/thoughtbot/scss.yml
@@ -86,7 +86,7 @@ linters:
     severity: warning
   SelectorDepth:
     enabled: true
-    max_depth: 2
+    max_depth: 4
     severity: warning
   SelectorFormat:
     enabled: true


tysongach:
Although we don’t have anything specific in our guides about selector depth, I think we should keep this at 2.

These two linters are similar, but very different in purpose and as the SCSS-Lint docs mention, it’s generally bad to couple your CSS so closely to the HTML structure. I think having Hound bark at this would be a good reminder and an author can use their best judgement to not listen and explain why.

christoomey:
@tysongach thanks for the update. Does your summary map to our guides as they stand? The logic you laid out makes sense, although I've personally felt this (SelectorDepth 2) was overly restrictive and have as such ignored it with a comment pointing at the guides. The styling in question for this case was:

``` scss
.questions-show {
  .confidence form {
    display: inline;
  }
}
```

To your comment, would 3 (the scsslint default) be a good compromise?

Just to be clear, I'm fine with keeping this as is, but wanted to open this PR so I could at least gain clarity on our current stance re: Style guide & Hound.

tysongach:
Totally; clarity is always key.

Currently, the guides don’t cover selector depth at all, only Sass nesting. Nesting ultimately compiles and leads to selector depth, but nesting can also help build complex selectors. Some authors like it use Sass’s parent selector to do just that:

```scss
.component {
  &__child {
    font-size: modular-scale(2);
  }
}
```

This would compile to:

```css
.component__child {
    font-size: modular-scale(2);
}
```

So here, the Sass is nested two levels deep, but the CSS output is only a selector depth of one.

In your example above, I would try as much as possible to put a `class` on the `form` or at a minimum try to un-nest it from `.questions-show` (which seems like a `body` class?). Using components/patterns and modifier classes on those is a more systematic approach to accomplishing that:

```scss
.confidence-form {
  // form component styles

  &--special-version {
    color: blue;
  }
}

Hope that clarifies it a little?

wardpenney:
I would rather the depth stayed at 2. I think more depth is a detrimental way for this practice to trend.

In the example above, the `confidence` class is a victim of the illusion of encapsulation with Sass. The ancestral nesting prevents `confidence` from causing problems elsewhere in the codebase. However it does not shield its effects from:

* any deeper ancestral instances of `confidence`, or
* uses of `confidence` elsewhere in the codebase

Also, suppose we need to modify the pattern with a patten class for some reason (for example, a `display: inline-block` rule) it would not win in the Cascade against selector depth in the first declaration.

christoomey:
This all makes sense to me. I'll close this now and leave it to someone on @thoughbot/design to update the style guide (if that is in fact needed?). Thanks for the clarifications!

PR #785, opened 2015-05-26
Remove unreachable repo from user's repos

gylaz:
When a repo cannot be reached with a user's token,
causing a 404 Not Found error,
that means that user has been barred access to it.

By removing the repo from user's associated repos,
the job will be retried with a different user's token,
or fallback to Hound's.

This should fix numerous 404 errors that are causing jobs to fail.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -1,5 +1,6 @@
 class BuildRunner
-  class ExpiredToken < StandardError; end
+  HOUND_TOKEN = ENV.fetch("HOUND_GITHUB_TOKEN")
+  ExpiredToken = Class.new(StandardError)
 
   pattr_initialize :payload
 
@@ -16,6 +17,11 @@ class BuildRunner
     else
       raise
     end
+  rescue Octokit::NotFound
+    if token != HOUND_TOKEN
+      remove_repo_from_user
+    end
+    raise
   end
 
   private
@@ -53,7 +59,7 @@ class BuildRunner
   end
 
   def token
-    @token ||= user_token || ENV["HOUND_GITHUB_TOKEN"]
+    @token ||= user_token || HOUND_TOKEN
   end
 
   def user_token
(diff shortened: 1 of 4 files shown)


thorncp:
Do we want to do this when we see `Octokit::Unauthorized` as well? I'm unclear to the difference between the two errors, but it *seems* we want to do the same thing in both cases.

gylaz:
I believe those are two different situations.

`Octokit::Unauthorized` means that the user's token has been revoked/expired. Thus, we want to blow away the token.

`Octokit::NotFound` means the user no longer has access to the repo, but his token might still be good.

thorncp:
Ah, gotcha. :+1:

app/services/build_runner.rb
```diff
@@ -113,6 +122,10 @@ def upsert_owner
     repo.update(owner: owner)
   end
 
+  def current_token_user
```
teoljungberg: how about `current_user_token`?

gylaz: That name implies the method returns a `token`, however the method returns a `user found from the current token`.

app/services/build_runner.rb
```diff
@@ -16,6 +17,11 @@ def run
     else
       raise
     end
+  rescue Octokit::NotFound
+    if token != HOUND_TOKEN
+      remove_repo_from_user
+    end
+    raise
```
salbertson: There's a lot of exception handling here along with code needed to recover from exceptions. What do you think about moving some of the code out of `run` and closer to the source of the exception?

Another thought I had was to raise custom exceptions that have a `recover` method or something that would contain the code inside these `rescue` blocks to help clean things up a bit.

app/services/build_runner.rb
```diff
@@ -73,9 +79,12 @@ def repo
   end
 
   def reset_token
-    token_user = repo.users.detect { |user| user.token == token }
-    token_user.update_columns(token: nil)
+    current_token_user.update_columns(token: nil)
+    @token = nil
```
salbertson: What do you think about avoiding this `nil` assignment by removing the memoization?

app/services/build_runner.rb
gylaz: Talked in person, and decided to address this later.

PR #786, opened 2015-05-27
Store FileReview records per reviewed file 

gylaz:
Violations now belong to FileReview record,
and builds have many file reviews.

File reviews have a completed status which can be used for jobs.


app/models/style_guide/coffee_script.rb
```diff
@@ -4,18 +4,18 @@ class CoffeeScript < Base
...
-
-        Violation.new(
-          filename: file.filename,
-          line: line,
-          patch_position: line.patch_position,
-          line_number: violation["lineNumber"],
-          messages: [violation["message"]]
-        )
+    def file_review(file)
+      FileReview.new(filename: file.filename).tap do |file_review|
```
thorncp: Don't need `tap` here, `new` will accept the block.

app/models/style_guide/coffee_script.rb
```diff
@@ -4,18 +4,18 @@ class CoffeeScript < Base
...
+        content = content_for_file(file)
+        lint(content).map do |violation|
```
thorncp: `each` instead of `map` since we're saving the return value?

app/models/violations.rb
```diff
@@ -1,33 +0,0 @@
-# Container of unique violations
-class Violations
```
thorncp: :scissors: :+1:

db/migrate/20150526214750_associate_violations_with_file_reviews.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  private
+
+  def associate_violations_with_file_reviews
+    Violation.find_each do |violation|
+      file_review = FileReview.find_or_create_by!(
+        build_id: violation.build_id,
+        filename: violation.filename,
+        completed_at: Time.current
```
thorncp: Seems like including this in the query will end up creating a file review for every violation.

Should we do something like this?

```ruby
attrs = { build_id: violation.build_id, filename: violation.filename }
file_review = FileReview.find_or_create_by!(attrs) do |file_review|
  file_review.completed_at = Time.current
end
```

app/models/style_guide/coffee_script.rb
teoljungberg: I think it could be good to have `map` over `each` for debugging purposes.

app/models/style_guide/ruby.rb
```diff
@@ -3,20 +3,16 @@ module StyleGuide
...
-          line = file.line_at(violation.line)
-
-          Violation.new(
-            filename: file.filename,
-            patch_position: line.patch_position,
-            line: line,
-            line_number: violation.line,
-            messages: [violation.message]
-          )
+        FileReview.new(filename: file.filename).tap do |file_review|
```
teoljungberg: Can this be extracted into a private method for the style guide classes? `conduct_file_review` or something along those lines

spec/models/file_review_spec.rb
```diff
@@ -5,8 +5,47 @@
...
-    it { should validate_presence_of :build }
+  describe "#build_violation" do
+    context "when line has been changed" do
+      it "builds a violations" do
+        line = double("Line", changed?: true, patch_position: 121)
+        file_review = FileReview.new
+        expected_violation = Violation.new(
+          patch_position: 121,
+          line_number: 1,
+          messages: ["hello"]
```
teoljungberg: How about a trailing comma?

spec/models/style_checker_spec.rb
```diff
@@ -1,17 +1,18 @@
 require "rails_helper"
 
-describe StyleChecker, "#violations" do
-  it "returns a collection of computed violations" do
+describe StyleChecker, "#file_reviews" do
```
teoljungberg: How about having the `describe "#file_reviews"` on a separate line?

spec/models/style_guide/coffee_script_spec.rb
```diff
@@ -31,26 +31,30 @@
     end
   end
 
-  describe "#violations_in_file" do
+  describe "#file_review" do
+    it "returns completed file review" do
+      repo_config = double("RepoConfig", enabled_for?: true, for: {})
+      style_guide = StyleGuide::CoffeeScript.new(repo_config, "bob")
+      file = build_file("foo")
```
teoljungberg: How about having the filename being a keyword argument for easier readability?

spec/models/style_guide/coffee_script_spec.rb
```diff
@@ -31,26 +31,30 @@
     end
   end
 
-  describe "#violations_in_file" do
+  describe "#file_review" do
+    it "returns completed file review" do
+      repo_config = double("RepoConfig", enabled_for?: true, for: {})
```
teoljungberg: We're creating a lot of repo configs in this test, how about if we add a `build_repo_config` factory?

spec/models/style_guide/coffee_script_spec.rb
teoljungberg: Oh, it seems like this is the content

spec/models/style_guide/coffee_script_spec.rb
```diff
@@ -179,12 +170,21 @@ def violations_in(content, repository_owner_name: "ralph")
...
-    def build_file(content)
-      line = double("Line", content: "blah", number: 1, patch_position: 2)
-      double(:file, content: content, filename: "test.coffee", line_at: line)
+    def build_file(content, filename = "test.coffee")
+      line = double(
+        "Line",
+        changed?: true,
+        content: "blah",
+        number: 1,
+        patch_position: 2
```
teoljungberg: How about adding a trailing comma?

spec/models/style_guide/coffee_script_spec.rb
```diff
@@ -179,12 +170,21 @@ def violations_in(content, repository_owner_name: "ralph")
...
+      )
+      double(:file, content: content, filename: filename, line_at: line)
```
teoljungberg: Can we use strings as the name of the double for consistency?

teoljungberg: Can this be split up into several lines for consistency with the `Line` double?

spec/models/style_guide/java_script_spec.rb
```diff
@@ -3,26 +3,32 @@
 describe StyleGuide::JavaScript do
   include ConfigurationHelper
 
-  describe "#violations_in_file" do
+  describe "#file_review" do
+    it "returns completed file review" do
+      repo_config = double("RepoConfig", enabled_for?: true, for: {})
```
teoljungberg: Could we do the same with the repo config factory method here?

spec/models/style_guide/ruby_spec.rb
```diff
@@ -1,8 +1,17 @@
 require "rails_helper"
 
-describe StyleGuide::Ruby, "#violations_in_file" do
+describe StyleGuide::Ruby, "#file_review" do
```
teoljungberg: Can we move the `describe "#file_review"` into a line of it's own?

spec/models/style_guide/ruby_spec.rb
```diff
@@ -1,8 +1,17 @@
...
   include ConfigurationHelper
 
+  it "returns completed file review" do
```
teoljungberg: `a completed`?

spec/models/style_guide/scss_spec.rb
```diff
@@ -1,7 +1,15 @@
 require "rails_helper"
 
 describe StyleGuide::Scss do
-  describe "#violations_in_file" do
+  describe "#file_review" do
+    it "returns completed file review" do
```
teoljungberg: `a completed`?

spec/models/violation_spec.rb
```diff
@@ -1,12 +1,7 @@
 require "rails_helper"
 
 describe Violation do
-  it { should belong_to(:build) }
-
-  describe "validations" do
-    it { should validate_presence_of(:build) }
-    it { should validate_presence_of(:filename) }
-  end
+  it { should belong_to(:file_review) }
```
teoljungberg: How about adding so we validate the presence of the file_review aswell?

spec/services/build_report_spec.rb
```diff
@@ -6,36 +6,31 @@
       it "comments a maximum number of times" do
         stub_const("::BuildReport::MAX_COMMENTS", 1)
         commenter = stubbed_commenter(comment_on_violations: true)
-        build = create(:build, violations: build_list(:violation, 2))
+        file_review = create(
+          :file_review,
+          violations: build_list(:violation, 2)
```
teoljungberg: Could we add a trailing comma here?

app/models/style_guide/coffee_script.rb
gylaz: Huh, really? TIL. Is that a AR thing?

app/models/style_guide/coffee_script.rb
gylaz: I agree that `each` is probably what should be used. `map` communicates a certain intent that we're not using here. IMO, "for debugging purposes" is YAGNI.

db/migrate/20150526214750_associate_violations_with_file_reviews.rb
gylaz: Good catch!

app/models/style_guide/coffee_script.rb
thorncp: [Yup](http://api.rubyonrails.org/classes/ActiveRecord/Base.html#class-ActiveRecord::Base-label-Creation)

app/models/style_guide/coffee_script.rb
thorncp: It would just be an array of `true` values anyway.

spec/models/violation_spec.rb
gylaz: It's already validated at the DB layer. There is no value of adding a Rails validation because users aren't directly creating validations.

spec/models/style_guide/ruby_spec.rb
gylaz: I think, I'm going to hold off on fixing this in this PR, as it will cause a lot more diff noise.

spec/models/style_checker_spec.rb
gylaz: This would cause everything to shift, adding more diff noise. I agree in general, but not in this PR :smile:

spec/models/style_guide/coffee_script_spec.rb
gylaz: If it fits on one line, I prefer it to keep it like that. I hear the consistency argument, but then this method length would grow -- we could go about extracting that, but this is just test helpers.

spec/models/style_checker_spec.rb
teoljungberg: You can remove the diff noise by adding `?w=1` in the url. But maybe it's a better fit for another PR, but my thinking if it's moved forward to another PR it will never be dealt with

app/models/style_guide/ruby.rb
```diff
@@ -3,26 +3,26 @@ module StyleGuide
   class Ruby < Base
     DEFAULT_CONFIG_FILENAME = "ruby.yml"
 
-    def violations_in_file(file)
+    def file_review(file)
       if config.file_to_exclude?(file.filename)
-        []
+        ExcludedFileReview.new
```
salbertson: Why not return a `FileReview` object without violations?

app/models/file_review.rb
```diff
@@ -1,7 +1,18 @@
 class FileReview < ActiveRecord::Base
   belongs_to :build
+  has_many :violations, dependent: :destroy
```
salbertson: When would file reviews be destroyed?

app/models/style_guide/coffee_script.rb
```diff
@@ -4,18 +4,18 @@ class CoffeeScript < Base
...
-        line = file.line_at(violation["lineNumber"])
-
-        Violation.new(
-          filename: file.filename,
-          line: line,
-          patch_position: line.patch_position,
-          line_number: violation["lineNumber"],
-          messages: [violation["message"]]
-        )
+    def file_review(file)
```
salbertson: Ha, busted! Is this check useful?

app/models/style_guide/unsupported.rb
```diff
@@ -1,8 +1,8 @@
 # Returns empty set of violations.
 module StyleGuide
   class Unsupported < Base
-    def violations_in_file(_)
-      []
+    def file_review(_)
+      ExcludedFileReview.new
```
salbertson: Same here, could we use a `FileReview` object that doesn't have any violations?

db/migrate/20150526214750_associate_violations_with_file_reviews.rb
```diff
@@ -0,0 +1,60 @@
...
+  private
+
+  def associate_violations_with_file_reviews
+    Violation.find_each do |violation|
+      attributes = {
+        build_id: violation.build_id,
+        filename: violation.filename,
+      }
+
+      file_review = FileReview.find_or_create_by!(attributes) do |review|
```
salbertson: Are there `FileReview` records to find?

spec/factories.rb
```diff
@@ -4,18 +4,16 @@
...
   factory :file_review do
-    build
-
     trait :completed do
       completed_at Time.zone.now
     end
+
+    build
+
+    filename "the_thing.rb"
```
salbertson: Should we use a unique string for filename?

app/models/style_checker.rb
```diff
@@ -7,20 +7,16 @@ def initialize(pull_request)
     @style_guides = {}
   end
 
-  def violations
-    @violations ||= Violations.new.push(*violations_in_checked_files).to_a
```
salbertson: Is `Violations` no longer used?

app/models/file_review.rb
```diff
@@ -1,7 +1,18 @@
...
 
-  validates :build, presence: :true
+  def build_violation(line, line_number, message)
```
salbertson: Can we send `line` and get `line_number` from that object instead of passing two arguments?

app/models/style_checker.rb
```diff
@@ -7,20 +7,16 @@ def initialize(pull_request)
...
+  def file_reviews
+    files_to_check.map do |file|
```
teoljungberg: How about using `flat_map` to avoid `nil` values?

teoljungberg:
@gylaz I weighed in and updated the PR with the newly merged HAML support

app/models/style_guide/coffee_script.rb
gylaz: It hasn't been too bad yet, and causes to extract smaller methods. However, the comment will almost never collapse, since this line isn't likely to change.

app/models/style_guide/scss.rb
```diff
@@ -2,31 +2,36 @@ module StyleGuide
...
-        []
+        ExcludedFileReview.new
       else
+        perform_file_review(file)
+      end
+    end
+
+    private
+
+    def perform_file_review(file)
```
gylaz: I'm not sure this one is helpful at all. I think it's best humans review and judge complexity.

app/models/style_guide/ruby.rb
gylaz: Great idea.

app/models/file_review.rb
gylaz: Likely never, but they could be manually. Are you opposed to having the `dependent: :destroy` here?

db/migrate/20150526214750_associate_violations_with_file_reviews.rb
gylaz: No. We'll create one for the first violation in a file, then re-use that record for the other violations.

spec/factories.rb
gylaz: The same filename could be reviewed multiple times. Is there a particular benefit to have it be unique?

app/models/style_checker.rb
gylaz: `flat_map` doesn't get rid of `nil`s. Are you thinking of `compact`?

There should never be a `nil` returned from `file_review`. Am I misunderstanding?

app/models/file_review.rb
```diff
@@ -10,4 +21,11 @@ def completed?
   def running?
     !completed?
   end
+
+  private
+
+  def find_or_build_violation(line)
+    violations.detect { |violation| violation.line_number == line.number } ||
+      violations.build(line_number: line.number)
```
thorncp: Could this be `violations.find_or_initialize_by(line_number: line.number)`?

app/models/style_guide/coffee_script.rb
thorncp: Yeah, I think the comment not collapsing is the biggest drawback. I like the check itself.

thorncp:
LGTM :+1:  I like how this cleans up the style guide classes.

app/models/file_review.rb
gylaz: Good idea, but unfortunately, I don't think so. The reason this is using `detect` is because the violations aren't in the database -- they are being built in memory as the file is being reviewed. I've added a test in 9c5f3ed to ensure this.

thorncp: Ah, yeah good point :+1:

app/models/style_checker.rb
teoljungberg: You're right, thanks.

app/models/style_guide/coffee_script.rb
salbertson: Ah, didn't think about collapsing, that's a bummer.

app/models/violations.rb
salbertson: Noice

app/models/style_guide/scss.rb
salbertson: Yea, go ahead and disable if it isn't useful.

app/models/file_review.rb
salbertson: Yea, it suggests that deleting could happen in the app. I think if they are deleted manually the person should be deleting associated records, but I'm not convinced violations should ever be deleted, even manually.

db/migrate/20150526214750_associate_violations_with_file_reviews.rb
salbertson: Ok, so should we be creating instead of finding or creating?

spec/factories.rb
salbertson: No particular benefit, that makes sense.

app/models/style_guide/ruby.rb
salbertson: Awesome, I like it.

app/models/file_review.rb
gylaz: FWIW, the foreign_key on the db will ensure cascade deleting, and I would keep that for data integrity. I'll remove it from the model.

db/migrate/20150526214750_associate_violations_with_file_reviews.rb
gylaz: Perhaps I don't follow your logic, but here's how I'm thinking about this:

* Get the first violation
* Create a file review record for it
* Get the second violation which happens to be for the same file and build as the first one
* Find the existing file review record and associate the two

gylaz: Basically, we don't want to create a new `file_review` record for each existing violation.

salbertson: Ah, that makes sense.

salbertson:
Looks good to me. :+1:

thorncp:
Updated the `up` migration to use SQL in d3d3168de95ab0d1f833ed151e05c52584e77ca2. Definitely needs :eyes: on it to make sure the data is correct.

I pulled the production database locally and ran this migration. After an hour and fifteen minutes I killed the killed the process. With SQL version it executes in about three minutes.

PR #787, opened 2015-06-01
Disable Raven in test environment

teoljungberg:


gylaz:
Is the benefit that the exceptions now get displayed in the console?

teoljungberg:
The main reason for me was to remove the logging to STDOUT, which made running tests with`vim-dispatch` not as easy

croaky:
How about moving Raven to only staging and production environments? e80491e

thorncp:
LGTM

PR #788, opened 2015-06-01
Require Bundler >= 1.10

teoljungberg:
As of Bundler 1.10, Bundler now includes an additional section in
Gemfile.lock, "BUNDLED WITH", that tells it which version was used to
run `bundle install`:

<bundler/bundler@a264e6b>

Unfortunately, on versions prior to 1.10, this section is missing, so
Gemfile.lock will change every time you run `bundle` until you upgrade
to Bundler 1.10.

For this reason, bump the dependency on Bundler to force people to
upgrade.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@ source "https://rubygems.org"
 
 ruby "2.2.2"
 
+gem "bundler", "~> 1.10"
+
 gem "active_model_serializers", "0.8.3"
 gem "analytics-ruby", "~> 2.0.0", require: "segment/analytics"
 gem "angularjs-rails"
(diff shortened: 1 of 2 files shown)


JuanitoFatas:
I remembered bundler author said specify bundler version in Gemfile does not work :sweat: :

> There is no such thing as a "required bundler version", yet. Right now (with the awkward exception of the ruby command), all Gemfiles and Gemfile.locks are compatible with all versions of Bundler from 1.0.0 up.
> https://github.com/bundler/bundler/issues/3327#issuecomment-70007750
> (Even if there is a Bundler version in the Gemfile or in the Gemfile.lock, it is ignored by Bundler internally.)
> https://github.com/bundler/bundler/issues/3327#issuecomment-70007848

teoljungberg:
@JuanitoFatas I think you have a point there, I remembered something similar but I was unsure. And after trying it out it seems to have zero merit.

JuanitoFatas:
:+1:

PR #790, opened 2015-06-01
Normalize Segment approach

croaky:
Use same styles and techniques as our other growth apps:

* Extract `identify_hash`, `intercom_hash`, and `campaign_hash`.
* Rename partial `_analytics`.
* JavaScript escape the page title in `page()` event names.
* Pass `campaign_hash` as properties into `page()` events.
* Use `SEGMENT_KEY` environment variable name.


gylaz:
:+1:

PR #802, opened 2015-06-12
Fetch all organizations using pagination.

arunagw:
Right now if there are more than 30 organizations in an account only first 30 get synced

This commit fixes that bug in octokit gem.
https://github.com/octokit/octokit.rb/commit/c15fa3464f55dd82bc5a8223d8c51fe8930b3e84


teoljungberg:
LGTM

salbertson:
We could avoid this by using the preview version of the API.

https://developer.github.com/v3/repos/#list-your-repositories

With the API changes all we need to do is request a user's repos, which will include organization owned repos. We won't have to fetch orgs just to fetch repos.

teoljungberg:
That sounds loke a good idea, Scott. Do you mind adding a trello card
for it?

PR #803, opened 2015-06-12
Disable all HAML rules for everyone except thoughtbot

salbertson:
We would like to slowly roll out HAML support.


croaky:
Looks like there might be some non-bubbles, Haml-related CI failures. Otherwise looks good.

croaky:
Merge it!

PR #804, opened 2015-06-12
Use ENV var instead of fixed value in repo_list_spec

jakecraige:
Having this value hard coded led to hard to debug failures
when your local `HOUND_GITHUB_USERNAME` was not "houndci".
This was because the `WebMock` stubs were stubbing
with "houndci" and the app makes requests
based on the `ENV` var.

### DIFF
--- a/spec/features/repo_list_spec.rb
+++ b/spec/features/repo_list_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 feature "Repo list", js: true do
-  let(:username) { 'houndci' }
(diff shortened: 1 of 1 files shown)


gylaz:
:+1:

croaky:
Looks like Travis is okay without this?

jakecraige:
@croaky In travis we first [copy the sample env](https://github.com/thoughtbot/hound/blob/master/.travis.yml#L15) which has the "correct" env var present [here](https://github.com/thoughtbot/hound/blob/master/.sample.env#L8). This came up on my machine because I had it set to something else in my local `.env`

thorncp:
LGTM

PR #805, opened 2015-06-12
Add API endpoint for updating account

malandrina:
* Support updating an account's billable email address
* Introduce `PATCH /account` endpoint
* Add `PaymentGatewayCustomer#update_email`
* Add `AccountPage#billable_email` and returns billable email in `GET
  /account` response

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -1,11 +1,39 @@
 class AccountsController < ApplicationController
   def show
-    @account_page = AccountPage.new(find_subscribed_repos)
+    @account_page = AccountPage.new(
+      repos: find_subscribed_repos,
+      billable_email: current_user.billable_email
+    )
+  end
+
+  def update
+    customer = PaymentGatewayCustomer.new(current_user)
+
+    respond_to do |format|
+      format.json do
+        if customer.update_email(new_billable_email)
+          render json: updated_account_page
+        else
+          head :bad_gateway
+        end
+      end
+    end
   end
 
   private
 
+  def new_billable_email
+    params.require(:billable_email)
+  end
+
   def find_subscribed_repos
     current_user.subscribed_repos.order(:full_github_name)
+  end
+
+  def updated_account_page
+    AccountPage.new(
+      repos: find_subscribed_repos,
+      billable_email: new_billable_email
+    )
   end
 end
--- a/spec/controllers/accounts_spec.rb
+++ b/spec/controllers/accounts_spec.rb
@@ -1,0 +1,49 @@
+require "rails_helper"
+
+describe AccountsController do
+  context "updating billable email" do
+    context "when email is not provided" do
+      it "raises" do
+        user = create(:user, stripe_customer_id: "customer-id")
+        stub_sign_in(user)
+
+        expect do
+          patch :update, billable_email: "", format: :json
+        end.to raise_error(ActionController::ParameterMissing)
+      end
+    end
+
+    context "when email is provided" do
+      context "on success" do
+        it "returns updated account" do
+          new_email = "new-email@example.com"
+          customer_id = "customer-id"
+          user = create(:user, stripe_customer_id: customer_id)
+          stub_sign_in(user)
+          stub_customer_find_request(customer_id)
+          stub_customer_update_request(email: new_email)
+
+          patch :update, billable_email: new_email, format: :json
+
+          response_body = JSON.parse(response.body)
+          expect(response_body["billable_email"]).to eq new_email
+        end
+      end
+    end
+
+    context "on failure" do
+      it "returns 502 Bad Gateway" do
+        new_email = "new-email@example.com"
+        customer_id = "customer-id"
+        user = create(:user, stripe_customer_id: customer_id)
+        stub_sign_in(user)
+        stub_customer_find_request(customer_id)
+        stub_failed_customer_update_request(email: new_email)
+
+        patch :update, billable_email: new_email, format: :json
+
+        expect(response).to have_http_status(:bad_gateway)
+      end
+    end
+  end
+end
--- a/app/models/payment_gateway_customer.rb
+++ b/app/models/payment_gateway_customer.rb
@@ -24,6 +24,14 @@ class PaymentGatewayCustomer
     customer.save
   end
 
+  def update_email(email)
+    customer.email = email
+    customer.save
+  rescue Stripe::APIError => e
+    Raven.capture_exception(e)
+    false
+  end
+
   private
 
   def default_card
--- a/app/presenters/account_page.rb
+++ b/app/presenters/account_page.rb
@@ -1,9 +1,10 @@
 class AccountPage
   include ActionView::Helpers::NumberHelper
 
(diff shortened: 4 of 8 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -1,11 +1,41 @@
...
+
+  def new_billable_email
+    params[:billable_email]
+  end
+
   def find_subscribed_repos
     current_user.subscribed_repos.order(:full_github_name)
   end
+
+  def user_has_paid_subscription?
```
graysonwright: does it make sense to move this method into the `User` class?

app/controllers/accounts_controller.rb
```diff
@@ -1,11 +1,41 @@
...
+      )
+    else
+      head 503
+    end
   end
 
   private
 
+  def billable_email
+    if user_has_paid_subscription?
```
graysonwright: can this whole method be moved onto Users, as well?

app/controllers/accounts_controller.rb
```diff
@@ -1,11 +1,45 @@
...
+
+    respond_to do |format|
+      format.json do
+        if customer.update_email(new_billable_email)
+          render json: AccountPage.new(
+            repos: find_subscribed_repos,
+            billable_email: new_billable_email
+          )
+        else
+          head 503
```
graysonwright: Why 503 (service unavailable)?

Also, how do you feel about using the Rails symbol for the code (`:service_unavailable`)?
http://teamco-anthill.blogspot.com/2010/04/rails-http-status-code-to-symbol.html

app/controllers/accounts_controller.rb
malandrina: Good point. I actually think I can get rid of `AccountsController#billable_email` and `AccountsController#user_has_paid_subscription?` entirely because `PaymentGatewayCustomer` handles the case where there's no customer for a given user.

spec/controllers/accounts_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        new_email = "new-email@example.com"
+        customer_id = "customer-id"
+        user = create(:user, stripe_customer_id: customer_id)
+        stub_sign_in(user)
+        stub_customer_find_request(customer_id)
+        stub_failed_customer_update_request(email: new_email)
+
+        patch :update, billable_email: new_email, format: :json
+
+        expect(response.status).to eq 503
```
graysonwright: I think `).to be_service_unavailable` might work?

app/controllers/accounts_controller.rb
graysonwright: nice! :hammer::exclamation::nose::rice_cracker:

app/controllers/accounts_controller.rb
malandrina: Oops, I confused `503` with `502 Bad Gateway` (which is what I really want).

app/controllers/accounts_controller.rb
```diff
@@ -1,11 +1,45 @@
...
+  def update
+    customer = PaymentGatewayCustomer.new(current_user)
+
+    respond_to do |format|
+      format.json do
+        if customer.update_email(new_billable_email)
+          render json: AccountPage.new(
+            repos: find_subscribed_repos,
+            billable_email: new_billable_email
+          )
```
gylaz: Thoughts on pulling `AccountPage.new` into a var or a method?

app/controllers/accounts_controller.rb
malandrina: Removed in 29da038ed95e32d2acf95c77de80e76f5683ff03.

app/controllers/accounts_controller.rb
malandrina: Updated in 29da038ed95e32d2acf95c77de80e76f5683ff03.

spec/controllers/accounts_spec.rb
malandrina: I couldn't get the matcher `to be_bad_gateway` to work. Maybe because this is a controller spec, not a request spec? I'll do a bit more research.

app/controllers/accounts_controller.rb
malandrina: @gylaz -- done in 29da038ed95e32d2acf95c77de80e76f5683ff03.

spec/controllers/accounts_spec.rb
malandrina: @graysonwright -- figured it out: a4d7a7088fd51aa8e932b8b874ac28a6798beb4c

malandrina:
@graysonwright, @gylaz -- ready for another look when you have time.

app/controllers/accounts_controller.rb
```diff
@@ -1,11 +1,39 @@
...
+
+  def update
+    customer = PaymentGatewayCustomer.new(current_user)
+
+    respond_to do |format|
+      format.json do
+        if customer.update_email(new_billable_email)
+          render json: updated_account_page
+        else
+          head :bad_gateway
```
graysonwright: Cool - this makes more sense. :smile:

graysonwright:
Looks good to me!
:aries::construction::phone::ramen:

malandrina:
Thanks @graysonwright! I'll merge once Travis is green.

app/controllers/accounts_controller.rb
```diff
@@ -1,11 +1,39 @@
...
+        end
+      end
+    end
   end
 
   private
 
+  def new_billable_email
+    params[:billable_email]
```
jakecraige: Should this come in through strong params?

malandrina: Yep, good idea. Updated / test added in https://github.com/thoughtbot/hound/commit/edf6e4f4ceea67f4f7d48bfb59f5dce226cb9fa1.

app/controllers/accounts_controller.rb
thorncp: Does it make sense to include an error message with the response? Can some errors be actionable by the user?

malandrina: I don't think so. `customer.update_email` will only fail if the Stripe API is down for whatever reason, and in that case there's nothing the user can do.

It's likely we'll want to validate the format of the email address before calling `customer.update_email`, but that should probably be done in a `before_filter`. Don and I are planning on handling email format validation in a separate PR when we hook up the front end to this endpoint.

thorncp: What if they attempt to change to an email taken by another user? I'm not sure what Stripe does for that. Do they just accept it and send a confirmation, or will they fail fast?

malandrina: Good point. I'm not sure, either. I'll look into it.

malandrina: Just to follow up on this, I looked through the Stripe FAQ and API reference and didn't see any indication that they validate the uniqueness of customer email addresses. [This Stack Overflow post](http://stackoverflow.com/questions/27588258/stripe-check-if-a-customer-exists) explicitly says they don't (FWIW). So I think we're good.

PR #807, opened 2015-06-12
Always save completed file reviews

salbertson:
We noticed that new file review records are being created without `completed_at` values. This should avoid that and not save any file reviews without `completed_at`.

### DIFF
--- a/app/models/style_guide/base.rb
+++ b/app/models/style_guide/base.rb
@@ -1,4 +1,3 @@
-# Base to contain common style guide logic
 module StyleGuide
   class Base
     pattr_initialize :repo_config, :repository_owner_name
@@ -8,8 +7,10 @@ module StyleGuide
       repo_config.enabled_for?(name)
     end
 
-    def file_included?(_file)
-      true
+    def file_included?(*)
+      raise StandardError.new(
+        "Implement #file_included? in your StyleGuide class"
+      )
     end
 
     private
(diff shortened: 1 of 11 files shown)


croaky:
Are SCSS, CoffeeScript, and JavaScript all set with the completed bit? Thought I'd see specs for those like Ruby and Haml.

Also, those Ruby style specs are really noisy, seem like they're testing Rubocop. Do we need them or is it better to delete them for less maintenance?

app/models/style_guide/base.rb
```diff
@@ -8,7 +8,7 @@ def enabled?
       repo_config.enabled_for?(name)
     end
 
-    def file_included?(_file)
+    def file_included?(*)
```
gylaz: Should we remove this from base altogether? That way a new linter doesn't forget to implement it?

gylaz:
:+1:

salbertson:
@croaky, yup, all the style guide classes should response to `#file_review` with completed file reviews. I went through the style guide specs, looks like we covered them all.

Yea, there are a lot of tests covering RuboCop cases, but mostly because we are using RuboCop in an abnormal way.

app/models/style_guide/base.rb
salbertson: What do you think about raising? Check out the latest commit.

PR #808, opened 2015-06-12
Fix bundler caching on Travis to speed up builds

croaky:
Trying technique noted on:

https://github.com/thoughtbot/suspenders/pull/567#issuecomment-109330176

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 language: ruby
 sudo: false
-cache: bundler
 rvm: 2.2.2
+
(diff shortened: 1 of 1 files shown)


croaky:
Not entirely sure if this works on an open source repo. Will see what the results are.

croaky:
Looks like we're switching to Circle instead.

iwz:
Did you switch to circle because of this issue?

croaky:
@iwz No, more because Hound is open source, it's on the free OSS tier for Travis, which gets backed up. Circle has been faster to start builds.

PR #811, opened 2015-06-12
Listen for link clicks correctly

croaky:
* Listen for link clinks on all `auth` elements on the page,
  not just the first one in the header.
* Don't add `trackClick` listener for each
  `_auth_button` partial render.

Refactoring:

* Remove unnecessary `ANALYTICS` environment variable.
  This is in line with our other apps.

### DIFF
--- a/app/views/application/_analytics.haml
+++ b/app/views/application/_analytics.haml
@@ -13,6 +13,17 @@
       #{raw identify_hash.to_json},
       #{raw intercom_hash.to_json}
     );
+- else
+  :javascript
+    var links = document.getElementsByClassName("auth");
+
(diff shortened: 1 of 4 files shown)


salbertson:
Looks good to me. :+1:

app/views/application/_analytics.haml
```diff
@@ -13,6 +13,17 @@
       #{raw identify_hash.to_json},
       #{raw intercom_hash.to_json}
     );
+- else
+  :javascript
+    var links = document.getElementsByClassName("auth");
```
teoljungberg: Could this be extracted into a partial later?

PR #812, opened 2015-06-12
Store utm_source on user

croaky:
The goal of this change is to make it easier
to query the customer acquisition channel
where the user came from.


croaky:
End result chart on the [North Star Metric](https://robots.thoughtbot.com/north-star-metric) spreadsheet will look like this in a month or two:

screen shot 2015-06-12 at 4 48 53 pm

croaky:
Got an in-person "looks good" from @salbertson

PR #813, opened 2015-06-12
Add ability to exclude SCSS files

gylaz:
Monkey patch the SCSSLint gem to expand the file path
to the same directory as the config file.
Since the config file is a Tempfile,
that directory can be acquired by creating another Tempfile

This only makes the top level `exclude:` option work
in the SCSSLint configuration file.

`exclude` at each individual SCSS linters level will not work
because we are passing SCSSLint string (file content),
and thus file path is not available when SCSSLint needs it.

### DIFF
--- a/lib/ext/scss-lint/config.rb
+++ b/lib/ext/scss-lint/config.rb
@@ -1,0 +1,15 @@
+# Patch SCSSLint::Config to expand file_path
+# to the same directory as the config file,
+# which is a Tempfile itself.
+module SCSSLint
+  class Config
+    def excluded_file?(file_path)
+      config_file_tmp_dir = Dir.tmpdir
+      abs_path = File.expand_path(file_path, config_file_tmp_dir)
+
+      @options.fetch("exclude", []).any? do |exclusion_glob|
+        File.fnmatch(exclusion_glob, abs_path)
+      end
+    end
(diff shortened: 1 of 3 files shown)


lib/ext/scss-lint/config.rb
```diff
@@ -0,0 +1,15 @@
+# Patch SCSSLint::Config to expand file_path
+# to the same directory as the config file,
+# which is a Tempfile itself.
+module SCSSLint
+  class Config
+    def excluded_file?(file_path)
+      config_file_tmp_dir = File.dirname(Tempfile.new(""))
```
thorncp: Would it be cleaner to use [`Dir.tmpdir`](http://ruby-doc.org/stdlib-2.2.2/libdoc/tmpdir/rdoc/Dir.html) here? It [looks like](http://ruby-doc.org/stdlib-2.2.2/libdoc/tempfile/rdoc/Tempfile.html#method-c-new) that's what `Tempfile` is using, so it seems like it should return what we need.

thorncp:
LGTM :+1:

lib/ext/scss-lint/config.rb
gylaz: Nice!

PR #814, opened 2015-06-12
Validate billable email when updating account

malandrina:
* Update `AccountsController#update` to validate billable email
* Add `email_validator` to Gemfile
* Rename `AccountsController` spec file
* Don't use `strong_parameters` in `AccountsController` because it doesn't not return a JSON-encoded response with nicely formatted error messages, per this issue: https://github.com/rails/strong_parameters/issues/157


thorncp:
LGTM :+1:

app/controllers/accounts_controller.rb
```diff
@@ -22,8 +24,15 @@ def update
 
   private
 
+  def validate_billable_email
+    unless EmailValidator.valid?(new_billable_email)
+      error_message = I18n.t("account.email.invalid")
```
Koronen: `error_message = I18n.t("account.billable_email.invalid")`?

teoljungberg: I prefer `billable_email` over just plain 'old `email` aswell

malandrina: Good call. Changed in my most recent commit.

PR #815, opened 2015-06-13
Add "BUNDLED WITH" section to Gemfile.lock

Koronen:
It's a thing now. See
* <https://github.com/bundler/bundler/pull/3546>
* <https://github.com/bundler/bundler/issues/3697>

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -390,3 +390,6 @@ DEPENDENCIES
   uglifier (>= 1.0.3)
   unicorn
   webmock
+
+BUNDLED WITH
+   1.10.3


teoljungberg:
We tried this approach in #788, the problem is that if you run `bundle` with `1.9.*` it will remove the `BUNDLED_WITH` section from `Gemfile.lock`. Which will cause, unneccessary diff-noise and comments.

If there is a way to solve so everyone _must_ run bundler 1.10, that would be great. Sadly I can't see any way around it now

Koronen:
Sorry, I failed to notice #788.

PR #817, opened 2015-06-16
Save subscription metadata as string

gylaz:
Stripe metadata does not support arrays.
The metadata must be a string.

Fixes these errors from Stripe:

    {
      error:
      {
        type: "invalid_request_error"
        message: "Invalid val: ["161936", "326736"] must be a string under 500 characters"
        param: "metadata"
      }
    }


salbertson:
Looks good to me. :+1:

PR #818, opened 2015-06-16
Use new GitHub API endpoint for user's repos

gylaz:
As per: https://developer.github.com/v3/repos/#list-your-repositories
we should be using the new API endpoint to fetch user's repositories.


spec/support/helpers/github_api_helper.rb
```diff
@@ -262,20 +256,7 @@ def stub_contents_request(options = {})
...
 
-  def stub_orgs_request(token)
-    stub_request(
-      :get,
-      "https://api.github.com/user/orgs?per_page=100"
-    ).with(
-      headers: { "Authorization" => "token #{token}" }
-    ).to_return(
-      status: 200,
-      body: File.read('spec/support/fixtures/github_orgs_response.json'),
```
salbertson: Should we remove these fixtures if they aren't used?

salbertson:
What do you think about removing the fixtures that are no longer being used?

Looks good to me. :+1:

spec/support/helpers/github_api_helper.rb
gylaz: Great idea.

spec/support/fixtures/github_repos_response_for_jimtom_page2.json
```diff
@@ -32,6 +32,6 @@
   "created_at": "2011-01-26T19:01:12Z",
   "updated_at": "2011-01-26T19:14:43Z",
   "permissions": {
-    "admin": false
+    "admin": true
```
salbertson: Should a test change if the fixture changes?

salbertson:
@gylaz, one questions, but looks good to me. :+1:

spec/support/fixtures/github_repos_response_for_jimtom_page2.json
gylaz: The test was using different fixtures the ones that aren't needed anymore and were removed in this PR. The to make the test pass using these pagination fixtures, I had to make this change.

salbertson:
@gylaz is this good to go?

https://developer.github.com/changes/2015-06-10-breaking-changes-to-organization-permissions-coming-on-june-24/

gylaz:
:shipit:

PR #819, opened 2015-06-16
Remove .gitkeep from app/models folder [ci skip]

JuanitoFatas:
Plenty of files under `app/models` now, no need to use `.gitkeep` anymore. :relaxed:.


JuanitoFatas:
Failure is not related with the change in this pull request.

arunagw:
I think this is structure created by `rails` or `suspenders` 

Do we really need to delete this file?

JuanitoFatas:
> Do we really need to delete this file?

No need.

PR #826, opened 2015-06-19
Fix intermittent test failures

gylaz:
There are intermittent test failures,
due to us setting the `github_id` on repo.
Since the repo factory sets the `github_id` via sequence,
sometimes there are conflicts with what we set.
No tests depend on explicit `github_id`, so this commit kills it.


thorncp:
:+1:

PR #827, opened 2015-06-20
Add ability to signup using bitbucket

bronislav:
I know that this question is asked in #222 and #285, but those questions were without pull request.

I just want to add bitbucket support to this awesome service, because at my work I'm mostly using Bitbucket.

Any thoughts?


config/initializers/omni_auth.rb
```diff
@@ -1,8 +1,13 @@
...
   provider(
     :github,
-    ENV['GITHUB_CLIENT_ID'],
-    ENV['GITHUB_CLIENT_SECRET'],
+    ENV["GITHUB_CLIENT_ID"],
+    ENV["GITHUB_CLIENT_SECRET"],
     scope: 'user:email,repo'
   )
+  provider(
+    :github,
```
Koronen: `:bitbucket`?

teoljungberg:
thanks for the pr @bronislav, maybe I'm missing something here. But this only allows the user to signup using Bitbucket. But not to use the actual service. So what value does this add?

bronislav:
This is not completed yet. I just want to here feedback.
Is such contribution (I'm going to add full bitbucket support) are acceptable or I should create own fork?

Actually, add bitbucket support in one pull request is quite complicated and it will be a lot of changes. I'd prefer to add things incrementally.

teoljungberg:
I figured as much. We're planning on adding support for multiple providers in the future, our current focus is to supporting more languages.

We would love to work with you later on to add BitBucket support

/cc @thoughtbot/hound

bronislav:
@teoljungberg, when do you plan to add other providers? Is is better to start working on it or have separate fork?

/cc @thoughtbot/hound

teoljungberg:
@bronislav It's not on the current roadmap. And as far as the discussions have gone so far, it won't be for a while longer.

PR #828, opened 2015-06-23
Extract PR status setting to CommitStatus class

gylaz:
This solves the issue where the app was always setting
the final success status with `houndci`'s token,
instead of user's token (if available).

This also gets all that code out of `BuildRunner`
and into its own `CommitStatus` class.

### DIFF
--- a/app/jobs/completed_file_review_job.rb
+++ b/app/jobs/completed_file_review_job.rb
@@ -35,7 +35,11 @@ class CompletedFileReviewJob
     payload = Payload.new(build.payload)
     pull_request = PullRequest.new(payload, ENV.fetch("HOUND_GITHUB_TOKEN"))
 
-    BuildReport.run(pull_request, build)
+    BuildReport.run(
+      pull_request: pull_request,
+      build: build,
+      token: build.user_token,
+    )
   rescue ActiveRecord::RecordNotFound, Resque::TermException
     Resque.enqueue(self, attributes)
   end
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -1,5 +1,6 @@
 class Build < ActiveRecord::Base
   belongs_to :repo
+  belongs_to :user
   has_many :file_reviews, dependent: :destroy
   has_many :violations, through: :file_reviews
 
@@ -7,8 +8,18 @@ class Build < ActiveRecord::Base
 
   validates :repo, presence: true
 
+  delegate :name, to: :repo, prefix: true
+
   def completed?
     file_reviews.where(completed_at: nil).empty?
+  end
+
+  def violation_count
+    violations.map(&:messages_count).sum
+  end
+
+  def user_token
+    (user && user.token) || Hound::GITHUB_TOKEN
   end
 
   private
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -1,0 +1,33 @@
+class CommitStatus
+  def initialize(repo_name:, sha:, token:)
+    @repo_name = repo_name
+    @sha = sha
+    @token = token
+  end
+
+  def set_pending
+    github.create_pending_status(repo_name, sha, I18n.t(:pending_status))
+  end
+
+  def set_success(violation_count)
+    message = I18n.t(:success_status, count: violation_count)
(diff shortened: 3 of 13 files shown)


app/models/commit_status.rb
```diff
@@ -0,0 +1,29 @@
+class CommitStatus
+  def initialize(repo_name:, sha:, github:)
+    @repo_name = repo_name
+    @sha = sha
+    @github = github
```
JuanitoFatas: What do you think about use [`pattr_initialize`](https://github.com/barsoom/attr_extras#pattr_initialize) here?

JuanitoFatas: Sorry did not see you use keyword arguments, ignore me.

app/services/build_runner.rb
```diff
@@ -28,10 +28,11 @@ def run
 
   def review_pull_request
     track_subscribed_build_started
-    create_pending_status
+    commit_status.set_pending
     upsert_owner
     build = create_build
     BuildReport.run(pull_request, build)
+    commit_status.set_success(build.violation_count)
```
thorncp: When the SOA stuff lands, this'll have to go back into `BuildReport` (or somewhere else) since we'll have to wait for all the workers to respond before setting the success status. Worth moving it in there now?

gylaz: Hmm, the whole point behind this PR was to move the status manipulation out of `BuildReport`. Do we have a pretty clear idea on how we'll handle finalizing the PR review?

teoljungberg: I'll prefer having it moved into `BuildReport` aswell

gylaz: Can you guys give a little more explanation behind this? Right now, both `BuildRunner` and `BuildReport` are concerned with setting PR status. That doesn't follow SRP, and two classes know how to do this. What is the reason behind wanting to move just the success status in there?

thorncp: It should be moved out because we'll be firing off the workers asynchronously and setting the success status in `BuildRunner` would be premature, e.g. we mark it as success then a linter fails for some reason.

thorncp: Hmm, if an asynchronous linter bombs, we'll need a way to set the failure state after-the-fact as well. Now I think this is good as-is and we can solve those problems as the SOA changes come in.

config/initializers/constants.rb
```diff
@@ -0,0 +1,3 @@
+module Hound
+  GITHUB_TOKEN = ENV.fetch("HOUND_GITHUB_TOKEN")
```
thorncp: :+1: I like where this is

app/models/build.rb
```diff
@@ -7,10 +8,20 @@ class Build < ActiveRecord::Base
 
   validates :repo, presence: true
 
+  delegate :name, to: :repo, prefix: true
+
   def completed?
     file_reviews.where(completed_at: nil).empty?
   end
 
+  def violation_count
```
thorncp: Should we add tests for this?

app/models/build.rb
```diff
@@ -7,10 +8,20 @@ class Build < ActiveRecord::Base
...
+    violations.map(&:messages_count).sum
+  end
+
+  def user_token
```
thorncp: Should we add tests for this?

thorncp:
:+1: I think this makes things quite a bit cleaner.

app/jobs/completed_file_review_job.rb
```diff
@@ -35,7 +35,11 @@ def self.perform(attributes)
     payload = Payload.new(build.payload)
     pull_request = PullRequest.new(payload, ENV.fetch("HOUND_GITHUB_TOKEN"))
 
-    BuildReport.run(pull_request, build)
+    BuildReport.run(
+      pull_request: pull_request,
+      build: build,
+      token: build.user_token
```
gylaz: @teoljungberg It's working!

app/models/build.rb
gylaz: Yep.

app/models/build.rb
gylaz: Will do.

PR #831, opened 2015-06-25
Don't hard-fail on repo update during sync

gylaz:
There are situations when both github_id and github_name are taken
by other repos. We don't want to fail the sync in that case,
but we do want to get notified, and address those repos manually.

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -22,7 +22,11 @@ class Repo < ActiveRecord::Base
       find_by(github_id: attributes[:github_id]) ||
       Repo.new
 
-    repo.update!(attributes)
+    begin
+      repo.update!(attributes)
+    rescue ActiveRecord::RecordInvalid => error
+      report_update_failure(error, attributes)
+    end
 
     repo
   end
@@ -73,4 +77,15 @@ class Repo < ActiveRecord::Base
       full_github_name.split("/").first
     end
   end
+
+  def self.report_update_failure(error, attributes)
+    Raven.capture_exception(
+      error,
+      extra: {
+        github_id: attributes[:github_id],
+        full_github_name: attributes[:full_github_name],
+      }
+    )
+  end
+  private_class_method :report_update_failure
 end
--- a/spec/models/repo_spec.rb
+++ b/spec/models/repo_spec.rb
@@ -144,6 +144,24 @@ describe Repo do
         expect(repo.reload).to be_present
       end
     end
+
+    context "when one repo has taken the github name and another taken id" do
+      it "repors update failure" do
+        github_name = "foo/bar"
+        github_id = 40023
+        _repo_with_id = create(:repo, github_id: github_id)
+        _repo_with_name = create(:repo, full_github_name: github_name)
+        new_attributes = { github_id: github_id, full_github_name: github_name }
+        allow(Raven).to receive(:capture_exception)
+
+        Repo.find_or_create_with(new_attributes)
+
+        expect(Raven).to have_received(:capture_exception).with(
+          instance_of(ActiveRecord::RecordInvalid),
+          extra: { github_id: github_id, full_github_name: github_name }
+        )
+      end
+    end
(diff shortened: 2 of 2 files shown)


spec/models/repo_spec.rb
```diff
@@ -144,6 +144,24 @@
         expect(repo.reload).to be_present
       end
     end
+
+    context "when one repo has take the github name and another taken id" do
+      it "send a raven" do
```
croaky: Three-eyed or regular?

spec/models/repo_spec.rb
```diff
@@ -144,6 +144,24 @@
         expect(repo.reload).to be_present
       end
     end
+
+    context "when one repo has take the github name and another taken id" do
```
croaky: take -> taken

app/models/repo.rb
```diff
@@ -22,7 +22,14 @@ def self.find_or_create_with(attributes)
       find_by(github_id: attributes[:github_id]) ||
       Repo.new
 
-    repo.update!(attributes)
+    begin
+      repo.update!(attributes)
+    rescue ActiveRecord::RecordInvalid => error
+      Raven.capture_exception(
+        error,
+        extra: { github_id: repo.github_id, full_github_name: repo.name }
```
croaky: Should we access attribute here rather than repo?

croaky:
Looks good to me.

spec/models/repo_spec.rb
```diff
@@ -144,6 +144,24 @@
...
+        _repo_with_id = create(:repo, github_id: github_id)
+        _repo_with_name = create(:repo, full_github_name: github_name)
+        new_attributes = { github_id: github_id, full_github_name: github_name }
+        allow(Raven).to receive(:capture_exception)
+
+        Repo.find_or_create_with(new_attributes)
+
+        expect(Raven).to have_received(:capture_exception).with(
+          instance_of(ActiveRecord::RecordInvalid),
+          extra: { github_id: github_id, full_github_name: github_name }
```
teoljungberg: Should we add a trailing comma here?

spec/models/repo_spec.rb
```diff
@@ -144,6 +144,24 @@
         expect(repo.reload).to be_present
       end
     end
+
+    context "when one repo has take the github name and another taken id" do
+      it "send a raven" do
+        github_name = "foo/bar"
+        github_id = 40023
+        _repo_with_id = create(:repo, github_id: github_id)
```
teoljungberg: what do you think of ommitting these local-variables entirely?

malandrina: @teoljungberg out of curiosity, what do you think is the benefit of removing the variables? I like them because I think they make the setup phase easier to grok.

spec/models/repo_spec.rb
gylaz: I'm a fan of trailing comma when args/params are on their own line. Is there a benefit to adding a trailing coma when they are on the same line?

teoljungberg: What happeneds if another argument is passed to `capture_exception`? We're going to get a line-diff here since we're adding a comma

gylaz: Do you mean, if the line changes from
    extra: { github_id: github_id, full_github_name: github_name }
to
   extra: { github_id: github_id, full_github_name: github_name, test: something }
?

I think, in that case, it won't matter if the trailing comma is there or not -- the line will be equally flagged as changed.

If you mean, what happens if another argument is added that forces each hash key-value pair on its own line, then I think at that time the trailing comma can be added.

What does RuboCop report? I think we should enable that trailing comma rule, and see if this get flagged. If yes, then I think it's good to start adding trailing commas in all situations.

teoljungberg: Sounds like a plan, let's leave this in for now and activate the Rubocop rule and see where that takes us

PR #832, opened 2015-06-26
SCSS review as a service

salbertson:
This change is the first style guide we are breaking out of Hound and moving to a service. The service will share Hound queue and use incomplete `FileReview`s to keep track of SCSS reviews.


app/jobs/scss_review_job.rb
```diff
@@ -0,0 +1,3 @@
+class ScssReviewJob
+  @queue = :scss_review
```
seanpdoyle: whoa how does this work without a mixin or inheritance?

brittballard:
Failed code review.

app/jobs/completed_file_review_job.rb
```diff
@@ -0,0 +1,37 @@
+class CompletedFileReviewJob
+  @queue = :high
+
+  def self.perform(attributes)
+    # filename
+    # commit_sha
+    # patch
+    # violations
+    #   [{ line: 123, message: "WAT" }]
```
brittballard: What's up with the comments, dawg?

brittballard:
I don't have a lot of context, and one nitpic, but LGTM.

app/jobs/completed_file_review_job.rb
salbertson: They "document" the interface between Hound and style services. These are the values expected to be included in the Resque job. I was thinking about leaving the comments until we have a better way to document and/or enforce the interface.

app/jobs/scss_review_job.rb
salbertson: The class is used directly by `Resque` in https://github.com/thoughtbot/hound/blob/sa-ct-review-scss-in-service/app/models/style_guide/scss.rb.

teoljungberg:
I'm so excited for this! Great work pulling this together @salbertson @arunagw @gylaz @thorncp

PR #833, opened 2015-06-26
Reschedule job if build hasn't been created yet

salbertson:
I believe there are cases where a SCSS review can finish before the build is saved. This change handles that case by rescheduling the `CompletedFileReview` job when the build doesn't exist.


thorncp:
LGTM :+1:

gylaz:
Does it make sense to create (save) the build _before_ we send out the review jobs?

PR #834, opened 2015-06-26
Handle Resque::TermException in CompletedReviewJob

salbertson:
This is recommended by Heroku and is necessary since we aren't extending `ApplicationJob`. We aren't extending `ApplicationJob` because the style review services serialize jobs using the job class, not `ActiveJob`s wrapper class.


thorncp:
LGTM

salbertson:
Thanks for the quick reviews @thorncp!

PR #835, opened 2015-06-29
Refactor CommitFile

salbertson:
* Fixes an issue where multiple builds could be created with the same `commit_sha` and the correct one might not be found in `CompletedFileReview`.
* Makes it easier to use `CommitFile`, it lets you initialize `CommitFile` without having to create `file` and `commit` objects
* Gets rid of an `OpenStruct` in `CompletedFileReview`
* Makes it clear what type of "file" object we are dealing with. `CommitFile` was often confused with what returned by `GithubApi` and `Octokit` when fetching pull request files
* Avoids confusing `patch_body` and `patch`

### DIFF
--- a/app/jobs/completed_file_review_job.rb
+++ b/app/jobs/completed_file_review_job.rb
@@ -4,22 +4,25 @@ class CompletedFileReviewJob
   def self.perform(attributes)
     # filename
     # commit_sha
+    # pull_request_number
     # patch
     # violations
     #   [{ line: 123, message: "WAT" }]
 
-    build = Build.find_by!(commit_sha: attributes.fetch("commit_sha"))
+    build = Build.find_by!(
+      pull_request_number: attributes.fetch("pull_request_number"),
+      commit_sha: attributes.fetch("commit_sha")
+    )
     file_review = build.file_reviews.find_by(
       filename: attributes.fetch("filename")
     )
-
-    file = OpenStruct.new(
+    commit_file = CommitFile.new(
       filename: file_review.filename,
-      patch: attributes.fetch("patch")
+      content: "",
+      patch: attributes.fetch("patch"),
+      pull_request_number: attributes.fetch("pull_request_number"),
+      sha: build.commit_sha
     )
-
-    commit = Commit.new(build.repo.full_github_name, build.commit_sha, nil)
-    commit_file = commit_file = CommitFile.new(file, commit)
 
     attributes.fetch("violations").each do |violation|
       line = commit_file.line_at(violation.fetch("line"))
--- a/app/models/commit_file.rb
+++ b/app/models/commit_file.rb
@@ -1,12 +1,12 @@
 class CommitFile
-  pattr_initialize :file, :commit
+  attr_reader :filename, :content, :patch, :pull_request_number, :sha
 
-  def filename
-    file.filename
-  end
-
-  def content
-    @content ||= commit.file_content(filename)
+  def initialize(filename:, content:, patch:, pull_request_number:, sha:)
+    @filename = filename
+    @content = content
+    @patch = patch
+    @pull_request_number = pull_request_number
+    @sha = sha
   end
 
   def line_at(line_number)
@@ -14,21 +14,9 @@ class CommitFile
       UnchangedLine.new
   end
 
-  def sha
-    commit.sha
-  end
-
-  def patch_body
-    file.patch
-  end
-
   private
 
   def changed_lines
-    @changed_lines ||= patch.changed_lines
-  end
-
-  def patch
-    Patch.new(file.patch)
+    @changed_lines ||= Patch.new(patch).changed_lines
   end
 end
--- a/app/models/pull_request.rb
+++ b/app/models/pull_request.rb
@@ -7,8 +7,8 @@ class PullRequest
     @comments ||= user_github.pull_request_comments(full_repo_name, number)
   end
 
-  def pull_request_files
-    @pull_request_files ||= changed_pull_request_files
+  def commit_files
+    @commit_files ||= modified_commit_files
   end
 
   def comment_on_violation(violation)
@@ -39,15 +39,24 @@ class PullRequest
 
   private
 
-  def build_commit_file(file)
-    CommitFile.new(file, head_commit)
+  def modified_commit_files
+    modified_github_files.map do |github_file|
+      CommitFile.new(
+        filename: github_file.filename,
+        content: head_commit.file_content(github_file.filename),
+        patch: github_file.patch,
+        pull_request_number: number,
+        sha: head_commit.sha
+      )
+    end
   end
 
-  def changed_pull_request_files
-    user_github.
-      pull_request_files(full_repo_name, number).
-      reject { |file| file.status == FILE_REMOVED_STATUS }.
-      map { |file| build_commit_file(file) }
+  def modified_github_files
+    github_files = user_github.pull_request_files(full_repo_name, number)
+
+    github_files.select do |github_file|
+      github_file.status != FILE_REMOVED_STATUS
+    end
   end
 
   def user_github
--- a/spec/jobs/completed_file_review_job_spec.rb
+++ b/spec/jobs/completed_file_review_job_spec.rb
@@ -57,12 +57,39 @@ describe CompletedFileReviewJob do
         )
       end
     end
+
+    context "when there are two builds with the same commit_sha" do
+      it "finds the correct build by pull request number" do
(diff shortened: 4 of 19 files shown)


app/jobs/completed_file_review_job.rb
```diff
@@ -4,23 +4,26 @@ class CompletedFileReviewJob
...
+    # pull_request_number
     # patch
     # violations
     #   [{ line: 123, message: "WAT" }]
 
-    build = Build.find_by!(commit_sha: attributes.fetch("commit_sha"))
+    build = Build.find_by!(
+      pull_request_number: attributes.fetch("pull_request_number"),
+      commit_sha: attributes.fetch("commit_sha")
+    )
```
gylaz: Thoughts on cleaning this up by extracting this into a `find_build` method that takes `attributes`?

app/jobs/completed_file_review_job.rb
```diff
@@ -4,23 +4,26 @@ class CompletedFileReviewJob
...
     file_review = build.file_reviews.find_by(
       filename: attributes.fetch("filename")
     )
-
-    file = OpenStruct.new(
+    commit_file = commit_file = CommitFile.new(
```
gylaz: `commit_file` twice.

app/jobs/completed_file_review_job.rb
```diff
@@ -4,23 +4,26 @@ class CompletedFileReviewJob
...
       filename: file_review.filename,
-      patch: attributes.fetch("patch")
+      content: "",
+      patch: attributes.fetch("patch"),
+      pull_request_number: attributes.fetch("pull_request_number"),
+      sha: build.commit_sha
```
gylaz: Could this be extracted into a private method to clean this method up?

app/models/commit_file.rb
```diff
@@ -1,12 +1,12 @@
 class CommitFile
-  pattr_initialize :file, :commit
-
-  def filename
-    file.filename
-  end
-
-  def content
-    @content ||= commit.file_content(filename)
+  attr_reader :filename, :content, :patch, :pull_request_number, :sha
```
gylaz: Should these be private?

app/models/pull_request.rb
```diff
@@ -39,15 +39,22 @@ def head_commit
...
 
-  def build_commit_file(file)
-    CommitFile.new(file, head_commit)
-  end
+  def modified_commit_files
+    github_files = user_github.pull_request_files(full_repo_name, number)
+
+    modified_github_files = github_files.select do |github_file|
+      github_file.status != FILE_REMOVED_STATUS
+    end
```
gylaz: Thoughts on extracting this to a `modified_github_files` method?

spec/jobs/completed_file_review_job_spec.rb
```diff
@@ -57,12 +57,38 @@
         )
       end
     end
+
+    context "when there are two builds with the same commit_sha" do
+      it "finds the correct build by pull request" do
```
gylaz: `by pull request number`?

spec/jobs/completed_file_review_job_spec.rb
```diff
@@ -71,11 +97,15 @@ def attributes
   end
 
   def create_build
-    @build ||= build(:build, commit_sha: attributes.fetch("commit_sha"))
+    @build ||= build(
```
gylaz: Looks like `create_build` is called once per test (from another helper method). Thoughts on removing the memoization?

spec/models/pull_request_spec.rb
```diff
@@ -80,19 +80,34 @@
...
+        patch: "patch"
+      )
+      modified_file = double(
+        :github_file,
+        filename: "baz.rb",
+        status: "modified",
+        patch: "patch"
+      )
+      removed_file = double(
+        :github_file,
```
gylaz: Use strings for consistency?

gylaz:
LGTM

spec/models/style_guide/java_script_spec.rb
```diff
@@ -159,26 +159,26 @@
...
-  def build_file(content = "foo")
+  def build_commit_file(content = "foo")
     filename = "some-file.js"
     line = double("Line", number: 1, patch_position: 1, changed?: true)
-    double("File", filename: filename, line_at: line, content: content)
+    double("CommitFile", filename: filename, line_at: line, content: content)
   end
 
-  def violations_in(file, repo_config, repository_owner_name: "not_thoughtbot")
+  def violations_in(commit_file, repo_config, repository_owner_name: "not_thoughtbot")
```
teoljungberg: What do you think of using keyword args here, this order is getting crazy.

app/models/pull_request.rb
teoljungberg: That seems like a good abstraction, it's also what we had before

spec/jobs/completed_file_review_job_spec.rb
```diff
@@ -57,12 +57,38 @@
...
       end
     end
+
+    context "when there are two builds with the same commit_sha" do
+      it "finds the correct build by pull request" do
+        create(:build, commit_sha: "abc123", pull_request_number: 1)
+        correct_build = create(
+          :build,
+          commit_sha: "abc123",
+          pull_request_number: 123
```
teoljungberg: trailing comma?

spec/jobs/completed_file_review_job_spec.rb
```diff
@@ -57,12 +57,38 @@
...
+        )
+        create(
+          :file_review,
+          build: correct_build,
+          filename: attributes.fetch("filename")
```
teoljungberg: trailing comma?

spec/jobs/completed_file_review_job_spec.rb
```diff
@@ -57,12 +57,38 @@
...
+        )
+        stub_build_report_run
+        pull_request = stub_pull_request
+        stub_payload
+
+        CompletedFileReviewJob.perform(attributes)
+
+        expect(BuildReport).to(
+          have_received(:run).with(pull_request, correct_build)
```
teoljungberg: I usually prefer splitting up the `with` like so:

```ruby
expect(BuildReport).to have_received(:run).with(
  pull_request,
  correct_build,
) 
```

spec/models/patch_spec.rb
```diff
@@ -5,8 +5,7 @@
 describe Patch do
   describe "#changed_lines" do
     it 'returns lines that were modified' do
-      patch_body = File.read('spec/support/fixtures/patch.diff')
-      patch = Patch.new(patch_body)
+      patch = Patch.new(File.read('spec/support/fixtures/patch.diff'))
```
teoljungberg: I think this reads less clear than before, thoughts?

spec/models/style_guide/java_script_spec.rb
salbertson: Good idea, done.

spec/models/patch_spec.rb
salbertson: Yea, maybe a little dense. I extracted a local variable.

app/jobs/completed_file_review_job.rb
salbertson: Good catch, fixed.

app/jobs/completed_file_review_job.rb
salbertson: Yea, this entire job class needs to be broken up. I think we should extract a service object or something in a different PR.

app/models/commit_file.rb
salbertson: No, they are used in app/models/style_guide/scss.rb.

app/models/pull_request.rb
salbertson: Good idea, done.

spec/jobs/completed_file_review_job_spec.rb
salbertson: Yes, updated.

spec/jobs/completed_file_review_job_spec.rb
salbertson: I like it, done. I also combined the two helper methods since it wasn't really "creating" a build and the `create_build` method was only used once.

spec/models/pull_request_spec.rb
salbertson: I left them as symbols because the object type is determined by Octokit. Maybe remove the double type and use a nicely named variable?

spec/jobs/completed_file_review_job_spec.rb
salbertson: I like that too, done.

PR #836, opened 2015-06-30
Show email update form if user is Stripe customer

malandrina:
* Only show the form for updating the billable email on an account if the
  current user has a Stripe customer ID. Users who aren't in Stripe's
  system shouldn't have the ability to update their billable email
  address.


gylaz:
LGTM.

PR #838, opened 2015-07-02
Don't comment until the build is finished

thorncp:
This will allow us to respect the MAX_COMMENTS environment variable
across the whole build and should help alleviate some abuse warnings.

As-is, if an external worker fails for whatever reason, the build will
not be marked as complete and comments will never go out. I *think*
we'll want to get that working by having the workers be retryable and we
can mark the build as complete if it fails too many times. Maybe this is
good enough for now, though?

### DIFF
--- a/app/services/build_report.rb
+++ b/app/services/build_report.rb
@@ -10,9 +10,8 @@ class BuildReport
   pattr_initialize :pull_request, :build
 
   def run
-    commenter.comment_on_violations(priority_violations)
-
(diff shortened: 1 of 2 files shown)


gylaz:
We need to mitigate the abuse errors, and this helps with that. So, :+1:

salbertson:
:shipit:

teoljungberg:
Good catch @thorncp !

PR #839, opened 2015-07-03
Reduces usage of ENABLE_HTTPS

arunagw:
We can remove the usage of `ENABLE_HTTPS` variable and that can be removed in later commits.

We can just us what rails is generating for url. 
So sending a URL from controller can be http or https based on what is application using/running. 

We can force the https in `production.rb` instead of having this check in before_filter.


gylaz:
Would you mind adding a commit description of explaining why this is a good idea?

arunagw:
@gylaz Yes. I have updated the PR description.

Card is here https://trello.com/c/3xft1Ylq

app/controllers/activations_controller.rb
```diff
@@ -23,7 +23,11 @@ def check_repo_plan
   end
 
   def activator
-    @activator ||= RepoActivator.new(repo: repo, github_token: github_token)
+    @activator ||= RepoActivator.new(
+      repo: repo,
+      github_token: github_token,
+      builds_url: builds_url
```
gylaz: Instead of passing this in, how about just including the route helpers in the `RepoActivator`?

arunagw:
@gylaz I have updated PR. Please have a look.

gylaz:
Looks good!

lib/routing.rb
```diff
@@ -0,0 +1,10 @@
+module Routing
+  extend ActiveSupport::Concern
+  include Rails.application.routes.url_helpers
+
+  included do
+    def default_url_options
+      ActionMailer::Base.default_url_options
+    end
+  end
+end
```
gylaz: I'm interested about this approach. What does it provide over doing `include Rails.application.routes.url_helpers` in `RepoActivator`?

arunagw: The only thing we need to do is set `host` for routes as well. 

Basically 

```Rails.application.routes.default_url_options[:host] = ENV['host']```

arunagw:
@gylaz what you think about this PR?

lib/routing.rb
arunagw: We can set this in all of our environment files or can just read from mailer options like I did in this concern.

gylaz: Could we just set it once in `application.rb` for all envs? If possible, that approach (plus mixing in `Rails.application.routes.url_helpers`) seems simpler.

gylaz:
I had one more question regarding the mixin. Make the call on what makes sense. If this works on staging, then lets merge it.

arunagw:
@gylaz I have removed that module for now. Seems like too much as well.

Will check once on staging before merging.

PR #844, opened 2015-07-06
Upgrade RuboCop

peritpatrio:
Upgrade RuboCop to version 0.32.1: https://github.com/bbatsov/rubocop/releases/tag/v0.32.1


salbertson:
@peritpatrio do you want to address the failing tests?

peritpatrio:
Fixed the failing tests.

Note that there are style rules for this version of RuboCop that are not defined in `config/style_guides/ruby.yml`, such as:

- Style/BlockDelimiters (replaces Style/Blocks)
- Style/CommandLiteral
- Style/EmptyElse
- Style/IndentationConsistency
- Style/SpaceAroundOperators
- Style/SpaceInsideStringInterpolation
- Rails/Date
- Rails/FindBy
- Rails/FindEach
- Rails/TimeZone

Did not really know what should be done with those.

gylaz:
@peritpatrio We usually disable new rules, as to not introduce unexpected violations on existing users.
We also verify that none of the existing rules in default RuboCop config got disabled, or enabled -- and make the appropriate changes in Hound's default `ruby.yml` config.

gylaz:
We should consider always upgrading to RuboCop's latest default config. It would be very easy on us during upgrade to generate that config and swap it for our default -- maybe we could do away with our default. Users have raised the issue several times that they did not expect Hound to have its own default config that is different from RuboCop's default.

The downside is that users could begin receiving unexpected violations comments -- though the rules that get changed are usually for rare/uncommon violations. You can spin this into a "feature" -- Hound is getting better, more up to date.

Thoughts @salbertson @thorncp @teoljungberg?

teoljungberg:
I would expect our default config being rubocop's. If the default config we currently have differs from that, we should migrate over. This might cause some minor inconveniences, but altleast the upgrades from then on will be consistent. 
Our thoughtbot guide will be a bit trickier to handle, but that's something we can live with.

Using Rubocops config get's a :+1: from me

thorncp:
Using Rubocop's default makes sense to me and it should make maintenance easier :+1:

jondkinney:
I think it would be really great if you started using the default rubocop config! There have been several releases since 0.29.1 that I'd love to take advantage of but we have to lock our local rubocop back to an old version to keep sync'd with Hound.

peritpatrio:
I switched to RuboCop's default config now. Wasn't sure what would be the correct way to refactor some of the tests so it would be nice if someone could take a look at 'em.

app/models/user.rb
```diff
@@ -18,11 +18,11 @@ def billable_email
     payment_gateway_customer.email
   end
 
-  def has_repos_with_missing_information?
-    repos.where("in_organization IS NULL OR private IS NULL").count > 0
+  def repos_with_missing_information?
+    repos.where('in_organization IS NULL OR private IS NULL').count > 0
```
teoljungberg: Why did this need to change? It's outside the scope of this PR so could we revert these changes for a slimmer PR?

app/models/user.rb
```diff
@@ -33,14 +33,12 @@ def token=(value)
...
-    unless encrypted_token.nil?
-      crypt.decrypt_and_verify(encrypted_token)
-    end
+    crypt.decrypt_and_verify(encrypted_token) unless encrypted_token.nil?
   end
 
   def access_to_private_repos?
     if token_scopes
-      token_scopes.split(",").include? "repo"
+      token_scopes.split(',').include? 'repo'
```
teoljungberg: Same as https://github.com/thoughtbot/hound/pull/844/files#r37756435

app/models/user.rb
```diff
@@ -18,11 +18,11 @@ def billable_email
...
   end
 
-  def has_repos_with_missing_information?
-    repos.where("in_organization IS NULL OR private IS NULL").count > 0
+  def repos_with_missing_information?
+    repos.where('in_organization IS NULL OR private IS NULL').count > 0
   end
 
-  def has_active_repos?
+  def active_repos?
```
teoljungberg: Same as https://github.com/thoughtbot/hound/pull/844/files#r37756435

teoljungberg:
Hound itself doesn't use nor follow the default rubocop config, we're using the thoughtbot styleguide. So the changes that was made outside of upgrading rubocop is out of the scope of this PR

app/models/user.rb
peritpatrio: When running tests with `RAILS_ENV=test rake` I got the following:

```
An error occurred while Style/InitialIndentation cop was inspecting /Users/partiop/GitHub/hound/app/models/user.rb.
To see the complete backtrace run rubocop -d.
```

How should I approach this error in a way that won't involve making these changes?

jondkinney:
@teoljungberg you commented earlier on this PR saying you supported migrating to use the default rubocop config. So I'm confused...

Perhaps you're suggesting upgrading rubocop here with the existing thoughtbot config, then in a separate PR doing the work to migrate to using the default rubocop config?

teoljungberg:
@jondkinney I maybe have missunderstood things - I thought the discussion we had in the earlier PR was to have the default style guide being what's used to be rubocops. Not the semi-oppinionated one we have now.
All repos under the thoughtbot organization will use the thoughtbot styleguide, which can be found under `config/style_guides/thoughtbot`

peritpatrio:
@teoljungberg So how should we proceed with this? If I revert the changes made to .rb files tests won't pass as they're run against the new default RuboCop config.

teoljungberg:
I don't think I understand your point.

What we, as the maintainers of Hound, want is to ease our maintenance burden with updating the configurations for the linters. Which is why we want to change the default configuration(found in `config/style_guides/ruby.yml` to be the default one for Rubocop.

But, the codebase of Hound itself should not change. In the codebase we adhear to the thoughtbot stylegudie(https://github.com/thoughtbot/guides <https://github.com/thoughtbot/guides>. And that is not going to change. That styleguide can be found in `config/style_guides/thoughtbot/ruby.yml`

Does that answer any unanswered questions you had?

app/controllers/repos_controller.rb
```diff
@@ -4,7 +4,7 @@ def index
       format.html
 
       format.json do
-        if current_user.has_repos_with_missing_information?
+        if current_user.repos_with_missing_information?
```
gylaz: I think `has_` prefix reads better here. `current_user` is not a `repos_with_missing_information`.

salbertson:
@peritpatrio thank you for the time and effort put into this. We are planning on removing Hound configs all together, so I'm closing this for now. Check out https://github.com/thoughtbot/hound/tree/use-linter-configs to see progress.

tansaku:
hi @salbertson that link https://github.com/thoughtbot/hound/tree/use-linter-configs doesn't seem to go anywhere ...  I tried searching the code base of "linter-configs" but no joy :-(

Apologies if this has already been answered (I've re-read the thread and I'm not sure), but is there now a pathway that will see hound using a more recent version of rubocop? (now at 0.34.1)

teoljungberg:
@tansaku that links works just fine, that's a branch we're working on.

teoljungberg:
@tansaku For the pathway of upgrading rubocop - it's in our backlog and what's blocking it is that we're using a custom styleguide for Hound that's no the rubucop default. Which makes upgrading difficult. So, what the branch @salbertson linked is going to add the rubocop defaults to the config so our upgrading process can be easier

tansaku:
thanks for clarifying @teoljungberg - any idea what the timeline is for us to see the latest rubocop into hound then?

teoljungberg:
@tansaku Not yet - I'll circle back to this thread once we know more

jondkinney:
@teoljungberg any progress on the effort to stay more in sync with rubocop's releases and use their config by default?

teoljungberg:
@jondkinney We should always stay in sync with Rubocop's release cycle.

We are upgrading Rubocop in #844 

As far as the configuration issue, that is something that @gylaz and @thorncp are currently working on.

jondkinney:
@teoljungberg this is #844 :) I asked because even with the upgrade to 0.32.1 hound is still behind. Current rubocop version is: 0.36.0.

teoljungberg:
My bad - the upgrade to 0.36 is done in https://github.com/thoughtbot/hound/pull/1034 (I was to quick to copy-paste there)

jondkinney:
:+1: Thanks for the heads up @teoljungberg

Issue #845, opened 2015-07-07
Puma?


beedub:
Looks like your unicorn config is tailored for/from heroku. As they now recommend Puma over Unicorn, are you guys willing to switch? I'm open to pull requesting if so.

gylaz:
A PR would be definitely welcome. We've done a lot of research in https://github.com/thoughtbot/suspenders/pull/501, so please have a look at it.

croaky:
Closing in favor of https://github.com/thoughtbot/hound/pull/846. Thanks, @beedub!

PR #846, opened 2015-07-07
Replace unicorn with puma

beedub:
This essentially follows heroku and suspenders' guide for using puma.

Fixes https://github.com/thoughtbot/hound/issues/845


beedub:
I attempted to fix the timeouts in the specs with e452540, but it looks like they're still there :( The issue is described here https://github.com/heroku/rack-timeout/issues/55#issuecomment-75450863

Not sure where you guys want to go from here, but let me know and I can update the PR accordingly

croaky:
@beedub Can you move Rack Timeout to staging and production environments? Both the `Gemfile` groups and the `config/initializers/timeout.rb` file to `config/environments/production.rb`?

Example: https://github.com/thoughtbot/suspenders/commit/a470912e9df01fda62f0c23ce032a1c9b5493b69

spec/rails_helper.rb
```diff
@@ -27,6 +27,13 @@
   DatabaseCleaner.strategy = :deletion
 end
 
+# This is a fix from https://github.com/heroku/rack-timeout/issues/55#issuecomment-75450863
+# for timeouts described in the same link
+Capybara.server do |app, port|
+  require "rack/handler/thin"
+  Rack::Handler::Thin.run(app, Port: port)
+end
```
croaky: I think this can be deleted once Rack Timeout is moved to staging/production.

config/puma.rb
```diff
@@ -0,0 +1,17 @@
...
+port(ENV["PORT"] || 3000)
+environment(ENV["RACK_ENV"] || "development")
+
+on_worker_boot do
+  # Worker specific setup for Rails 4.1+
+  # See: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#on-worker-boot
+  ActiveRecord::Base.establish_connection
+
+  # Reconnect resque's redis after worker boots
+  Resque.redis = ENV["REDIS_URL"] || "redis://127.0.0.1:6379"
```
croaky: Can you switch up any of the `ENV[] ||` in this green portions of this PR to be `ENV.fetch`?

> Use ENV.fetch for environment variables instead of ENV[]so that unset environment variables are detected on deploy.

https://github.com/thoughtbot/guides/tree/master/best-practices#rails

croaky:
Do we want `pool` and `reaping_frequency` options set in the `database.yml`?

```
development: &default
  pool: <%= ENV.fetch("DB_POOL", 2) %>
  reaping_frequency: <%= ENV.fetch("DB_REAPING_FREQUENCY", 10) %>

production: &deploy
  encoding: utf8
  min_messages: warning
  pool: <%= [ENV.fetch("MAX_THREADS", 2), ENV.fetch("DB_POOL", 2)].max %>
  timeout: 5000
  url:  <%= ENV.fetch("DATABASE_URL", "") %>

staging: *deploy
```

https://devcenter.heroku.com/articles/concurrency-and-database-connections#bad-connections
https://github.com/thoughtbot/suspenders/pull/234

croaky:
This is looking good to me. I'll leave it to @gylaz or @salbertson to give it a final review, squash, and merge.

config/database.yml
```diff
@@ -2,8 +2,13 @@ development: &default
...
+  pool: <%= ENV.fetch("DB_POOL", 5) %>
+  reaping_frequency: <%= ENV.fetch("DB_REAPING_FREQUENCY", 10) %>
 
 test:
   <<: *default
   database: hound_test
+
+production:
+  pool: <%= [ENV.fetch("MAX_THREADS", 5), ENV.fetch("DB_POOL", 5)].max %>
+  url:  <%= ENV.fetch("DATABASE_URL", "") %>
```
beedub: i'm not sure about this one.. why default to ""?

beedub:
I've pushed commits addressing @croaky's comments

config/database.yml
teoljungberg: What does the heroku guides say?

teoljungberg: Also, could you add a trailing newline?

beedub: > What does the heroku guides say?

I haven't found anything on heroku's guides relate dto this, but I think it'd produce a non-obvious err if `DATABASE_URL` defaulted to `""`. I vote to remove it and fail-fast.

> Also, could you add a trailing newline?

Sure

croaky: +1 on removing it to fail fast.

salbertson:
@beedub other than following Heroku's recommendation, what is the reason to use Puma for Hound?

We don't have file uploading and most of our larger requests come from GitHub directly. Are we worried about large requests from GitHub and Unicorn not handling those efficiently?

croaky:
Heroku's recommendation seems to be partially based on security, or preventing one type of denial of service: https://devcenter.heroku.com/changelog-items/594

Procfile
```diff
@@ -1,3 +1,3 @@
-web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
+web: bundle exec puma -C config/puma.rb
```
gylaz: We've lost port passing here. Should we add it back?

config/environments/production.rb
```diff
@@ -53,9 +53,6 @@
   # config.action_mailer.raise_delivery_errors = false
   config.action_mailer.default_url_options = { :host => 'houndci.com' }
 
-  # Enable threaded mode
-  # config.threadsafe!
-
```
gylaz: Any reason to remove this comment?

Procfile
croaky: It's in `config/puma.rb` as `port ENV.fetch("PORT", 3000)`

config/environments/production.rb
croaky: I think it was removed from Rails. http://tenderlovemaking.com/2012/06/18/removing-config-threadsafe.html

config/puma.rb
```diff
@@ -0,0 +1,17 @@
+workers Integer(ENV.fetch("WEB_CONCURRENCY", 3))
+threads_count = Integer(ENV.fetch("MAX_THREADS", 5))
+threads(threads_count, threads_count)
+
+preload_app!
+
+port ENV.fetch("PORT", 3000)
```
gylaz: Heroku's example calls `rackup DefaultRackup` right before `port`. Do we need that?

config/environments/production.rb
```diff
@@ -67,3 +64,5 @@
   # with SQLite, MySQL, and PostgreSQL)
   # config.active_record.auto_explain_threshold_in_seconds = 0.5
 end
+
+Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
```
gylaz: Should this be larger, maybe 20?

gylaz:
I had a few small suggestions. Looks really good!

config/environments/production.rb
beedub: I used the suspenders default https://github.com/thoughtbot/suspenders/blob/ac3924da64bb2c71f8611919e5f43b85261da600/lib/suspenders/app_builder.rb#L245. I have no strong feelings about which the value of this as long as it's reasonably high.

config/puma.rb
beedub: I removed it because it's the default if one isn't specified https://github.com/puma/puma/blob/v2.12.2/lib/puma/configuration.rb#L63-L65

config/environments/production.rb
beedub: (for the reason @croaky specified)

config/environments/production.rb
beedub: Actually, I've changed it to 20 to remain consistent with https://github.com/thoughtbot/hound/pull/846/files#diff-f579cccc964135c7d644c7b2d3b0d3ecR15, which already specifies 20

beedub:
I've rebased, fixed merge conflicts, and addressed the comments

@gylaz @croaky

gylaz:
Do you mind squashing it down too?

beedub:
Squashed

gylaz:
Thanks for all your work, @beedub!

Merged as 93d5644e55672cf1122d4c8b60f7a2afe9b43214.

PR #848, opened 2015-07-08
Enable trailing comma for thoughtbot

teoljungberg:
https://trello.com/c/aQnzuYg3

### DIFF
--- a/.ruby-style.yml
+++ b/.ruby-style.yml
@@ -424,8 +424,8 @@ Style/TrailingBlankLines:
 Style/TrailingComma:
   Description: Checks for trailing comma in parameter lists and literals.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas
-  Enabled: false
-  EnforcedStyleForMultiline: no_comma
+  Enabled: true
+  EnforcedStyleForMultiline: comma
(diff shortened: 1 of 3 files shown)


thorncp:
LGTM

.ruby-style.yml
```diff
@@ -424,8 +424,8 @@ Style/TrailingBlankLines:
 Style/TrailingComma:
   Description: Checks for trailing comma in parameter lists and literals.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas
-  Enabled: false
-  EnforcedStyleForMultiline: no_comma
+  Enabled: true
```
teoljungberg: Why do we have `.ruby-style.yml` here? Shouldn't this read the thoughtbot styleguide?

gylaz:
:moneybag:!

.ruby-style.yml
teoljungberg: We use this to test things in hound before rolling it out to thoughtbot itself, nevermind

gylaz: We override some styles for Hound -- alpha testing certain settings. It could probably be paired down to just overrides, but I don't have any original context. Maybe check blame?

PR #850, opened 2015-07-10
Include .env in the repo

thorncp:
Moves .sample.env to .env. This file has all the config necessary to get
running in development and `cp .sample.env .env` was in `bin/setup`. If
this file just exists as `.env` in the repo, then the setup step is not
needed and things should Just Work™ out of the box. The dotenv [docs][1]
even recommend doing this to ease new developers onto the project.

Additionally, our heavy use of `ENV.fetch` is now causing the test suite
to [fail][2] on CircleCI, because the database is being setup before the
`pre` hook fires. With `.env` the test suite will work fine without this
setup.

Local customizations can be added to `.env.local`, which is ignored.

[1]: https://github.com/bkeepers/dotenv#should-i-commit-my-env-file
[2]: https://circleci.com/gh/thoughtbot/hound/202

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -6,12 +6,6 @@ bundle install
 # Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
 mkdir -p .git/safe
 
-# Set up development environment config
-if [ ! -f .env ]; then
-  cp .sample.env .env
-  cp .sample.env .env.test
-fi
-
 # Set up database
 bundle exec rake db:setup
 
@@ -28,10 +22,9 @@ if ! grep -qs 'port' .foreman; then
 fi
 
 # Stripe credentials
-if ! grep -qs 'STRIPE' .env; then
-  heroku config --app hound-staging | grep "STRIPE" > .env
-  sed -i '' 's/\:/=/' .env
-  sed -i '' 's/ //g' .env
+if ! grep -qs 'STRIPE' .env.local; then
+  heroku config:get STRIPE_API_KEY --remote staging --shell >> .env.local
+  heroku config:get STRIPE_PUBLISHABLE_KEY --remote staging --shell >> .env.local
 fi
 
 # GitHub
--- a/.env
+++ b/.env
@@ -1,3 +1,4 @@
+# Don't modify this file. Create `.env.local` and put your overrides in there.
 CHANGED_FILES_THRESHOLD=300
 ENABLE_HTTPS=no
 EXEMPT_ORGS=organisation_name,user_name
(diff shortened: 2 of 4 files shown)


teoljungberg:
@thorncp How would `.env.local` be read?

thorncp:
@teoljungberg It's supported by [dotenv-rails](https://github.com/bkeepers/dotenv#multiple-rails-environments)

bin/setup
```diff
@@ -28,10 +22,10 @@ if ! grep -qs 'port' .foreman; then
 fi
 
 # Stripe credentials
-if ! grep -qs 'STRIPE' .env; then
-  heroku config --app hound-staging | grep "STRIPE" > .env
-  sed -i '' 's/\:/=/' .env
-  sed -i '' 's/ //g' .env
+if ! grep -qs 'STRIPE' .env.local; then
+  heroku config --app hound-staging | grep "STRIPE" >> .env.local
```
gabebw: The sed isn't necessary - `heroku config` can output env variables in shell format with `--shell`:

```
heroku config --app hound-staging --shell | grep STRIPE >> .env.local
```

gabebw: Or, even better, since it looks like we only reference `ENV["STRIPE_API_KEY"]` in the app:

```
heroku config:get STRIPE_API_KEY --app hound-staging --shell >> .env.local
```

gabebw: Ah, never mind, there are two STRIPE variables. Maybe be more explicit?

```
heroku config:get STRIPE_API_KEY --app hound-staging --shell >> .env.local
heroku config:get STRIPE_PUBLISABLE_KEY --app hound-staging --shell >> .env.local
```

gylaz:
:+1:

bin/setup
thorncp: Done in 82e08ce10065db05ba6ece8dbd8ce10bffa9ff7d

bin/setup
```diff
@@ -28,10 +22,9 @@ if ! grep -qs 'port' .foreman; then
 fi
 
 # Stripe credentials
-if ! grep -qs 'STRIPE' .env; then
-  heroku config --app hound-staging | grep "STRIPE" > .env
-  sed -i '' 's/\:/=/' .env
-  sed -i '' 's/ //g' .env
+if ! grep -qs 'STRIPE' .env.local; then
+  heroku config:get STRIPE_API_KEY --app hound-staging --shell >> .env.local
```
sikachu: I'd do `--remote staging` instead, as if for some reason we change the app name in the future we could just change line 13-14.

sikachu:
:+1:, but maybe add some comment on the top of `.env` file to say something like:

```
# Don't modify this file. Create `.env.local` and put your overrides in there.
```

bin/setup
thorncp: Good idea, done in d49031282c8333af164668ee3b0d3398639fd0cf

sikachu:
:sheep: :it:

PR #852, opened 2015-07-18
Add Go as a new language supported by Hound.

bernerdschaefer:
This adds basic support to Hound for Go,
by integrating against [hound-go]
via Hound's external review Resque API.

This should be considered an early release feature.
As the review service matures
and the feature-set stabilizes,
we can define supported configuration options
and the necessary documentation
to promote this to an officially supported language.

  [hound-go]: github.com/thoughtbot/hound-go

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -1,7 +1,8 @@
 # Load and parse config files from GitHub repo
 class RepoConfig
   HOUND_CONFIG = ".hound.yml"
-  LANGUAGES = %w(ruby coffeescript javascript scss haml)
+  BETA_LANGUAGES = %w(go)
+  LANGUAGES = %w(ruby coffeescript javascript scss haml go)
   FILE_TYPES = {
     "ruby" => "yaml",
     "javascript" => "json",
@@ -15,8 +16,7 @@ class RepoConfig
   pattr_initialize :commit
 
   def enabled_for?(language)
-    options = options_for(language)
-    options.nil? || !disabled?(language)
+    !disabled?(language)
   end
 
   def for(language)
@@ -55,8 +55,18 @@ class RepoConfig
 
   private
 
+  def beta?(language)
+    BETA_LANGUAGES.include?(language)
+  end
+
+  def defualt_options_for(language)
+    { "enabled" => !beta?(language) }
+  end
+
   def options_for(language)
-    hound_config[language] || hound_config[language_camelize(language)]
+    hound_config[language] ||
+      hound_config[language_camelize(language)] ||
+      defualt_options_for(language)
   end
 
   def disabled?(language)
--- a/app/models/style_guide/go.rb
+++ b/app/models/style_guide/go.rb
@@ -1,0 +1,32 @@
+module StyleGuide
+  class Go < Base
+    LANGUAGE = "go"
+
+    def file_review(commit_file)
+      Resque.enqueue(
+        GoReviewJob,
+        filename: commit_file.filename,
+        commit_sha: commit_file.sha,
+        pull_request_number: commit_file.pull_request_number,
+        patch: commit_file.patch,
+        content: commit_file.content,
+        config: repo_config.raw_for(LANGUAGE),
+      )
+
+      FileReview.new(filename: commit_file.filename)
+    end
+
+    def file_included?(commit_file)
+      !vendored?(commit_file.filename)
+    end
+
+    private
+
+    def vendored?(filename)
+      path_components = Pathname(filename).each_filename
+
+      path_components.include?("vendor") ||
+        path_components.take(2) == ["Godeps", "_workspace"]
+    end
+  end
+end
--- a/spec/models/repo_config_spec.rb
+++ b/spec/models/repo_config_spec.rb
@@ -6,20 +6,31 @@ require "app/models/repo_config"
 describe RepoConfig do
   describe "#enabled_for?" do
     context "with invalid config" do
-      it "returns true for all languages" do
+      it "returns true for all default languages" do
         commit = double("Commit", file_content: <<-EOS.strip_heredoc)
           hello world!
         EOS
         repo_config = RepoConfig.new(commit)
 
-        RepoConfig::LANGUAGES.each do |language|
+        default_languages.each do |language|
           expect(repo_config).to be_enabled_for(language)
+        end
+      end
+
+      it "returns false for all beta languages" do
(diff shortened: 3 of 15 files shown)


bernerdschaefer:
I copied what existed for SCSS support. Can you think of any reason this might not "just work"?

gylaz:
This looks good.

teoljungberg:
Love seeing the new workers moving forward, fix the violations and :ship:it

teoljungberg:
As for a release-plan, maybe we should de-activate it per default so we can battle-test it. Thoughts?

thorncp:
Need to update the [StyleChecker](https://github.com/thoughtbot/hound/blob/1cc5f1bcf21227136f7abbef2643ab6f2988559c/app/models/style_checker.rb#L35) so it knows about .go files.

bernerdschaefer:
@teoljungberg Agreed on deactivated by default -- where should I look to do that?

thorncp:
I think historically we've disabled new linters at the [linter config level](https://github.com/thoughtbot/hound/commit/7c7d44ada5113bfd74d18f4c811e174ade4c6084). Since there's no config for golint (I think?) and the [default with no config](https://github.com/thoughtbot/hound/blob/1cc5f1bcf21227136f7abbef2643ab6f2988559c/app/models/repo_config.rb#L19) is to enable the linter, it seems like we'd have to make some tweaks to allow for an opt-in approach.

spec/models/style_guide/go_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      patch_position: 2
+    )
+    double(
+      "CommitFile",
+      content: "codes",
+      filename: "lib/a.go",
+      line_at: line,
+      sha: "abc123",
+      patch: "patch",
+      pull_request_number: 123
```
salbertson: I've seen this commented on in a couple PRs. Should we disable this check?

spec/models/style_guide/go_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    it "returns true" do
+      style_guide = build_style_guide
+
+      expect(style_guide.file_included?(double)).to eq true
+    end
+  end
+
+  private
+
+  def build_style_guide(config = "config")
```
salbertson: What do you think about moving these methods to a spec "helper" so they can be reused?

spec/models/style_guide/go_spec.rb
gylaz: I don't think we should disable it. It's in our guides. Like everything else, we need to get better at not triggering this violation.

salbertson:
@bernerdschaefer since there's no config I think what @thorncp suggested is what we should do; take a look at `app/models/style_checker.rb` since that's where we check if a language is enabled.

Having a mechanism to turn languages off by default for early release is something we will continue to do, so I think we should solve that now.

salbertson:
@bernerdschaefer What do you think about updating the [docs](https://github.com/thoughtbot/hound/blob/6ee13e926723dce0bf7814dc1c124ac3e242bfe4/app/views/pages/configuration.haml) and [security page](https://github.com/thoughtbot/hound/blob/6ee13e926723dce0bf7814dc1c124ac3e242bfe4/doc/SECURITY.md)?

spec/models/style_guide/go_spec.rb
bernerdschaefer: I couldn't find a good way of reusing this one in particular, but cc194ed20d87d1aae91f06ec469998ad29c0b6a8 shares a build_commit_file helper method.

bernerdschaefer:
I've updated based on some of the feedback:

* df1f18d94e97cdcba952ecfda0b39b32b939678d adds a `BETA_LANGUAGES`, whose purpose is currently to disable linting by default.

* 862aafb1a4b86021b262b78d81ef058cb2034269 ignores vendored Go files

* 457bf2e adds some basic information to the configuration and security docs for Go.

spec/models/style_guide/go_spec.rb
salbertson: :+1:

app/models/repo_config.rb
```diff
@@ -1,7 +1,8 @@
 # Load and parse config files from GitHub repo
 class RepoConfig
   HOUND_CONFIG = ".hound.yml"
-  LANGUAGES = %w(ruby coffeescript javascript scss haml)
+  BETA_LANGUAGES = %w(go)
```
salbertson: Awesome, this is going to be helpful.

app/models/style_guide/go.rb
```diff
@@ -0,0 +1,32 @@
...
+        content: commit_file.content,
+        config: repo_config.raw_for(LANGUAGE),
+      )
+
+      FileReview.new(filename: commit_file.filename)
+    end
+
+    def file_included?(commit_file)
+      !vendored?(commit_file.filename)
+    end
```
salbertson: What do you think about moving this logic to the service? It would be nice if Hound didn't know about vendored Go files.

The `file_included?` method was used for style review in Hound before services were extracted. It should just return true for service based review.

salbertson:
@bernerdschaefer one question about excluding files, but looks good to me. :+1: 

Thanks for adding support for beta languages!

app/models/style_guide/go.rb
bernerdschaefer: I agree that it would be nice if Hound didn't know about Go's vendoring policy.

That said, it also would be nice not to fetch content and enqueue jobs for code which should never be checked for style. From checking w/ @gylaz, hooking into `file_included?` means we can exclude these vendored files before fetching their content.

In the end, I'm okay going either way.

croaky:
Excited to see this go in. Seems like today's the rollout day?

app/models/style_guide/go.rb
salbertson: Oh, right, fetching file content is an issue. This is fine with me for now.

spec/models/repo_config_spec.rb
```diff
@@ -6,38 +6,31 @@
 describe RepoConfig do
   describe "#enabled_for?" do
     context "with invalid config" do
-      it "returns true for all languages" do
+      it "returns true for all official languages" do
```
gylaz: What do you think about `default` or `live` instead of `official`?

bernerdschaefer: Done.

PR #754, opened 2015-04-11
[wip] Add python linting.

jmcarp:
Work in progress. Begins support for python linting with flake8.

[addresses #463]

PR #853, opened 2015-07-19
Add Python linting via hound-python service.

jmcarp:
Add support for Python linting via https://github.com/jmcarp/hound-python. Replaces #754.

### DIFF
--- a/app/models/style_checker.rb
+++ b/app/models/style_checker.rb
@@ -46,6 +46,8 @@ class StyleChecker
       StyleGuide::Scss
     when /.+\.go\z/
       StyleGuide::Go
+    when /.+\.py\z/
+      StyleGuide::Python
     else
       StyleGuide::Unsupported
     end
--- a/app/models/style_guide/python.rb
+++ b/app/models/style_guide/python.rb
@@ -1,0 +1,28 @@
+module StyleGuide
+  class Python < Base
+    LANGUAGE = "python"
+
+    def file_review(commit_file)
+      Resque.push(
+        "python_review",
+        {
+          class: "review.PythonReviewJob",
+          args: [
+            commit_file.filename,
+            commit_file.sha,
+            commit_file.pull_request_number,
+            commit_file.patch,
+            commit_file.content,
+            repo_config.raw_for(LANGUAGE),
+          ],
+        }
+      )
+
+      FileReview.new(filename: commit_file.filename)
+    end
+
+    def file_included?(_)
+      true
+    end
+  end
+end
--- a/spec/models/style_checker_spec.rb
+++ b/spec/models/style_checker_spec.rb
@@ -167,6 +167,17 @@ describe StyleChecker, "#file_reviews" do
     end
   end
 
+  context "for a Python file" do
+    it "does not immediately return violations" do
+      commit_file = stub_commit_file("test.py", "import this")
+      pull_request = stub_pull_request(commit_files: [commit_file])
+
+      violation_messages = pull_request_violations(pull_request)
+
+      expect(violation_messages).to be_empty
+    end
+  end
+
   context "for a Haml file" do
     context "with style violations" do
       it "returns no violations" do
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -1,8 +1,8 @@
 # Load and parse config files from GitHub repo
 class RepoConfig
   HOUND_CONFIG = ".hound.yml"
-  BETA_LANGUAGES = %w(go)
-  LANGUAGES = %w(ruby coffeescript javascript scss haml go)
+  BETA_LANGUAGES = %w(go python)
+  LANGUAGES = %w(ruby coffeescript javascript scss haml go python)
   FILE_TYPES = {
     "ruby" => "yaml",
     "javascript" => "json",
--- a/spec/models/repo_config_spec.rb
+++ b/spec/models/repo_config_spec.rb
@@ -42,6 +42,8 @@ describe RepoConfig do
             enabled: true
           go:
             enabled: true
+          python:
+            enabled: true
         EOS
         repo_config = RepoConfig.new(commit)
 
@@ -122,6 +124,8 @@ describe RepoConfig do
             SCSS:
               Enabled: true
             Go:
+              Enabled: true
+            Python:
               Enabled: true
           EOS
           repo_config = RepoConfig.new(commit)
--- a/spec/models/style_guide/python_spec.rb
+++ b/spec/models/style_guide/python_spec.rb
@@ -1,0 +1,71 @@
+require "rails_helper"
+
+describe StyleGuide::Python do
+  describe "#file_review" do
+    it "returns an incompleted file review" do
+      style_guide = build_style_guide
+      commit_file = build_commit_file
+
+      result = style_guide.file_review(commit_file)
+
+      expect(result).not_to be_completed
+    end
+
+    it "schedules a review job" do
+      allow(Resque).to receive(:push)
+      style_guide = build_style_guide("config")
+      commit_file = build_commit_file
+
+      style_guide.file_review(commit_file)
+
+      expect(Resque).to have_received(:push).with(
+        "python_review",
+        {
+          class: "review.PythonReviewJob",
+          args: [
+            commit_file.filename,
+            commit_file.sha,
+            commit_file.pull_request_number,
+            commit_file.patch,
+            commit_file.content,
+            "config",
+          ],
+        }
+      )
+    end
+  end
+
+  describe "#file_included?" do
+    it "returns true" do
+      style_guide = build_style_guide
+
+      expect(style_guide.file_included?(double)).to eq true
+    end
+  end
+
+  private
+
+  def build_style_guide(config = "config")
+    repo_config = double("RepoConfig", raw_for: config)
+    StyleGuide::Python.new(repo_config, "ralph")
+  end
+
+  def build_commit_file
+    line = double(
+      "Line",
+      changed?: true,
+      content: "blah",
+      number: 1,
+      patch_position: 2,
+    )
+    double(
+      "CommitFile",
+      content: "codes",
+      filename: "lib/a.py",
+      line_at: line,
+      sha: "abc123",
+      patch: "patch",
+      pull_request_number: 123,
+    )
+  end
+end


app/models/style_checker.rb
```diff
@@ -44,6 +44,8 @@ def style_guide_class(filename)
       StyleGuide::Haml
     when /.+\.scss\z/
       StyleGuide::Scss
+    when /.+\.py\z/
```
teoljungberg: Are we missing a test for this?

teoljungberg: We don't, nevermind

teoljungberg:
This PR looks good, though I have some ideas for the linter itself.

One of the goals we had for this new achitechture was to write the linters in their own language. Like done in https://github.com/thoughtbot/hound-go

Could we do that for your python linter aswell? I'm not a python developer myself, so maybe we can get someone else from our side to review the code? @mike-burns have talked about implementing the same linter and should be able to help you.

Any other ideas @thoughtbot/hound ?

jmcarp:
Writing the go linter in go is straightforward because there's already a resque client written in go. I looked around but haven't found a python resque worker implementation--probably because most python devs use celery or rq. I could write the linter in python, but would have to implement the interface to resque. Which wouldn't be a whole lot of work, but is definitely more complicated than shelling out to `flake8`, like I'm doing now.

Is that more like what you had in mind @teoljungberg?

jmcarp:
Here's what a python implementation of the python linter might look like. I'd be happy to flesh this out it seems like the right direction for the task.
```python
# resque.py

import time
import collections

def _get_key(queue):
	return 'resque:queue:{0}'.format(queue)

class Resque(object):

	def __init__(self, conn, interval=5):
		self.conn = conn
		self.interval = interval
		self._registry = collections.defaultdict(dict)

	def work(self):
		while True:
			for queue in self._registry:
				key = _get_key(queue)
				raw = self.conn.lpop(key)
				if raw:
					body = json.loads(task)
					task = self._registry[body['class']]
					try:
						task(*body['args'])
					except Exception as error:
						pass
			time.sleep(self.interval)

	def enqueue(self, queue, klass, *args):
		key = _get_key(queue)
		value = {
			'class': klass,
			'args': args,
		}
		self.conn.rpush(key, json.dumps(value))

	def task(self, queue, klass):
		def wrapper(func):
			self._registry[queue][klass] = func
			return func
		return wrapper

# tasks.py

import pep8
import flake8.main as flake8

import redis

import resque

conn = redis.StrictRedis()
resque = resque.Resque(conn)

@resque.task('python_review', 'PythonReviewJob')
def python_review(filename, commit_sha, pull_request_number, patch, content, config):
	violations = [
		{
			'line': each[0],
			'message': each[3]
		}
		for each in check_code(content)
	]
	resque.enqueue(
		'high',
		'CompletedFileReviewJob',
		filename,
		commit_sha,
		pull_request_number,
		patch,
		violations,
	)

def check_code(code, **kwargs):
    flake8_style = flake8.get_style_guide(**kwargs)
    flake8_style.options.report = QuietReport(flake8_style.options)
    return flake8_style.input_file(None, lines=code.splitlines(True))

class QuietReport(pep8.StandardReport):
	def get_file_results(self):
		return sorted(self._deferred_print)

if __name__ == '__main__':
	resque.work()
```

jmcarp:
Actually, I spoke too soon. There *is* a python clone of resque called pyres. It doesn't appear to be actively maintained, and there are a few differences from the original resque implementation, but it looks close enough that it might work here. I'll take a crack at rewriting the python linter with pyres and see how it goes.

mike-burns:
@teoljungberg this PR is for making use of an existing external service, like what we do with SCSS. I have not been following Hound closely, but isn't this what we want?

teoljungberg:
@mike-burns it is, but we would prefer having the linters written in their own language. So a JS linter would be written in JS, a Go linter would be written in Golang, and a Python linter would be written in Python.

If it's impossible to write the linter in it's own language, like for SCSS(if someone writes a Resque implementation for SCSS, I'm impressed), we would fallback to Ruby.

salbertson:
@jmcarp what do you think about disabling Python by default until we know it's stable in production?

I did this recently with SCSS review by disabling all rules in the config. Is there something similar we could do in https://github.com/jmcarp/hound-python?

salbertson:
@jmcarp I'm totally fine with the service being written mostly in Ruby.

jmcarp:
Now that I've thought about it a bit, it makes sense to write linters in their own languages. Plus, I'm better with Python than Ruby. I put together a Python linter in Python at https://github.com/jmcarp/hound-python (I still have the old code if you wind up preferring Ruby). I'll document the linter this week, but for now, it's working in development, and tests are passing.

@salbertson I'll add a default config with all rules disabled.

jmcarp:
The Python linter has been updated with basic documentation, and now ignores all files by default. To enable, users would pass a flake8 ini-style configuration file like this:
```
[flake8]
exclude=
```
If anybody has time to look over the Python linter, I'd be happy to revise as necessary.

.env
```diff
@@ -16,3 +16,8 @@ SECRET_KEY_BASE=some_really_really_long_super_secret_key_base
 SPLIT_ADMIN_PASSWORD=split_password
 STRIPE_API_KEY=sk_test_123
 STRIPE_PUBLISHABLE_KEY=pk_test_123
+GITHUB_CLIENT_ID=43d612c5c4251c68574e
+GITHUB_CLIENT_SECRET=d163570955e4c95012b796911d7abb4729b32dc7
+HOST=jmc-hound.ngrok.com
+HOUND_GITHUB_TOKEN=f0c80f95dd2d4770b8a54e8d8ab3ab5ee3f0b96f
+HOUND_GITHUB_USERNAME=jmcarp
```
mike-burns: Can you change these values to one that look more obviously like dummy data?

mike-burns:
I'm glad this is a light wrapper around flake8. I've no concrete feedback on the Python code. CC @thoughtbot/python .

.env
jmcarp: Sorry, reverted.

salbertson:
> but we would prefer having the linters written in their own language

@teoljungberg I don't think I agree. Why would we prefer having the linters written in their own language?

I would rather have the style services written in Ruby, if that's a realistic option, since the services will most likely be maintained by thoughtbot and Ruby is the language most of us are more familiar with.

jmcarp:
Having written hound-python in both Ruby and Python, I have a slight preference for the native approach. Most linters will be implemented in the language they're meant to check, so writing all the services in Ruby means managing cross-language dependencies and talking to linters via subprocess. Which means writing potentially brittle code to prepare arguments for subprocess, parse output strings, handle errors, etc.

Regarding maintenance, linters should be trivial--the Python implementation of hound-python is something like 50 lines, plus tests and some utility code. So writing native linters wouldn't saddle you with maintaining large or complicated non-Ruby codebases. Plus, I would guess that you'd get more help from the community on maintenance with native linters--lots of folks in the Python community would help fix bugs in a linter written in Python, but not so many would be able to help with a Ruby implementation.

All that said, feel free to use the Ruby implementation of the Python linter if you prefer. It's at https://github.com/jmcarp/hound-python-ruby.

salbertson:
Did you run into many issues with brittle code while writing the Ruby version?

jmcarp:
Some. The code is so simple that I'm not too concerned, but here's a few examples of what I mean:
```ruby
def lint(content, config = {})
  opts = config.to_a.map { |key, value| "--#{key} #{value}" }.join(" ")
  report = `echo "#{content}" | flake8 - #{opts}`
  report.split("\n").map do |line|
    parse_flake8_line(line)
  end
end

def parse_flake8_line(line)
  parts = line.split(":")
  {
    line: parts[1].to_i,
    message: parts.drop(3).join(":").strip
  }
end
```
This passes all the options from the config file to the `flake8` executable, which has to be on the path. If the config file has invalid options, the subprocess call will fail. Then we have to parse the output of `flake8` as a string and munge it into line number and message. If for some reason the delimiting characters or position of different components change in `flake8`, we'll munge the errors incorrectly. Neither if these issues comes up with a native implementation. Anyway, now that we're talking about it, we would need to add some error handling to the Ruby implementation if we go in that direction.

croaky:
> Why would we prefer having the linters written in their own language?

I'm in favor of preferring native, if only for the community reason:

> get more help from the community on maintenance with native linters--lots of folks in the Python community would help fix bugs in a linter written in Python, but not so many would be able to help with a Ruby implementation.

Our job is to keep the linter API simple and well-defined so running these on Heroku and accessing the Redis queue is straightforward to operationalize, but make each linter as comfortable as possible for its language community.

salbertson:
@jmcarp ok, sounds like Python based service is a better choice. Would you mind using double quotes and then squashing your commits. This looks good to me. :+1:

app/models/style_guide/python.rb
```diff
@@ -0,0 +1,26 @@
+module StyleGuide
+  class Python < Base
+    LANGUAGE = "python"
+
+    def file_review(commit_file)
+      Resque.push(
+        'python_review',
+        'class' => 'review.PythonReviewJob',
+        'args' => [
```
gylaz: Mind changing this to ruby 1.9 style hashes?

spec/models/style_checker_spec.rb
```diff
@@ -167,6 +167,17 @@
     end
   end
 
+  context "for a Python file" do
+    it "does not immediately return violations" do
+      commit_file = stub_commit_file("test.py", "* { color: red; }")
```
gylaz: How about updating the content string to be something Python?

spec/models/style_checker_spec.rb
```diff
@@ -167,6 +167,17 @@
...
+      pull_request = stub_pull_request(commit_files: [commit_file])
+
+      violation_messages = pull_request_violations(pull_request)
+
+      expect(violation_messages).to be_empty
+    end
+  end
```
gylaz: I wonder what the value of this test (and the one it was based on) is? Maybe a better test would be to check that the associated `file_review` is pending. I'm also okay if we handle this in a different PR.

jmcarp:
Commits squashed, merge conflicts resolved.

salbertson:
Merged! Thanks @jmcarp for pushing this along!

salbertson:
@jmcarp Would you mind transferring the https://github.com/jmcarp/hound-python repo to the thoughtbot org and I'll make you a contributor?

salbertson:
@jmcarp, what do you think about transferring the service repo?

thorncp:
@jmcarp Any updates on transferring the hound-python repo to the thoughtbot org? We'd like to have some control over it since we'll be running the service and maintaining it. As mentioned previously, we'll happily add you as a collaborator.

jmcarp:
@salbertson @thorncp sorry, didn't see this. Sure, I'd be happy to transfer the repo. But it looks like I can't because I don't have admin rights to the organization. Does GitHub offer a way around this?

teoljungberg:
@jmcarp If you add either one of us on @thoughtbot/hound (@salbertson, @thorncp, @gylaz, or myself) - we can transfer it to the thoughtbot organization

gylaz:
@jakecraige Any progress on this? We'd like to try and get this all hooked up.

croaky:
Does the repo need to be transferred or would it suffice to clone and push to the new org? Benefit being we'd remove a blocker to getting this out.

gylaz:
> Does the repo need to be transferred or would it suffice to clone and push to the new org? Benefit being we'd remove a blocker to getting this out.

I assume you mean the proposed `houndci` org. This is certainly an alternative (or forking), if @jmcarp is okay with that rather than transferring.

thorncp:
I think the current blocker for us is that the repo doesn't have a license attached to it, so it defaults to @jmcarp having full copyrights and we can't fork it.

croaky:
If the blocker is legal stuff, can we provide some suggested text for the LICENSE file (or other docs) as a GitHub comment?

jmcarp:
Added @teoljungberg as a contributor to hound-python.

thorncp:
Awesome, thanks @jmcarp !

mike-burns:
@jmcarp https://github.com/thoughtbot/hound-python/blob/master/review.py#L9 - can you explain this TODO a bit more, especially so that we know when we should take action?

PR #855, opened 2015-07-23
Add option to fail the build

thorncp:
Adds a configuration option to set the PR status to 'error' when
violations are found.

*.hound.yml*
```yaml
fail_on_violations: true
```

more_dumb_stuff_with_custom_config_inheriting_by_thorncp_ _pull_request__59_ _thornco_dumb-ruby

https://trello.com/c/MtnDjmfF

### DIFF
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -10,11 +10,16 @@ class CommitStatus
   end
 
   def set_success(violation_count)
-    message = I18n.t(:success_status, count: violation_count)
+    message = I18n.t(:complete_status, count: violation_count)
     github.create_success_status(repo_name, sha, message)
   end
 
-  def set_failure
+  def set_failure(violation_count)
+    message = I18n.t(:complete_status, count: violation_count)
+    github.create_error_status(repo_name, sha, message)
+  end
+
+  def set_config_error
     message = I18n.t(:config_error_status)
     github.create_error_status(repo_name, sha, message, configuration_url)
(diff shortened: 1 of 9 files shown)


app/models/commit_status.rb
```diff
@@ -14,7 +14,12 @@ def set_success(violation_count)
     github.create_success_status(repo_name, sha, message)
   end
 
-  def set_failure
+  def set_failure(violation_count)
+    message = I18n.t(:success_status, count: violation_count)
```
gylaz: Should we rename `success_status` to something more general?

thorncp:
I just realized this doesn't check the violation count, so it'll always set the error status if it's enabled.

salbertson:
Why are we making this an option?

thorncp:
People have asked for it. I think it'd be an interesting experiment to try, even if it doesn't stick around.

https://trello.com/c/MtnDjmfF

thorncp:
Updated to not fail with 0 violations in fb0c4836a3567f74718d093c78a9b695fc8c2a0d.

app/models/commit_status.rb
thorncp: Changed to `complete_status` in 917d100409fe3514d7dd13d03b873d3e9d1415b9

thorncp:
Enabled this on the Hound main repo in 87543b7810d80e389fef0752c411c0f2f425f258

PR #856, opened 2015-07-24
Ensure checkout modal can be displayed

malandrina:
* Users without a credit card on file are currently unable to activate
  paid repos when using Safari on an iOS device: When the user attempts
  to activate a private repo in her repo list by clicking the "Activate"
  button, the Stripe Checkout popup does not appear and the user is
  unable to proceed with the transaction.

  This is because the trigger for displaying the popup is currently
  invoked in a callback. Per Stripe's documentation, this implementation
  causes "mobile devices and some version of Internet Explorer [to]
  block the popup and prevent users from checking out". Reference:
  https://stripe.com/docs/checkout#integration-more-runloop

* This commit changes the `Repo` and `RepoList` scripts to load user
  payment information earlier in the run loop, thereby allowing us to
  decouple fetching the current user and checking for an existing card
  from showing the popup.

https://trello.com/c/r2iCkXVA/617-stripe-credit-card-modal-fails-to-display-in-safari-on-iphones

### DIFF
--- a/app/assets/javascripts/directives/repo_list.js.coffee.erb
+++ b/app/assets/javascripts/directives/repo_list.js.coffee.erb
@@ -13,6 +13,12 @@ App.directive 'repoList', ['Repo', 'Sync', 'User', '$timeout', (Repo, Sync, User
         alert('Your repos failed to load.')
       )
 
+    loadUserPaymentInfo = ->
+      user = User.get()
+      user.$promise
+        .then (result) -> scope.userCardExists = result.card_exists
+        .catch -> alert("Failed to load current user.")
+
     pollSyncStatus = ->
       successfulSync = (user) ->
         if user.refreshing_repos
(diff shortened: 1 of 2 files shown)


thorncp:
LGTM

app/assets/javascripts/directives/repo_list.js.coffee.erb
```diff
@@ -13,6 +13,13 @@ App.directive 'repoList', ['Repo', 'Sync', 'User', '$timeout', (Repo, Sync, User
         alert('Your repos failed to load.')
       )
 
+    loadUserPaymentInfo = ->
+      user = User.get()
+      user.$promise.then((result) ->
+        scope.userCardExists = result.card_exists
+      , -> alert("Failed to load current user.")
```
gylaz: How about using `catch` instead of failure callback? The benefit is, that it always fires, as opposed to if there is an error in the failure callback, it can fail.
```coffee
User.get().user.$promise
  .then (result) -> scope.userCardExists = result.card_exists
  .catch -> alert("Failed to load current user.")
```

gylaz:
One comment, :+1:

malandrina:
Thanks @thorncp and @gylaz !

app/assets/javascripts/directives/repo_list.js.coffee.erb
malandrina: Good call. Changed in https://github.com/thoughtbot/hound/commit/6e2e28316db8524268024e5a9e676f250fe12b8b.

PR #857, opened 2015-07-24
Consider Haml review as beta

salbertson:
* Now that we have `BETA_LANGUAGES`, thanks @bernerdschaefer, we don't have to disable all rules in config to disable Haml review.
* If you want Haml review you will need to enable it in `.hound.yml`.


gylaz:
:+1:

thorncp:
:+1:

gylaz:
I want to point out that HAML linter seems to be working well after the gem upgrade.

PR #858, opened 2015-07-24
Upgrade Haml linter

gylaz:
We are quite a few verions behind.

The linter has had a few issues,
and there's a good possibility that they have been fixed.

Also, improved code/tests around excluding files.


thorncp:
LGTM

PR #859, opened 2015-07-24
Include filename in config parse error

thorncp:
Adds a `filename` field to the `RepoConfig::ParserError` exception class
and uses it when setting the GitHub status.

https://trello.com/c/vCW2P0eX

more_dumb_stuff_with_custom_config_inheriting_by_thorncp_ _pull_request__62_ _thornco_dumb-ruby

### DIFF
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -19,8 +19,8 @@ class CommitStatus
     github.create_error_status(repo_name, sha, message)
   end
 
-  def set_config_error
-    message = I18n.t(:config_error_status)
+  def set_config_error(filename)
+    message = I18n.t(:config_error_status, filename: filename)
     github.create_error_status(repo_name, sha, message, configuration_url)
   end
 
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -11,8 +11,6 @@ class RepoConfig
     "haml" => "yaml"
   }
 
-  class ParserError < StandardError; end
-
   pattr_initialize :commit
 
   def enabled_for?(language)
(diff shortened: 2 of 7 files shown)


app/models/commit_status.rb
```diff
@@ -19,8 +19,8 @@ def set_failure(violation_count)
     github.create_error_status(repo_name, sha, message)
   end
 
-  def set_config_error
-    message = I18n.t(:config_error_status)
+  def set_config_error(exception)
```
gylaz: What do you think about passing the filename instead of exception?

app/models/repo_config.rb
```diff
@@ -11,7 +11,14 @@ class RepoConfig
...
 
-  class ParserError < StandardError; end
+  class ParserError < StandardError
+    attr_reader :filename
+
+    def initialize(message, filename:)
+      super(message)
+      @filename = filename
+    end
+  end
```
gylaz: Time to move this to own file, or not?

gylaz:
Couple of questions, LGTM.

app/models/repo_config.rb
thorncp: I had that thought, but didn't act on it. Sounds good to me, but not real sure where to put it. `app/models`, `app/exceptions`, `app/models/exceptions`? It's also currently nested under `RepoConfig`. Should it stay that way (`app/models/repo_config/parser_error.rb`)?

app/models/commit_status.rb
thorncp: Good idea, done in a82d54193f194d6ffb5bb99a107f56989f60a746

app/models/repo_config.rb
gylaz: I think that last one makes the most sense since it's namespaced.

thorncp: Done ad3cf78e36ad9995dc628f384b22e7b629123f9b

gylaz:
:shipit:

PR #860, opened 2015-07-25
Store bulk customer data in the database

gylaz:
This makes it easier to add and manage new bulk customers.

We can build on this to make it even simpler.

### DIFF
--- a/app/models/bulk_customer.rb
+++ b/app/models/bulk_customer.rb
@@ -1,0 +1,21 @@
+class BulkCustomer < ActiveRecord::Base
+  def self.unpaid_repos
+    bulk_orgs = pluck(:org)
+    Repo.
+      includes(:subscription).
+      where(subscriptions: { repo_id: nil }).
+      where(private: true).
+      where("full_github_name !~* ?", "(#{bulk_orgs.join('|')})").
+      active
+  end
+
+  def update_current_repos
+    repo_count = Repo.
+      where("full_github_name ILIKE ?", "#{org}/%").
+      where(private: true).
+      active.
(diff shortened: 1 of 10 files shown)


app/models/repo.rb
```diff
@@ -59,8 +59,9 @@ def stripe_subscription_id
     end
   end
 
-  def exempt?
-    ENV["EXEMPT_ORGS"] && ENV["EXEMPT_ORGS"].split(",").include?(organization)
+  def bulk?
+    BulkCustomer.where(org: organization).any? ||
+      ENV.fetch("EXEMPT_ORGS", "").split(",").include?(organization)
```
gylaz: We can remove this once we move data from `EXEMPT_ORGS` to db.

app/models/bulk_customer.rb
```diff
@@ -0,0 +1,21 @@
+class BulkCustomer < ActiveRecord::Base
+  def self.unpaid
```
thorncp: What do you think about naming this `unpaid_repos` to be explicit that it returns `Repo` instances?

db/migrate/20150725013530_create_bulk_customers.rb
```diff
@@ -0,0 +1,14 @@
+class CreateBulkCustomers < ActiveRecord::Migration
+  def change
+    create_table :bulk_customers do |t|
+      t.timestamps null: false
+
+      t.string :org, index: { unique: true }
```
thorncp: `null: false` ?

thorncp:
LGTM

PR #862, opened 2015-07-29
Use separate resque processes for high and medium/low queues

salbertson:
This should prevent a busy queue from blocking medium and low queue items.

### DIFF
--- a/.env
+++ b/.env
@@ -8,7 +8,6 @@ HOST=your-initials-hound.ngrok.com
 HOUND_GITHUB_TOKEN=your_github_token
 HOUND_GITHUB_USERNAME=your_github_username
 MAX_COMMENTS=10
-QUEUE=high,medium,low
 REDIS_URL=redis://localhost:6379
 RESQUE_ADMIN_PASSWORD=password
 RESQUE_JOB_TIMEOUT=120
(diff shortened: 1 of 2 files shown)


gylaz:
I would rather just use one queue name than add more worker types. I don't think we have a good reason to use differently named queues, and it's just one more thing to manage in Heroku.

salbertson:
We introduced priority to prevent large build jobs from blocking other jobs like repo refreshes, org invites and small builds. Are you suggesting we lump all jobs together?

thorncp:
When we introduced the SOA, we put the `CompletedFileReview` as high priority. Since we only have one procfile entry for all the queues, the works will focus on finalizing reviews, and will not spend time reviewing files. This is effectively turning our asynchronous architecture into a synchronous one, because the workers review a chunk of builds, then finalize a chunk of builds. If we split out workers for each queue, we can always be working instead of getting clogged up:

resque_

thorncp:
What do you think about a config like this?

```
resque_high: ... QUEUE=high,medium ...
resque_medium: ... QUEUE=medium ...
resque_low: ... QUEUE=low,medium ...
```

This would allow us to always be working off jobs on each queue and the `high` and `low` workers can handle `medium` jobs if there's no pending `high` or `low` jobs.

PR #864, opened 2015-07-31
Only fetch DATABASE_URL in production env

gylaz:
Why:

* `DATABASE_URL` only needs to be defined in production environment
* Raises an error if not set in dev/test environments


thorncp:
:+1:

PR #865, opened 2015-07-31
Remove repo_limit null constraint on bulk_customers

gylaz:
Why:

* Some customers, like thoughtbot, may not have a limit.


thorncp:
LGTM

PR #866, opened 2015-08-01
Ensure build exists before scheduling reviews

gylaz:
Why:

* There is currently a problem where reviews have been schedules
  but the assosiated build cannot be found.
* Possibly the build creation failed for some reason,
  but, as is, the review jobs get scheduled anyways.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -30,6 +30,7 @@ class BuildRunner
     commit_status.set_pending
     upsert_owner
     build = create_build
+    review_files(build)
     BuildReport.run(pull_request: pull_request, build: build, token: token)
   end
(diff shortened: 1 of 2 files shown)


thorncp:
LGTM

Do you think a test would be beneficial?

gylaz:
I do think a test would be beneficial, but I couldn't think of an easy way to test the situation we encountered. I think it's important to plug the hole, then think of a way to test the regression.

thorncp:
How about stubbing out the build create to raise, then asserting no file reviews were created?

gylaz:
@thorncp I pushed up a test.

thorncp:
:+1:

PR #867, opened 2015-08-01
Wrap review related jobs in ApplicationJob

gylaz:
Why:

* `ApplicationJob` has logic to retry then give up
* Avoid re-scheduling jobs infinitely

### DIFF
--- a/app/jobs/completed_file_review_job.rb
+++ b/app/jobs/completed_file_review_job.rb
@@ -1,14 +1,14 @@
-class CompletedFileReviewJob
-  @queue = :high
+class CompletedFileReviewJob < ApplicationJob
+  queue_as :high
 
-  def self.perform(attributes)
-    # filename
-    # commit_sha
-    # pull_request_number
-    # patch
-    # violations
-    #   [{ line: 123, message: "WAT" }]
-
+  # attributes is a hash with the following required keys:
+  # - filename
+  # - commit_sha
+  # - pull_request_number
+  # - patch
+  # - violations (an array)
+  #   Example: [{ line: 123, message: "WAT" }]
+  def perform(attributes)
     build = Build.find_by!(
       pull_request_number: attributes.fetch("pull_request_number"),
       commit_sha: attributes.fetch("commit_sha")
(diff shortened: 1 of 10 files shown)


app/jobs/completed_file_review_job.rb
```diff
@@ -40,9 +40,5 @@ def self.perform(attributes)
       build: build,
       token: build.user_token,
     )
-  rescue ActiveRecord::RecordNotFound
-    Resque.enqueue_in(30, self, attributes)
-  rescue Resque::TermException
```
thorncp: Can we get rid of this? `Resque::TermException` [inherits from `SignalException`](https://github.com/resque/resque/blob/a3845fe4a5bcac0e6107be6f0d494b72c958c3c0/lib/resque/errors.rb#L12) which is *not* a `StandardError`, so `Retryable` [won't re-enqueue](https://github.com/thoughtbot/hound/blob/2dfbfba24c4914767b7c93f71c64609e5b849abd/app/jobs/retryable.rb#L7) it.

Perhaps `Retryable` should also rescue `Resque::TermException`?

gylaz: Looks like our [`ApplicationJob`](https://github.com/thoughtbot/hound/blob/master/app/jobs/application_job.rb) handles it.

thorncp: Ah, I see! :+1:

salbertson:
@gylaz I don't think this is going to work since the services rely on the class name being serialized in Redis. If this change is made I believe the `ActiveJob` job wrapper class will be used.

https://github.com/thoughtbot/hound-scss/blob/master/lib/jobs/scss_review_job.rb

gylaz:
@salbertson @thorncp You were correct, the services can't handle `ActiveJob` jobs. What are your thoughts about removing `ActiveJob`? We are ties to Resque now via the services, so I don't see any wins using `ActiveJob`.

gylaz:
Closing this for now. Going to look into reverting to just using `Resque`, or at least retrying `Resque` jobs properly.

PR #871, opened 2015-08-06
Update and use token in database when creating a session

salbertson:
This change updates the user's token each time they sign in. The token used to be saved once, when the token in the database was null, and never updated again.

This change also requires a token be saved in the database and will redirect to the landing page if token is null. We used to allow users to sign in without saving a token to prevent currently authenticated users from going back through the auth flow.

This also sets us up for having two levels of access in Hound since we are updating the token during auth flow.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,7 +36,7 @@ class ApplicationController < ActionController::Base
   end
 
   def signed_in?
-    current_user.present?
+    current_user.present? && current_user.token.present?
   end
 
   def current_user
--- a/app/controllers/repo_syncs_controller.rb
+++ b/app/controllers/repo_syncs_controller.rb
@@ -3,10 +3,7 @@ class RepoSyncsController < ApplicationController
     unless current_user.refreshing_repos?
       current_user.update_attribute(:refreshing_repos, true)
 
-      RepoSynchronizationJob.perform_later(
-        current_user,
-        session[:github_token]
-      )
+      RepoSynchronizationJob.perform_later(current_user)
     end
 
     head 201
--- a/app/jobs/repo_synchronization_job.rb
+++ b/app/jobs/repo_synchronization_job.rb
@@ -1,8 +1,8 @@
 class RepoSynchronizationJob < ApplicationJob
   queue_as :high
 
-  def perform(user, github_token)
-    synchronization = RepoSynchronization.new(user, github_token)
+  def perform(user)
+    synchronization = RepoSynchronization.new(user)
     synchronization.start
     user.update_attribute(:refreshing_repos, false)
   end
--- a/spec/features/repo_list_spec.rb
+++ b/spec/features/repo_list_spec.rb
@@ -2,9 +2,8 @@ require "rails_helper"
 
 feature "Repo list", js: true do
   let(:username) { ENV.fetch("HOUND_GITHUB_USERNAME") }
-  let(:token) { ENV.fetch("HOUND_GITHUB_TOKEN") }
 
-  scenario "user views landing page" do
+  scenario "signed in user views repo list" do
     user = create(:user)
     repo = create(:repo, full_github_name: "thoughtbot/my-repo")
     repo.users << user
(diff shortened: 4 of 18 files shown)


app/controllers/repo_syncs_controller.rb
```diff
@@ -3,10 +3,7 @@ def create
     unless current_user.refreshing_repos?
       current_user.update_attribute(:refreshing_repos, true)
 
-      RepoSynchronizationJob.perform_later(
-        current_user,
-        session[:github_token]
-      )
+      RepoSynchronizationJob.perform_later(current_user.id)
```
salbertson: This keeps us from storing unencrypted tokens in Redis.

app/jobs/repo_synchronization_job.rb
```diff
@@ -1,8 +1,9 @@
 class RepoSynchronizationJob < ApplicationJob
   queue_as :high
 
-  def perform(user, github_token)
-    synchronization = RepoSynchronization.new(user, github_token)
+  def perform(user_id)
+    user = User.find(user_id)
```
thorncp: What do you think about taking advantage of [ActiveJob/global id](http://guides.rubyonrails.org/active_job_basics.html#globalid) and passing the user record directly to the constructor?

app/controllers/repo_syncs_controller.rb
gylaz: Any particular reason to do `current_user.id` her? ActiveJob uses GlobalID, which deserializes AR records.

spec/features/repo_list_spec.rb
```diff
@@ -2,20 +2,35 @@
...
   let(:username) { ENV.fetch("HOUND_GITHUB_USERNAME") }
-  let(:token) { ENV.fetch("HOUND_GITHUB_TOKEN") }
 
-  scenario "user views landing page" do
-    user = create(:user)
-    repo = create(:repo, full_github_name: "thoughtbot/my-repo")
-    repo.users << user
-    sign_in_as(user)
+  context "authenticated user and auth token" do
+    scenario "views landing page" do
```
thorncp: Should this be "views repo list"?

app/controllers/application_controller.rb
```diff
@@ -36,7 +36,7 @@ def authenticate
   end
 
   def signed_in?
-    current_user.present?
+    current_user.present? && current_user.token.present?
```
gylaz: If the token isn't present doesn't it sign the user out?

spec/features/repo_list_spec.rb
gylaz: Do you find `context`s helpful in feature specs? Another way to word this scenario would be `authenticated user views landing page` and the one below could be `user without token views landing page`. But, I'm not sure, maybe `context`s should be used.

spec/features/repo_list_spec.rb
```diff
@@ -2,20 +2,35 @@
...
 feature "Repo list", js: true do
   let(:username) { ENV.fetch("HOUND_GITHUB_USERNAME") }
-  let(:token) { ENV.fetch("HOUND_GITHUB_TOKEN") }
 
-  scenario "user views landing page" do
-    user = create(:user)
-    repo = create(:repo, full_github_name: "thoughtbot/my-repo")
-    repo.users << user
-    sign_in_as(user)
+  context "authenticated user and auth token" do
```
gylaz: Is this context missing a `with` or `when`?

gylaz:
Looks great!

app/controllers/repo_syncs_controller.rb
salbertson: I had no idea `GlobalID` existed, that's awesome. I'll change the job signature back.

app/jobs/repo_synchronization_job.rb
salbertson: Good call, didn't know about it until you and @gylaz commented.

spec/features/repo_list_spec.rb
salbertson: > Should this be "views repo list"?

Yes, good catch. Done.

> Do you find contexts helpful in feature specs?

I don't think so. I updated the test descriptions similar to your suggestion.

app/controllers/application_controller.rb
salbertson: Yes, that's one of the main differences with this change. Check out the PR description.

spec/features/repo_list_spec.rb
salbertson: I removed the `context`.

PR #875, opened 2015-08-10
Use Hound::GITHUB_TOKEN constant everywhere

arunagw:
As we have this constant which we can use everywhere in code so we don't need to use `.fetch` on ENV


thorncp:
LGTM

PR #876, opened 2015-08-10
Set PR status as an error

arunagw:
When hound failed itself we should update PR status with a message

### DIFF
--- a/app/jobs/buildable.rb
+++ b/app/jobs/buildable.rb
@@ -4,4 +4,10 @@ module Buildable
     build_runner = BuildRunner.new(payload)
     build_runner.run
   end
+
+  def after_retry_exhausted
+    payload = Payload.new(*arguments)
+    build_runner = BuildRunner.new(payload)
+    build_runner.set_internal_error
+  end
 end
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -23,6 +23,10 @@ class BuildRunner
     raise
   end
 
+  def set_internal_error
+    commit_status.set_internal_error
+  end
+
   private
 
   def review_pull_request
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -14,6 +14,7 @@ en:
   config_error_status:
     "Error parsing config file: %{filename}. Click \"details\" for
     assistance."
+  hound_error_status: "We've encountered an error while reviewing your code."
 
   account:
     bulk_pricing:
--- a/spec/jobs/buildable_spec.rb
+++ b/spec/jobs/buildable_spec.rb
@@ -1,4 +1,5 @@
 require "spec_helper"
+require "active_job"
 require "app/jobs/buildable"
 require "app/models/payload"
 require "app/services/build_runner"
@@ -11,19 +12,31 @@ describe Buildable do
 
   describe "perform" do
     it 'runs build runner' do
-      stub_const("Owner", "foo")
       build_runner = double(:build_runner, run: nil)
-      payload_data = "some data"
       payload = double("Payload")
       allow(Payload).to receive(:new).with(payload_data).and_return(payload)
       allow(BuildRunner).to receive(:new).and_return(build_runner)
-      allow(Owner).to receive(:upsert)
 
       BuildableTestJob.perform_now(payload_data)
 
       expect(Payload).to have_received(:new).with(payload_data)
       expect(BuildRunner).to have_received(:new).with(payload)
       expect(build_runner).to have_received(:run)
+    end
+  end
+
+  describe "#after_retry_exhausted" do
+    it "sets internal server error on github" do
+      build_runner = double(:build_runner, set_internal_error: nil)
+      payload = double("Payload")
+      allow(Payload).to receive(:new).with(payload_data).and_return(payload)
+      allow(BuildRunner).to receive(:new).and_return(build_runner)
+
+      BuildableTestJob.new(payload_data).after_retry_exhausted
+
+      expect(Payload).to have_received(:new).with(payload_data)
+      expect(BuildRunner).to have_received(:new).with(payload)
+      expect(build_runner).to have_received(:set_internal_error)
     end
   end
 
--- a/spec/jobs/retryable_spec.rb
+++ b/spec/jobs/retryable_spec.rb
@@ -1,7 +1,4 @@
-require "spec_helper"
-require "active_job"
-require "lib/ext/active_job/base"
-require "app/jobs/retryable"
+require "rails_helper"
 
 describe Retryable do
   context "a successful job" do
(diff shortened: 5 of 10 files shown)


app/services/build_runner.rb
```diff
@@ -21,6 +21,9 @@ def run
       remove_repo_from_user
     end
     raise
+  rescue
+    commit_status.set_internal_error
+    raise
```
gylaz: Is this the right place for this? Our jobs will retry 3 times. Would https://github.com/thoughtbot/hound/blob/master/app/jobs/retryable.rb#L9 be a better place for the error status?

config/locales/en.yml
```diff
@@ -12,6 +12,8 @@ en:
   config_error_status:
     "Error parsing config file: %{filename}. Click \"details\" for
     assistance."
+  hound_error_status:
+    "Something went wrong"
```
gylaz: How about `We've encountered an error while reviewing your code...` or something of the like?

app/services/build_runner.rb
arunagw: But that will get back to this block right? As this whole method gets into the queue.

Also I don't think that we have access to repo and other things at that place.

config/locales/en.yml
arunagw: Yes. We need to have a better message for sure

app/services/build_runner.rb
gylaz: It would get to this block, since it will retry the job 3 times. Ideally, I was thinking after the last attempt we'd set the status. However, if that's not possible then this is the next best thing. The only side-effect I can think of is the error status will be set on every retry.

arunagw: So it will make 3 calls to GitHub in "hound" context which will be replaced each other. In UI it will be only one message.

thorncp: We are pretty consistently running into API usage issues, so I think we don't want to make any requests to GitHub that can be avoided. I agree it would be best to make a single request after the last job attempt fails.

config/locales/en.yml
```diff
@@ -12,6 +12,9 @@ en:
   config_error_status:
     "Error parsing config file: %{filename}. Click \"details\" for
     assistance."
+  hound_error_status:
+    "We've encountered an error while reviewing your code. Please email hound@thoughtbot.com
+    assistance."
```
thorncp: Should this be `for assistance.` ?

config/locales/en.yml
```diff
@@ -12,6 +12,9 @@ en:
   config_error_status:
     "Error parsing config file: %{filename}. Click \"details\" for
     assistance."
+  hound_error_status:
+    "We've encountered an error while reviewing your code.
+    Please email hound@thoughtbot.com for assistance."
```
gylaz: I do believe there is a character limit on the status. See the screenshot for example of CC's error status:
screenshot 2015-08-13 06 50 57

spec/jobs/retryable_spec.rb
```diff
@@ -69,5 +69,7 @@ def perform
 
       expect(job).to have_received(:retry_job).with(wait: 10)
     end
+
+    it "calls after_retry_exhausted if exists"
```
arunagw: @thorncp I have updated code. But I am not sure how I can write test for this.

config/locales/en.yml
arunagw: Didn't find anything in docs. Can we ask someone?

spec/jobs/buildable_spec.rb
```diff
@@ -27,6 +25,21 @@ class BuildableTestJob < ActiveJob::Base
     end
   end
 
+  describe "after_retry_exhausted" do
+    it "will set internal error on github" do
```
teoljungberg: `it "sets internal server error on github"`?

config/locales/en.yml
gylaz: Not that I know of. I think it's likely only the first sentence will be able to fit.

spec/jobs/buildable_spec.rb
```diff
@@ -27,6 +25,21 @@ class BuildableTestJob < ActiveJob::Base
     end
   end
 
+  describe "after_retry_exhausted" do
```
gylaz: `#` or `.` before the method name?

gylaz:
Looking great! Let's make sure to give it a good spin on staging.

spec/jobs/buildable_spec.rb
teoljungberg: It should be `#`

spec/jobs/retryable_spec.rb
arunagw: @teoljungberg I need a bit of help here to complete this spec. Can you help?

teoljungberg: I imagine something like this rouch sketch:

```ruby
context "when `after_retry_exhausted` exists" do
  it "calls `after_retry_exhausted` after retrying" do
    job = double("Job")
    allow(job).to receive(:after_retry_exhausted).and_return(true)

    perform_the_job_so_it_retries

    expect(job).to have_received(:after_retry_exhausted)
  end
end

context "when `after_retry_exhausted` does not exists" do
  it "does not call it" do
    job = double("Job")

    perform_the_job_so_it_retries

    expect(job).not_to have_received(:after_retry_exhausted)
  end
end
```

arunagw:
@teoljungberg See this for spec https://github.com/thoughtbot/hound/commit/2ffaae84bccd080beb87649d6ea1adc3d0cc6ced

gylaz:
Looks good on staging!
screenshot 2015-09-10 15 27 26

app/jobs/buildable.rb
```diff
@@ -4,4 +4,10 @@ def perform(payload_data)
     build_runner = BuildRunner.new(payload)
     build_runner.run
   end
+
+  def after_retry_exhausted
+    payload = Payload.new(*arguments)
```
thorncp: Is `arguments` an ActiveJob thing? It seems a little that it's being splatted here but not in `#perform`.

thorncp: I guess what I mean is, do we need the whole splat, or only the first element?

thorncp:
This looks great!

app/jobs/buildable.rb
gylaz: I'm guessing it is. At the very least it seems to be a "magic" variable that holds data like `[{"repository"=>{"owner"=>{"id"=>1234, "login"=>"test", "type"=>"Organization"}}}]`. Because of the array, it needs to be splatted.

gylaz:
Merged as 03273b0557e74c05887d11076c3f2cc2e37d2088.

PR #878, opened 2015-08-11
Remove unnecessary Haml docs

salbertson:
It should be clear how to disable Haml since it's disabled by default and we show instructions for enabling.


teoljungberg:
LGTM

PR #882, opened 2015-08-17
Update repo attributes on each build

arunagw:
The idea is to keep repo attributes updated always so we can track public/private flag change on repo from GitHub.

### DIFF
--- a/app/jobs/buildable.rb
+++ b/app/jobs/buildable.rb
@@ -1,6 +1,8 @@
 module Buildable
   def perform(payload_data)
     payload = Payload.new(payload_data)
+
+    UpdateRepoStatus.new(payload).run
     build_runner = BuildRunner.new(payload)
     build_runner.run
   end
--- a/app/services/update_repo_status.rb
+++ b/app/services/update_repo_status.rb
@@ -1,0 +1,22 @@
+class UpdateRepoStatus
+  pattr_initialize :payload
+
+  def run
+    if repo
+      repo.update(repo_attributes)
+    end
+  end
+
+  private
+
+  def repo
+    @repo ||= Repo.active.find_by(github_id: payload.github_repo_id)
+  end
+
+  def repo_attributes
+    {
+      full_github_name: payload.full_repo_name,
+      private: payload.private_repo?,
+    }
+  end
+end
--- a/spec/jobs/buildable_spec.rb
+++ b/spec/jobs/buildable_spec.rb
@@ -1,19 +1,14 @@
-require "spec_helper"
-require "active_job"
-require "app/jobs/buildable"
-require "app/models/payload"
-require "app/services/build_runner"
-require "raven"
+require "rails_helper"
 
 describe Buildable do
   class BuildableTestJob < ActiveJob::Base
     include Buildable
   end
 
-  describe "perform" do
+  describe "#perform" do
     it 'runs build runner' do
       build_runner = double(:build_runner, run: nil)
-      payload = double("Payload")
+      payload = double("Payload", github_repo_id: 1)
       allow(Payload).to receive(:new).with(payload_data).and_return(payload)
       allow(BuildRunner).to receive(:new).and_return(build_runner)
 
@@ -23,11 +18,24 @@ describe Buildable do
       expect(BuildRunner).to have_received(:new).with(payload)
       expect(build_runner).to have_received(:run)
     end
+
+    it "runs repo updater" do
+      repo_updater = double("RepoUpdater", run: nil)
+      payload = double("Payload", github_repo_id: 1)
+      allow(Payload).to receive(:new).with(payload_data).and_return(payload)
+      allow(UpdateRepoStatus).to receive(:new).and_return(repo_updater)
+
+      BuildableTestJob.perform_now(payload_data)
+
+      expect(Payload).to have_received(:new).with(payload_data)
+      expect(UpdateRepoStatus).to have_received(:new).with(payload)
+      expect(repo_updater).to have_received(:run)
+    end
   end
 
   describe "#after_retry_exhausted" do
     it "sets internal server error on github" do
-      build_runner = double(:build_runner, set_internal_error: nil)
+      build_runner = double("BuildRunner", set_internal_error: nil)
       payload = double("Payload")
       allow(Payload).to receive(:new).with(payload_data).and_return(payload)
       allow(BuildRunner).to receive(:new).and_return(build_runner)
--- a/spec/models/payload_spec.rb
+++ b/spec/models/payload_spec.rb
@@ -150,6 +150,34 @@ describe Payload do
     end
   end
 
+  describe "#private_repo?" do
+    context "when repo is private" do
+      it "returns true" do
+        payload_json = {
+          "repository" => {
+            "private" => true,
+          }
+        }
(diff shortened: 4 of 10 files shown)


app/jobs/buildable.rb
```diff
@@ -1,6 +1,9 @@
 module Buildable
   def perform(payload_data)
     payload = Payload.new(payload_data)
+
+    RepoUpdater.new(payload).run
```
arunagw: I am not sure what to call this. This class will mostly be handling repo updates and deactivating repo based on conditions.

app/jobs/buildable.rb
```diff
@@ -1,6 +1,9 @@
 module Buildable
   def perform(payload_data)
     payload = Payload.new(payload_data)
+
+    repo_updater = RepoUpdater.new(payload)
```
gylaz: How about we get away from the "Land of Nouns" and call this `UpdateRepoStatus`?

arunagw: Yes. Will update this.

gylaz:
@arunagw Do you mind adding a descriptive commit message about why something was done?

app/services/update_repo_status.rb
```diff
@@ -0,0 +1,34 @@
...
+  def repo_attributes
+    {
+      full_github_name: payload.full_repo_name,
+      private: payload.private_repo?,
+    }
+  end
+
+  def deactive_repo(repo)
+    if repo.previous_changes[:private] == change_public_to_private
+      repo.deactivate
```
gylaz: I just went through some of the repos that went from public to private, and manually reached out to folks. One person converted and started paying. Based on that, I'm not sure it's a good idea to disable the repo right away. I think it might be best to track such repos (maybe via Administrate dashboard), and reach out to them separately.

Thoughts @thoughtbot/hound?

app/jobs/buildable.rb
```diff
@@ -1,6 +1,8 @@
 module Buildable
   def perform(payload_data)
     payload = Payload.new(payload_data)
+
+    UpdateRepoStatus.new(payload).run
```
gylaz: What do you think about `BuildRunner` calling this? This way we wouldn't need to find `repo` in two places.

spec/models/payload_spec.rb
```diff
@@ -150,6 +150,34 @@
...
+    context "when repo is private" do
+      it "returns true" do
+        payload_json = {
+          "repository" => {
+            "private" => true,
+          }
+        }
+        payload = Payload.new(payload_json)
+
+        expect(payload.private_repo?).to be_truthy
```
gylaz: What do you think about `expect(payload.private_repo?).to eq true`?

spec/models/payload_spec.rb
```diff
@@ -150,6 +150,34 @@
...
+    context "when repo is public" do
+      it "returns false" do
+        payload_json = {
+          "repository" => {
+            "private" => false,
+          }
+        }
+        payload = Payload.new(payload_json)
+
+        expect(payload.private_repo?).to be_falsy
```
gylaz: What do you think about `expect(payload.private_repo?).to eq false`?

gylaz:
@arunagw I've rebased latest master and force pushed.

app/services/update_repo_status.rb
thorncp: Could also make it easy to deactivate those repos in Administrate, if we don't get a response. :+1:

app/jobs/buildable.rb
arunagw: I think finding repo is fine at two places because it's anyway in background. Previously we were updating repo when finding that in `BuildRunner`.

I think updating repo needs to come before `BuildRunner` so we make sure that it's updated.

What you think?

app/services/update_repo_status.rb
arunagw: We can also add another flag like `pending_deactivation` and that can generate a report to know.

We can avoid deactivation in this PR and just update a new status and then we can do another PR for report or admin handle.

app/jobs/buildable.rb
gylaz: Sounds good.

app/services/update_repo_status.rb
gylaz: There's a fairly easy query we can run to see which repos are private, but don't have a subscription (and are not bulk customers). So, an additional flag might be unnecessary.

arunagw: Should we not deactivate then? And just change status based on GitHub payload?

gylaz: I'm leaning towards not activating, then we can run a query (or add an admin dashboard) to see who's not paying.

arunagw: Taken care here https://github.com/thoughtbot/hound/commit/1d200eebd3f57a95916e8aaef45ec337dc93a7dd 

I think you meant not deactivating right?

arunagw:
In this PR we are just updating repo attributes whenever a new build comes in.

app/services/update_repo_status.rb
gylaz: Yep.

app/services/update_repo_status.rb
```diff
@@ -0,0 +1,22 @@
+class UpdateRepoStatus
+  pattr_initialize :payload
+
+  def run
+    if repo
+      repo.update_attributes(repo_attributes)
```
gylaz: What do you think about using just `update` (`repo.update(repo_attributes)`)?

gylaz:
:+1:

spec/jobs/buildable_spec.rb
```diff
@@ -23,6 +18,19 @@ class BuildableTestJob < ActiveJob::Base
       expect(BuildRunner).to have_received(:new).with(payload)
       expect(build_runner).to have_received(:run)
     end
+
+    it "runs repo updater" do
+      repo_updater = double(:repo_updater, run: nil)
```
teoljungberg: We've consistently used strings as the names for the `double`, what do you think about doing so here too?

spec/services/update_repo_status_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expected_repo_name = "foo/bar"
+      repo = create(:repo, :active)
+      payload = double(
+        "Payload",
+        github_repo_id: repo.github_id,
+        full_repo_name: expected_repo_name,
+        private_repo?: repo.private,
+      )
+
+      UpdateRepoStatus.new(payload).run
```
teoljungberg: What do you think about having the `repo` being returned by this service? That would avoid the `.reload` call and make the code a bit clearer, I think.

teoljungberg:
A few ideas and comments, but this looks good to me

spec/services/update_repo_status_spec.rb
arunagw: I usually don't prefer anything to be returned from a `.run` method. But what if repo not found? It might return nil.

spec/jobs/buildable_spec.rb
arunagw: Yes.

arunagw:
Will merge this in when CI is green. Thanks for reviews.

Issue #701, opened 2015-03-03
Allow users to try Hound without asking for access to all repos


salbertson:
Would it benefit Hound users if read/write access to private repos was requested after the user had a chance to authenticate and try Hound on a public repo?

https://twitter.com/DinisCruz/status/572682201978048512

DinisCruz:
I think this would make a massive difference, and in my case it is a deal breaker.

Question: Is there a wiki page that has a direct mapping to the 'privs requested to github' to the corresponding 'hound features'

Maybe having 'sliding scale of capabilities' which would depend on how much privs you are assigned?

salbertson:
Here's the branch where this work is being done, https://github.com/thoughtbot/hound/tree/sa-github-scopes.

fpietrosanti:
+1000 Let us know when it's ready, GlobaLeaks Anonymous Whistleblowing project and OpenPGP.js encryption library will immediately start using it!

salbertson:
Related: https://github.com/thoughtbot/hound/issues/597

fpietrosanti:
Any time estimate for delivery of that feature? :*

thorncp:
This is now live!

salbertson:
@thorncp spoke too soon, my bad.

fpietrosanti:
Related to #895

salbertson:
@fpietrosanti yea, sorry about that. Here's a [new PR](https://github.com/thoughtbot/hound/pull/900) that will hopefully be wrapped up soon.

fpietrosanti:
@salbertson ack, will test once avilable! Super thanks! :)

salbertson:
@fpietrosanti it's in production, again!

salbertson:
Done. https://github.com/thoughtbot/hound/pull/900

fpietrosanti:
@salbertson I've tried and now it ask me to authorized "organization access"  but at the same time it keep asking the WRITE privilege permission to the Code, Wiki and everything else.

At GlobaLeaks we would not be able to use it by giving to @houndci such high level of privileges.

Would you be able to make so that for Public Repository it does only ask for READ privileges but not for WRITE on the Code, Wiki and other part of the software project that does not need high privileges?

Below screenshot of the inquiry for read and write privileges to all:
[...]

Issue #597, opened 2015-01-30
github token gives too liberal access


ashleygwilliams:
hiya! first off: cool product. i saw it and wanted to share it with my team.

HOWEVER, just to sign in, one is req'd to use github and provide read/write to all repos the user owns.

there are people on my team with ownership over the whole jquery CDN, among other less 'shareable', more 'sensitive' projects. they aren't keen on giving you access, even though they'd like to try out your product on other repos.

do ya'll have any thoughts on making the sign in access a bit more subtle?

thanks again for making and sharing this!

-a

ashleygwilliams:
i also realize this could be a github issue, as well. but i'll leave this here anyways.

salbertson:
Hi @ashleygwilliams! I'm glad you brought this up, others have asked about it as well.

Would a progressive approach to asking for access help in your case?

I'm thinking Hound could ask for access to public repos to being with, then once you are authenticated Hound would allow users to give more access based on organization.

Does that make sense?

salbertson:
@ashleygwilliams, any thoughts on the approach I mentioned?

rogierslag:
Same here, I'd like to give Hound access to some repos, but certainly not all private repositories. Even within an organization I'd like to make the access level a bit more fine grained than "you go girl!"

jpalumickas:
I think this is an GitHub issue to not letting us choose to which private/organization repos give permissions.

rogierslag:
You could opt to do it like eg Shippable and Codeclimate. They set a deploy token on the repository so you can access it

salbertson:
Closing in favor of https://github.com/thoughtbot/hound/issues/701.

salbertson:
@rogierslag that's a good idea. We are going to try out allowing users to opt into private repo access and see how that works for everyone.

salbertson:
Hound now has optional private repo access. https://github.com/thoughtbot/hound/pull/900

PR #883, opened 2015-08-17
Allow users to try Hound without giving access to private repos

salbertson:
Addresses https://github.com/thoughtbot/hound/issues/597 and https://github.com/thoughtbot/hound/issues/701.

hound private access

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -2,9 +2,13 @@ class SessionsController < ApplicationController
   skip_before_action :authenticate, only: [:create]
 
   def create
-    user = find_user || create_user
-    create_session_for(user)
-    user.update(token: github_token)
+    create_session
+
+    if github_token
+      update_token
+      update_scopes
+    end
+
     finished("auth_button")
     redirect_to repos_path
   end
@@ -15,6 +19,10 @@ class SessionsController < ApplicationController
   end
 
   private
+
+  def user
+    @user ||= find_user || create_user
+  end
 
   def find_user
     if user = User.where(github_username: github_username).first
@@ -34,12 +42,16 @@ class SessionsController < ApplicationController
     user
   end
 
-  def create_session_for(user)
+  def create_session
     session[:remember_token] = user.remember_token
   end
 
   def destroy_session
     session[:remember_token] = nil
+  end
+
+  def github
+    GithubApi.new(github_token)
   end
 
   def github_username
@@ -52,5 +64,24 @@ class SessionsController < ApplicationController
 
   def github_token
     request.env["omniauth.auth"]["credentials"]["token"]
+  end
+
+  def scopes_changed?
+    user.token_scopes != token_scopes
+  end
+
+  def token_scopes
+    @token_scopes ||= github.scopes
+  end
+
+  def update_token
+    user.update!(token: github_token)
+  end
+
+  def update_scopes
+    if scopes_changed?
+      user.update!(token_scopes: token_scopes)
+      user.repos.clear
+    end
   end
 end
--- a/app/models/github_auth_options.rb
+++ b/app/models/github_auth_options.rb
@@ -1,0 +1,13 @@
+class GithubAuthOptions
+  def initialize(env)
+    @request = Rack::Request.new(env)
+  end
+
+  def to_hash
+    if @request.params["access"] == "full"
+      { scope: "repo,user:email" }
+    else
+      { scope: "user:email" }
+    end
+  end
+end
--- a/db/migrate/20150728154011_add_token_scopes_to_users.rb
+++ b/db/migrate/20150728154011_add_token_scopes_to_users.rb
@@ -1,0 +1,10 @@
+class AddTokenScopesToUsers < ActiveRecord::Migration
+  def up
+    add_column :users, :token_scopes, :string
+    execute "UPDATE users SET token_scopes = 'repo,user:email' WHERE token IS NOT NULL"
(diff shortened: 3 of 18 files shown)


db/migrate/20150728154011_add_token_scopes_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTokenScopesToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :token_scopes, :string
```
thorncp: Do you think it would be beneficial to backfill this for existing records and make the column not nullable?

app/controllers/sessions_controller.rb
```diff
@@ -53,4 +62,25 @@ def github_email_address
   def github_token
     request.env["omniauth.auth"]["credentials"]["token"]
   end
+
+  def scopes_changed?
+    user.token_scopes.present? && (user.token_scopes != token_scopes)
+  end
+
+  def token_scopes
+    @token_scopes ||= github.scopes
```
gylaz: Do we need this memoization?

gylaz:
Looks great!

db/migrate/20150728154011_add_token_scopes_to_users.rb
salbertson: Well, some users don't have a token saved so I'm not sure if it makes sense to backfill the scopes in those cases.

app/controllers/sessions_controller.rb
salbertson: I think so, it's used in `scopes_changed?` and `update_scopes`.

db/migrate/20150728154011_add_token_scopes_to_users.rb
thorncp: Ah, that makes sense. We could still backfill for those that do have a token, even if it doesn't make sense to make it non-null.

I guess there's no real need to do it, just feels like something is missing.

app/controllers/sessions_controller.rb
gylaz: `GithubApi#scopes` is already memoized. What do you think about memorizing `#github` method instead?

croaky:
Whoa, big time.

app/models/github_auth_options.rb
```diff
@@ -0,0 +1,13 @@
+class GithubAuthOptions
```
arunagw: Can this file be in `lib` ?

spec/features/repo_list_spec.rb
```diff
@@ -96,7 +86,7 @@
     stub_repos_requests(token)
     sign_in_as(user)
 
-    expect(page).to have_content I18n.t("syncing_repos")
+    expect(page).to have_content "Sign Out"
```
arunagw: Use this from Locale?

lib/github_api.rb
```diff
@@ -18,6 +17,10 @@ def client
     @client ||= Octokit::Client.new(access_token: token, auto_paginate: true)
   end
 
+  def scopes
+    @scopes ||= client.scopes(@token).join(",")
```
arunagw: `@token` => `token` as we have attr_reader

app/controllers/sessions_controller.rb
```diff
@@ -53,4 +62,25 @@ def github_email_address
...
+  def update_token
+    if github_token
+      user.update!(token: github_token)
+    end
+  end
+
+  def update_scopes
+    if scopes_changed?
+      user.update!(token_scopes: token_scopes)
+      user.repos.clear
```
arunagw: Should we also start repo sync job ? Also that job is already clearing repos.

app/controllers/sessions_controller.rb
```diff
@@ -53,4 +62,25 @@ def github_email_address
   def github_token
     request.env["omniauth.auth"]["credentials"]["token"]
   end
+
+  def scopes_changed?
```
arunagw: Can most of these methods move into a service class by supplying `token` and dirty `@user` ?

db/migrate/20150728154011_add_token_scopes_to_users.rb
salbertson: Good idea, done!

app/controllers/sessions_controller.rb
salbertson: Ah, good call. I removed the memoization from `GithubApi` since it's not a common pattern in that class and left the optimization in the controller.

app/models/github_auth_options.rb
salbertson: It's app specific, so I'd vote no.

spec/features/repo_list_spec.rb
salbertson: Great idea, done.

lib/github_api.rb
salbertson: Good catch, done.

app/controllers/sessions_controller.rb
salbertson: I don't think so. I want the repos cleared out right away so users don't see their old list after their GitHub access is changed. If we enqueue the job there's a good chance repos won't be cleared in time.

Clearing the repos will trigger a sync job when the repos page loads and repo count == 0.

app/controllers/sessions_controller.rb
salbertson: Probably, but I'll leave that for a separate PR since this one is getting large.

PR #888, opened 2015-08-19
CoffeeScript: Remove 'Empty function' violation

mcmire:
Oftentimes it's useful in tests to stub out methods on objects with
empty functions.

For instance you might make a helper function in a test file:

``` coffee
buildSomeFakeObject = (overrides = {}) ->
  _.defaults overrides,
   foo: ->
   bar: ->
```

and then later on in the test, you might say:

``` coffee
fakeObject = buildSomeFakeObject()
spyOn(fakeObject, "foo")

...

expect(fakeObject.foo).toHaveBeenCalled()
```

Currently, Hound will complain within `buildSomeFakeObject` that `foo`
and `bar` are empty functions. This creates a lot of noise.

I don't think this rule is necessary in the general case anyway -- I've
never seen anyone make an empty function accidentally.

### DIFF
--- a/spec/models/style_guide/coffee_script_spec.rb
+++ b/spec/models/style_guide/coffee_script_spec.rb
@@ -88,6 +88,21 @@ describe StyleGuide::CoffeeScript do
       end
     end
 
+    context "with thoughtbot configuration" do
+      context "for an empty function" do
+        it "returns a file review without violations" do
+          code = <<-CODE.strip_heredoc
+            class FooBar
+              foo: ->
+          CODE
+
+          violations = violations_in(code, repository_owner_name: "thoughtbot")
+
+          expect(violations).to be_empty
+        end
+      end
+    end
+
     context "with violation on unchanged line" do
       it "finds no violations" do
         file = double(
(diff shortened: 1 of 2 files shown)


gylaz:
While I've seen Hound catch an empty function once, I have no objections to merging this.

spec/models/style_guide/coffee_script_spec.rb
```diff
@@ -87,6 +87,21 @@ class FooBar
...
+      context "for an empty function" do
+        it "returns a file review without any violations" do
+          code = <<-CODE.strip_heredoc
+            class FooBar
+              foo: ->
+          CODE
+
+          violations = violations_in(code, repository_owner_name: "thoughtbot")
+
+          expect(violations).not_to be_any { |m| m =~ /Empty function/ }
```
teoljungberg: Can this assertion be more explicit? Are there any other violations than the one we care about?

mcmire: There aren't any other violations than the one we care about -- what do you mean?

teoljungberg: Oh, I miss-read the assertion.

How about: `expect(violatiotions).to be_empty`?

salbertson:
Looks good to me. :+1:

PR #890, opened 2015-08-21
Update ruby version for security fix

camelmasa:
See. https://www.ruby-lang.org/en/news/2015/08/18/ruby-2-2-3-released/


thorncp:
Merged as 50cb7aee44a95f6242c596ff00d8af68c72c219c. Thanks!

PR #892, opened 2015-08-21
Revert "Structure files according to used version of rails"

salbertson:
This reverts commit 9988a31724e61a7a4363bc582bc7ef776e91d428.

This change fixes this

screenshot 2015-08-21 07 40 51

and

screenshot 2015-08-21 07 41 05

PR #893, opened 2015-08-21
Revert "Structure files according to used version of rails"

thorncp:
This reverts commit 9988a31724e61a7a4363bc582bc7ef776e91d428.

Changing cookie serialization to JSON causes the app to fail when attempting to parse old cookies that aren't JSON. See also: #892


salbertson:
:+1:

PR #894, opened 2015-08-21
Revert SSL refactoring

thorncp:
Reverts aaef09bb2ee10d5d7370ce62b77c93063bafa626 and 3b22d42c1ded47a485e2f8bf6bc9e6138784aa58.

The SSL refactorings were forcing all incoming traffic to use `https`, but we have old webhooks in the wild that still use `http`. GitHub's webhooks do not honor redirects, so those payloads will fail and users will not have their PRs reviews.

webhook_-_https___houndci_com_builds

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,12 +1,25 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery
 
+  before_action :force_https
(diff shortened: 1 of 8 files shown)


thorncp:
Since this are `git revert`s I think I don't want to address Hound's complaints.

salbertson:
:+1:

PR #896, opened 2015-08-25
Add Swift Support

jakecraige:
Why:

* Requested by Users
* We want to use it internally :)

This change addresses the need by:

* Adding Swift support so that Hound queues jobs for Swift
  that the worker can pick up.

Swift Service Repo: https://github.com/thoughtbot/hound-swift


teoljungberg:
LGTM - thanks for pulling this through!

salbertson:
:raised_hands:

PR #897, opened 2015-08-25
Add an admin dashboard through Administrate

graysonwright:
- Admin handles can be set through an environment variable, `ADMIN_GITHUB_USERNAMES`.

See individual commit messages for more detailed descriptions of changes

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 ruby "2.2.3"
 
 gem "active_model_serializers", "0.8.3"
+gem "administrate", "~> 0.0.8"
 gem "analytics-ruby", "~> 2.0.0", require: "segment/analytics"
 gem "angularjs-rails"
 gem "angular_rails_csrf"
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -1,0 +1,27 @@
+class Admin::ApplicationController < Administrate::ApplicationController
+  before_action :authenticate_admin
+
+  private
+
+  def controller_path
+    params[:resource_class].to_s
+  end
+
+  def authenticate_admin
+    unless github_admin?
+      redirect_to :root
+    end
+  end
+
+  def github_admin?
+    current_user && admin_github_handles.include?(current_user.github_username)
+  end
+
+  def current_user
+    @current_user ||= User.find_by(remember_token: session[:remember_token])
+  end
+
+  def admin_github_handles
+    ENV.fetch("ADMIN_GITHUB_USERNAMES", "").split(",")
+  end
+end
--- a/app/dashboards/bulk_customer_dashboard.rb
+++ b/app/dashboards/bulk_customer_dashboard.rb
@@ -1,0 +1,31 @@
+require "administrate/base_dashboard"
+
+class BulkCustomerDashboard < Administrate::BaseDashboard
+  READ_ONLY_ATTRIBUTES = [
+    :id,
+    :created_at,
+    :updated_at,
+  ]
+
+  ATTRIBUTE_TYPES = {
+    id: Field::String,
+    created_at: Field::String,
+    updated_at: Field::String,
+    org: Field::String,
+    description: Field::String,
+    interval: Field::String,
+    repo_limit: Field::String,
+    current_repos: Field::String,
+    subscription_token: Field::String,
+  }
+
+  TABLE_ATTRIBUTES = [
+    :org,
+    :description,
+    :repo_limit,
+    :current_repos,
+  ]
+
+  SHOW_PAGE_ATTRIBUTES = ATTRIBUTE_TYPES.keys
+  FORM_ATTRIBUTES = ATTRIBUTE_TYPES.keys - READ_ONLY_ATTRIBUTES
+end
--- a/app/dashboards/dashboard_manifest.rb
+++ b/app/dashboards/dashboard_manifest.rb
@@ -1,0 +1,5 @@
+class DashboardManifest
+  DASHBOARDS = [
+    :bulk_customers,
+  ]
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,16 @@
 Houndapp::Application.routes.draw do
+  namespace :admin do
+    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
+      resources(
+        dashboard_resource,
+        controller: "application",
+        resource_class: dashboard_resource,
+      )
+    end
+
+    root to: "application#index", resource_class: :bulk_customers
+  end
+
   mount Resque::Server, at: "/queue"
   mount JasmineRails::Engine => "/specs" if defined?(JasmineRails)
   mount Split::Dashboard, at: "split"
--- a/spec/features/admin_authorization_spec.rb
+++ b/spec/features/admin_authorization_spec.rb
@@ -1,0 +1,39 @@
+require "rails_helper"
+
+feature "Admin authorization" do
+  scenario "admin accesses dashboard" do
+    stub_repos_requests(token)
+    stub_admin_github_usernames("admin_user,other_admin_user")
+    admin = create(:user, github_username: "admin_user")
+
+    sign_in_as(admin, token)
+    visit admin_bulk_customers_path
+
+    expect(page).to have_admin_bulk_customers_header
+  end
+
+  scenario "non-admin cannot access dashboard" do
+    stub_repos_requests(token)
+    stub_admin_github_usernames("admin_user,other_admin_user")
+    non_admin = create(:user, github_username: "not_admin_user")
+
+    sign_in_as(non_admin, token)
+    visit admin_bulk_customers_path
+
+    expect(page).not_to have_admin_bulk_customers_header
+  end
+
+  def stub_admin_github_usernames(usernames)
+    allow(ENV).to receive(:fetch).
+      with("ADMIN_GITHUB_USERNAMES", "").
+      and_return(usernames)
+  end
+
+  def have_admin_bulk_customers_header
+    have_css("h1", text: "Bulk Customers")
+  end
+
+  def token
+    "usergithubtoken"
+  end
+end
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 # Don't modify this file. Create `.env.local` and put your overrides in there.
+ADMIN_GITHUB_USERNAMES=your_github_handle,another_github_handle
 CHANGED_FILES_THRESHOLD=300
 DB_POOL=5
 DB_REAPING_FREQUENCY=10
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,6 +46,12 @@ GEM
       thread_safe (~> 0.3, >= 0.3.4)
       tzinfo (~> 1.1)
     addressable (2.3.7)
+    administrate (0.0.8)
+      neat (~> 1.1)
+      normalize-rails (~> 3.0)
+      rails (~> 4.2)
+      sass (~> 3.4)
+      selectize-rails (~> 0.6)
     analytics-ruby (2.0.6)
     angular_rails_csrf (1.0.3)
       rails (>= 3, < 5)
@@ -158,6 +164,7 @@ GEM
     newrelic_rpm (3.9.9.275)
     nokogiri (1.6.6.2)
       mini_portile (~> 0.6.0)
+    normalize-rails (3.0.3)
     oauth2 (1.0.0)
       faraday (>= 0.8, < 0.10)
       jwt (~> 1.0)
@@ -282,6 +289,7 @@ GEM
     sawyer (0.6.0)
       addressable (~> 2.3.5)
       faraday (~> 0.8, < 0.10)
+    selectize-rails (0.12.1)
     sentry-raven (0.9.4)
       faraday (>= 0.7.6)
       hashie (>= 1.1.0)
@@ -332,6 +340,7 @@ PLATFORMS
 
 DEPENDENCIES
   active_model_serializers (= 0.8.3)
+  administrate (~> 0.0.8)
   analytics-ruby (~> 2.0.0)
   angular_rails_csrf
   angularjs-rails
--- a/app/models/bulk_customer.rb
+++ b/app/models/bulk_customer.rb
@@ -18,4 +18,8 @@ class BulkCustomer < ActiveRecord::Base
 
     update(current_repos: repo_count)
   end
+
+  def to_s
+    org
+  end
 end


app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,36 @@
+# All Administrate controllers inherit from this `ApplicationController`,
```
thorncp: What do you think about removing these generated comments, to cut down on the diff noise?

spec/features/admin_authorization_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "rails_helper"
+
+feature "Admin authorization" do
+  scenario "admin accesses dashboard" do
+    token = "usergithubtoken"
+    user = create(:user, github_username: "admin_user")
+    stub_repos_requests(token)
+    allow(ENV).to receive(:fetch).
```
teoljungberg: We're not stubbing `ENV` anywhere else in the codebase. It's done What do you think of doing the same convention here?

graysonwright: @teoljungberg sorry, looks like part of your comment got cut off. How would you suggest changing it?

app/controllers/admin/application_controller.rb
graysonwright: Sure, we can do that.

spec/features/admin_authorization_spec.rb
teoljungberg: You're right. We're setting the env variables in the spec/rails helpers. I think we should continue to do that

spec/features/admin_authorization_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      with("ADMIN_GITHUB_HANDLES", "").
+      and_return("admin_user,other_admin_user")
+
+    sign_in_as(user, token)
+    visit "/admin"
```
teoljungberg: What do you think about using the named routes instead?

spec/features/admin_authorization_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    expect(page).to have_content "Administrate"
```
teoljungberg: Should this be translated?

spec/features/admin_authorization_spec.rb
graysonwright: Good call, will do.

spec/features/admin_authorization_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    sign_in_as(admin, token)
+    visit admin_users_path
+
+    expect(page).to have_admin_users_header
+
+    ENV["ADMIN_GITHUB_HANDLES"] = nil
+  end
+
+  scenario "non-admin cannot access dashboard" do
+    ENV["ADMIN_GITHUB_HANDLES"] = "admin_user,other_admin_user"
```
graysonwright: @teoljungberg I looked through the `spec` folder for instances of `ENV`, and found this pattern: https://github.com/thoughtbot/hound/blob/master/spec/helpers/analytics_helper_spec.rb#L5

Does this look good, or were you thinking of something different?

app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  before_action :authenticate_admin
+
+  def index
+    super
+
+    flash[:alert] =
+      "For performance, Administrate limits the index page to show 20 items.
+      Customize this action to update/remove the limit,
+      or implement the pagination library of your choice."
+    @resources = @resources.limit(20)
```
calebthompson: 1. Does this happen on every hit to the index page?
2. Should this only happen when `count` > 20?

app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def authenticate_admin
+    unless current_user &&
+        admin_github_handles.include?(current_user.github_username)
+      redirect_to :root
+    end
+  end
+
+  def current_user
+    @current_user ||= User.where(remember_token: session[:remember_token]).first
```
calebthompson: `find_by(remember_token:)` is the equivalent of `where(*).first)`

app/controllers/admin/violations_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Admin::ViolationsController < Admin::ApplicationController
```
calebthompson: Any chance Administrate can provide a default controller? Empty classes are an antipattern, IMO

Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 ruby "2.2.3"
 
 gem "active_model_serializers", "0.8.3"
+gem "administrate", "~> 0.0.7"
```
teoljungberg: Do we need to lock the version?

graysonwright:
Screenshots:

## Index
administrate

## Show
administrate

## Edit
administrate

app/dashboards/bulk_customer_dashboard.rb
```diff
@@ -0,0 +1,31 @@
...
+  def table_attributes
+    ATTRIBUTES.first(4)
+  end
+
+  def show_page_attributes
+    ATTRIBUTES
+  end
+
+  def form_attributes
+    ATTRIBUTES - READ_ONLY_ATTRIBUTES
```
calebthompson: This can be a constant as well: `FORM_ATTRIBUTES = ATTRIBUTES - READ_ONLY_ATTRIBUTES` as long as both other constants have already been defined.

app/models/repo.rb
```diff
@@ -71,6 +71,10 @@ def total_violations
       count
   end
 
+  def to_s
```
teoljungberg: Could these be moved into a concern? For `Administratable` models?

app/dashboards/dashboard_manifest.rb
```diff
@@ -0,0 +1,22 @@
...
+    :owners,
+    :repos,
+    :style_configs,
+    :subscriptions,
+    :users,
+    :violations,
+  ]
+
+  def dashboards
+    DASHBOARDS
```
calebthompson: Can we just use the constant?

spec/features/admin_authorization_spec.rb
teoljungberg: Looks good to me, let's see if we can find a better convention and do that in a seperate PR

Gemfile
graysonwright: Good call, forgot about that. We're still pre-1.0, so there could be breaking changes. I'll lock it down to `"0.0.7"`.

calebthompson:
All of the Admin::BaseDashboard subclasses share a similar pattern.

1. I think we shouldn't be defining methods at all and just use constants for these, but:
2. If we must keep the methods, they should all be memoized.

app/controllers/admin/application_controller.rb
jakecraige: What bout Default or GET param for more? Ideally pagination at some point but until then that feels like a nice in between

Gemfile
calebthompson: Ideally, we'd find the breaking changes immediately by _not_ locking the version and dogfooding whatever code we've released.

teoljungberg: I agree with @calebthompson

app/models/repo.rb
calebthompson: :+1:

config/routes.rb
```diff
@@ -1,4 +1,12 @@
 Houndapp::Application.routes.draw do
+  namespace :admin do
+    DashboardManifest.new.dashboards.each do |dashboard_resource|
```
calebthompson: It seems like it would be useful to have:

```
Administrator.dashboards.each { ... }
```

exposed for this. It hides internal implementation details I shouldn't need to care about as an end-user.

app/controllers/admin/violations_controller.rb
graysonwright: yeah, we went back and forth on this. There's a bit of discussion here: https://github.com/thoughtbot/administrate/pull/29

It would be nice to not need these subclasses, but the metaprogramming required (overriding `Object.const_get`) is a bit extreme.

Thoughts?

spec/features/admin_authorization_spec.rb
calebthompson: Oof, I'd much rather we didn't set ENV. We should stub it instead.

Here's the most important reason: If these tests fail, the reset line never gets called.

app/controllers/admin/application_controller.rb
graysonwright: 1. yep.
2. that makes sense - I'll change this up.

calebthompson:
The style with the logo + "Admin" (I assume) text is wonky. Can we make the logo a block element so that it doesn't cause the text to overflow?

app/controllers/admin/violations_controller.rb
calebthompson: I honestly don't have time to catch up with the discussion, so you'll have to excuse me if this has already been discussed, but what about just having the default implementation use the base controller for every model? Maybe add a blacklist configuration for models not to include?

app/dashboards/dashboard_manifest.rb
graysonwright: @calebthompson the Administrate API looks for the `DashboardManifest#dashboards` method. The only reason I extracted a constant was because Hound was complaining about overly long methods.

See [the generated code] for an idea of Administrate's API.

[the generated code]: https://github.com/thoughtbot/hound/commit/7005353bb813dd136415473ef6134ce09a2c63a9#diff-97c4bbd36554739ad38ce3d1f90830c7R3

app/models/repo.rb
graysonwright: will do.

graysonwright:
@calebthompson agreed on the logo + text. I'll remove the text completely in the next release.

app/dashboards/dashboard_manifest.rb
calebthompson: 1. Should we be generating methods Hound will disapprove of?
2. Are we okay with a method like this? I'd never let a dev get away with it, so should we be building Administrate to use the constant? Is there a solid use case for having a different set of dashboards each time you ask?

app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+      Customize this action to update/remove the limit,
+      or implement the pagination library of your choice."
+    @resources = @resources.limit(20)
+  end
+
+  private
+
+  def authenticate_admin
+    unless current_user &&
+        admin_github_handles.include?(current_user.github_username)
```
JoelQ: Could this be moved to a private method? The indentation is a little confusing here. Something like:

```ruby
unless current_user && github_admin?
  redirect_to :root
end
```

config/routes.rb
graysonwright: oh, that's interesting.

I assumed that devs would like the transparency of seeing exactly where these dashboards came from (`DashboardManifest` is a generated file, so the dev has full control over its contents).

I've been wanting to change up the routing system a bit anyway - let's discuss more on https://trello.com/c/iXv6MLU3/52-config-routes-api.

calebthompson: I didn't realize this was internal API.

graysonwright:
@calebthompson:

> All of the Admin::BaseDashboard subclasses share a similar pattern.

> 1. I think we shouldn't be defining methods at all and just use constants for these, but:
> 2. If we must keep the methods, they should all be memoized.

What would memoization get us in this case? All of the methods are pretty simple, and none have heavy computation. I'm not opposed to memoization if it'll benefit us, but I try to avoid it when it has insignificant benefits. I don't know enough about the performance of Ruby constant lookups, etc. to make a call on that.

calebthompson:
I lack clarity into how many times these things are called, but if it's for every page load on the admin side, I'd say we might want the memoization (trade memory for a couple of stack hits).

I suspect the difference is small overall so it may not matter, just a gut reaction.

graysonwright:
also, to respond to your first point:
> I think we shouldn't be defining methods at all and just use constants for these

I'm a bit hesitant to create an API that relies only on constants.
1. I've never seen it done before.
2. It seems like it would limit the flexibility of the API substantially.

But, you're right - so far we're only using these methods to return constant data, and it does look like a bit of a smell to have so many methods acting simply as wrappers around constants.

> Is there a solid use case for having a different set of dashboards each time you ask?

The only one I can think of would be displaying different dashboards to users with different roles (admin, owner, etc).

- - -

I'd like to get a bit more information about how Administrate will be used before making big changes to this API. I'd hate to switch it only to have someone need to dynamically change dashboards.

app/controllers/admin/application_controller.rb
graysonwright: Good call.

config/routes.rb
graysonwright: yeah, I like that. I'll add it to the Trello card for a future release.

calebthompson:
> I'd like to get a bit more information about how Administrate will be used before making big changes to this API. I'd hate to switch it only to have someone need to dynamically change dashboards.

Isn't that why we're here?

I think we should default to a simpler solution (constants, assume no differences).

For the owner/admin difference, we could do different routing right? I feel like I would prefer:

```
class OwnerUserDashboard
  FIELDS=ALL_OF_THEM
end

class UserDashboard
  FIELDS=SOME_OF_THEM
end
```

to

```
class UserDashboard
  def fields
    if current_user.owner?
      ALL_OF_THEM
    else
      SOME_OF_THEM
    end
  end
end
```

That's the point of polymorphism, one of the big benefits to using the classes provided by Administrate, right?

spec/features/admin_authorization_spec.rb
graysonwright: Good point. reverting to the original stubs.

app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Admin::ApplicationController < Administrate::ApplicationController
+  before_action :authenticate_admin
+
+  def index
+    super
+
+    flash[:alert] =
+      "For performance, Administrate limits the index page to show 20 items.
+      Customize this action to update/remove the limit,
+      or implement the pagination library of your choice."
```
gylaz: Could this flash be removed?

app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  def github_admin?
+    current_user && admin_github_handles.include?(current_user.github_username)
+  end
+
+  def current_user
+    @current_user ||= User.find_by(remember_token: session[:remember_token])
+  end
+
+  def admin_github_handles
+    ENV.fetch("ADMIN_GITHUB_HANDLES", "").split(",")
```
gylaz: We call the field `github_username`, so for consistency we should either go with `ADMIN_GITHUB_HANDLES` or maybe just `ADMINS`?

app/controllers/admin/style_configs_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Admin::StyleConfigsController < Admin::ApplicationController
+end
```
gylaz: How about we remove this one? Not really any data here yet.

app/controllers/admin/memberships_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Admin::MembershipsController < Admin::ApplicationController
+end
```
gylaz: Probably not useful to have this tab.

app/dashboards/build_dashboard.rb
```diff
@@ -0,0 +1,55 @@
+require "administrate/base_dashboard"
```
gylaz: This doesn't get autoloaded?

app/dashboards/build_dashboard.rb
```diff
@@ -0,0 +1,55 @@
...
+    file_reviews: Field::HasMany,
+    violations: Field::HasMany,
+  }
+
+  def attribute_types
+    ATTRIBUTE_TYPES
+  end
+
+  def table_attributes
+    ATTRIBUTES.first(4)
```
gylaz: What is the significance of this?

gylaz:
@graysonwright There is a lot of code that is generated. Is it possible for Administrate to use sane defaults, and if there is need to override then a file(s) is generated for the developer?

thorncp:
I feel like the only model that's useful for us to have in an admin dashboard is `BulkCustomer`. Is anyone manually editing other records than that?

gylaz:
@thorncp I reset users that are stuck refreshing too.

graysonwright:
The most recent commit addresses all of the comments about constants & memoization.

app/dashboards/build_dashboard.rb
graysonwright: At the moment, it doesn't. We can look into autoloading it in the future.

thorncp: I don't think this is useful to have in a dashboard. As far as I know we don't manually view/edit these records and there's 712043 of them.

app/dashboards/file_review_dashboard.rb
```diff
@@ -0,0 +1,24 @@
+require "administrate/base_dashboard"
+
+class FileReviewDashboard < Administrate::BaseDashboard
```
thorncp: Same as the builds, but we have 3274641 of them; I don't think this is useful.

app/dashboards/membership_dashboard.rb
```diff
@@ -0,0 +1,23 @@
+require "administrate/base_dashboard"
+
+class MembershipDashboard < Administrate::BaseDashboard
```
thorncp: Same as builds, but we have 474381 of them; I don't think this is useful.

app/dashboards/repo_dashboard.rb
```diff
@@ -0,0 +1,32 @@
+require "administrate/base_dashboard"
+
+class RepoDashboard < Administrate::BaseDashboard
```
thorncp: Same as builds, but we have 351463 of them; I don't think this is useful.

app/dashboards/style_config_dashboard.rb
```diff
@@ -0,0 +1,22 @@
+require "administrate/base_dashboard"
+
+class StyleConfigDashboard < Administrate::BaseDashboard
```
thorncp: This isn't used yet, so it's probably not useful to have a dashboard for it.

app/dashboards/owner_dashboard.rb
```diff
@@ -0,0 +1,24 @@
+require "administrate/base_dashboard"
+
+class OwnerDashboard < Administrate::BaseDashboard
```
thorncp: This could be useful if it included some statistics for the owners. Active repo count... maybe something else?

app/dashboards/violation_dashboard.rb
```diff
@@ -0,0 +1,25 @@
+require "administrate/base_dashboard"
+
+class ViolationDashboard < Administrate::BaseDashboard
```
thorncp: Same as builds, but we have 10873298 of them; I don't think this is useful.

config/routes.rb
```diff
@@ -1,4 +1,16 @@
...
+  namespace :admin do
+    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
+      resources(
+        dashboard_resource,
+        controller: "application",
+        resource_class: dashboard_resource,
+      )
+    end
+
+    root to: "application#index", resource_class: :users
```
graysonwright: What is the most useful dashboard to see at the root?

Options are:
- :bulk_customers,
- :owners,
- :subscriptions,
- :users,

app/dashboards/owner_dashboard.rb
graysonwright: `active_repo_count` added in https://github.com/thoughtbot/hound/commit/f20ea0faa4b361b1960a6f8b50d1d499f30f237c

app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,38 @@
+class Admin::ApplicationController < Administrate::ApplicationController
+  before_action :authenticate_admin
+
+  def index
+    super
+
+    if params[:show] != "all" && @resources.count > 20
```
teoljungberg: Could we extract this into a local and well-named method?

app/models/owner.rb
```diff
@@ -9,4 +11,8 @@ def self.upsert(github_id:, name:, organization:)
     owner.save!
     owner
   end
+
+  def active_repo_count
```
teoljungberg: Is there a test for this?

app/models/owner.rb
```diff
@@ -9,4 +11,8 @@ def self.upsert(github_id:, name:, organization:)
...
+    repos.where(active: true).count
```
teoljungberg: `active` is already a defined scope on `Repo`. Could we use that?

config/routes.rb
teoljungberg: I think all but `owners`.

spec/features/admin_authorization_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "rails_helper"
+
+feature "Admin authorization" do
```
teoljungberg: All of these feature specs releates to the dashboard, could we rename the feature and the spec to reflect that?

app/models/owner.rb
graysonwright: `active` is defined as a class method, not as a scope. I can switch it to a scope, though.

teoljungberg: They are the same, you can defined scopes as class methods aswell as with the `scope` dsl syntax.

spec/features/admin_authorization_spec.rb
graysonwright: @teoljungberg all of the tests have `admin` in their names. We're not testing thee behaviors of any dashboard, just the `Admin::ApplicationController`.

I'd be alright with leaving them as-is. Can you suggest names you'd like better?

teoljungberg: I'm fine with leaving it as is in that case

app/models/owner.rb
graysonwright: There is now! https://github.com/thoughtbot/hound/pull/897/files#diff-80e069b9638d8d95fd3b901786602050R41

config/routes.rb
graysonwright: ok - It's currently set to `users`, let me know if I should change that.

app/models/repo.rb
```diff
@@ -13,9 +15,7 @@ class Repo < ActiveRecord::Base
   validates :full_github_name, uniqueness: true, presence: true
   validates :github_id, uniqueness: true, presence: true
 
-  def self.active
-    where(active: true)
-  end
+  scope :active, -> { where(active: true) }
```
teoljungberg: I think you missed my comment. `def self.active` is the same as `scope :active ...`. And our stylegudie says that we should not use the `scope` dsl

app/dashboards/bulk_customer_dashboard.rb
```diff
@@ -0,0 +1,16 @@
...
+  READ_ONLY_ATTRIBUTES = [
+    :id,
+    :created_at,
+    :updated_at,
+  ]
+
+  ATTRIBUTE_TYPES = {
+  }
+
+  TABLE_ATTRIBUTES = ATTRIBUTE_TYPES.keys.first(4)
```
jakecraige: These bottom 3 are the same for every dashboard, is there a way to extract that out?

app/dashboards/bulk_customer_dashboard.rb
```diff
@@ -0,0 +1,16 @@
+require "administrate/base_dashboard"
+
+class BulkCustomerDashboard < Administrate::BaseDashboard
+  READ_ONLY_ATTRIBUTES = [
+    :id,
+    :created_at,
+    :updated_at,
+  ]
+
+  ATTRIBUTE_TYPES = {
```
jakecraige: What does this do when it's empty?

app/dashboards/bulk_customer_dashboard.rb
```diff
@@ -0,0 +1,16 @@
...
+  }
```
thorncp: Can we add fields to this? I think these would be useful

```
    t.string   "org",                            null: false
    t.string   "description"
    t.string   "interval",                       null: false
    t.integer  "repo_limit"
    t.integer  "current_repos",      default: 0, null: false
```

config/routes.rb
thorncp: I think `bulk_customers` will probably see the most traffic.

app/dashboards/bulk_customer_dashboard.rb
graysonwright: yeah, this should have been filled in by the generator. Not sure why it was left out.

I'll add those fields.

gylaz:
@graysonwright What do you think about just adding one or two sections/panels for starters? `BulkCustomer` and `Users` that have `refreshing_repos` flag set to `true` would be the most useful right now. I think this would cut down the size of this PR considerable, and we can add more sections as we need them.

graysonwright:
@gylaz I talked to Chris and Scott, and we're cutting it back to just the `BulkCustomers` dashboard. The other tables are too big to use until we implement filtering.

app/dashboards/bulk_customer_dashboard.rb
graysonwright: Attributes added in https://github.com/thoughtbot/hound/commit/d626c482dc20bae472ab7fb2c386df8139bb080b

app/models/build.rb
```diff
@@ -1,4 +1,6 @@
 class Build < ActiveRecord::Base
+  include Administratable
```
thorncp: Should we remove these from the models we don't have dashboards for?

graysonwright: good call.

thorncp:
> Users that have refreshing_repos flag set to true

This is a neat idea for working around not having a filter.

graysonwright:
> >Users that have refreshing_repos flag set to true

> This is a neat idea for working around not having a filter.

Unfortunately, the query is currently hard-coded at `Model.all`. I'm planning to add support for Rails scopes soon, but let's leave users out until that gets added.

thorncp:
:+1: Looks great!

app/controllers/admin/application_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Admin::ApplicationController < Administrate::ApplicationController
+  before_action :authenticate_admin
+
+  private
+
+  def controller_path
+    params[:resource_class].to_s
+  end
```
gylaz: I'm curious why this isn't the default in Administrate?

spec/features/admin_authorization_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def stub_admin_github_usernames(usernames)
+    allow(ENV).to receive(:fetch).
+      with("ADMIN_GITHUB_USERNAMES", "").
+      and_return(usernames)
+  end
+
+  def have_admin_bulk_customers_header
+    have_selector("h1", text: "Bulk Customers")
```
gylaz: How about `have_css` instead?

gylaz:
A couple of minor things. Looks good to me!

app/controllers/admin/application_controller.rb
graysonwright: This is required to avoid subclassing `Admin::ApplicationController` for each resource class.

After looking at this use case, I think we will switch it so this is the default. We were playing with the idea before, but before seeing this in action it didn't seem like it added much value.

PR #899, opened 2015-08-25
Revert "Allow users to opt into more GitHub access"

### LINKED ISSUE #898: Default public level of access doesn't have privileges for webhooks or commit statuses

thorncp:
This reverts commit cdfcf5dafab24a3941d714fb52aca67759b2907d.

The `user:email` scope doesn't allow us to manage webhooks or commit
statuses, which means public-only users cannot enable Hound. We're gonna
have to experiment more to find a good balance of scopes and privacy,
but for now we need Hound to work.

Closes #898


thorncp:
Ignoring Hound since this is a revert.

teoljungberg:
Makes sense to me

PR #900, opened 2015-08-26
Reduce GitHub access necessary to start using Hound

salbertson:
Addresses https://github.com/thoughtbot/hound/issues/597 and https://github.com/thoughtbot/hound/issues/701.

The first time around we were missing `public_repo`. This change includes that scope and the original work.

Here are some screen shots taken during testing:

screenshot 2015-08-26 07 56 49
screenshot 2015-08-26 07 58 54
screenshot 2015-08-26 07 58 59
screenshot 2015-08-26 07 59 10
screenshot 2015-08-26 08 00 02
screenshot 2015-08-26 08 02 00
screenshot 2015-08-26 08 02 26

### DIFF
--- a/app/assets/javascripts/directives/repo_list.js.coffee.erb
+++ b/app/assets/javascripts/directives/repo_list.js.coffee.erb
@@ -3,6 +3,8 @@ App.directive 'repoList', ['Repo', 'Sync', 'User', '$timeout', (Repo, Sync, User
   templateUrl: '/templates/repo_list'
 
   link: (scope, element, attributes) ->
+    scope.includePrivateReposButtonText = '<%= I18n.t('include_private_repos') %>'
+
     loadRepos = ->
       scope.syncingRepos = false
 
--- a/app/services/repo_activator.rb
+++ b/app/services/repo_activator.rb
@@ -8,7 +8,13 @@ class RepoActivator
   end
 
   def activate
-    activate_repo.tap { enqueue_org_invitation }
+    activated = activate_repo
+
+    if activated
+      enqueue_org_invitation
+    end
+
+    activated
   end
 
   def deactivate
@@ -21,13 +27,20 @@ class RepoActivator
 
   def activate_repo
     change_repository_state_quietly do
-      add_hound_to_repo && create_webhook && repo.activate
+      if repo.private?
+        add_hound_to_repo && create_webhook && repo.activate
+      else
+        create_webhook && repo.activate
+      end
     end
   end
 
   def deactivate_repo
     change_repository_state_quietly do
-      remove_hound_from_repo
+      if repo.private?
+        remove_hound_from_repo
+      end
+
       delete_webhook && repo.deactivate
     end
   end
@@ -59,7 +72,7 @@ class RepoActivator
   end
 
   def enqueue_org_invitation
-    if repo.in_organization?
+    if repo.private? && repo.in_organization?
       AcceptOrgInvitationsJob.perform_later
     end
   end
--- a/spec/services/repo_activator_spec.rb
+++ b/spec/services/repo_activator_spec.rb
@@ -2,69 +2,194 @@ require "rails_helper"
 
 describe RepoActivator do
   describe "#activate" do
-    context "with org repo" do
-      it "will enqueue org invitation job" do
-        allow(AcceptOrgInvitationsJob).to receive(:perform_later)
-        repo = create(:repo, in_organization: true)
-        stub_github_api
-        activator = build_activator(repo: repo)
+    context "when repo is public" do
+      context "when repo belongs to an org" do
+        it "will not add Hound to repo" do
+          allow(AddHoundToRepo).to receive(:run)
+          repo = create(:repo, private: false, in_organization: true)
+          stub_github_api
+          activator = build_activator(repo: repo)
 
-        activator.activate
+          activator.activate
 
-        expect(AcceptOrgInvitationsJob).to have_received(:perform_later)
+          expect(AddHoundToRepo).not_to have_received(:run)
+        end
+
+        it "will not enqueue org invitation job" do
+          allow(AcceptOrgInvitationsJob).to receive(:perform_later)
+          repo = create(:repo, private: false, in_organization: true)
+          stub_github_api
+          activator = build_activator(repo: repo)
+
+          activator.activate
+
+          expect(AcceptOrgInvitationsJob).not_to have_received(:perform_later)
+        end
+
+        it "marks repo as active" do
+          repo = create(:repo, private: false, in_organization: true)
+          stub_github_api
+          activator = build_activator(repo: repo)
+
+          activator.activate
+
+          expect(repo.reload).to be_active
+        end
+
+        it "returns true" do
+          repo = create(:repo, private: false, in_organization: true)
+          stub_github_api
+          activator = build_activator(repo: repo)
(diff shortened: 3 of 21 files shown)


.ruby-style.yml
```diff
@@ -424,7 +424,7 @@ Style/TrailingBlankLines:
 Style/TrailingComma:
   Description: Checks for trailing comma in parameter lists and literals.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas
-  Enabled: true
+  Enabled: false
```
jakecraige: ?

thorncp:
For public repos under an organization, Hound is still trying create a Team to add @houndci to. The `public_repo` scope doesn't include privileges for that, so it fails to activate org repos. Chatted with @salbertson and we think we can skip creating the team for public repos.

your_repos___hound

.ruby-style.yml
gylaz: Nice try, Scott!

gylaz:
Is this pretty much the same PR as the one that was reviewed and merged, plus `public_repo`?

.ruby-style.yml
teoljungberg: Nice try, but that's not going to fly

salbertson:
@gylaz yea, pretty much. Like @thorncp said we talked about an issue with teams. I'm going to try and add Hound as a collaborator for open source repos, instead of creating team etc.

gylaz:
>  I'm going to try and add Hound as a collaborator for open source repos, instead of creating team etc.

I think some older organizations might not support this feature -- adding a user to an org repo as a collaborator. However, if I'm wrong, and GitHub has rolled this out to everyone, the I don't see a reason not to always add Hound as a collaborator to **all** repos, and rip out the team code.

spec/support/helpers/github_api_helper.rb
```diff
@@ -254,6 +254,21 @@ def stub_contents_request(options = {})
...
+      with(
+        headers: {
+          "Accept" => "application/vnd.github.v3+json",
+          "Authorization" => "token #{token}",
+          "Content-Type" => "application/json",
+          "User-Agent" => "Octokit Ruby Gem 4.0.0"
+        }
+      ).
+      to_return(
+        status: 200, body: "", headers: { "X-OAuth-Scopes" => scopes }
```
salbertson: I don't agree.

.ruby-style.yml
salbertson: Fine, removed.

spec/services/repo_activator_spec.rb
```diff
@@ -2,69 +2,194 @@
...
+          expect(AcceptOrgInvitationsJob).not_to have_received(:perform_later)
+        end
 
-        expect(AcceptOrgInvitationsJob).to have_received(:perform_later)
+        it "marks repo as active" do
+          repo = create(:repo, private: false, in_organization: true)
+          stub_github_api
+          activator = build_activator(repo: repo)
+
+          result = activator.activate
```
thorncp: Lies.

spec/services/repo_activator_spec.rb
```diff
@@ -2,69 +2,194 @@
...
+
+          expect(AcceptOrgInvitationsJob).not_to have_received(:perform_later)
+        end
+
+        it "marks repo as active" do
+          repo = create(:repo, private: false, in_organization: false)
+          stub_github_api
+          activator = build_activator(repo: repo)
+
+          result = activator.activate
```
thorncp: More lies.

thorncp:
LGTM and works for me on staging :+1:

spec/services/repo_activator_spec.rb
gylaz: But... why?

thorncp: I tried to repro locally with the rubocop cli but didn't succeed. Seems like a rubocop bug though.

salbertson:
@fotinakis here's the PR I was talking about.

app/assets/javascripts/directives/repo_list.js.coffee.erb
```diff
@@ -3,6 +3,8 @@ App.directive 'repoList', ['Repo', 'Sync', 'User', '$timeout', (Repo, Sync, User
   templateUrl: '/templates/repo_list'
 
   link: (scope, element, attributes) ->
+    scope.includePrivateReposButtonText = '<%= I18n.t('include_private_repos') %>'
```
gylaz: I'm not proposing this change in this PR, just wanted to share why using ERB for I18n like this is not a good idea (I found this out the hard way).

If we were to change this string in `en.yml` this coffeescript file would not pick up that change, because this file hasn't changed, so it doesn't re-compiled by the asset pipeline. Thus, the old string will still be there, until we modify this file, so it gets re-compiled.

Just something to keep in mind.

app/services/repo_activator.rb
```diff
@@ -59,7 +72,7 @@ def create_webhook
   end
 
   def enqueue_org_invitation
-    if repo.in_organization?
+    if repo.private && repo.in_organization?
```
gylaz: `private?`

gylaz:
:seal_of_approval:

app/services/repo_activator.rb
salbertson: Yes, done.

app/assets/javascripts/directives/repo_list.js.coffee.erb
salbertson: Ugh, good call. What's a better way to deal with this to keep text centralized in translations?

fotinakis:
Cool, thanks for the ping @salbertson!

Gave this a glance over, I think it will introduce the weird problem we talked about. The flow is: user logs in, you get a token with low permissions, user clicks a button and upgrade to more scopes, user logs out then logs back in—since you can't know who the user is before they login, you have to send them to GitHub with your default lower-permissions scopes, so they get yet another GitHub login screen that now says "Removed permissions" and they click authorize, then your higher-permissions token gets revoked and they give you another new token with the original lower scopes.

This may or may not be an issue for you, depending on if you care about retaining the higher-scoped token or just using it once. It can also introduce a bit of UI confusion because users see lots of GitHub authorizations screens, some adding permissions, then removing, then adding again, etc. Separate GitHub apps for authentication and for private repo access might be the way to go to avoid these problems.

app/assets/javascripts/directives/repo_list.js.coffee.erb
gylaz: The alternative I used is exporting variables from the view to JS/Coffee, like:
```haml
# app/views/repos/index.html.haml
...
:coffeescript
  Hound.Translations.privateReposText = "#{I18n.t('include_private_repos')}"
```

There's also https://github.com/fnando/i18n-js which I've heard good things about, but haven't tried.

salbertson:
@fotinakis, yea, there's still the weirdness with signing out. We are counting on the fact that users don't interact with houndci.com much once the service is enabled for private repos, but we'll see.

The private access token being invalidated when they sign in again isn't an issue since we fall back to our houndci user to make the requests, which is added as a collaborator during setup. This is not ideal though, since making requests from houndci count against our rate limit and we went through the trouble to introduced user specific tokens to avoid our rate limit.

Thanks for taking a look!

PR #901, opened 2015-08-26
Update 'contributing.md' to explain how to set up without using ngrok

ldamon:


CONTRIBUTING.md
```diff
@@ -19,6 +19,32 @@ Here are a few technical guidelines to follow:
...
+    `./bin/setup`
+
+1. Log into your GitHub account and go to your
+   [application settings].
+
+1. Under the Developer applications panel - Click on "Register new
+   application" and fill in the details:
+
+    * Application Name: Hound Development
+    * Homepage URL: `http://localhost:5000`
```
jakecraige: Are you able to authenticate with Github when setting it to this? Sub question, if this works, why are we using ngrok?

ldamon: Yes, I authenticated with Github.
@salbertson - can you explain why ngrok is needed?

gylaz: @jakecraige ngrok is so GitHub can reach a local machine.

gylaz:
This duplicates a lot of `Configure Hound on Your Local Development Environment` section.

`ngrok` is not a requirement. I was likely a mistake to make it seem that way. We should fix the existing section, and only add the optional `ngrok` steps.

CONTRIBUTING.md
jakecraige: Ah so you can authenticate via it, but PRs won't trigger a build?

ldamon:
Ok I'll take another stab at this!

CONTRIBUTING.md
salbertson: Right, authentication works but GitHub cannot make requests via webhook to start a build unless ngrock is used.

gylaz:
:+1:

PR #905, opened 2015-08-28
Upgrade RSpec to 3.3

teoljungberg:
- Specify the error that's expected to be raised

To quell warning:

> Using the `raise_error` matcher without providing a specific
> error or message risks false positives, since `raise_error`
> will match when Ruby raises a `NoMethodError`, `NameError` or
> `ArgumentError`, potentially allowing the expectation to pass
> without even executing the method you are intending to call.
> Instead consider providing a specific error class or message.
> This message can be supressed by setting:
>
> `RSpec::Expectations.configuration.warn_about_potential_false_positives =
false`


arunagw:
LGTM

PR #906, opened 2015-08-28
Fix pricing link

salbertson:
> Go to houndci.com, click on docs or FAQ link at the top, then click pricing link at the top. The #pricing gets added to the URL but that doesn't go anywhere unless you're on the initial houndci.com URL.


thorncp:
:+1:

PR #907, opened 2015-08-28
Change pricing description

salbertson:
Hopefully this makes pricing more clear. We've had questions about whether Hound subscription is for a single private repo or all private repos.

before:

screenshot 2015-08-28 08 22 39

after:

screenshot 2015-08-28 08 22 50


croaky:
Looks good

teoljungberg:
LGTM

PR #908, opened 2015-08-28
Set environment variables as constants

teoljungberg:
- Replace ENV with constants where applicable. Where there are
 fallbacks, I left ENV intact.
- Add missing env variables
- Setting constants from ENV vars in an initializer catches a missing ENV variable early (at boot time as opposed to runtime).
- This makes testing easier, since you can stub out an environment variable
by leveraging `stub_const`. i.e `stub_const("Hound::MY_ENV_VARIABLE",
"value")`


config/initializers/constants.rb
```diff
@@ -1,3 +1,5 @@
-module Hound
-  GITHUB_TOKEN = ENV.fetch("HOUND_GITHUB_TOKEN")
-end
+require "homeostasis"
+
+module Hound; end
```
teoljungberg: I thought about this:

To make the error messages clearer, we can define `const_missing` to give back an actionable error message like:
`BLAHBLAH is not defined, please define it in `env` or something along those lines.

Which would mean that Homeostasis would need to define that constant and have that being meta-programmatically generated.

config/initializers/constants.rb
```diff
@@ -1,3 +1,5 @@
...
+
+Homeostasis.call(Hound)
```
teoljungberg: This api doesn't fell 100%, any ideas on a clearer API?

lib/homeostasis.rb
```diff
@@ -0,0 +1,40 @@
...
+  def already_defined?(constant)
+    if target.const_get(constant)
+      true
+    end
+  rescue NameError
+    false
+  end
+
+  def load_dotenv
+    Dotenv.load
```
thorncp: Dotenv is only in the development and test groups in Gemfile, so this will fail in staging/production. Also, I _think_ this will only return variables defined in the various .env files and not those that exist in the environment, which would not pick up the variables defined on Heroku.

lib/homeostasis.rb
```diff
@@ -0,0 +1,40 @@
+require "dotenv"
+
+class Homeostasis
```
gylaz: Curious why this name was chosen? Doesn't seem to be descriptive, and there is already a gem by this name.

config/initializers/constants.rb
gylaz: Can we infer the root namespace of the app (in our case `Houndapp`) and nest the constants under that?

teoljungberg: Good call, lets!

lib/homeostasis.rb
teoljungberg: Oh, I didn't think off that. Good call, thanks

lib/homeostasis.rb
teoljungberg: Homeostasis means that the environment(actually the internal environment of the human body) is stable and constant. Which I found to be funny.

If you have a better-name. I'm all ears

gylaz: > Homeostasis means that the environment(actually the internal environment of the human body) is stable and constant. 

That's pretty clever!

I was thinking something simple and straight-forward like `EnvVars` or `Constants`.

salbertson:
What benefits besides using `stub_const` in tests do we get with this?

gylaz:
In general, setting constants from ENV vars in an initializer catches a missing ENV var early (at boot time as opposed to runtime).

I'm partial to the [current implementation](https://github.com/thoughtbot/hound/blob/1cc5f1bcf21227136f7abbef2643ab6f2988559c/config/initializers/constants.rb) because it makes it obvious how and where the constants are set. Even though we have to manually list them out, I don't expect there to be much churn in that file.

teoljungberg:
I'm fine with using the current implementation and hard-specifying the env. varaibles as constants. This was more an interesting idea that I thought I would share.

Should I modify the PR to use that instead?

teoljungberg:
When I hard-code the constants into `config/initializers/constants.rb` I get warnings like `warning: already initialized constant Hound::INTERCOM_API_SECRET`. 

We did not have that behaviour before, so what changed?

teoljungberg:
Setting the constants with `||=` instead of `=` solved it

gylaz:
@teoljungberg My guess would be that it only happens in development? Could it be Rails' autoloading reloading that file?

teoljungberg:
@gylaz That's my thought too. But we still don't want warnings everytime we run our tests. So I'll leave `||=` in

teoljungberg:
@gylaz Turns out - by removing the loading of the constants initializer - that warning went away. Updated in  5eb9dd1

gylaz:
@teoljungberg Do focused specs (ones that rely just on `spec_helper`) still work?

app/jobs/accept_org_invitations_job.rb
```diff
@@ -2,7 +2,7 @@ class AcceptOrgInvitationsJob < ApplicationJob
   queue_as :high
 
   def perform
-    github = GithubApi.new(Hound::GITHUB_TOKEN)
+    github = GithubApi.new(Hound::HOUND_GITHUB_TOKEN)
```
gylaz: Do you think it's redundant to have `Hound::HOUND_...`?

gylaz:
:+1:

teoljungberg:
@gylaz Turns out - they do not. And it feels redundant to stub the constants in each of the specs. 
I'll revert back to setting the constants with `||=`

app/controllers/builds_controller.rb
```diff
@@ -23,7 +23,7 @@ def ignore_confirmation_pings
   end
 
   def build_job_class
-    if payload.changed_files < ENV['CHANGED_FILES_THRESHOLD'].to_i
+    if payload.changed_files < Hound::CHANGED_FILES_THRESHOLD.to_i
```
thorncp: Since we're putting knowledge of env variables into a single file, what do you think about moving the `.to_i` in there as well?

app/helpers/analytics_helper.rb
```diff
@@ -1,6 +1,6 @@
 module AnalyticsHelper
   def analytics?
-    ENV["SEGMENT_KEY"]
+    Hound::SEGMENT_KEY.present?
```
thorncp: The key is `fetch`'ed without a default, so this `present?` check will always be true. I don't know if it'd be better to always try to log analytics (with bogus dev/test keys) or to not use `fetch` and keep the check. Thoughts?

app/controllers/builds_controller.rb
thorncp: I guess this depends on what we want `constants.rb` to do. If it's the basis for a potential gem that just does direct env variables to constants, then maybe it doesn't make sense to move it there. But if it is the keeper of env config then I think it does make sense. Another example of config logic is the `Hound::ENABLE_HTTPS == "yes"` that's in a few spots. If that's in the constants file we can do the `== "yes"` check there and put the result in the constant, giving us a simpler `if Hound::ENABLE_HTTPS` check. Thoughts

thorncp:
I brain-dumped some thoughts that may be irrelevant. LGTM

app/helpers/analytics_helper.rb
teoljungberg: Yes and no - This will allow the segment key to be set to `""`.

app/controllers/builds_controller.rb
teoljungberg: I like both of those ideas.

app/helpers/analytics_helper.rb
thorncp: Ah, makes sense. Should it be set to that in `.env` then?

spec/support/helpers/github_api_helper.rb
```diff
@@ -35,7 +35,7 @@ def stub_repo_request(repo_name, token)
     )
   end
 
-  def stub_repo_with_org_request(repo_name, token = hound_token)
+  def stub_repo_with_org_request(repo_name, token)
```
gylaz: Is this change related to this PR?

spec/lib/github_api_spec.rb
```diff
@@ -323,7 +323,7 @@
 
   describe "#add_repo_to_team" do
     it "makes a request to GitHub" do
-      token = "some_token"
+      token = Hound::GITHUB_TOKEN
```
gylaz: Changes in this file seem unrelated. The token doesn't have to be `Hound::GITHUB_TOKEN`, right?

spec/helpers/analytics_helper_spec.rb
```diff
@@ -1,17 +1,21 @@
...
-  it 'is true when ENV["SEGMENT_KEY"] is present' do
-    ENV["SEGMENT_KEY"] = "anything"
+describe AnalyticsHelper do
+  describe "#analytics?" do
+    context "when SEGMENT_KEY is present" do
+      it "returns true" do
+        stub_const("Hound::SEGMENT_KEY", "anything")
 
-    expect(analytics?).to be_truthy
+        expect(analytics?).to be_truthy
```
gylaz: The method now returns a boolean. Would `expect(analytics?).to eq true` be more explicit?

app/services/repo_activator.rb
```diff
@@ -84,11 +84,11 @@ def delete_webhook
   end
 
   def builds_url
-    URI.join("#{protocol}://#{ENV["HOST"]}", "builds").to_s
+    URI.join("#{protocol}://#{Hound::HOST}", "builds").to_s
   end
 
   def protocol
-    if ENV.fetch("ENABLE_HTTPS") == "yes"
+    if Hound::ENABLE_HTTPS == "yes"
```
gylaz: Sounded like we wanted to move the `== "yes"` into the constants initializer.

app/helpers/analytics_helper.rb
gylaz: @thorncp Are you proposing setting `SEGMENT_KEY` to `""` in `.env`?

I think what this change allows is *if* this key is set to `""` then we want to return `false`.

gylaz:
Added a few comments. This is looking good!

app/helpers/analytics_helper.rb
thorncp: Yeah, setting the value to `""` in `.env` makes sense to me. If it's a supported feature, it seems like we should do that instead of having a fake `"segment_key"` value in development.

gylaz: That would imply we want analytics to be disabled by default. Do we want that?

spec/support/helpers/github_api_helper.rb
teoljungberg: It's not - good call. Reverted

spec/lib/github_api_spec.rb
teoljungberg: You're right, reverted

spec/helpers/analytics_helper_spec.rb
teoljungberg: I agree, I fixed the formatting of the spec. I'll fix this in the same whirl

app/services/repo_activator.rb
teoljungberg: We do, I'll fix that. Thanks

app/controllers/application_controller.rb
```diff
@@ -10,7 +10,7 @@ class ApplicationController < ActionController::Base
   private
 
   def force_https
-    if ENV['ENABLE_HTTPS'] == 'yes'
+    if Hound::ENABLE_HTTPS == "yes"
```
thorncp: This can be `if Hound::ENABLE_HTTPS` now, right?

app/models/repo.rb
```diff
@@ -61,7 +61,7 @@ def stripe_subscription_id
 
   def bulk?
     BulkCustomer.where(org: organization).any? ||
-      ENV.fetch("EXEMPT_ORGS", "").split(",").include?(organization)
+      Hound::EXEMPT_ORGS.split(",").include?(organization)
```
thorncp: In the same vein as doing the enable https comparison in constants.rb, what do you about spitting this into an array in there as well?

app/views/application/_analytics.haml
```diff
@@ -1,6 +1,6 @@
 :javascript
   window.analytics||(window.analytics=[]),window.analytics.methods=["identify","track","trackLink","trackForm","trackClick","trackSubmit","page","pageview","ab","alias","ready","group","on","once","off"],window.analytics.factory=function(t){return function(){var a=Array.prototype.slice.call(arguments);return a.unshift(t),window.analytics.push(a),window.analytics}};for(var i=0;i<window.analytics.methods.length;i++){var method=window.analytics.methods[i];window.analytics[method]=window.analytics.factory(method)}window.analytics.load=function(t){var a=document.createElement("script");a.type="text/javascript",a.async=!0,a.src=("https:"===document.location.protocol?"https://":"http://")+"d2dq2ahtl5zl1z.cloudfront.net/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(a,n)},window.analytics.SNIPPET_VERSION="2.0.8",
-  window.analytics.load("#{ENV['SEGMENT_KEY']}");
+  window.analytics.load("#{Hound::SEGMENT_KEY}");
```
thorncp: Does this need interpolation?

app/models/repo.rb
gylaz: I'd say this is legacy, needs to be removed, and likely not worth any more time spent on it.

thorncp: :+1:

thorncp:
I like how this turned out!

app/views/application/_analytics.haml
gylaz: Yep, otherwise it will get compiled to `window.analytics.load(abcdeft123456);`

config/initializers/constants.rb
```diff
@@ -1,3 +1,14 @@
 module Hound
-  GITHUB_TOKEN = ENV.fetch("HOUND_GITHUB_TOKEN")
+  CHANGED_FILES_THRESHOLD ||= ENV.fetch("CHANGED_FILES_THRESHOLD").to_i
+  ENABLE_HTTPS ||= ENV.fetch("ENABLE_HTTPS") == "yes"
```
gylaz: Worth renaming to `HTTPS_ENABLED`, otherwise it kind of sounds like an action.

spec/support/constants.rb
```diff
@@ -0,0 +1,10 @@
+require "dotenv"
+Dotenv.load
+
+begin
+  if Hound.constants.none?
```
gylaz: Where does `.constants` method come from?

app/views/application/_analytics.haml
thorncp: Oh, haml :(

spec/support/constants.rb
thorncp: http://ruby-doc.org/core-2.2.3/Module.html#method-c-constants

app/controllers/application_controller.rb
teoljungberg: It sure can, done

config/initializers/constants.rb
teoljungberg: Good call, done

app/helpers/analytics_helper.rb
gylaz: @thorncp Any thoughts here?

thorncp: I feel like analytics should be disabled in dev/test and if the only way to do that is an empty env variable, then it should default to empty. It's not a strong feeling though, and if the library can gracefully handle a junk value then it's fine staying like it is.

gylaz: :+1: to `""`.

teoljungberg:
Thanks for reviewing @gylaz and @thorncp

PR #910, opened 2015-08-29
Configure OmniAuth to use Rail's logger

thorncp:
The default is STDOUT, which is inconsistent with other logs and was
filling test output with irrelevant messages.


gylaz:
:+1:

PR #912, opened 2015-08-31
Save file reviews before queueing a review job

gylaz:
Why:

* SCSS worker is very fast and sometimes performs a review before the
  build and file reviews are saved.
  This will case the complete job to fail, since it won't find the
  file review.

Addresses: https://app.getsentry.com/hound-1/production/group/70334013

### DIFF
--- a/app/models/style_guide/base.rb
+++ b/app/models/style_guide/base.rb
@@ -1,7 +1,12 @@
 module StyleGuide
   class Base
-    pattr_initialize :repo_config, :repository_owner_name
     attr_implement :file_review
+
+    def initialize(repo_config:, build:, repository_owner_name:)
+      @repo_config = repo_config
+      @build = build
+      @repository_owner_name = repository_owner_name
+    end
 
     def enabled?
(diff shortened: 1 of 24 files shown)


app/models/style_guide/base.rb
```diff
@@ -1,8 +1,25 @@
 module StyleGuide
   class Base
-    pattr_initialize :repo_config, :repository_owner_name
```
thorncp: :+1:

thorncp:
The refactorings are kinda hiding the bug fix. Would it be too much trouble to separate it?

Looks great :+1:

gylaz:
@thorncp Good point. I realized it once it was done :grin:. I'll take a quick stab at separating.

gylaz:
@thorncp Another look? I've left the refactoring of tests because it made it easier to make the change in tests (not as many places to update for new params).

thorncp:
Definitely easier to see the fix now. I only re-reviewed the /app code, but LGTM

PR #913, opened 2015-09-01
Extract duplicate file_review method into Base

gylaz:
Why:

* All service-based style guides enqueue almost identical jobs.
* We can derive what is different from the class itself.
* Less code to add when we add more service-based linters.

### DIFF
--- a/app/models/style_guide/base.rb
+++ b/app/models/style_guide/base.rb
@@ -8,6 +8,18 @@ module StyleGuide
       @repository_owner_name = repository_owner_name
     end
 
+    def file_review(commit_file)
+      attributes = build_review_job_attributes(commit_file)
+      file_review = FileReview.create!(
+        build: build,
+        filename: commit_file.filename,
+      )
+
+      Resque.enqueue(job_class, attributes)
+
+      file_review
+    end
+
     def enabled?
       repo_config.enabled_for?(name)
     end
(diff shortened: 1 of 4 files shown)


app/models/style_guide/base.rb
```diff
@@ -22,6 +34,25 @@ def file_included?(*)
 
     attr_reader :repo_config, :build, :repository_owner_name
 
+    def build_attributes(commit_file)
```
salbertson: What do you think about using a more descriptive method name? Maybe `build_review_job_attributes`?

salbertson:
I like it. :+1:

thorncp:
This makes a lot of sense for us moving to everything-as-a-service :+1:

app/models/style_guide/base.rb
gylaz: Done.

app/models/style_guide/scss.rb
```diff
@@ -2,25 +2,6 @@ module StyleGuide
...
-        pull_request_number: commit_file.pull_request_number,
-        patch: commit_file.patch,
-        content: commit_file.content,
-        config: repo_config.raw_for(LANGUAGE)
-      )
-
-      file_review
-    end
-
     def file_included?(_)
```
jakecraige: Thoughts on setting a default for this to `true` so these subclasses don't have to define it? Go seems to have an exceptional case where it needs to override it.

gylaz: I'd welcome that PR :)

PR #914, opened 2015-09-01
Remove the NewRelic gem

gylaz:
Why:

* There is indication that NewRelic is the cause of our workers to get
  stuck. [source](https://github.com/resque/resque/issues/180)
* We haven't had the need for NewRelic in quite some time.
* We can try alternatives like Skylight, when the need arises.


thorncp:
:sheep:

PR #918, opened 2015-09-03
Fetch file contents when file is being reviewed

gylaz:
Why:

* We were fetching content for all changed files,
  even ones that we won't review because of their unsupported extension.
* Because of this we were wasting API requests on such files.
* There was an issue where a file was named `[:facebook]`
  and we tried to fetch its content.
  Because, by GitHub's design, file path is part of th URL,
  it was raising an error, because that is not a valid URL

### DIFF
--- a/app/jobs/completed_file_review_job.rb
+++ b/app/jobs/completed_file_review_job.rb
@@ -17,11 +17,9 @@ class CompletedFileReviewJob
       filename: attributes.fetch("filename")
     )
     commit_file = CommitFile.new(
-      filename: file_review.filename,
-      content: "",
       patch: attributes.fetch("patch"),
-      pull_request_number: attributes.fetch("pull_request_number"),
-      sha: build.commit_sha
+      filename: nil,
+      commit: nil,
     )
 
     attributes.fetch("violations").each do |violation|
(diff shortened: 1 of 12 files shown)


app/jobs/completed_file_review_job.rb
```diff
@@ -17,11 +17,9 @@ def self.perform(attributes)
       filename: attributes.fetch("filename")
     )
     commit_file = CommitFile.new(
-      filename: file_review.filename,
-      content: "",
       patch: attributes.fetch("patch"),
-      pull_request_number: attributes.fetch("pull_request_number"),
-      sha: build.commit_sha
+      filename: nil,
```
thorncp: I don't like these `nil`s here. What do you think about using `Patch` directly instead of creating a partial `CommitFile`? Looks like all that's being used is `#line_at` which is mostly a delegation to `Patch` with graceful `UnchangedLine` handling.

thorncp:
One question, but LGTM

app/jobs/completed_file_review_job.rb
gylaz: That is exactly the thought I had (moving `#line_at` into `Patch`), but I didn't want to muddle this PR with more re-factors. What do you think about a followup PR for this work?

thorncp: :+1:

gylaz:
Looks good on staging.

PR #920, opened 2015-09-05
Delete config/newrelic.yml

JuanitoFatas:
Follow up of #914.


thorncp:
Nice catch, thanks! :eyes: :mag:

PR #924, opened 2015-09-12
Simplify adding new linters

teoljungberg:
- Rename Styleguides to Linters
- Determine which Linter to use with `MyLinter.lint?(filename)`
- Update javascript regexp to disallow `foo.coffee.js`


app/models/style_guide/base.rb
```diff
@@ -41,14 +41,14 @@ def build_review_job_attributes(commit_file)
       }
     end
 
-    def job_class
-      "#{language.capitalize}ReviewJob".constantize
-    end
-
```
gylaz: What were the issues with this method as is?

gylaz:
`custom_linter` could imply that user can configure any linter he'd like. In reality, users choose from a few that we support. Perhaps a clearer name would be something like `specified_linter` or `configured_linter`?

app/models/style_guide.rb
```diff
@@ -0,0 +1,24 @@
+module StyleGuide
+  def self.for(filename)
```
teoljungberg: This was moved from `StyleChecker` into it's own class.

We can have a each linter determine if it should lint that specific file. Which
I think would make it easier for new languages to be added.
I imagine something like:

```ruby
module StyleGuide
  class Javascript < Base
    def self.lint?(filename)
      # check if `filename` should be linted
    end
  end
end

# In StyleChecker

private

LINTERS = [
  StyleGuide::Javascript,
  # ... more linters
]

def style_guide_class(filename)
  LINTERS.detect { |linter| linter.lint?(filename) }
end
```

Thoughts?

app/models/style_guide/base.rb
teoljungberg: It felt wrong specifying the "language` as i.e "es_lint", it's a custom linter - not it' own language.
Then I thought it made more sense for each styleguide to specifh which job-class it should use.

What is your take in this?

teoljungberg:
@gylaz Agreed, `custom_linter` felt off. What do you think about sticking to just `linter`?

thorncp:
>     javascript:
>       linter: custom_linter

What do you think about inverting this structure? E.g.

```yaml
    custom_linter:
      file_patterns:
        - *.js
	- *.es6

**Comment by gylaz** [comment]
I'm a fan of the idea of having the linter being the top-level! However, I'm not sure I like having the users specify the patterns. What happens if they need to configure Ruby on multiple repos?
```yaml
rubocop:
  file_patterns:
    - *.rb
    - *.ru
    - Gemfile
    - Rakefile
``` 
In each project?

What happens if they add a non-ruby file accidentally to that pattern? Will RuboCop raise exceptions? How will other linters handle such a situation?

Is there some middle-ground where Hound decides what file extensions to lint for each linter, but the user still specifies which linter to use?

```yaml
eslint:
  enabled: true
  config_file: eslint_config.json

rubocop:
  enabled: true
  config_file: .rubocop.yml
```

This would imply we don't enable everyone on by default. That would take away from out-of-the-box value experience Hound provides to new users. Perhaps we could enable some linters by default, and users can disable the defaults and enable what they prefer?

app/models/style_guide.rb
gylaz: I think this makes sense for each linter class to know wether or not it cares about a file. In the case of Markdown there are a bunch of extensions that are valid and listing all of these in `StyleGuide` doesn't feel as clean as listing them in `StyleGuide::Markdown`.

app/models/style_guide/base.rb
gylaz: I think we should move to `Linter::EsLint` structure now that we are not dealing with languages anymore.

gylaz: It would be nice to reduce the amount of variables/constants/methods specified each time a new linter is added.

spec/models/style_checker_spec.rb
```diff
@@ -1,236 +1,242 @@
...
-    violated_commit_file = stub_commit_file("bad.rb", "def bad( a ); a; end  ")
-    pull_request = stub_pull_request(
-      commit_files: [stylish_commit_file, violated_commit_file]
-    )
-    expected_violations = [
-      "Unnecessary spacing detected.",
-      "Space inside parentheses detected.",
-      "Trailing whitespace detected.",
-    ]
+describe StyleChecker do
```
gylaz: I've merge this reformatting into master as a stand alone PR. So, a rebase should quiet some of this noise.

spec/models/style_guide/java_script_spec.rb
```diff
@@ -163,16 +163,16 @@
...
       style_guide = build_style_guide(repo_config: repo_config)
       commit_file1 = double(
         "CommitFile",
-        filename: "app/assets/javascripts/bar.js"
+        filename: "app/assets/javascripts/bar.js",
       )
       commit_file2 = double(
         "CommitFile",
-        filename: "vendor/assets/javascripts/foo.js"
+        filename: "vendor/assets/javascripts/foo.js",
```
gylaz: These aren't really related. What do you think about taking them out to reduce the noise for this PR?

spec/models/style_guide_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        filename = "hello.rb"
+
+        klass = build_style_guide(filename: filename)
+
+        expect(klass).to eq StyleGuide::Ruby
+      end
+    end
+
+    context "given a swift file" do
+      it "returns `StyleGuide::Swift`" do
```
gylaz: Thoughts on keeping these doc strings more generic, like `returns the swift class`? Otherwise, it ties these docs to the implementation -- more places to rename if we rename the module, or prefix the class name.

thorncp:
> Is there some middle-ground where Hound decides what file extensions to lint for each linter, but the user still specifies which linter to use?

I think that's a good middle ground. After putting more thought into it, my biggest reservation was how to handle multiple linters linting the same file types, which can be accomplished without the user specifying the file patterns.

> Perhaps we could enable some linters by default, and users can disable the defaults and enable what they prefer?

We could tell them what's enabled by default when they activate a repo, so it's not a surprised.

app/models/style_guide/base.rb
teoljungberg: @gylaz Sadly, I think that's inevitable. Currently - you have to define 2 constants(which I think is fair)

app/models/linter/base.rb
```diff
@@ -1,5 +1,13 @@
-module StyleGuide
+module Linter
   class Base
+    def self.lint?(filename)
+      if filename =~ self::FILE_REGEXP
```
thorncp: Shorten this to `filename =~ self::FILE_REGEXP` ?

app/models/linter/coffee_script.rb
```diff
@@ -1,8 +1,9 @@
 # Determine CoffeeScript style guide violations per-line.
-module StyleGuide
+module Linter
   class CoffeeScript < Base
     DEFAULT_CONFIG_FILENAME = "coffeescript.json"
     ERB_TAGS = /<%.*%>/
+    FILE_REGEXP = /.+\.coffee(\.js)?(\.erb)?\z/
```
thorncp: What's the benefit of the leading `.+`?

app/models/linter/markdown.rb
```diff
@@ -0,0 +1,7 @@
+module Linter
+  class Markdown < Base
+    FILE_REGEXP = /.+(?:\.md|\.markdown)\z/
```
thorncp: Do we need to specify the non-capturing group? The CoffeeScript one doesn't use it and it doesn't seem like we look at any capture groups.

app/models/linter/unsupported.rb
```diff
@@ -1,8 +1,12 @@
 # Returns empty set of violations.
-module StyleGuide
+module Linter
   class Unsupported < Base
     class CannotReviewUnsupportedFile < StandardError; end
 
+    def self.lint?(*)
+      true
```
thorncp: Should this be `false`? It seems like a class called `Unsupported` would not respond to true to this.

app/models/style_checker.rb
```diff
@@ -2,61 +2,49 @@
 # Builds style guide based on file extension.
 # Delegates to style guide for line violations.
 class StyleChecker
-  def initialize(pull_request, build)
-    @pull_request = pull_request
-    @build = build
-    @style_guides = {}
+  pattr_initialize :pull_request, :build do
```
thorncp: I much prefer an explicit initialize method over this. What does `pattr` mean?

app/models/style_checker.rb
```diff
@@ -2,61 +2,49 @@
...
-      StyleGuide::Python
-    when /.+\.swift\z/
-      StyleGuide::Swift
-    when /.+(?:\.md|\.markdown)\z/
-      StyleGuide::Markdown
-    else
-      StyleGuide::Unsupported
-    end
+  def find_linter_class(filename)
+    LINTERS.detect { |linter_class| linter_class.lint?(filename) }
```
thorncp: What do you think about having this return all linters that can lint the file, so we're ready to handle multiple linters per file in the future?

app/models/style_checker.rb
teoljungberg: pattr == "private attr"
Using `attr_extras` is something that we're already using in Hound.

Do you think we should move away from it?

app/models/linter/markdown.rb
teoljungberg: I really just stole this from the regexp that's on master - I'll try simplifying it

app/models/linter/coffee_script.rb
teoljungberg: I really just stole this from the regexp that's on master - I'll try simplifying it

app/models/linter/base.rb
teoljungberg: Since it's a predicate method - I want it to explicitly return true or false.

What would make this clearer? Is there something nice regexpy we can do for that to be clearer?

app/models/style_checker.rb
thorncp: I'm not a fan of it, so my vote would be to move away from it. That should probably be a team decision though, and what you have is fine for now.

app/models/style_checker.rb
teoljungberg: Good call, sounds good

app/models/linter/base.rb
thorncp: Ah, ok. Could use [`Regexp#===`](http://ruby-doc.org/core-2.1.1/Regexp.html#method-i-3D-3D-3D).

teoljungberg: Awesome, didn't know of `#===`. Thank you

app/models/linter/coffee_script.rb
gylaz: To make sure we're not linting a file named just `.coffee`? Not sure if there's any value in that though.

app/models/linter/markdown.rb
gylaz: The idea was to be explicit that nothing is meant to be done with the capture group.

app/models/linter/python.rb
```diff
@@ -9,7 +11,7 @@ def file_review(commit_file)
       )
 
       Resque.push(
-        "python_review",
+        JOB_CLASS,
```
gylaz: Using `JOB_CLASS` here is misleading, this is the queue name.

app/models/style_checker.rb
gylaz: I don't mind the simple case scenario of using `attr_extras`, like `pattr_intitialize :foo, :bar`. It's all the other use-cases that quickly start to add extra methods to remember.

app/models/style_checker.rb
```diff
@@ -2,61 +2,57 @@
...
+    Linter::JavaScript,
+    Linter::Python,
+    Linter::Ruby,
+    Linter::Scss,
+    Linter::Swift,
+    Linter::Unsupported,
+  ]
+
+  def lint(commit_file)
+    build_linters(commit_file.filename).each do |linter|
```
gylaz: We'll run `build_linters` for each file. Do we want to memoize?

app/models/style_checker.rb
```diff
@@ -2,61 +2,57 @@
...
+      linter.file_review(commit_file)
+    end
+  end
 
   def commit_files_to_check
     pull_request.commit_files.select do |file|
-      file_style_guide = style_guide(file.filename)
-      file_style_guide.enabled? && file_style_guide.file_included?(file)
+      build_linters(file.filename).any? do |linter|
```
gylaz: Does it makes sense to lint every file, and have the linter decide not to do anything on a non-relevant file?

app/models/style_checker.rb
```diff
@@ -2,61 +2,57 @@
...
-  def style_guide(filename)
-    style_guide_class = style_guide_class(filename)
-    style_guides[style_guide_class] ||= style_guide_class.new(
-      repo_config: config,
-      build: build,
-      repository_owner_name: pull_request.repository_owner_name,
-    )
+  def build_linters(filename)
+    find_linter_classes(filename).map do |linter_class|
+      linters[linter_class] ||= linter_class.new(
```
gylaz: What do you think about not setting the linters dictionary inside of a `map`?

spec/models/linter/base_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  describe "#file_included?" do
+    it "returns true" do
+      style_guide = Linter::Test.new(
+        repo_config: double,
+        build: double,
+        repository_owner_name: "foo",
+      )
+
+      expect(style_guide.file_included?(double)).to eq true
```
gylaz: Is this test providing any value?

spec/models/linter/coffee_script_spec.rb
```diff
@@ -1,8 +1,42 @@
 require "rails_helper"
 
-describe StyleGuide::CoffeeScript do
+describe Linter::CoffeeScript do
   include ConfigurationHelper
 
+  describe ".lint?" do
+    context "given a `coffee` file" do
```
gylaz: What do you think about `given a file with .coffee extension`?

gylaz:
There is a lot of changes here. I don't understand why we need these changes. We've only talked about supporting JSHint and JSCS, but we are focusing on ESLint for now. Is this PR premature?

app/models/repo_config.rb
```diff
@@ -48,6 +48,14 @@ def for(language)
     end
   end
 
+  def custom_linter?(language)
+    enabled_for?(language) && custom_linter(language).present?
+  end
+
+  def custom_linter(language)
```
thorncp: Are these two methods still used?

app/models/style_guide/base.rb
gylaz: How about renaming language to `linter`? Then we can derive the job class like `"#{linter.capitalize}ReviewJob".constantize`.

spec/models/linter/base_spec.rb
teoljungberg: The default return value of `file_included?` is true, and in some sub-classes of `Base` we're not overriding it. So I would say it is.

Also, this test was already present in `spec/models/style_guide/test_spec.rb`. I just renamed it to follow naming conventions

app/models/style_checker.rb
teoljungberg: @thorncp I'll leave the current behaviour in for now

app/models/linter/unsupported.rb
teoljungberg: `Unsupported` is meant to be run if no other linters are found. So we want it to return `true` for that to take effect.

app/models/linter/base.rb
```diff
@@ -41,16 +45,20 @@ def build_review_job_attributes(commit_file)
...
-    end
-
     def enqueue_job(attributes)
       Resque.enqueue(job_class, attributes)
     end
 
     def language
-      self.class::LANGUAGE
+      self.class::NAME
+    end
```
gylaz: How about `alias_method` here instead of adding `linter_name`?

app/models/linter/markdown.rb
gylaz: @thorncp Is there a way to write this `or` regex without using a capture group?

app/models/style_checker.rb
```diff
@@ -2,61 +2,49 @@
...
-    @style_guides = {}
+  pattr_initialize :pull_request, :build do
+    @linters = {}
   end
+  attr_private :linters
 
   def review_files
     commit_files_to_check.each do |commit_file|
-      style_guide(commit_file.filename).file_review(commit_file)
+      build_linter(commit_file.filename).file_review(commit_file)
```
gylaz: I think it might be a smell that we're calling `build_linter` here and on line 32.
What do you think about something like this?
```rb
pull_request.commit_files.each do |file|
  linter = build_linter(file.filename)
  if linter.enabled? && linter.file_included?(file)
    linter.file_review(commit_file)
  end
end
```

app/models/style_checker.rb
gylaz: Will this loop be run for every changed file in a commit? Is that going to be a lot less efficient than the previous `case` statement? If so, do we even care about the loss of performance?

spec/models/linter/coffee_script_spec.rb
```diff
@@ -1,8 +1,42 @@
 require "rails_helper"
 
-describe StyleGuide::CoffeeScript do
+describe Linter::CoffeeScript do
   include ConfigurationHelper
 
+  describe ".lint?" do
+    context "given a .coffee file" do
+      it "returns true" do
+        result = Linter::CoffeeScript.lint?("foo.coffee")
```
gylaz: Worth using `described_class` here?

app/models/linter/markdown.rb
thorncp: I think `/\.md|\.markdown\z/` would work.

app/models/linter/python.rb
```diff
@@ -8,7 +9,6 @@ def enqueue_job(attributes)
       Resque.push(
         "python_review",
         {
-          class: "review.PythonReviewJob",
```
gylaz: Is this intentional? I think we still want this for Python's Resque sake.

teoljungberg: A miss from a merge-conflict, thank you

spec/models/linter/coffee_script_spec.rb
teoljungberg: The use of `described_class` is not per-se against our styleguide, but it got voted against in https://github.com/thoughtbot/guides/pull/256.

I would prefer not having it here

app/models/style_checker.rb
teoljungberg: I like it, it simplifies the class alot. Thank you

app/models/linter/base.rb
teoljungberg: Another merge conflict miss, I'll remove it in favor of `linter_name`

app/models/style_checker.rb
teoljungberg: After doing a quick benchmark I got this result:

```
Calculating -------------------------------------
      case-statement     6.335k i/100ms
              detect     8.275k i/100ms
-------------------------------------------------
      case-statement     82.364k (±19.1%) i/s -    386.435k
              detect    137.070k (± 7.1%) i/s -    686.825k

Comparison:
              detect:   137069.6 i/s
      case-statement:    82364.1 i/s - 1.66x slower
```

With this benchmark:
```ruby
$: << "."
require "benchmark/ips"
require "config/environment"

Benchmark.ips do |bench|
  bench.report("case-statement") do
    style_checker.send(:find_linter_class, filename)
  end

  bench.report("detect") do
    style_checker.send(:case_statement, filename)
  end

  bench.compare!

  def style_checker
    pull_request = "pull_request"
    build = "build"
    StyleChecker.new(pull_request, build)
  end

  def filename
    "foo.rb"
  end
end
```

Freezing `StyleChecker::LINTERS` brought it down a bit more, but it's still
slower.

Is this a blocker? I think that it's a major win for the architecture of the
code and it's maintainability.

gylaz: Not a blocker, just wanted to open up a discussion to see if there's reason for concert here.

You have `detect` report using `case_statement` method, is that backwards?

teoljungberg: Hah, I mixed up the stats in my head. `.detect` is faster than the case-statement.

app/models/linter/scss.rb
```diff
@@ -0,0 +1,6 @@
+module Linter
+  class Scss < Base
+    FILE_REGEXP = /.+\.scss\z/
```
seanpdoyle: Do we not support SASS?

teoljungberg: As far as I know, this is a limitation of the linter we use - https://github.com/brigade/scss-lint <https://github.com/brigade/scss-lint>

app/models/linter/coffee_script.rb
mcmire: Do these need to be regexes? What do you think about making this an array of extensions and then using the regex later?

app/models/linter/java_script.rb
```diff
@@ -1,6 +1,7 @@
-module StyleGuide
+module Linter
   class JavaScript < Base
     DEFAULT_CONFIG_FILENAME = "javascript.json"
+    FILE_REGEXP = /.+(?<!\.coffee)\.js\z/
```
mcmire: What's the `(?<!\.coffee)` part for?

app/models/style_checker.rb
mcmire: I am a big fan of attr_extras, but I might just use a straight `initialize` here. I realize if you know about attr_extras you may know that `pattr_initialize` takes a block form, but I'm not sure that's easy to remember.

app/models/linter/java_script.rb
mcmire: Oh I see, there's a test for this that says this won't catch `*.coffee.js` files. Why?

app/models/linter/base.rb
```diff
@@ -1,5 +1,9 @@
-module StyleGuide
+module Linter
   class Base
+    def self.lint?(filename)
```
gabebw: What do you think about calling this something else? `lint?` doesn't immediately mean "this can lint this file" to me.

Idea: `can_lint?`

app/models/linter/coffee_script.rb
mcmire: The only case for using a regex here seems to be `/(?>!\.coffee)\.js\z/` but see my note on that.

mcmire:
I think that renaming "style guide" to "linter" makes a lot of sense. This is the same terminology that Syntastic uses.

The only issue I have is with the JavaScript linter and the regex it is using, see my note above.

Other than that I am :+1: on this.

spec/models/linter/base_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+module Linter
+  class Test < Base
+    FILE_REGEXP = /.+\.yes\z/
+  end
+end
+
+describe Linter::Base do
+  describe ".lint?" do
+    it "uses the FILE_REGEXP to determine the match" do
+      result1 = Linter::Test.lint?("foo.yes")
```
gabebw: Let's use more descriptive variable names like `yes_lint` and `no_lint`.

Or, no variables:

```ruby
expect(Linter::Test.lint?("foo.yes")).to eq true
```

spec/models/linter/base_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "spec_helper"
```
mcmire: What do you feel about moving the tests for the instance methods here (so far just `file_included?`) into a shared example group and then mixing it into all example groups for the subclasses?

app/models/linter/base.rb
mcmire: Yeah, I actually thought the same thing too.

app/models/linter/java_script.rb
gylaz: Because that's would imply an invalid pre-processor extension. JS process it first, then CoffeeScript?

thorncp: I'm making an assumption, but it's probably because those files are better suited to be handled by the CoffeeScript linter. JSHint will probably fall over if we feed it a .coffee file.

mcmire: If this is a reference to files in the asset pipeline, then CoffeeScript files used to be named `*.js.coffee` (not `*.coffee.js`) but are no longer named that.

I think we should make the assumption that if a file ends in `*.js` it is a JavaScript file and should be linted that way. I'm pretty sure that editors follow this logic as well.

gylaz: Historically, we've seen `*.coffee.js` files coming in (probably because a few users mixed up the extensions), and it caused JSHint to throw exceptions. I believe this regex was put in place, to avoid that situation. In retrospect, a better fix would've been to handle the exception properly.

teoljungberg: @mcmire `*.coffee.js` is catched and linted by the Coffescript linter already, and since the current implementation is bound to the order of list of linters. And I didn't want the two collide.

Do you have any other suggestion on handling this?

spec/models/linter/base_spec.rb
teoljungberg: I like variables, and I like your suggestion. Thank you

spec/models/linter/base_spec.rb
teoljungberg: I could go either way, but I lean awway from shared examples because I find the repetition to help useful to find better abstractions

app/models/linter/base.rb
teoljungberg: I'm wary about `.lint?` as a name, `.can_lint?` sounds better. Let's go with that

app/models/linter/java_script.rb
thorncp: I think it makes sense to leave the restriction in, at least for the time being. Since the current flow is to find _the_ linter for a given file, it doesn't make sense to have two linters claim they can lint it.

teoljungberg:
Thanks for the extensive reviews @gylaz, @thorncp, @mcmire, and @gabebw

PR #927, opened 2015-09-16
Format style checker spec to conform to our guides

gylaz:
Why:

* Because fitting such changes into PRs when we touch them generates
  a lot of noise, and thus doing them separately is an easier task.


thorncp:
:+1:

gylaz:
Strange that Hound didn't report on the PR's status.

PR #928, opened 2015-09-16
Add support for Markdown service

gylaz:
Why:

* Reviewing Markdown for style violations is neat.
* We should support well-know Markdown file extensions.

https://trello.com/c/mzlQMOCl

### DIFF
--- a/app/models/repo_config.rb
+++ b/app/models/repo_config.rb
@@ -1,14 +1,31 @@
 # Load and parse config files from GitHub repo
 class RepoConfig
   HOUND_CONFIG = ".hound.yml"
-  BETA_LANGUAGES = %w(go haml python swift)
-  LANGUAGES = %w(ruby coffeescript javascript scss haml go python swift)
+  BETA_LANGUAGES = %w(
+    go
+    haml
+    markdown
+    python
+    swift
+  )
+  LANGUAGES = %w(
+    coffeescript
+    go
+    haml
+    javascript
+    markdown
+    python
+    ruby
+    scss
+    swift
+  )
   FILE_TYPES = {
-    "ruby" => "yaml",
-    "javascript" => "json",
     "coffeescript" => "json",
+    "haml" => "yaml",
+    "javascript" => "json",
+    "markdown" => "rb",
+    "ruby" => "yaml",
     "scss" => "yaml",
-    "haml" => "yaml"
   }
 
   pattr_initialize :commit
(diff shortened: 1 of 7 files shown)


gylaz:
screenshot 2015-09-16 11 21 57

teoljungberg:
@gylaz What does the prefix mean? I find it to be a bit confusing, what do you think about removing it before sending the payload back to Hound?

gylaz:
@teoljungberg The prefix is actually something we added to make it easier to debug. It made things easier during development.

I'm okay with removing it, but that is on the https://github.com/thoughtbot/hound-markdown side.

app/models/repo_config.rb
```diff
@@ -1,14 +1,31 @@
...
+    scss
+    swift
+  )
   FILE_TYPES = {
-    "ruby" => "yaml",
-    "javascript" => "json",
     "coffeescript" => "json",
+    "haml" => "yaml",
+    "javascript" => "json",
+    "markdown" => "rb",
```
teoljungberg: `rb`?

teoljungberg:
One question, besides that - codewise this seems fine to me

app/models/repo_config.rb
gylaz: Yep. `mdl`'s configuration is Ruby code:

```ruby
rule "MD002"
rule "MD030", ul_single: 2
```

Here are all of their rules: https://github.com/mivok/markdownlint/blob/master/lib/mdl/rules.rb

teoljungberg: Neat. Thanks for clarifying

teoljungberg: Besides - I can't find a `parse_rb` helper. How are these rules parsed?

gylaz: I assume you're asking on `mdl`'s side. This is the method that starts it all: https://github.com/mivok/markdownlint/blob/master/lib/mdl/ruleset.rb#L46

teoljungberg: I'm not, I'm asking about https://github.com/thoughtbot/hound/blob/env-variables-as-constants/app/models/repo_config.rb#L115-L123 <https://github.com/thoughtbot/hound/blob/env-variables-as-constants/app/models/repo_config.rb#L115-L123>

gylaz: Service-based linters aren't parsed. They use `raw_for` which passes raw config to the service.

gylaz:
Staging works!
screenshot 2015-09-16 15 17 51

PR #932, opened 2015-09-18
Standardize Python service attributes

thorncp:
* Updates the Python service to be serialized with a hash argument, rather
  than an array.
* Adds a `StyleGuide::Base#enqueue_job` private method that handles the
  Resqueue details. The Python linter overrides this because it has a
  non-standard class name, `review.PythonReviewJob`. As far as I can
  tell, this name is required because the Python Resuque library, Pyres,
  uses the python import system, which needs a module name. I'd love to
  hear more about this from anyone with more knowledge.

### DIFF
--- a/spec/models/style_guide/python_spec.rb
+++ b/spec/models/style_guide/python_spec.rb
@@ -25,12 +25,12 @@ describe StyleGuide::Python do
         {
           class: "review.PythonReviewJob",
           args: [
-            commit_file.filename,
-            build.commit_sha,
-            build.pull_request_number,
-            commit_file.patch,
-            commit_file.content,
-            "config",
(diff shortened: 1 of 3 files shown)


spec/models/style_guide/python_spec.rb
```diff
@@ -24,14 +24,14 @@
...
-            "config",
-          ],
+          args: [{
+            filename: commit_file.filename,
+            commit_sha: build.commit_sha,
+            pull_request_number: build.pull_request_number,
+            patch: commit_file.patch,
+            content: commit_file.content,
+            config: "config",
+          }],
```
gylaz: Can kill the `{}`

gylaz:
:+1:

Issue #879, opened 2015-08-11
Hound config not respecting config files


elwayman02:
Our `.hound.yml` config is as follows:

```
scss:
  enabled: true
  config_file: .scss-lint.yml

javascript:
  enabled: true
  config_file: .jshintrc
```

The `enabled` flags are being respected (ie, if we set them to false, Hound doesn't make any comments), but the `config_file` flags are not. The default Hound rules are being used, even in the case where we have set our own rule to override it.

gylaz:
Can you provide your `.scss-lint.yml` and `.jshintrc` files?

monfresh:
I'm running into this as well. Hound is not respecting the file exclusion that I specified in my `.scss-style.yml`. I tried both of these, and neither worked:
```yaml
linters:
  SelectorFormat:
    exclude:
      - '/assets/stylesheets/custom.scss'
```
```yaml
scss:
  linters:
    SelectorFormat:
      exclude:
        - '/assets/stylesheets/custom.scss'
```

elwayman02:
scss-lint:

```
scss_files: 'app/styles/**/*.scss'
exclude:
- 'app/styles/dazzler/font-icons.scss'
- 'app/styles/dazzler/social-icons.scss'
linters:
  BangFormat:
    enabled: true
  BorderZero: #Don't care
    enabled: false
  ColorKeyword:
    enabled: true
  ColorVariable: #TODO: Fix violations
    enabled: false
  Comment:
    enabled: true
  DebugStatement:
    enabled: true
  DeclarationOrder: #TODO: Fix violations
    enabled: false
  DisableLinterReason:
    enabled: true
  DuplicateProperty:
    enabled: true
  ElsePlacement:
    enabled: true
  EmptyLineBetweenBlocks: #TODO: Determine chosen behavior/Fix violations
    enabled: false
  EmptyRule:
    enabled: true
  ExtendDirective:
    enabled: false
  FinalNewline:
    enabled: true
  HexLength: #Not worth enforcing this rule, though we tend to prefer long
    enabled: false
    style: long
  HexNotation: #Not worth enforcing this rule
    enabled: false
    style: uppercase
  HexValidation:
    enabled: true
  IdSelector:
    enabled: true
  ImportantRule: #TODO: Not sure we can ever turn this one, but we should definitely reduce our use of !important
    enabled: false
  ImportPath:
    enabled: true
  Indentation:
    enabled: true
[...]

PR #933, opened 2015-09-21
Use correct glob for SCSS exclude example

thorncp:
SCSS-Lint compares the `exclude` entries to the fully expanded path, so
it is expecting a glob that matches the whole thing.

https://github.com/thoughtbot/hound/issues/879

### DIFF
--- a/app/views/pages/configuration.haml
+++ b/app/views/pages/configuration.haml
@@ -198,7 +198,7 @@
           :preserve
             scss:
               exclude:
-                - "app/assets/stylesheets/plugins/**"
+                - "**/app/assets/stylesheets/plugins/**"
 
       %p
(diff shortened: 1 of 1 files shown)


teoljungberg:
Good catch

gylaz:
We probably want to port https://github.com/thoughtbot/hound/pull/813 to `hound-scss`. `scss-lint` supports globs like `app/assets/stylesheets/*` and its likely users will still configure it so, since we rely on `scss-lint`'s configuration.

thorncp:
Totally forgot about that feature. I agree porting it to the service is a good move.

thorncp:
Closing in favor of https://github.com/thoughtbot/hound-scss/pull/14

Issue #725, opened 2015-03-14
prefer `x` over `x` in Hound, but works in Rubocop


lowjoel:
After #717, Rubocop configs are inherited (somewhat).

coursemology/coursemology2#120 is a PR where we use `reduce`. That is the Rubocop default, and was overridden in our configuration (via an `inherits_from` directive). Running Rubocop doesn't give any warnings with that line; however the PR does have Hound commenting on it (search the comments for 'houndci').

lowjoel:
And after more attempts, in Coursemology/coursemology2#128, it appears that the configuration merging ignores hashes more than one level deep. I'm not sure if this is something involving Rubocop, or is because of Hound's extra settings.

gylaz:
There hasn't been any recent reports of this issue. Closing.

PR #940, opened 2015-09-23
Update Rubocop to 0.34.2

### LINKED ISSUE #903: bump Rubocop version to 0.33
Any chance we could bump the Rubocop version to the latest? i.e. 0.33.  From the Gemfile it looks like hound is on "0.29.1"

There are various config options in 0.33 such as DisabledByDefault, which I think might help us get round the various confusions arising from the way in which hound overrides the rubocop settings in each project:

https://github.com/thoughtbot/hound/issues/576

Many thanks in advance

thorncp:
* Renames `Style/Blocks` to `Style/BlockDelimiters`
* Renames `Style/UnneededPercentX` to `Style/CommandLiteral`
* Disables all new cops for Hound and thoughtbot config.
* Does not use the new `DisabledByDefault` config value, which only
  enables cops you specify. I left this off because we have all the cops
  explicitly defined in the config already.
* Updates expected violation messages to reflect new text
* Tweaks a couple tests that were no longer seeing the
  "Unnecessary spacing detected." message. I'm not sure what this was
  even covering, given that other cops in those tests are completely
  handling the whitespace (I think).
* `Style/RegexpLiteral` no longer supports a `MaxSlashes` parameter, so
  it's removed.

Will close #903

Potentially solves #725


config/style_guides/ruby.yml
```diff
@@ -20,7 +20,12 @@ AllCops:
   - "db/schema.rb"
   RunRailsCops: false
   DisplayCopNames: false
+  DisplayStyleGuide: false
   StyleGuideCopsOnly: false
+  DisabledByDefault: false
```
gylaz: What is this set to by default in RuboCop?

config/style_guides/ruby.yml
```diff
@@ -923,7 +1010,11 @@ Style/UnneededPercentQ:
   Description: Checks for %q/%Q when single quotes or double quotes would do.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#percent-q
   Enabled: true
-Style/UnneededPercentX:
```
gylaz: This was removed (or rather became `CommandLiteral`) but what about its options?

gylaz:
Thoughts on **not** disabling new rules? They are fairly uncommon violation rules, and it might give us a reality check about the config changes we are doing -- given that in the future we'll just be rolling with linter defaults, thus each time we'll update the rules will change on our users.

config/style_guides/ruby.yml
thorncp: [false](https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L51)

config/style_guides/ruby.yml
thorncp: The old options are in the same place, but looks like I added two `CommandLiteral`s by accident. Removed one in b36763283a0c0a8ca7c228839e12b67360328f69

thorncp:
I'm fine using the Rubocop defaults for all the new cops. I just went into this PR with thought of having the least resistance.

thorncp:
Pulled in Rubocop defaults for new inters in a6c8bda29d5c7e5ce432231c0c738f8b5e706d27

gylaz:
Looks good to me. @salbertson Thoughts?

teoljungberg:
Looks good to me

salbertson:
Looks good to me! :+1:

PR #941, opened 2015-09-24
Add Markdown documentation

thorncp:
It was announced in an email, so we should have information about it.

configuration___hound


teoljungberg:
:+1:

PR #944, opened 2015-09-25
Add configuration page docs for Swift

jakecraige:
This can remain open while it's in beta
but I figured I'd open a PR now
since I have the time.

Why:

* Users would like to know how to add Swift to their projects.

This change addresses the need by:

* Adding the Swift section to the configuration page.

screen shot 2015-09-25 at 2 40 18 pm

### DIFF
--- a/app/helpers/configuration_helper.rb
+++ b/app/helpers/configuration_helper.rb
@@ -27,6 +27,10 @@ module ConfigurationHelper
     "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/haml.yml"
   end
 
+  def swift_config_url
+    config_url("thoughtbot/hound-swift", "config/default.yml")
+  end
+
   private
 
   def config_url(slug, config_file)
--- a/app/views/pages/configuration.haml
+++ b/app/views/pages/configuration.haml
@@ -21,6 +21,8 @@
         = link_to "Go config", "#go"
       %li
         = link_to "Python config", "#python"
+      %li
+        = link_to "Swift config", "#swift"
 
   %section.doc-content
     %article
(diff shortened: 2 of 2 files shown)


thorncp:
LGTM

jakecraige:
Waiting till merge till we complete some internal testing to make sure things work as expected.

salbertson:
@jakecraige what isn't working as expected?

app/helpers/configuration_helper.rb
```diff
@@ -22,4 +22,9 @@ def scss_config_url
   def haml_config_url
     "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/haml.yml"
   end
+
+  def swift_config_url
+    "https://raw.githubusercontent.com/thoughtbot/hound-swift" \
```
teoljungberg: What do you think about merging this to one line? If that's too long, how about extracting a helper to genearate the url based on the repository and the filepath?

salbertson:
:+1:

salbertson:
@jakecraige is this ready?

jakecraige:
@salbertson We're still testing internally. @gfontenot and @sidraval have been using it on a project this week I believe. Would like to hear their feedback if it's working as expected for them before merging.

gfontenot:
We should add explicit credit to Realm somewhere on here for creating/maintaining SwiftLint

jakecraige:
@gfontenot I think this was discussed briefly with @jpsim before but I don't remember where. We're linking to their repo in 2 places in these docs, do you think we need more?

gfontenot:
When I talked to JP, the request was to have an explicit note about SwiftLint being a Realm project like CocoaDocs is doing with Jazzy (see the bottom of the page on [Argo's docs](http://cocoadocs.org/docsets/Argo/2.2.0/index.html).

jpsim:
> @gfontenot I think this was discussed briefly with @jpsim before but I don't remember where. We're linking to their repo in 2 places in these docs, do you think we need more?

I appreciate linking to the SwiftLint source! I think that's sufficient.

jpsim:
I'd love to see this go live :smile:

jakecraige:
Deployed to prod :) https://houndci.com/configuration#swift

PR #947, opened 2015-10-05
Replace `style_guide` with `linter`

teoljungberg:
To follow up on https://github.com/thoughtbot/hound/pull/924

### DIFF
--- a/spec/models/linter/go_spec.rb
+++ b/spec/models/linter/go_spec.rb
@@ -21,10 +21,10 @@ describe Linter::Go do
 
   describe "#file_review" do
     it "returns a saved and incomplete file review" do
-      style_guide = build_style_guide
+      linter = build_linter
       commit_file = build_commit_file(filename: "a.go")
 
-      result = style_guide.file_review(commit_file)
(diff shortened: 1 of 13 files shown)


spec/models/linter/go_spec.rb
```diff
@@ -21,10 +21,10 @@
 
   describe "#file_review" do
     it "returns a saved and incomplete file review" do
-      style_guide = build_style_guide
+      linter = build_linter
       commit_file = build_commit_file(filename: "a.go")
 
-      result = style_guide.file_review(commit_file)
+      result = linter.file_review(commit_file)
```
salbertson: What do you think about inlining `build_linter` and getting rid of `linter = build_linter`? We could rename `build_linter` to something like `linter` as well.

salbertson:
@teoljungberg one question; looks good to me. :+1:

spec/models/linter/go_spec.rb
teoljungberg: The pattern of `build_linter` is already established in the code-base, and I would prefer to stick to it

PR #950, opened 2015-10-06
Introduce a `Config` class for each `Linter`

teoljungberg:
To decouple the different kinds of configuration options Hound supports,
let each linter decide on how the specific kind of configuration it
needs. Both in what format, how to parse it, and how to access it.

Changes:

- Add `HoundConfig` to be the object representation of `.hound.yml`.
- Add `Config::LINTER` to represent the configuration for the specific `LINTER`.
- Each subclass of `Config::Base` implements their own `parser` method, that takes the content of the file it should parse, and do so accordingly.
- Replace, and remove, `RepoConfig` with `HoundConfig` and/or
  `Config::LINTER` (where applicable).
-  Remove `NAME` constant from `Linter::*`   
    `NAME` was used to set `Linter::Base#_linter_name`, which can be
    determined by `Linter::Base#name` instead.

https://trello.com/c/PoRwwKIw


spec/models/config/parser_spec.rb
```diff
@@ -0,0 +1,238 @@
...
+              "enabled" => true,
+              "config_file" => "config/jshint.json",
+            },
+          },
+        )
+        parser = Config::Parser.new(hound_config)
+
+        expect do
+          parser.load_json("config/jshint.json")
+        end.to raise_error do |exception|
```
teoljungberg: I actually don't have a problem with this. Is this a rule we care about?

spec/models/linter/ruby_spec.rb
```diff
@@ -717,18 +737,26 @@ def violations_in(content, config: nil, repository_owner_name: "joe")
     end
 
     def build_linter(
-      repo_config: build_repo_config,
+      hound_config: build_hound_config,
+      config: stub_ruby_config,
```
teoljungberg: This was fixed in d65928cdc8d406bb927b78202c2a2622a4a13fcd, don't know why it's not reflected here

spec/models/linter/java_script_spec.rb
```diff
@@ -212,29 +207,40 @@ def build_js_file(content = "foo")
...
     )
     linter.file_review(commit_file).violations
   end
 
   def build_linter(
-    repo_config: default_repo_config,
-    repository_owner_name: "not_thoughtbot"
+    hound_config: default_hound_config,
+    repository_owner_name: "not_thoughtbot",
+    config: stub_javascript_config
```
teoljungberg: This was fixed in, I'm unsure why it's not reflected here

spec/models/linter/coffee_script_spec.rb
```diff
@@ -238,17 +232,29 @@ def spy_on_coffee_lint
   end
 
   def build_linter(
-    repo_config: default_repo_config,
+    hound_config: default_hound_config,
+    config: stub_coffeescript_config,
```
teoljungberg: This was also fixed in d65928cdc8d406bb927b78202c2a2622a4a13fcd

spec/models/config/parser_spec.rb
teoljungberg: It's not something we have specified on our styleguide

app/models/config/ruby.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    private
+
+    def legacy?
+      (configured_languages & HoundConfig::LANGUAGES).empty?
+    end
+
+    def configured_languages
+      hound_config.content.keys
```
mike-burns: Can `hound_config.content` produce `nil`?

spec/models/config/swift_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      allow(Config::Parser).to receive(:new).and_return(parser)
+
+      result = swift_config.content
+
+      expect(Config::Parser).to have_received(:new)
+      expect(parser).to have_received(:load_raw).with("config/swiftlint.yml")
+      expect(result).to eq(raw_config)
+    end
+  end
+end
```
mike-burns: This (and Scss) are long tests for short files. How about a shared example group?

app/models/hound_config.rb
```diff
@@ -0,0 +1,55 @@
...
+  def content
+    @content ||= parser.load_yaml(HOUND_CONFIG)
+  end
+
+  def enabled_for?(name)
+    supported_language?(name) || legacy_key?(name) || enabled?(name)
+  end
+
+  def fail_on_violations?
+    !!(content["fail_on_violations"])
```
mike-burns: Why the cast to Boolean?

app/models/linter/ruby.rb
```diff
@@ -35,8 +40,8 @@ def parsed_source(commit_file)
       RuboCop::ProcessedSource.new(commit_file.content, absolute_filepath)
     end
 
-    def config
-      @config ||= RuboCop::Config.new(merged_config, "")
+    def rubocop_config
+      @rubocop_config ||= RuboCop::Config.new(merged_config, "")
```
mike-burns: Why is this memoized?

app/models/linter/ruby.rb
```diff
@@ -61,7 +66,7 @@ def custom_config
     # This is deprecated in favor of RuboCop's DisplayCopNames option.
     # Let's track how often we see this and remove it if we see fit.
     def rubocop_options
-      if config.delete("ShowCopNames")
+      if rubocop_config.delete("ShowCopNames")
```
mike-burns: I'm confused about the order of operations here: when `RuboCop::Cop::Team.new` gets a `rubocop_config`, which has been memoized, will it have the `ShowCopNames` key?

app/models/config/parser.rb
```diff
@@ -0,0 +1,59 @@
...
+      YAML.safe_load(file_content, [Regexp]) || {}
+    rescue Psych::Exception => exception
+      raise_repo_config_parser_error(exception, file_path)
+    end
+
+    def content_for(file_path)
+      if file_path
+        commit.file_content(file_path)
+      else
+        "{}"
```
tabfugnic: Why is this a string? Am I missing something?

tabfugnic: Nevermind I see it, this is being eventually parsed by JSON. Thoughts on turning this into a constant like `BLANK_JSON_STRING` or something that represents what it is?

app/models/config/python.rb
```diff
@@ -0,0 +1,4 @@
+module Config
+  class Python < Base
```
tabfugnic: We lint Python!?! Or is this just being hopeful?

app/models/linter/coffee_script.rb
```diff
@@ -4,6 +4,7 @@ class CoffeeScript < Base
     DEFAULT_CONFIG_FILENAME = "coffeescript.json"
     ERB_TAGS = /<%.*%>/
     FILE_REGEXP = /.+\.coffee(\.js)?(\.erb)?\z/
+    NAME = "coffee_script"
```
tabfugnic: Isn't this the legacy name for the `coffeescript` linter? I think this should just be `coffeescript`

spec/models/config/parser_spec.rb
tabfugnic: I think some people prefer `{}` for blocks that return things

app/models/config/python.rb
teoljungberg: We do lint Python! Currently it's in a beta-phase, so you have to opt-in to use it. But yes!

app/models/config/parser.rb
teoljungberg: Turning it into a constant sounds like a great idea!

app/models/linter/coffee_script.rb
teoljungberg: Since the class is named `CoffeeScript`, which translates into `coffee_script` in ruby-lingo, I consider `coffee_script` to be a more precise name than `coffeescript`. The `NAME` constant is used to set the `linter_class` in `Linter::Base`. But thinking about it now - we actually don't need to set it, since we can deduce that from the `Linter::Base#name` instead.

I'll remove the `NAME` constant, thanks!

app/models/config/ruby.rb
teoljungberg: No, it will return the parsed JSON/YAML, the fallback value of `{}` Or raise a parsing exception. So no

mike-burns: So if the JSON is `[]` this will also raise a `MethodNotFound`?

teoljungberg: That's correct, I didn't think of that. I'm unsure if that's an edge-case that will ever happen though.

What do you think is the best solution here? We can type-case `[]` into `{}`, but `["hello"]` can't be converted(raises a `TypeError`)

app/models/linter/coffee_script.rb
tabfugnic: I'm glad that it ultimately came to that conclusion

app/models/config/parser.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    private
+
+    BLANK_JSON_STRING = "{}"
+
+    def parse_json(file_path, file_content)
+      JSON.parse(file_content)
+    rescue JSON::ParserError => exception
+      raise_repo_config_parser_error(exception, file_path)
+    end
```
mike-burns: This method needs to produce a hash. If it produces something else (e.g. an array), it should raise instead.

So far as I can see this is the only requirement on this JSON: it must be a JSON object. It might be good to validate it against a more complete schema down the road, though.

spec/models/config/swift_spec.rb
teoljungberg: Good call, extracted into a shared example in b995768b763b9ccf182124f50782a280a9582e62

spec/models/config/parser_spec.rb
teoljungberg: That read weirder and weirder, so I'll keep this as is

app/models/hound_config.rb
teoljungberg: I want predicate methods to always return a boolean, do you think it would be clearer to do `content["fail_on_violations"].present?` instead?

app/models/config/parser.rb
teoljungberg: I agree that a schema validation makes sense when and if the schema grows more complex. For now, we validate the type in 5da14f0

app/models/linter/ruby.rb
teoljungberg: I didn't write the initial implementation of this, I just renamed the method.

I think we can safely remove the memozation, I atleast see no purpose of having it. Thoughts @thoughtbot/hound ?

app/models/hound_config.rb
mike-burns: As you please.

(In Ruby, I just use truthy values instead of Booleans.)

app/models/config/ruby.rb
teoljungberg: Resolved in 5da14f0

app/models/config/java_script.rb
```diff
@@ -0,0 +1,42 @@
+module Config
+  class JavaScript < Base
+    DEFAULT_EXCLUDED_FILES = %(vendor/*)
```
seanpdoyle: What makes this different than `"vendor/*"`?

teoljungberg: Nothing, I just wrote `%()` out of habit. Do you think it should be changed into a list instead?

seanpdoyle: isn't a list denoted by `%w`?

teoljungberg: `%()` will turn it into a string.

seanpdoyle: Right.

Is it more obvious that it's a string when wrapped in `"`?

Will this conversation and confusion be repeated by new developers as they become acquainted with the codebase?

teoljungberg: I don't have a strong preference, changing to `""`

app/models/config/java_script.rb
```diff
@@ -0,0 +1,42 @@
+module Config
+  class JavaScript < Base
+    DEFAULT_EXCLUDED_FILES = "vendor/*"
```
seanpdoyle: :+1: 

> Use %{} for single-line strings needing interpolation and double-quotes.

From [thoughtbot/guides](https://github.com/thoughtbot/guides/tree/master/style/ruby#ruby)

gylaz: I believe this should be an array. That way we don't need to split it on newline.

app/models/config/base.rb
```diff
@@ -0,0 +1,19 @@
+module Config
+  class Base
+    pattr_initialize :hound_config, :linter
```
gylaz: It looks like we only ever use linter's name. What do you think about passing `linter_name`?

app/models/config/coffee_script.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    private
+
+    def file_path
+      linter_config["config_file"]
+    end
+
+    def linter_config
+      hound_config.content[linter.name] ||
```
gylaz: Thoughts on moving this into `Base` as `linter_config`? Then this method could override that:
```rb
def linter_config
  super || hound_config.content[alternate_linter_name]
end
```
and you wouldn't need to override `file_path` here?

app/models/config/parser.rb
```diff
@@ -0,0 +1,78 @@
...
+    def self.load_json(hound_config, file_path)
+      file_content = load_raw(hound_config, file_path)
+      parse_json(file_path, file_content)
+    end
+
+    def self.load_raw(hound_config, file_path)
+      content_for(hound_config, file_path)
+    end
+
+    private
```
gylaz: I believe this does nothing in a Module. Does it make sense to break some of this out into a class where we can take advantage of `private`?

app/models/config/parser.rb
```diff
@@ -0,0 +1,78 @@
...
+        hound_config.commit.file_content(file_path)
+      else
+        BLANK_JSON_STRING
+      end
+    end
+
+    def self.raise_parser_error(
+      exception,
+      file_path,
+      message: "#{exception.class}: #{exception.message}"
```
gylaz: It looks like you're expecting an instance here, but on line 47 a class is passed. Are there no tests to cover this?

app/models/config/parser.rb
```diff
@@ -0,0 +1,78 @@
...
+      raise_parser_error(
+        TypeError,
+        file_path,
+        message: "`#{file_path}` must be a Hash",
+      )
+    end
+
+    def self.content_for(hound_config, file_path)
+      if file_path
+        hound_config.commit.file_content(file_path)
```
gylaz: What do you think about removing this LoD violation?

app/models/config/parser.rb
```diff
@@ -0,0 +1,78 @@
...
+      else
+        BLANK_JSON_STRING
```
gylaz: Why is this a string representation of a JSON object? YAML parsing also hits this method, right?

app/models/config/parser.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def self.raise_parser_error(
+      exception,
+      file_path,
+      message: "#{exception.class}: #{exception.message}"
+    )
+      raise Error.new(message, filename: file_path)
+    end
+
+    class Error < StandardError
```
gylaz: Thoughts on calling this `ParseError` or something descriptive?

app/models/config/parser.rb
```diff
@@ -0,0 +1,78 @@
...
+      attr_reader :filename
+
+      def initialize(message, filename:)
+        super(message)
+        @filename = filename
```
gylaz: Do we use this anywhere?

app/models/hound_config.rb
```diff
@@ -0,0 +1,51 @@
+class HoundConfig
+  HOUND_CONFIG = ".hound.yml"
```
gylaz: What do you think about naming this `CONFIG_FILENAME`? It's already in `HoundConfig` class, so we know it's specific to Hound.

app/models/hound_config.rb
```diff
@@ -0,0 +1,51 @@
...
+  )
+
+  attr_reader_initialize :commit
+
+  def content
+    @content ||= Config::Parser.load_yaml(self, HOUND_CONFIG)
+  end
+
+  def enabled_for?(name)
+    supported_language?(name) || legacy_key?(name) || enabled?(name)
```
gylaz: This seems not quite correct. If we find a supported language (or legacy language) that doesn't mean it's enabled, does it?

app/models/linter/base.rb
```diff
@@ -50,15 +54,15 @@ def enqueue_job(attributes)
...
 
     def job_class
-      "#{linter_name.classify}ReviewJob".constantize
+      "#{name.classify}ReviewJob".constantize
     end
 
-    def linter_name
-      self.class::NAME
+    def config
+      @_config ||= config_class.new(hound_config, self)
```
gylaz: Why an underscore?

app/models/linter/coffee_script.rb
```diff
@@ -37,8 +37,8 @@ def erb?(file)
       file.filename.ends_with? ".erb"
     end
 
-    def config
-      default_config.merge(repo_config.for(name))
+    def coffeescript_config
```
gylaz: What is the reasoning behind renaming this?

gylaz: Don't we know which config it it based on this class's name? Also, this adds extra noise in the diffs.

app/models/linter/ruby.rb
gylaz: This method gets called in multiple places and configuration merging logic is complex. That was why it was memoized.

app/models/style_checker.rb
```diff
@@ -44,7 +44,7 @@ def find_linter_class(filename)
     LINTERS.detect { |linter_class| linter_class.can_lint?(filename) }
   end
 
-  def config
-    @config ||= RepoConfig.new(pull_request.head_commit)
+  def hound_config
+    @_hound_config ||= HoundConfig.new(pull_request.head_commit)
```
gylaz: Why an underscore here?

spec/models/config/base_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      linter = double("Linter", name: "linter")
+      config = Config::Test.new(hound_config, linter)
+      allow(Config::Parser).to receive(:load_raw)
+
+      config.content
+
+      expect(Config::Parser).to have_received(:load_raw).with(
+        hound_config,
+        "config-file.txt",
+      )
```
gylaz: What do you think of asserting on the result of `config.content` instead of testing implementation?

spec/models/config/coffee_script_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+              "enabled" => true,
+              "config_file" => "config/coffeescript.json",
+            },
+          },
+        )
+        linter = double("Linter", name: "coffee_script")
+        coffee_config = Config::CoffeeScript.new(hound_config, linter)
+        allow(Config::Parser).to receive(:load_json).and_return(
+          "arrow_spacing": { "level": "error" },
+        )
```
gylaz: Since we require `Config::Parser` does it make sense to let it do its work, and not stub? Otherwise, is it worth defining `Config::Parser` at the top of the test block, and not require the actual file?

spec/models/config/coffee_script_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+              "enabled" => true,
+              "config_file" => "config/coffeescript.json",
+            },
+          },
+        )
+        linter = double("Linter", name: "coffeescript")
+        coffee_config = Config::CoffeeScript.new(hound_config, linter)
+        allow(Config::Parser).to receive(:load_json).and_return(
+          "arrow_spacing": { "level": "error" },
+        )
```
gylaz: There's lot of duplication in these tests across different languages. What do you think about extracting this into a helper?

spec/models/linter/coffee_script_spec.rb
```diff
@@ -38,27 +38,19 @@
...
   describe "enabled?" do
-    context "with legacy coffee_script key" do
-      it "is not enabled" do
-        commit = double("Commit", file_content: <<-EOS.strip_heredoc)
-          coffee_script:
-            enabled: false
-        EOS
-        repo_config = RepoConfig.new(commit)
-        linter = build_linter(repo_config: repo_config)
+    context "when the config has enabled_for to true" do
```
gylaz: This reads strangely.

gylaz: Maybe generalize it to `when configuration is enabled`?

spec/models/hound_config_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+  describe "#enabled_for?" do
+    context "given a supported language" do
+      it "returns true for all of them" do
+        commit = stubbed_commit(".hound.yml" => "")
+        hound_config = HoundConfig.new(commit)
+
+        supported_languages = (
+          HoundConfig::LANGUAGES - HoundConfig::BETA_LANGUAGES
+        )
```
gylaz: Is this part of the verification phase?

spec/models/linter/java_script_spec.rb
gylaz: It's not reflected because this line isn't affected by that commit. GitHub only collapses comments when the line that was commented on changed.

spec/models/linter/ruby_spec.rb
```diff
@@ -358,9 +358,9 @@ def two
       context "for method definitions with optional named arguments" do
         it "does not return violations" do
           expect(violations_in(<<-CODE.strip_heredoc)).to be_empty
-  def register_email(email:)
-    register(email)
-  end
+            def register_email(email:)
+              register(email)
+            end
```
gylaz: These would've been a good candidate for a separate PR. This PR is already large as is :)

spec/models/linter/ruby_spec.rb
```diff
@@ -472,21 +475,33 @@ def foo
 
     context "with custom configuration" do
       it "finds only one violation" do
-        config = {
+        content = {
```
gylaz: `content` seems like the wrong variable name in a linter spec, as it could be confused with content that is being linted. What do you think about `config_content`?

spec/models/linter/ruby_spec.rb
```diff
@@ -556,21 +577,11 @@ def foo
...
+            hoge
           CODE
         end
       end
-
-      def violations_with_config(config)
-        content = <<-TEXT.strip_heredoc
-          def test_method
-            { :foo => "hello world" }
-          end
```
gylaz: Why did you opt to remove this and duplicate the code content in related tests?

spec/models/linter/ruby_spec.rb
```diff
@@ -609,11 +619,15 @@ def test_method
         thoughtbot_violations_in(code)
 
         expect(RuboCop::ConfigLoader).to(
-          have_received(:configuration_from_file).with(config_file)
+          have_received(:configuration_from_file).with(
+            thoughtbot_configuration_file(Linter::Ruby),
+          ).at_least(:once),
```
gylaz: Does it now receive this method more than once?

spec/support/helpers/linter_helper.rb
```diff
@@ -0,0 +1,63 @@
...
+      allow(commit).to receive(:file_content).with(filename).
+        and_return(contents)
+    end
+
+    commit
+  end
+
+  RSpec.configure do |config|
+    config.include LinterHelper
+  end
```
gylaz: Should this RSpec block be outside of the module?

thorncp:
Is there any way we can reduce the size of this diff? Would it be possible to
introduce these changes one linter-config at a time?

teoljungberg:
@thorncp - I don't see the point of making N number of PRs(one for adding config for each language). But I agree, this is a big PR. But with a refactoring of this size, I don't see an easier way.

But, I'll work on extracting smaller PRs out of this one, hopefully that will decrease the size of the diff.

spec/models/linter/ruby_spec.rb
teoljungberg: Good point, I'll extract smaller PRs out of this

spec/models/hound_config_spec.rb
teoljungberg: It is, since it's a a variable to make the verification phase easier to parse.

spec/models/config/coffee_script_spec.rb
teoljungberg: I wanted to get to a state where everything works and makes sense from a logical point of view before making those abstractions. But yes, a helper would indeed make this clearer

spec/models/config/coffee_script_spec.rb
teoljungberg: I wanted to stub it out since the Parser fetches the file-content aswell. But I think the suggestion you made in https://github.com/thoughtbot/hound/pull/950/files#r41442577 will solve this problem too

spec/models/config/base_spec.rb
teoljungberg: Good call

app/models/style_checker.rb
teoljungberg: It's common to define private instance variables with underscores, to tell the user that inspects this object that it's not meant to be used publicly

app/models/config/coffee_script.rb
teoljungberg: That makes sense, good call

app/models/config/base.rb
teoljungberg: Good call1

app/models/linter/coffee_script.rb
teoljungberg: It got renamed since `Config::Base#config` returns the corresponding config object for the linter. In this case that would be `Config::CoffeeScript`.
And I thought it would be good to clarify to the reader that this is the underlaying linter's config, so maybe `coffeelint_config` is a better fit. What do you think?

spec/models/linter/ruby_spec.rb
teoljungberg: Good call!

spec/models/linter/ruby_spec.rb
teoljungberg: I opt to un-dry most of my test to see where the troubles lie. But maybe it makes sense to have it here. I'll revert this change

spec/support/helpers/linter_helper.rb
teoljungberg: It should, I missed that. Thanks

spec/models/linter/ruby_spec.rb
teoljungberg: We're calling `configuration_from_file` in this test(https://github.com/thoughtbot/hound/blob/linters-have-configs/spec/models/linter/ruby_spec.rb#L685), and in the actual implementation(https://github.com/thoughtbot/hound/blob/linters-have-configs/app/models/linter/ruby.rb#L52-L54)

app/models/config/parser.rb
teoljungberg: We are https://github.com/thoughtbot/hound/blob/linters-have-configs/app/services/build_runner.rb#L10-L11

app/models/config/parser.rb
teoljungberg: I thought it was going to read nicely as `Config::Parser::Error`. But `ParserError` is much better

app/models/config/parser.rb
teoljungberg: Good catch

app/models/config/parser.rb
teoljungberg: You're right. It was a leftover from when I refactored it from a class to a module in c0c986c. But thinking about it more, I think I'll move it back to a class and follow something like this:

```ruby
parser = Config::Parser.new(file_path: some_file_path, commit: hound_config.commit)
parser.to_json
# => returns the parsed in file_path as json
parser.to_yml
# => returns the parsed in file_path as yaml
parser.to_s
# => returns the raw content for the file_path
```

app/models/style_checker.rb
gylaz: This makes sense in controllers.

This method is already private. Isn't the convention when memoizing is to use the same name variable as the method?

app/models/linter/coffee_script.rb
gylaz: What about `linter_config`? Repeating `coffeelint_` doesn't sound great, since that can be inferred from the class name.

app/models/config/parser.rb
gylaz: Actually, I missed the namespace. I agree `Config::Parser::Error` is probably enough.

app/models/style_checker.rb
mike-burns: I've always used an underscore prefix when memoizing to denote that it's not a shared instance variable but instead a sort of memory management hack to get memoization.

(But, this method does not need to be memoized, so perhaps this is all a moot point.)

spec/models/linter/ruby_spec.rb
teoljungberg: Done in 5f049c7

app/models/config/base.rb
```diff
@@ -0,0 +1,57 @@
...
+    def ensure_correct_type(result)
+      if result.is_a? Hash
+        result
+      else
+        raise_type_error
+      end
+    end
+
+    def raise_type_error
+      raise_parser_error(TypeError.new("`#{file_path}` must be a Hash"))
```
gylaz: We're not using the exception itself anymore. Perhaps we should be passing just the message?

app/models/config/base.rb
```diff
@@ -0,0 +1,57 @@
+module Config
+  class Base
+    pattr_initialize :hound_config, :linter_name
+
+    def content
+      @content ||= parser(load_content)
```
gylaz: Is it worth adding a `parser` definition that raises a not-implemented exception? Perhaps we can use that helper from `attr_extras` for this.

app/models/config/coffee_script.rb
```diff
@@ -0,0 +1,19 @@
+module Config
+  class CoffeeScript < Base
+    private
+
+    def parser(file_content)
+      result = JSON.parse(file_content)
+
+      ensure_correct_type(result)
```
gylaz: What do you think about moving this into `Base` and keeping only the actual parsing call in here? That way there is no duplication across child classes.

app/models/config/java_script.rb
```diff
@@ -0,0 +1,36 @@
+module Config
+  class JavaScript < Base
+    DEFAULT_EXCLUDED_FILES = %w(vendor/*)
+
+    def excluded_files
+      @excluded_files ||= (
+        load_javascript_ignore.presence ||
```
gylaz: TIL about `Object#presence`.

app/models/config/coffee_script.rb
gylaz: I just realized that service based classes don't need to do this.

app/models/hound_config.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  private
+
+  LEGACY_LANGUAGE_KEY = %w(
+    coffee_script
+    java_script
+  )
+
+  def parse(file_content)
+    YAML.safe_load(file_content, [Regexp]) || {}
```
gylaz: Did you scrap the `Parser` class idea? The reason I ask, is that this is the third instance of calling `safe_load`. If there was a bug or a security issue with this (again), we'd need to remember to track down all the instances of this invocation. If it was in `Parser` we'd only need to do it once.

app/models/config/base.rb
gylaz: I also think we need to standardize on the names we are going to use for our memoization technique. Are we using the underscore or not?

spec/models/config/base_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def parser(file_content)
+    file_content
+  end
+end
+
+describe Config::Base do
+  describe "#content" do
+    context "when `parser` is defined" do
+      it "uses the defined parser method to parse the given content" do
```
gylaz: From this, it sounds like we're just testing inheritance here. Testing the fact that the method just calls `parse` doesn't seem valuable. What do you think about not testing this, since it's covered by child classes?

spec/models/config/base_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+          },
+        )
+        config = Config::Test.new(hound_config, "linter")
+
+        expect(config.content).to eq "config"
+        expect(commit).to have_received(:file_content).with("config-file.txt")
+      end
+    end
+
+    context "when there is not specified filepath" do
```
gylaz: `when filepath is not specified`?

spec/models/config/coffee_script_spec.rb
```diff
@@ -0,0 +1,88 @@
+require "spec_helper"
+require "app/models/config/base"
+require "app/models/config/parser_error"
+require "app/models/config/coffee_script"
+
+describe Config::CoffeeScript do
+  describe "#content" do
+    context "with a modern coffeescript config" do
+      it "gets the content from GitHub as a hash" do
```
gylaz: What do you think about saying `returns` instead of `gets`?

spec/models/config/java_script_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+          "HoundConfig",
+          commit: commit,
+          content: {
+            "javascript" => {
+              "enabled" => true,
+              "config_file" => "config/jshint.json",
+            },
+          },
+        )
+        config = Config::JavaScript.new(hound_config, "javascript")
```
gylaz: What do you think about pulling out lines 36,46 into a `build_config` helper?

spec/models/hound_config_spec.rb
gylaz: What do you think about removing the parens?

spec/models/style_checker_spec.rb
```diff
@@ -251,6 +251,35 @@ def stub_pull_request(options = {})
...
+        config_file: config/javascript.json
+
+      scss:
+        enabled: true
+        config_file: config/scss.yml
+
+      haml:
+        enabled: true
+        config_file: config/haml.json
+    EOS
```
gylaz: Can we use `LinterHelper#raw_hound_config` for this?

gylaz:
This is shaping up to look awesome!

app/models/config/java_script.rb
teoljungberg: ActiveSupport is filled with magical goodies, for bad or for worse

app/models/config/base.rb
teoljungberg: Way ahead of you Greg - https://github.com/thoughtbot/hound/blob/linters-have-configs/app/models/config/base.rb#L19 <https://github.com/thoughtbot/hound/blob/linters-have-configs/app/models/config/base.rb#L19>

teoljungberg: It is standardized - `@_content` will be the memoized instance variable if we're memoizing the private method `content`. And the same goes for `@content` for the public method `content`

spec/models/config/base_spec.rb
teoljungberg: Sounds fair, I'll keep the other two contexts in place since they aren't tested anywhere else

app/models/hound_config.rb
teoljungberg: Good idea, I'll add a simple parser class that wraps the different methods we're using for parsing

spec/models/config/java_script_spec.rb
teoljungberg: Sounds great!

app/models/config/base.rb
thorncp: The name `parser` sounds like it's returning a thing that will parse something in the future, but seems to be returning some already parsed content. Is the latter true? What do you think about naming this `parse`?

gylaz: I don't understand the need for differentiation between two different naming schemes for private vs public memoization. What does it benefit does it provide us?

teoljungberg: I think we can safely stick to `@_METHOD_NAME`.

teoljungberg: Good call, I'll rename it to `parse`

teoljungberg: Then again, we have the convention of doing `@current_user = ...` in the application already. I'll stick to that

teoljungberg:
Confirmed that everything works as expected with https://github.com/teoljungberg/houndradish/pull/4

teoljungberg:
And Hound can still be configured, shown in https://github.com/teoljungberg/houndradish/pull/5#discussion_r41737969

PR #952, opened 2015-10-06
Support custom configuration outside of GitHub

salbertson:
This could be used to centralize style guides.


Gemfile
```diff
@@ -14,8 +14,8 @@ gem "coffeelint"
 gem "email_validator"
 gem "faraday"
 gem "font-awesome-rails"
-gem "haml_lint"
 gem "haml-rails"
+gem "haml_lint"
```
gylaz: This change seems unrelated?

salbertson: Yes, just sorted the gems.

app/models/repo_config.rb
```diff
@@ -134,6 +134,14 @@ def load_file(file_path, file_type)
     end
   end
 
+  def get_file_content(file_path_or_url)
+    if file_path_or_url.match("^http")
```
gylaz: What do you think about using `URI::regexp(%w(http https)).match(file_path_or_url )` to check for url?

gylaz:
LGTM

teoljungberg:
I'm super into this idea!

app/models/repo_config.rb
```diff
@@ -134,6 +134,14 @@ def load_file(file_path, file_type)
...
+      Faraday.new.get(file_path_or_url).body
```
gylaz: As discussed IRL, this would only work for public repos/gists. What if we supported private repos too? We could use Octokit library to fetch those files.

This way users could paste the link they grab from GitHub:
```yaml
ruby:
  config_file: https://github.com/thoughtbot/hound/blob/master/.ruby-style.yml
```
Thoughts?

teoljungberg: Also, what happens if the URL is a 404, do we want to guard against that somehow? I would like it to follow the same scheme as it does now; to mark the build as a failure with an error message saying that the provided config is invalid.

teoljungberg: I'm into the idea @gylaz suggested, that allows for private style guides which I can see being attractive to some

salbertson: I like the idea. It will take some reworking of `RepoConfig` so I'll try it in a different branch.

teoljungberg: What do you think about waiting on doing that until #950 has landed?

---
Teo Ljungberg

spec/models/repo_config_spec.rb
```diff
@@ -196,21 +196,75 @@
...
+          )
+
+          result = repo_config.for("ruby")
+
+          expect(result).to eq("LineLength" => { "Max" => 90 })
+        end
+      end
+
+      context "when config is an invalid URL" do
+        it "returns parsed config at URL" do
```
teoljungberg: Should this be `it "raises an error"`?

salbertson: Yes, thank you.

PR #953, opened 2015-10-07
Allow multiline block chaining

teoljungberg:
Allow multiline block chaining

For testing that an exception occured, multiline block chaining
might be needed to follow other guides we have.

Example:

```ruby
expect do
  Config::Parser.load_yaml(hound_config, "config/rubocop.yml")
end.to raise_error do |exception|
   expect(exception).to be_a Config::Parser::Error
   expect(exception.message).to eq(
     "`config/rubocop.yml` must be a Hash",
   )
   expect(exception.filename).to eq "config/rubocop.yml"
end
```

- To keep the line under 80 characters, it sometimes can't be re-written
  using `{}` instead of `do/end`.
- `{}` shouldn't be used for multiline blocks.
- If you want to assert more things on the exception, you have to yield
  a block to `raise_error` to do so.

I propose to relax this rule by allowing to chain blocks that spans
over multiple lines.

### DIFF
--- a/.ruby-style.yml
+++ b/.ruby-style.yml
@@ -757,10 +757,6 @@ Style/ModuleFunction:
   Description: Checks for usage of `extend self` in modules.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#module-function
   Enabled: false
-Style/MultilineBlockChain:
-  Description: Avoid multi-line chains of blocks.
-  StyleGuide: https://github.com/bbatsov/ruby-style-guide#single-line-blocks
-  Enabled: true
 Style/MultilineBlockLayout:
   Description: Ensures newlines after multiline block do statements.
   Enabled: true
--- a/config/style_guides/thoughtbot/ruby.yml
+++ b/config/style_guides/thoughtbot/ruby.yml
@@ -356,3 +356,8 @@ TimeZone:
 
 Validation:
   Enabled: false
+
+Style/MultilineBlockChain:
+  Description: 'Avoid multi-line chains of blocks.'
+  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#single-line-blocks'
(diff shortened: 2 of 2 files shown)


mike-burns:
Works for me.

config/style_guides/thoughtbot/ruby.yml
```diff
@@ -356,3 +356,8 @@ TimeZone:
 
 Validation:
   Enabled: false
+
+Style/MultilineBlockChain:
+  Description: 'Avoid multi-line chains of blocks.'
+  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#single-line-blocks'
+  Enabled: true
```
gylaz: Should this be `false`?

.ruby-style.yml
```diff
@@ -760,7 +760,7 @@ Style/ModuleFunction:
 Style/MultilineBlockChain:
   Description: Avoid multi-line chains of blocks.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#single-line-blocks
-  Enabled: true
+  Enabled: false
```
gylaz: We can probably remove this, if we make this change to thoughtbot styles.

gylaz:
I'm not against this.

.ruby-style.yml
teoljungberg: I thought `.ruby-style.yml` was used by Hound for the Hound codebase. So if we change the thoughtbot style, should we remove this?

config/style_guides/thoughtbot/ruby.yml
teoljungberg: Good catch, too much copy pasting. Thanks

arunagw:
What's stopping this to be merged?

PR #954, opened 2015-10-08
Properly indent here-docs

teoljungberg:
As suggested in
https://github.com/thoughtbot/hound/pull/950#discussion_r41444874


gylaz:
:+1:

PR #955, opened 2015-10-08
Remove legacy way of white-listing bulk customers

gylaz:
Why:

* We now have a `BulkCustomer` model that is backed by database.


salbertson:
Looks good to me. :+1:

Koronen:
Remove `EXEMPT_ORGS` from `.env` too?

teoljungberg:
We probably have dead env-variables in both staging and production aswell, we should clean those up

PR #957, opened 2015-10-09
Order language logos alphabetically

thorncp:
I assume Ruby was first because it was supported first and is still has
the largest percentage of builds. However, with the move towards equal
opportunity linters, it's awkward to have Ruby up front. Sorting
alphabetically seems good to me.

In c0ea6270a5b7aaccc0c8499fa047b7b06836543e, I forgot to create a
feature branch and accidentally pushed to master :frowning: . Since it
doesn't include any breaking changes, I think I don't want to muddy
the history further by introducing a revert commit, then going through
our PR flow. I'm hoping for this one to serve in its place.

hound

### DIFF
--- a/app/views/home/index.haml
+++ b/app/views/home/index.haml
@@ -22,20 +22,8 @@
         %ul.languages
           %li.language
             .language-logo
-              = image_tag "languages/ruby.svg"
-            %h5 Ruby
-          %li.language
-            .language-logo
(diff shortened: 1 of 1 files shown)


salbertson:
:+1:

thorncp:
Removed Python in fd6ccae8be04e976bf5e55b55aad564ec1cc81af

hound

PR #958, opened 2015-10-09
Optimize language logos

thorncp:
These logos were added as-is after being exported from whatever program
created them (Sketch, I think). They contain extraneous information that
we don't need on the web, so we can save some bandwidth by removing it.

Also removed an unneeded float.

More about SVG optimization at:
https://robots.thoughtbot.com/organized-workflow-for-svg


gylaz:
I assume it looks the same still? :+1: in that case.

ldamon:
Checked in browser and it looks the same. :+1:

PR #960, opened 2015-10-12
Introduce a `Config` class for each `Linter`

teoljungberg:
I removed the merge of #950 when I remembered this comment
https://github.com/thoughtbot/hound/pull/950#discussion_r41439933.

So I'm re-opening this to make sure that it's answered and corrected.

To decouple the different kinds of configuration options Hound supports,
let each linter decide on the specific kind of configuration it needs.
Both in what format, how to parse it, and how to access it.

Changes:

- Add `HoundConfig` to be the object representation of `.hound.yml`.
- Add `Config::LINTER` to represent the configuration for the specific `LINTER`.
- Each subclass of `Config::Base` implements their own `parse` method,
  that takes the content of the file it should parse, and do so
  accordingly.
- Wrap parsers in `Config::Parser`.
- Replace, and remove, `RepoConfig` with `HoundConfig` and/or
  `Config::LINTER` (where applicable).
- Remove `NAME` constant from `Linter::*`. `NAME` was used to set
  `Linter::Base#_linter_name`, which can be determined by
  `Linter::Base#name` instead.

https://trello.com/c/PoRwwKIw


app/models/hound_config.rb
```diff
@@ -0,0 +1,55 @@
...
+  )
+
+  attr_reader_initialize :commit
+
+  def content
+    @content ||= parse(commit.file_content(CONFIG_FILE))
+  end
+
+  def enabled_for?(name)
+    supported_language?(name) || legacy_key?(name) || enabled?(name)
```
teoljungberg: @gylaz You raised concerns about this implementation in https://github.com/thoughtbot/hound/pull/950#discussion_r41439933. What can we do to make it correct?

teoljungberg: @gylaz I think [this] commit solves the issue

What do you think?

[this]: https://github.com/thoughtbot/hound/commit/68fce3e19b241c4f4598040dd8e2f3fb3a3714f6

teoljungberg: After sleeping on it, I ended up with 05913ba

PR #961, opened 2015-10-13
Add support for `Eslint` alongside `JShint`

teoljungberg:
Add support for `eslint` backed by `hound-eslint`

Eslint is disabled by default, alongside any other Beta linters.

To enabled it, simplify configure your `.hound.yml`:

```yaml
eslint:
  enabled: true
  config_file: config/.eslintrc
  ignore_file: config/.eslintignore
```

This is backed by [hound-eslint]

[hound-eslint]: https://github.com/thoughtbot/hound-eslint

Changes:

- Implement `Config::Eslint` and `Linter::Eslint`.
- Support `*.js`, `*.es6`, and `*.es6.js`, for file extensions.
- Replace `!collection.any?` with `collection.none?`

https://trello.com/c/2a6yQORa

### DIFF
--- a/app/models/config/eslint.rb
+++ b/app/models/config/eslint.rb
@@ -1,0 +1,9 @@
+module Config
+  class Eslint < Base
+    private
+
+    def parse(file_content)
+      Parser.raw(file_content)
+    end
+  end
+end
--- a/app/models/linter/eslint.rb
+++ b/app/models/linter/eslint.rb
@@ -1,0 +1,5 @@
+module Linter
+  class Eslint < Base
+    FILE_REGEXP = /.+((?<!\.coffee)\.js|(?:\.es6|\.es6\.js))\z/
+  end
+end
--- a/app/models/linter/java_script.rb
+++ b/app/models/linter/java_script.rb
@@ -19,7 +19,7 @@ module Linter
     end
 
     def file_included?(commit_file)
-      !excluded_files.any? do |pattern|
+      excluded_files.none? do |pattern|
         File.fnmatch?(pattern, commit_file.filename)
       end
     end
--- a/app/jobs/eslint_review_job.rb
+++ b/app/jobs/eslint_review_job.rb
@@ -1,0 +1,3 @@
+class EslintReviewJob
+  @queue = :eslint_review
+end
--- a/app/models/hound_config.rb
+++ b/app/models/hound_config.rb
@@ -1,6 +1,7 @@
 class HoundConfig
   CONFIG_FILE = ".hound.yml"
   BETA_LANGUAGES = %w(
+    eslint
     python
     swift
   )
--- a/app/models/linter/collection.rb
+++ b/app/models/linter/collection.rb
@@ -2,6 +2,7 @@ module Linter
   class Collection
     LINTERS = [
       Linter::CoffeeScript,
+      Linter::Eslint,
       Linter::Go,
       Linter::Haml,
       Linter::JavaScript,
--- a/spec/models/config/eslint_spec.rb
+++ b/spec/models/config/eslint_spec.rb
@@ -1,0 +1,23 @@
+require "spec_helper"
+require "app/models/config/base"
+require "app/models/config/eslint"
+
+describe Config::Eslint do
+  it_behaves_like "a service based linter" do
+    let(:raw_config) do
+      <<-EOS.strip_heredoc
+        rules:
+            quotes: [2, "double"]
+      EOS
+    end
+
+    let(:hound_config_content) do
+      {
+        "eslint" => {
+          "enabled" => true,
+          "config_file" => "config/.eslintrc",
+        },
+      }
+    end
+  end
+end
--- a/spec/models/linter/eslint_spec.rb
+++ b/spec/models/linter/eslint_spec.rb
@@ -1,0 +1,84 @@
+require "rails_helper"
+
+describe Linter::Eslint do
+  describe ".can_lint?" do
+    context "given an .es6 file" do
+      it "returns true" do
+        result = Linter::Eslint.can_lint?("foo.es6")
+
+        expect(result).to eq true
+      end
+    end
+
+    context "given an .es6.js file" do
+      it "returns true" do
+        result = Linter::Eslint.can_lint?("foo.es6.js")
+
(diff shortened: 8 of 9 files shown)


seanpdoyle:
@teoljungberg could we include JSCS (backed by https://github.com/thoughtbot/hound-jscs) as well?

teoljungberg:
@seanpdoyle My idea was to make separate PRs for JSCS and Eslint. Do you think it makes sense to have them all done in one PR?

seanpdoyle:
Oh, sorry, I read this as ESLint **and** the new JSHint (backed by https://github.com/thoughtbot/hound-jshint).

Carry on

app/models/config/eslint.rb
```diff
@@ -0,0 +1,9 @@
+module Config
+  class Eslint < Base
+    private
+
+    def parse(file_content)
+      Parser.raw(file_content)
```
salbertson: Couldn't we simply return `file_content`?

app/models/linter/eslint.rb
```diff
@@ -0,0 +1,8 @@
+module Linter
+  class Eslint < Base
+    FILE_REGEXP = Regexp.union(
+      JavaScript::FILE_REGEXP,
```
salbertson: What do you think about moving regexp to `Base` and naming it something like `JAVASCRIPT_FILE_REGEXP`? Is there a reason it should stay in `JavaScript`?

app/models/linter/java_script.rb
```diff
@@ -19,7 +19,7 @@ def file_review(commit_file)
     end
 
     def file_included?(commit_file)
-      !excluded_files.any? do |pattern|
+      excluded_files.none? do |pattern|
```
salbertson: :+1:

app/models/style_checker.rb
```diff
@@ -19,29 +19,42 @@ def review_files
 
   private
 
+  JAVASCRIPT_LINTERS = [
+    Linter::Eslint,
+    Linter::JavaScript,
+  ]
```
salbertson: Why are these broken out from `LINTERS`?

app/models/style_checker.rb
```diff
@@ -19,29 +19,42 @@ def review_files
...
-  ].freeze
+  ] + JAVASCRIPT_LINTERS
+
+  DEFAULT_LINTER = Linter::Unsupported
 
   def build_linter(filename)
-    linter_class = find_linter_class(filename)
-    linters[linter_class] ||= linter_class.new(
+    linter_classes = find_linter_classes(filename)
+    linter = linter_classes.
```
salbertson: Does this limit one linter per file?

app/models/config/eslint.rb
teoljungberg: I wanted the `parse` method to have the uniform sense of abstraction. `Parser.raw` returns what it's given

app/models/linter/eslint.rb
teoljungberg: I don't follow, what would be the benefit of that?

Every linter has its own `FILE_REGEXP` constant, which is used to determine which Linter to use. So this fits into that architecture.
I reused `JavaScript::FILE_REGEXP` to avoid duplicating the patterns that the JavaScript linter uses. Do you think that's a bad move?

app/models/style_checker.rb
teoljungberg: It does, and I'm a bit torn by it. My thinking is that it would be confusing for the user if they got comments from two different linters, which might collide. But that might be an edge-case. What do you think?

app/models/style_checker.rb
teoljungberg: They don't need to be, good catch.

app/models/config/eslint.rb
salbertson: Cool, works for me.

app/models/linter/eslint.rb
salbertson: Got it. I think if we are sharing the regexp it shouldn't be tied to a specific linter, that's all. Not a big deal, I'll leave it up to you.

app/models/style_checker.rb
salbertson: :+1:

app/models/style_checker.rb
salbertson: I think it would be awesome to have multiple linters to run for the same file, especially if the linters are complimentary, like the JavaScript linters.

teoljungberg: Let's try out having multiple linters run at the same time. If the reported violations are ambiguous, we can revert back to this behaviour

teoljungberg: I opened https://github.com/thoughtbot/hound/pull/963 to add support for multiple linters. To keep the diff of this PR even smaller

salbertson: :+1:

app/models/linter/eslint.rb
thorncp: Is duplicating the regex completely out of the question? We're dealing with two separate linters, which happen to share some implementation detail. I don't feel strongly about it, but my preference would be for each linter to stand on its own.

thorncp:
LGTM after incorporating #963

app/models/linter/eslint.rb
teoljungberg: Good point, let's move it back to having each linter be responsible for its own regexp.

teoljungberg:
I'll hold of incorporating the changes until #963 has been flushed out

teoljungberg:
I've updated the PR to reflect the new setup from #963.

I'll hold of merging this until we've setup staging and production servers for [hound-eslint]

[hound-eslint]: /thoughtbot/hound-eslint

thorncp:
I love how small of a change this is. :+1:

teoljungberg:
`hound-eslint` successfully works on staging: https://github.com/teoljungberg/houndradish/pull/14

teoljungberg:
I made a PR to [hound-eslint] to prepare it to go live. ping @seanpdoyle 

[hound-eslint]: https://github.com/thoughtbot/hound-eslint/pull/1

thorncp:
:shipit:

PR #962, opened 2015-10-13
Send `excluded_files` to services

teoljungberg:
This is a step closer to push the responsibilities down to each service instead of having it living in `Hound` itself.

Use case: If a user adds Bootstrap's stylesheets to a project in a PR, Hound will pickup on it and try to lint them with whichever linter is responsible for it. But, since they are "vendored" code - it's not something you want linted. 

Format, send it as comma-separated list.
```json
{
   "excluded_files": "pattern_1, pattern_2"
}
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,6 +147,8 @@ Linter jobs are created with the following arguments:
 
 * `config` - The configuration for the linter. This will be linter specific.
 * `content` - The source code to check for violations.
+* `excluded_files` - A comma-separated list of with patterns that should be
+  ignored by the linter. i.e `vendor/assets/stylesheets/*`
 * `filename` - The name of the source file for the code snippet. This should be
   passed through to the outbound queue.
 * `commit_sha` - The git commit SHA of the code snippet. This should be passed
--- a/app/models/linter/base.rb
+++ b/app/models/linter/base.rb
@@ -43,6 +43,7 @@ module Linter
         commit_sha: build.commit_sha,
         config: config.content,
         content: commit_file.content,
+        excluded_files: config.excluded_files.join(", "),
         filename: commit_file.filename,
         patch: commit_file.patch,
         pull_request_number: build.pull_request_number,
--- a/spec/models/linter/go_spec.rb
+++ b/spec/models/linter/go_spec.rb
@@ -40,12 +40,13 @@ describe Linter::Go do
 
       expect(Resque).to have_received(:enqueue).with(
         GoReviewJob,
-        filename: commit_file.filename,
         commit_sha: build.commit_sha,
-        pull_request_number: build.pull_request_number,
-        patch: commit_file.patch,
-        content: commit_file.content,
         config: Config::Go::DEFAULT_CONFIG,
+        content: commit_file.content,
+        excluded_files: "",
+        filename: commit_file.filename,
+        patch: commit_file.patch,
+        pull_request_number: build.pull_request_number,
       )
     end
   end
--- a/spec/models/linter/python_spec.rb
+++ b/spec/models/linter/python_spec.rb
@@ -35,7 +35,7 @@ describe Linter::Python do
       allow(Resque).to receive(:push)
       build = build(:build, commit_sha: "foo", pull_request_number: 123)
       linter = build_linter(build)
-      stub_python_config("config")
+      stub_python_config(content: "config")
       commit_file = build_commit_file
 
       linter.file_review(commit_file)
@@ -45,12 +45,13 @@ describe Linter::Python do
         {
           class: "review.PythonReviewJob",
           args: [
-            filename: commit_file.filename,
             commit_sha: build.commit_sha,
-            pull_request_number: build.pull_request_number,
-            patch: commit_file.patch,
-            content: commit_file.content,
             config: "config",
+            content: commit_file.content,
+            excluded_files: "",
+            filename: commit_file.filename,
+            patch: commit_file.patch,
+            pull_request_number: build.pull_request_number,
           ],
         }
       )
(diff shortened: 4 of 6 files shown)


app/models/linter/base.rb
```diff
@@ -43,6 +43,7 @@ def build_review_job_attributes(commit_file)
         commit_sha: build.commit_sha,
         config: config.content,
         content: commit_file.content,
+        excluded_files: config.excluded_files.join(", "),
```
teoljungberg: Is there a better format to send this as? I have no strong preference, but I want a standard way how the data is sent

seanpdoyle: Are these filenames, or the contents of the excluded files?

teoljungberg: @seanpdoyle That's depending on the linter. But commonly they are `fnmatch`-able patterns. Like `vendor/assets/stylesheets/*`

teoljungberg: That's atleast what I've seen on scss_lint, rubocop, and `eslint`

teoljungberg: Is the name confusing, @seanpdoyle? What could make it clearer?

seanpdoyle: Sorry, I think I'm confusing `excluded_files` with [`ignore_file`](https://github.com/thoughtbot/hound/pull/961#issue-111146532).

`excluded_files` is a Hound feature, while `ignore_file` is specific to the linters, right?

In that case, I'm fine with the globbing.

teoljungberg: `exclude_files` with be the contents of `ignore_file`, in some reasonable format.

CONTRIBUTING.md
```diff
@@ -147,6 +147,8 @@ Linter jobs are created with the following arguments:
 
 * `config` - The configuration for the linter. This will be linter specific.
 * `content` - The source code to check for violations.
+* `excluded_files` - A comma-separated list of with patterns that should be
+  ignored by the linter. i.e `vendor/assets/stylesheets/*`
```
arunagw: As talked in person we might need to specify that this is a optional arguments for Linters.

app/models/linter/base.rb
salbertson: I would suggest sending the content as plain text and let the linter handle the parsing. This is similar to what we do for configuration files.

salbertson:
@teoljungberg is the place to send `excluded_files` to each linter?

Do all linters support excluding files?

Is this a custom "Hound" option for supported by each linter differently?

thorncp:
Is this a responsibility we want to push to the services? If we do, Hound will have to fetch the contents of each file so it can schedule the job, then the service can make the decision to lint the file or not. It seems like we'd see a dramatic increase in our GitHub API usage. Can we handle it?

teoljungberg:
@salbertson

> is the place to send excluded_files to each linter?

I would say so, but am I missing something here? Can you elaborate?

> Do all linters support excluding files?

After looking at the different linters, I gathered this list:

- Coffeelint: no
- Rubocop: yes
- scss-lint: yes
- jshint: yes
- haml-lint: yes
- go: maybe (we exclude vendored files in Hound proper, but exclusion is not supported by the linter itself)
- swift: yes
- python: yes

Future linters:
- eslint: yes
- jscs: yes

So if this were to be implemented, Coffeelint and Go(maybe?) would just not care about the `excluded_files` options and do their work themselves.

> Is this a custom "Hound" option for supported by each linter differently?

This would be implemented by the different linters depending on their needs, as laid out above - some linters don't have support for file exclusion. So they would just not care about this parameter.

@thorncp 
> Is this a responsibility we want to push to the services? If we do, Hound will have to fetch the contents of each file so it can schedule the job, then the service can make the decision to lint the file or not. It seems like we'd see a dramatic increase in our GitHub API usage. Can we handle it?

We might have one more API call, if there is an `ignore_file` specified. If there isn't one, we wouldn't fetch it.
Some linters, i.e Rubocop and future JSCS have their exclusions specified in the `config_file`. So that would be another API call.

I know that @gylaz and I have talked about moving as much as possible out of Hound proper and into the linters. It has been done for hound-scss [here](https://github.com/thoughtbot/hound-scss/pull/14)

thorncp:
> We might have one more API call, if there is an ignore_file specified. If there isn't one, we wouldn't fetch it.

I'm confused. Won't we have to fetch the contents of the source code file? It's a requirement of enqueueing a review job, which will be scheduled if the linter itself is responsible for ignoring things.

I feel like I'm missing something :confused:

teoljungberg:
I feel like we're talking past each-other @thorncp, or I'm misunderstanding your misunderstandings.
This change doesn't mean that we will fetch the contents of every file, only that we will fetch the `ignore_file` (if one is specified).

Which parts are unclear to you?

thorncp:
Indeed, I think I'm making loops. :baby_chick: 

I'll try to start over.

> This is a step closer to push the responsibilities down to each service instead of having it living in Hound itself.

My understanding of this is we're moving towards the services deciding whether or not to lint a file based on _ignore_file_. In order for the service to do that, a review job must be enqueued. One of the required parameters for enqueueing a review job is the raw content of the source file in question. It seems to me that we will be fetching _that_ file's content (so we can enqueue the job), then the service will do nothing with it if it's ignored.

Questions from there:

* Am I understanding that correctly?
* If so, can we handle that increased API Load?
* Can we do _something else_ to mitigate that load?

teoljungberg:
> Am I understanding that correctly?

You are, now we're on the same page. Thanks for summarizing your questions.
Different linters might have different solutions on how the exclude files, some even don't have a specific `ignore_file`. 

Actually, as I'm thinking through this. There might be an alternative approach, if the linter doesn't have a way to exclude files, which `haml` and `go` doesn't, we can use the `ignore_file` specified in a more generic manor.  Then, it doesn't matter if the different linters have their own implementations on which files to exclude/include. Since we have a generic `ignore_file` pattern and use that. So maybe this is a non-issue.

What do you think?

gylaz:
I don't think we should be fetching ignored/excluded files from GitHub.

teoljungberg:
After discussing this, I feel like this isn't the right move. Closing

PR #963, opened 2015-10-16
Support multiple linters per file

teoljungberg:
`Linter::Collection` wraps a list of linters. To work as a replacement
for a single `Linter`, even though multiple linters are used.

### DIFF
--- a/app/models/linter/collection.rb
+++ b/app/models/linter/collection.rb
@@ -1,0 +1,40 @@
+module Linter
+  class Collection
+    LINTERS = [
+      Linter::CoffeeScript,
+      Linter::Go,
+      Linter::Haml,
+      Linter::JavaScript,
+      Linter::Python,
+      Linter::Ruby,
+      Linter::Scss,
+      Linter::Swift,
+    ].freeze
+
+    def self.for(filename:, **linter_args)
+      linter_classes = LINTERS.
+        select { |linter_class| linter_class.can_lint?(filename) }.
+        map { |linter_class| linter_class.new(**linter_args) }
+
+      new(linter_classes.presence || default_linter(**linter_args))
+    end
+
+    attr_reader :linters
+
+    def initialize(linters = [])
+      @linters = Array(linters)
+    end
+
+    def file_review(commit_file)
+      linters.
+        select(&:enabled?).
+        select { |linter| linter.file_included?(commit_file) }.
+        each { |linter| linter.file_review(commit_file) }
+    end
+
+    def self.default_linter(**linter_args)
+      Linter::Unsupported.new(**linter_args)
+    end
+    private_class_method :default_linter
+  end
+end
--- a/app/models/style_checker.rb
+++ b/app/models/style_checker.rb
@@ -2,46 +2,25 @@
 # Builds style guide based on file extension.
 # Delegates to style guide for line violations.
 class StyleChecker
-  pattr_initialize :pull_request, :build do
-    @linters = {}
-  end
-  attr_private :linters
+  pattr_initialize :pull_request, :build
 
   def review_files
     pull_request.commit_files.each do |commit_file|
-      linter = build_linter(commit_file.filename)
+      collection = build_linter_collection(commit_file.filename)
 
-      if linter.enabled? && linter.file_included?(commit_file)
-        linter.file_review(commit_file)
-      end
+      collection.file_review(commit_file)
     end
   end
 
   private
 
-  LINTERS = [
-    Linter::CoffeeScript,
-    Linter::Go,
-    Linter::Haml,
-    Linter::JavaScript,
-    Linter::Python,
-    Linter::Ruby,
-    Linter::Scss,
-    Linter::Swift,
-    Linter::Unsupported,
-  ].freeze
-
-  def build_linter(filename)
-    linter_class = find_linter_class(filename)
-    linters[linter_class] ||= linter_class.new(
+  def build_linter_collection(filename)
+    Linter::Collection.for(
+      filename: filename,
       hound_config: hound_config,
       build: build,
       repository_owner_name: pull_request.repository_owner_name,
     )
-  end
-
-  def find_linter_class(filename)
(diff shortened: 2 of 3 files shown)


app/models/linter/collection.rb
```diff
@@ -0,0 +1,29 @@
+module Linter
+  class Collection
+    attr_reader :entries
```
arunagw: Can this be `private` or `protected` ?

teoljungberg: I don't think it needs to be public. Done in f1d0cc9

arunagw:
Looks fine to me. But another set of eye will be good to have on this.

app/models/linter/collection.rb
```diff
@@ -0,0 +1,29 @@
+module Linter
+  class Collection
+    attr_private :entries
+
+    def initialize(entries = [])
+      @entries = Array(entries)
+    end
+
+    def presence
+      entries.presence && self
```
seanpdoyle: What is the significance of `&& self`?

teoljungberg: `entires.presence` will return either `nil` or `entries` itself. And calling `collection.presence` I expect to get back the `collection` or `nil`.

Since it's a bit obscure, Is there a better way to do this?

seanpdoyle: Will this expression ever evaluate to `nil`, or `entries`/`false`?

teoljungberg: It will if `entries` is empty

teoljungberg: It doesn't matter if it's `nil` or `false. We only want a truthy or falsey value to have this expression working: https://github.com/thoughtbot/hound/blob/multiple-linters/app/models/style_checker.rb#L38-L39

app/models/style_checker.rb
```diff
@@ -28,20 +28,35 @@ def review_files
...
-    linter_class = find_linter_class(filename)
-    linters[linter_class] ||= linter_class.new(
+  def build_linters(filename)
+    linters = find_linter_classes(filename).
+      map { |linter_class| build_class(linter_class) }
+    collection = build_collection(linters)
+
+    all_linters[filename] ||= collection.presence ||
+      build_collection(build_default_linter)
+  end
```
salbertson: Now that we have a collection class what do you think about moving this logic to that class?

salbertson:
```
linters = Linter::Collection.for(filename)
```

Something like that? Just move more of the "build linters" logic into the collection class so `StyleChecker` isn't responsible and the file reads better.

app/models/linter/collection.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def enabled?
+      entries.all?(&:enabled?)
```
thorncp: Does `all?` accurately represent the collection? If I'm following `StyleChecker#review_files` correctly, we build a list of possible linters, then check if _all_ of them are enabled. If any of them respond false, none of them will lint the file. This also seems to be the case with `file_included?`.

Am I understanding this correctly?

app/models/style_checker.rb
thorncp: :+1: I think this would be great in the collection class.

teoljungberg: That makes alot of sense. Thanks!

app/models/linter/collection.rb
teoljungberg: Chris - you're right. We should run `file_review` on those linters that are `enabled?`. Thanks

app/models/style_checker.rb
```diff
@@ -2,46 +2,35 @@
...
+  pattr_initialize :pull_request, :build
 
   def review_files
     pull_request.commit_files.each do |commit_file|
-      linter = build_linter(commit_file.filename)
+      collection = build_linter_collection(commit_file.filename)
 
-      if linter.enabled? && linter.file_included?(commit_file)
-        linter.file_review(commit_file)
+      if collection.enabled? && collection.file_included?(commit_file)
```
georgebrock: Is this a violation of “tell, don't ask”?

app/models/style_checker.rb
```diff
@@ -2,46 +2,35 @@
...
+  def build_linter_collection(filename)
+    Linter::Collection.for(
+      filename: filename,
+      class_builder: class_builder,
     )
   end
 
-  def find_linter_class(filename)
-    LINTERS.detect { |linter_class| linter_class.can_lint?(filename) }
+  def class_builder
```
georgebrock: This class builder seems a little complex, and is repeated here and in the specs. Any reason these keyword args can't just be passed to `Linter::Collection.for`?

Maybe you already tried this. Writing it out so we can see what it'd look like:

```ruby
Linter::Collection.for(
  filename: filename,
  hound_config: hound_config,
  build: build,
  repository_owner_name: pull_request.repository_owner_name,
)
```

```ruby
def self.for(filename:, **linter_args)
  linter_classes = LINTERS.
    select { |linter_class| linter_class.can_lint?(filename) }.
    map { |linter_class| linter_class.new(**linter_args) }

  new(linter_classes.presence || default_linter(**linter_args))
end
```

georgebrock:
@teoljungberg Looks good. I really like the consistent interface between the collection and the individual linters. The one part that seemed strange—and broke that consistency—was constructing the individual linters in a lambda.

app/models/style_checker.rb
teoljungberg: I went back and forth on this. And my reasoning is that it doesn't feel like the `Collection`'s responsibility to know about the different components it takes to build up a linter: it just wants to create one.

What do you think?

app/models/style_checker.rb
teoljungberg: I think you're right, since we now have moved filtering out the linters that are `enabled?` into `#file_review`. I think it makes sense to move `file_included?` there too and remove this if-check entirely.

Thank you

teoljungberg:
Thanks for the review @georgebrock, @salbertson, and @thorncp!

PR #964, opened 2015-10-16
Handle missing linter configuration

salbertson:
Should address https://app.getsentry.com/hound-1/production/group/88432966/.


thorncp:
:+1:

PR #966, opened 2015-10-19
Remove `StyleConfig`

teoljungberg:
We decided not to save the configuration in the database.

https://trello.com/c/pDvJ8BbS

### DIFF
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -1,6 +1,5 @@
 class Owner < ActiveRecord::Base
   has_many :repos
-  has_many :style_configs, dependent: :destroy
 
   def self.upsert(github_id:, name:, organization:)
     owner = find_or_initialize_by(github_id: github_id)
--- a/app/models/style_config.rb
+++ b/app/models/style_config.rb
@@ -1,3 +1,0 @@
-class StyleConfig < ActiveRecord::Base
-  belongs_to :owner
-end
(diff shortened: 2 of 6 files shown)


arunagw:
If this table is empty on production and not used anywhere else.. this looks fine to me.

teoljungberg:
The table is empty in production:

```
hound-production::SILVER=> select * from style_configs;
(No rows)
```

arunagw:
Also is `Owner` table used anywhere? I can see it gets filled but is it used anywhere?

arunagw:
But we can do `Owner` table in another PR or issue

teoljungberg:
@arunagw Since the `Owner` is pretty coupled with the `StyleConfig` - I opted to delete the owner here aswell.

arunagw:
Sounds good. Let's wait for more review here before merging this in master.

thorncp:
Yep, `Owner` was introduced as a precursor to organization level config. See #586. I don't think it was ever used for anything else, but perhaps someone is using it for statistics? @gylaz or @salbertson ? If not, is there meaningful data we can get from it that we aren't?

salbertson:
`Owner` is not being used for stats. Why are we removing the owner code? Is it causing issues?

teoljungberg:
I opted to delete it since it's dormant code, we're not using it for anything.

Will we? Or does it only make sense to delete the style-config?

---
Teo Ljungberg

gylaz:
`Owner` was added to back fill information about GitHub users/orgs. We will be adding org-level configuration soon-ish, so this is probably good to keep. As for `StyleConfig`, I believe it can be removed.

teoljungberg:
@gylaz Changed in 1f79b79

gylaz:
:+1:

PR #968, opened 2015-10-20
Add specs for User#token

arunagw:


arunagw:
@teoljungberg want to have a quick look?

spec/models/user_spec.rb
```diff
@@ -31,6 +31,44 @@
...
+    it "returns original value of token on call" do
+      user = build(:user)
+      user.token = "original-token"
+
+      user.save
+
+      expect(user.reload.token).to eq("original-token")
+    end
+
+    it "uses ActiveSupport::MessageEncryptor to encrypted token" do
```
teoljungberg: I feel like this PR isn't needed, since it deeply tests the implementation of the `crypt` method. If we change the implementation of `crypt` this will break.

But I would like to see a test where we are testing that the value is encrypted

arunagw:
@teoljungberg PR updated. Final look please.

teoljungberg:
LGTM( :moon: )

PR #970, opened 2015-10-21
Allow supported languages to be unconfigured

teoljungberg:
Linters that are supported[1] out of the box for Hound, should not be
enabled in the users `.hound.yml`[1]
Beta languages still needs to be enabled in the configuration[2]

[1]
```yaml
ruby:
  enabled: true
```

[2]
```yaml
python:
  enabled: true
```

### DIFF
--- a/spec/models/linter/base_spec.rb
+++ b/spec/models/linter/base_spec.rb
@@ -29,4 +29,45 @@ describe Linter::Base do
       expect(linter.file_included?(double)).to eq true
     end
   end
+
+  describe "#enabled?" do
+    context "when the hound config is enabled for the given language" do
+      it "returns true" do
+        hound_config = double("HoundConfig", enabled_for?: true)
+        linter = Linter::Test.new(
+          hound_config: hound_config,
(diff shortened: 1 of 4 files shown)


arunagw:
Looks fine to me. Make sure to verify on staging as well.

teoljungberg:
Verified that this works as expected in https://github.com/teoljungberg/houndradish/pull/9 using [this] configuration

[this]: https://github.com/teoljungberg/houndradish/blob/ruby-violation/.hound.yml#L12-L13

spec/models/linter/base_spec.rb
```diff
@@ -29,4 +29,45 @@ class Test < Base
...
+          hound_config: hound_config,
+          build: double,
+          repository_owner_name: "foo",
+        )
+
+        expect(linter.enabled?).to eq false
+      end
+    end
+
+    context "when the hound config is disabled for the given language" do
```
gylaz: This is the same string as line 47. What should be the difference?

PR #971, opened 2015-10-21
Support `inherit_from` from Rubocop

teoljungberg:
A regression was introduced in fec130b, where Hound didn't
respect the specified files to inherit its configuration from.

### DIFF
--- a/app/models/config/ruby.rb
+++ b/app/models/config/ruby.rb
@@ -4,7 +4,8 @@ module Config
       if legacy?
         hound_config.content
       else
-        super
+        result = super
+        parse_inherit_from(result)
       end
     end
 
@@ -14,6 +15,18 @@ module Config
       result = Parser.yaml(file_content)
 
       ensure_correct_type(result)
+    end
+
+    def parse_inherit_from(config)
+      inherit_from = Array(config.fetch("inherit_from", []))
+
+      inherited_config = inherit_from.reduce({}) do |result, ancestor_file_path|
+        raw_ancestor_config = commit.file_content(ancestor_file_path)
+        ancestor_config = parse(raw_ancestor_config)
+        result.merge(ancestor_config)
+      end
+
+      inherited_config.merge(config.except("inherit_from"))
     end
 
     def legacy?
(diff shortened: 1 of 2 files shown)


teoljungberg:
Verified that it works as expected in https://github.com/teoljungberg/houndradish/pull/10

app/models/config/ruby.rb
```diff
@@ -16,6 +17,18 @@ def parse(file_content)
...
+    def parse_inherit_from(config)
+      inherit_from = Array(config.fetch("inherit_from", []))
+
+      inherited_config = inherit_from.reduce({}) do |config, ancestor_file_path|
+        raw_ancestor_config = commit.file_content(ancestor_file_path)
+        ancestor_config = parse(raw_ancestor_config)
+        config.merge(ancestor_config)
+      end
+
+      inherited_config.merge(config.except("inherit_from"))
```
salbertson: @teoljungberg is this basically the code we had before?

spec/models/config/ruby_spec.rb
```diff
@@ -30,6 +30,40 @@
     end
   end
 
+  context "when the configuration uses `inherit_from`" do
+    it "returns the content from GitHub as a hash" do
```
salbertson: Is this test description accurate?

app/models/config/ruby.rb
teoljungberg: Basically, yes - Here's the old code if you want to diff between the two  https://github.com/thoughtbot/hound/blob/cb7e24e03fb13f9a0fa197af6003bc4376808f9d/app/models/repo_config.rb#L115-L125

spec/models/config/ruby_spec.rb
teoljungberg: Not really, I'm thinking something like:

```
    it "merges the config files from `inherit_from` and returns the sum" do
```

What do you think?

salbertson: :+1:

salbertson:
:+1: :ship:

PR #969, opened 2015-10-21
Upgrade Rubygems according to bundler-audit, add rake bundler:audit rake task

JuanitoFatas:
This Pull Request

- Adds `rake bundler:audit` from https://github.com/thoughtbot/suspenders/commit/e23157ede1477a6e3e1163516281cfda10b479c6
- Uses 1.9 hash syntax in Rakefile
- Upgrades some gems (jquery-rails, uglifier, sentry-raven, and explicitly adds rest-client)  to address the advices from [bundler-audit](https://github.com/rubysec/bundler-audit):

```
$ bundle-audit
Insecure Source URI found: git://github.com/octokit/octokit.rb.git
Name: jquery-rails
Version: 3.1.1
Advisory: CVE-2015-1840
Criticality: Medium
URL: https://groups.google.com/forum/#!topic/ruby-security-ann/XIZPbobuwaY
Title: CSRF Vulnerability in jquery-rails
Solution: upgrade to >= 4.0.4, ~> 3.1.3

Name: rest-client
Version: 1.7.2
Advisory: CVE-2015-1820
Criticality: Unknown
URL: https://github.com/rest-client/rest-client/issues/369
Title: rubygem-rest-client: session fixation vulnerability via Set-Cookie headers in 30x redirection responses
Solution: upgrade to >= 1.8.0

Name: rest-client
Version: 1.7.2
Advisory: CVE-2015-3448
Criticality: Unknown
URL: http://www.osvdb.org/show/osvdb/117461
Title: Rest-Client Gem for Ruby logs password information in plaintext
Solution: upgrade to >= 1.7.3

Name: sentry-raven
Version: 0.9.4
Advisory: CVE-2014-9490
Criticality: Medium
URL: http://osvdb.org/show/osvdb/115654
Title: sentry-raven Gem for Ruby contains a flaw that can result in a denial of service
Solution: upgrade to >= 0.12.2

Name: uglifier
Version: 2.5.3
Advisory: 126747
Criticality: Unknown
URL: https://github.com/mishoo/UglifyJS2/issues/751
Title: uglifier incorrectly handles non-boolean comparisons during minification
Solution: upgrade to >= 2.7.2

Vulnerabilities found!
```

PR #972, opened 2015-10-21
Use bundle audit gem

arunagw:
As we do this in our all new apps in https://github.com/thoughtbot/suspenders/commit/e23157ede1477a6e3e1163516281cfda10b479c6

We should also add this gem to Hound.

Related https://github.com/thoughtbot/hound/pull/969


arunagw:
Closing in favour of #969

PR #973, opened 2015-10-21
Use latest version of Octokit gem

arunagw:
This current version from master was used to have paginated repo support for
an organization. That feature has been released in higher version.


teoljungberg:
Merge away!

arunagw:
Tested on staging. Works fine.

PR #975, opened 2015-10-22
Add tests for `CommitStatus`

teoljungberg:


jakecraige:
LGTM

spec/models/commit_status_spec.rb
```diff
@@ -0,0 +1,132 @@
+require "rails_helper"
+
+describe CommitStatus do
+  describe "#set_pending" do
+    it "sets the pending status on GithubApi" do
+      github_api = stubbed_github_api(create_pending_status: true)
```
arunagw: I think `create_pending_status` can return `nil` as well.

arunagw:
One small comment for return value in stub.. otherwise looks fine to me.

spec/models/commit_status_spec.rb
teoljungberg: As discussed in person. I'll refactor `stubbed_github_api` to take a list of methods and not care about the return value since it's arbitrary in the test

teoljungberg:
Thanks for the review @jakecraige and @arunagw

PR #976, opened 2015-10-22
Allow services to report config files as invalid

teoljungberg:
Either by executing the service, `ReportInvalidConfig`. Or by enqueueing
the corresponding job, `ReportInvalidConfigJob`.

https://trello.com/c/42U5K0Bs

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -173,6 +173,16 @@ Once linting is complete, resulting violations should be posted to the outbound
 * `patch` - The patch content from GitHub for the file being reviewed. This is
   provided by the inbound queue.
 
+If the given `config` is invalid, the invalid file should be posted to the
+outbound `ReportInvalidConfigJob` queue:
+
+* `commit_sha` - The git commit SHA of the code snippet. This is provided by the
+  inbound queue..
+* `filename` - The name of the source file for the code snippet. This is
+  provided by the inbound queue.
+* `pull_request_number` - The GitHub pull request number. This is provided by
+  the inbound queue.
+
 ## Contributor License Agreement
 
 If you submit a Contribution to this application's source code, you hereby grant
--- a/app/jobs/report_invalid_config_job.rb
+++ b/app/jobs/report_invalid_config_job.rb
@@ -1,0 +1,11 @@
+class ReportInvalidConfigJob
+  @queue = :high
+
+  def self.perform(attributes)
+    ReportInvalidConfig.run(
+      pull_request_number: attributes.fetch("pull_request_number"),
+      commit_sha: attributes.fetch("commit_sha"),
+      filename: attributes.fetch("filename"),
+    )
+  end
+end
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -173,6 +173,14 @@ Once linting is complete, resulting violations should be posted to the outbound
 * `patch` - The patch content from GitHub for the file being reviewed. This is
   provided by the inbound queue.
 
+If the given `config` is invalid, queue invalid file the outbound
```
thorncp: What do you think about keeping the verbiage consistent with the sentence above?

Maybe

> If the given `config` is invalid, the invalid file should be posted to the outbound `InvalidConfigFileJob` queue:

CONTRIBUTING.md
```diff
@@ -173,6 +173,14 @@ Once linting is complete, resulting violations should be posted to the outbound
...
+`InvalidConfigFileJob` queue:
+
+* `commit_sha` - The git commit SHA of the code snippet.
+* `filename` - The name of the source file for the code snippet.
+* `pull_request_number` - The GitHub pull request number. This is provided by the
```
thorncp: All three of these bullet points are provided by the inbound queue.

app/jobs/invalid_config_file_job.rb
```diff
@@ -0,0 +1,11 @@
+class InvalidConfigFileJob
+  @queue = :high
+
+  def self.perform(attributes)
+    InvalidConfigFile.run(
```
thorncp: This reads a little odd to me. What does it mean for a file to run? I don't have a suggestions :trollface:

thorncp:
A few questions/suggestions, but looks good overall.

app/jobs/invalid_config_file_job.rb
teoljungberg: Yeah, good point.
I wanted to have the pattern where the job-class only enqueued a service. Would a better name than `InvalidConfigFile` make it clearer? I don't have any suggestions

teoljungberg: `InvalidateConfig` maybe?

CONTRIBUTING.md
teoljungberg: Good idea.

CONTRIBUTING.md
teoljungberg: Good call, I added the explanation to the docs

app/jobs/invalid_config_file_job.rb
thorncp: Hmm, _invalidate_ makes it sound like Hound is causing the error. Maybe a verb that indicates we're informing someone about the error? `ReportInvalidConfig`? `PostInvalidConfig`?

teoljungberg: `ReportInvalidConfig` sounds like a good name, thanks!

PR #979, opened 2015-10-24
Add beta support for JSCS linting via hound-jscs

seanpdoyle:
Enable JSCS linting by setting the `jscs` config key to `true`.

[hound-jscs]: https://github.com/thoughtbot/hound-jscs/

To enable:

```
jscs:
  enabled: true
  config_file: .jscsrc
```

https://trello.com/c/M6hL2wKC

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -2,12 +2,13 @@
 
 # Add new inflection rules using the following format
 # (all these examples are active by default):
-# ActiveSupport::Inflector.inflections do |inflect|
+ActiveSupport::Inflector.inflections do |inflect|
 #   inflect.plural /^(ox)$/i, '\1en'
 #   inflect.singular /^(ox)en/i, '\1'
 #   inflect.irregular 'person', 'people'
 #   inflect.uncountable %w( fish sheep )
-# end
+  inflect.uncountable %w(jscs)
+end
 #
 # These inflection rules are supported but not enabled by default:
 # ActiveSupport::Inflector.inflections do |inflect|
--- a/spec/models/style_checker_spec.rb
+++ b/spec/models/style_checker_spec.rb
@@ -142,6 +142,30 @@ describe StyleChecker do
         end
       end
 
+      context "when JSCS is enabled" do
+        it "does not immediately return violations" do
+          commit_file = stub_commit_file("test.js", "var test = 'test'")
+          head_commit = stub_head_commit(
+            HoundConfig::CONFIG_FILE => <<-EOS.strip_heredoc
+              jscs:
+                enabled: true
+                config_file: config/.jscsrc
+            EOS
+          )
+          pull_request = stub_pull_request(
+            commit_files: [commit_file],
+            head_commit: head_commit,
+          )
+
+          violation_messages = pull_request_violations(pull_request)
+
+          expect(violation_messages).to eq [
+            "Missing semicolon.",
+            "'test' is defined but never used.",
+          ]
+        end
+      end
+
       context "when eslint is disabled or unconfigured" do
         context "with style violations" do
           it "returns violations" do
--- a/app/jobs/jscs_review_job.rb
+++ b/app/jobs/jscs_review_job.rb
@@ -1,0 +1,3 @@
+class JscsReviewJob
+  @queue = :jscs_review
+end
--- a/app/models/config/jscs.rb
+++ b/app/models/config/jscs.rb
@@ -1,0 +1,9 @@
+module Config
+  class Jscs < Base
+    private
+
+    def parse(file_content)
+      Parser.raw(file_content)
+    end
+  end
+end
--- a/app/models/hound_config.rb
+++ b/app/models/hound_config.rb
@@ -2,6 +2,7 @@ class HoundConfig
   CONFIG_FILE = ".hound.yml"
   BETA_LANGUAGES = %w(
     eslint
+    jscs
     python
     swift
   )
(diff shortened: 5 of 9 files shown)


app/helpers/configuration_helper.rb
```diff
@@ -15,6 +15,11 @@ def javascript_ignore_url
     "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/.jshintignore"
   end
 
+  def jscs_config_url
+    "https://raw.githubusercontent.com/thoughtbot/hound-jscs" \
+      "/master/config/default.json"
```
seanpdoyle: This file does not yet exist.

teoljungberg: What is this file going to include?

teoljungberg: And, is it going to be used internally by the linter?

seanpdoyle: This will be the default `.jscsrc` when one isn't specified.

teoljungberg: Doesn't JSCS lint files if it doesn't have a configuration? We decided that we didn't want to to have a default configuration, and use the linters defaults.

Doesn't JSCS have that?

seanpdoyle: Interesting question.

It has an [idea of "presets"](http://jscs.info/overview#presets), but I'm not sure it picks one by default out of the box.

teoljungberg: A good way to test it, is to lint a file and not specify and configuration. What happens then?

I'm skeptical to pick a preset out of the box, since we've chosen to use the linters default settings.

seanpdoyle: womp womp

```
2.2.2 in hound-jscs/ on master
› node_modules/.bin/jscs index.js
No configuration found. Add a .jscsrc file to your project root or use the -c option.
```

teoljungberg: That's unfortunate, does the linter function itself return the same response?

seanpdoyle: womp womp

```
2.2.2 in hound-jscs/ on master
› node
> Linter = require("./lib/linter")
[Function: Linter]
> jobs = [];
[]
> linter = new Linter({ enqueue: function(job) { jobs.push(job) })
... ;
...
>
(^C again to quit)
>

2.2.2 in hound-jscs/ on master
›

2.2.2 in hound-jscs/ on master
› node
> Linter = require("./lib/linter")
[Function: Linter]
> jobs = []
[]
> linter = new Linter({ enqueue: function(job) { jobs.push(job); } })
Linter { outbound: { enqueue: [Function] } }
> linter.lint({ content: "// TODO" })
undefined
> jobs
[ { violations: [],
    filename: undefined,
    commit_sha: undefined,
    pull_request_number: undefined,
    patch: undefined } ]
>
```

teoljungberg: That's interesting!

In that case, I guess we need it

seanpdoyle: It might make sense to come up with our own in thoughtbot/guides, the way we have a rubocop config. I'm not sure how long that would take, given that the rubocop config has had a few years to come about organically.

Given all that, would it make sense to clone a preset?

teoljungberg: I think we can write a set of rules of what we already have in [thoughtbot/guides]. 

[thoughtbot/guides]: /thoughtbot/guides

thorncp: :smiling_imp: If we stick hard to our decision, we'd do nothing because the default does nothing. Is that completely out of the question?

Could we ease that providing documentation/messaging encouraging users to check out the existing presets?

app/views/pages/configuration.haml
```diff
@@ -12,6 +12,8 @@
       %li
         = link_to "Javascript config", "#javascript"
       %li
+        = link_to "JSCS config", "#jscs"
```
teoljungberg: What do you think about omitting the configuration documentation from this PR? Then, we can test JSCS out internally first before announcing the public beta? As we did for ESLint in #980.

teoljungberg:
This all seems totally reasonable to me!
A blocker seems to be the [discussion] around the default configuration for JSCS.

[discussion]: https://github.com/thoughtbot/hound/pull/979/files#r42957071

app/views/pages/configuration.haml
seanpdoyle: https://github.com/thoughtbot/hound/pull/981

config/initializers/inflections.rb
```diff
@@ -2,12 +2,13 @@
...
 # Add new inflection rules using the following format
 # (all these examples are active by default):
-# ActiveSupport::Inflector.inflections do |inflect|
+ActiveSupport::Inflector.inflections do |inflect|
 #   inflect.plural /^(ox)$/i, '\1en'
 #   inflect.singular /^(ox)en/i, '\1'
 #   inflect.irregular 'person', 'people'
 #   inflect.uncountable %w( fish sheep )
-# end
+  inflect.uncountable %w( jscs )
```
teoljungberg: This always bites me. Good catch!

teoljungberg: What do you think about removing the spaces around the parenthesis?

spec/models/style_checker_spec.rb
```diff
@@ -142,6 +142,30 @@
         end
       end
 
+      context "when Jscs is enabled" do
```
teoljungberg: Should this be written as `JSCS`?

teoljungberg:
Have you tested this on staging on a test-repo? If so - how did it go?

PR #980, opened 2015-10-26
Document how to configure EsLint

teoljungberg:
- Sort the Table of Contents for the different languages and linters.

PR #981, opened 2015-10-27
Add HTML Documentation for JSCS

seanpdoyle:
Split from [#979].

Omit the configuration documentation from [#979] so we can test JSCS out
internally before announcing the public beta.

[#979]: https://github.com/thoughtbot/hound/pull/979/files#r43117833

### DIFF
--- a/app/helpers/configuration_helper.rb
+++ b/app/helpers/configuration_helper.rb
@@ -15,11 +15,21 @@ module ConfigurationHelper
     "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/.jshintignore"
   end
 
+  def jscs_config_url
+    config_url("thoughtbot/guides", "style/javascript/.jscsrc")
+  end
+
   def scss_config_url
     "https://raw.githubusercontent.com/thoughtbot/hound-scss/master/config/default.yml"
   end
 
   def haml_config_url
     "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/haml.yml"
+  end
+
+  private
+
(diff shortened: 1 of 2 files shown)


teoljungberg:
I would wait with merging this as we're doing with #980, we can also use the refactoring made there to remedy the long lines.

app/helpers/configuration_helper.rb
```diff
@@ -15,6 +15,10 @@ def javascript_ignore_url
     "https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/.jshintignore"
   end
 
+  def jscs_config_url
+    config_url("thoughtbot/guides", "style/javascript/.jscsrc")
```
seanpdoyle: @teoljungberg is it ok that our default config would live in our style guides? would we be better off moving / duplicating this to `hound-jscs`?

teoljungberg: I prefer having all of the configs living in our styleguide repo. It feels like the right spot to have them in

seanpdoyle:
@teoljungberg https://github.com/thoughtbot/robots.thoughtbot.com/pull/863 has been merged.

Once that post is public, this branch is good to merge, right?

thorncp:
This should probably be merged before that post goes out. I've been out for a few weeks: do we think this is vetted enough to call it a stable release? If so, :sheep:

thorncp:
The Circle failure seemed unrelated. I restarted it.

thorncp:
I think the latest failure just needs a rebase on master to get a security patch.

thorncp:
:+1:

PR #985, opened 2015-10-28
Add beta support for `JSHint` linting via `hound-jshint`

teoljungberg:
Enable JSHint linting backed by [hound-jshint] by setting the `jshint`
config key to `true`.

[hound-jshint]: https://github.com/thoughtbot/hound-jshint/

To enable:

```
jshint:
  enabled: true
  config_file: .jshintrc

javascript:
  enabled: false
```

The current JavaScript linter is already backed by `JSHint`, so to use
the newer version we must disable the old one.

https://trello.com/c/Xk8StXIm

### DIFF
--- a/app/models/linter/jshint.rb
+++ b/app/models/linter/jshint.rb
@@ -1,0 +1,5 @@
+module Linter
+  class Jshint < Base
+    FILE_REGEXP = /.+\.js\z/
+  end
+end
--- a/spec/models/config/jshint_spec.rb
+++ b/spec/models/config/jshint_spec.rb
@@ -1,0 +1,24 @@
+require "spec_helper"
+require "app/models/config/base"
+require "app/models/config/jshint"
+
+describe Config::Jshint do
+  it_behaves_like "a service based linter" do
+    let(:raw_config) do
+      <<-EOS.strip_heredoc
+        {
+          "maxlen": 80,
+        }
+      EOS
+    end
+
+    let(:hound_config_content) do
+      {
+        "jshint" => {
+          "enabled" => true,
+          "config_file" => "config/.jshintrc",
+        },
+      }
+    end
+  end
+end
--- a/spec/models/style_checker_spec.rb
+++ b/spec/models/style_checker_spec.rb
@@ -166,7 +166,30 @@ describe StyleChecker do
         end
       end
 
-      context "when eslint is disabled or unconfigured" do
+      context "when JSHint is enabled" do
+        it "does not immediately return violations" do
+          commit_file = stub_commit_file("test.js", "var test = 'test'")
+          head_commit = stub_head_commit(
+            HoundConfig::CONFIG_FILE => <<-EOS.strip_heredoc
+              jshint:
+                enabled: true
+                config_file: config/.jshintrc
+              javascript:
+                enabled: false
+            EOS
+          )
+          pull_request = stub_pull_request(
+            commit_files: [commit_file],
+            head_commit: head_commit,
+          )
+
+          violation_messages = pull_request_violations(pull_request)
+
+          expect(violation_messages).to eq []
+        end
+      end
+
+      context "when a specifc linter is disabled or unconfigured" do
         context "with style violations" do
           it "returns violations" do
             commit_file = stub_commit_file("test.js", "var test = 'test'")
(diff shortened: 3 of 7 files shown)


seanpdoyle:
@teoljungberg are we planning a subsequent PR to make `javascript: enabled: true` use the new worker as well?

teoljungberg:
@seanpdoyle Yes, the plan is to test out `hound-jshint` internally - and then we'll remove the existing `JavaScript` linter and have `javascript: enabled:  true` use `hound-jshint`.

teoljungberg:
Correction of my last comment, we'll have a public beta before replacing the existing `JavaScript` linter with our new service `hound-jshint`

spec/models/config/jshint_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "spec_helper"
+require "app/models/config/base"
+require "app/models/config/jshint"
+
+describe Config::Jshint do
+  it_behaves_like "a service based linter" do
+    let(:raw_config) do
+      <<-EOS.strip_heredoc
+        rules:
```
thorncp: Should this be JSON? It looks like YAML.

thorncp:
LGTM

spec/models/config/jshint_spec.rb
teoljungberg: It should be, I just did a big 'ol copy from ESLint.

Thanks

spec/models/style_checker_spec.rb
```diff
@@ -166,7 +166,31 @@
         end
       end
 
-      context "when eslint is disabled or unconfigured" do
+      context "when JSHint is enabled" do
+        it "does not immediately return violations" do
```
gylaz: This description doesn't match the verification phase. Perhaps it should state something like `finds violations`?

app/models/linter/jshint.rb
```diff
@@ -0,0 +1,5 @@
+module Linter
+  class Jshint < Base
+    FILE_REGEXP = /.+(?<!\.coffee)\.js\z/
```
gylaz: I think the `coffee` part can be removed. It was a legacy thing to ensure we don't lint `*.coffee.js` files, but that was a user-error on improperly naming files (should be `*.js.coffee`).

gylaz:
Super happy this is finally a reality!

app/models/linter/jshint.rb
teoljungberg: Interesting, I opened https://github.com/thoughtbot/hound/pull/989 to fix the rest of the linters to reflect that.

spec/models/style_checker_spec.rb
teoljungberg: Sounds good, thanks

teoljungberg: Or, `returns violations` to keep it simple?

spec/models/style_checker_spec.rb
```diff
@@ -166,7 +166,31 @@
         end
       end
 
-      context "when eslint is disabled or unconfigured" do
+      context "when JSHint is enabled" do
+        it "return violations" do
```
gylaz: Typo `returns`.

The thing is `StyleChecker#review_files` does not return violations, it creates records in the database. I feel this description doesn't quite fit.

teoljungberg: I agree, and `does not return violations immediately` doesn't either, since we're creating the linting job and the files are being linted by another service.

The most generic one I can image is along the lines of `it "returns the found violations"`.

gylaz: Still, the method described doesn't return anything. How about something along the lines of `it "creates violations"` or `it "finds violations"`?

teoljungberg: In that case, I would want to rename all of the test-descriptions in the file to follow that idea. Do you think it's worth it?

gylaz: I think it's worth doing that in a small follow-up PR.

teoljungberg: Sounds good.

I'm going to leave this as it's currently us. And then make a global search and replace
What do you think of that approach?

gylaz: Seems good.

PR #988, opened 2015-11-04
Changed EnforcedStyle for StringLiterals to single_quotes

deepakm-webonise:
Changed EnforcedStyle for StringLiterals to single_quotes
https://github.com/bbatsov/ruby-style-guide#consistent-string-literals  - **(Option A)** is preferred


gylaz:
We can't be making changes to such common rules -- since Hound will start commenting on all of the existing user's code that are used to the `double_quotes` rule.

We're in the process of getting closer to using the linter defaults, but we'll need to notify (or provide a migration strategy to) users that are currently using Hound as is.

deepakm-webonise:
Ok, Sounds legible :+1:

PR #993, opened 2015-11-16
Remove preview media type constant, freeze constant string

JuanitoFatas:
This Pull Request:

- Remove moondragon media constant (not exists anymore)
- :snowflake: strings

### DIFF
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -3,8 +3,7 @@ require "octokit"
 require "base64"
 
 class GithubApi
-  ORGANIZATION_TYPE = "Organization"
-  PREVIEW_MEDIA_TYPE = "application/vnd.github.moondragon+json"
+  ORGANIZATION_TYPE = "Organization".freeze
 
   attr_reader :file_cache, :token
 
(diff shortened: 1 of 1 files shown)


lib/github_api.rb
```diff
@@ -22,7 +21,7 @@ def scopes
   end
 
   def repos
-    all_repos = client.repos(nil, accept: PREVIEW_MEDIA_TYPE)
+    all_repos = client.repos
```
JuanitoFatas: Removes default `nil` since it is the default: https://github.com/octokit/octokit.rb/blob/38a4dff/lib/octokit/client/repositories.rb#L69

thorncp:
This looks good to me. I'm going to batch this up with some other changes and do a smoke test/runthrough on staging before merging though. Thanks @JuanitoFatas !

teoljungberg:
@thorncp can you test this out to staging today?

thorncp:
It worked great! Merged as 49f94059a07638084adcbfda5066864d1509865a. Thanks @JuanitoFatas

JuanitoFatas:
:+1:

PR #1000, opened 2015-11-30
Support disabling cops with inline comments

### LINKED ISSUE #995: rubocop:disable comments not supported
I've setup Rubocop locally with Guard for faster feedback, and it respects the comments-

 ```   
 if raise_if_error && result.size > 0
      # rubocop:disable Rails/Output
      puts "---- JSON SCHEMA -----"
      ap schema
      raise result.to_s
    end
```

... but Hound is still complaining about the issue on the PR.

These pragmas are super useful in documenting intent, and filter out noise locally. Would be really nice for Hound to respect them.

### PR DESCRIPTION
**Pull request by thorncp**
Rubocop supports [disabling cops with inline comments][inline-disable],
but Hound does not. This is so because Rubocop still reports the
violations internally, then chooses to ignore those that have been
disabled via comment. Hound is not using the parts of Rubocop that do
this, resulting in comments on the violations. Fortunately there is a
`Offense#disabled?` method which returns true for violations disabled
by comments.

Fixes #995

[inline-disable]: https://github.com/bbatsov/rubocop/tree/2195f366bb47d8d8794943c247c00ba55213900b#disabling-cops-within-source-code

### DIFF
--- a/app/models/linter/ruby.rb
+++ b/app/models/linter/ruby.rb
@@ -16,7 +16,7 @@ module Linter
 
     def perform_file_review(commit_file)
       FileReview.create!(filename: commit_file.filename) do |file_review|
-        team.inspect_file(parsed_source(commit_file)).each do |violation|
+        permitted_rubocop_offenses(commit_file).each do |violation|
           line = commit_file.line_at(violation.line)
           file_review.build_violation(line, violation.message)
         end
@@ -24,6 +24,10 @@ module Linter
         file_review.build = build
         file_review.complete
       end
+    end
+
+    def permitted_rubocop_offenses(commit_file)
+      team.inspect_file(parsed_source(commit_file)).reject(&:disabled?)
     end
 
     def team
--- a/spec/models/linter/ruby_spec.rb
+++ b/spec/models/linter/ruby_spec.rb
@@ -580,6 +580,26 @@ describe Linter::Ruby do
       end
     end
 
+    context "with inline configuration" do
+      context "disabling a cop" do
+        it "does not return a violation" do
+          config = stub_ruby_config(
+            "StringLiterals" => {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by gylaz** [comment]
Looks gewd!

File: app/models/linter/ruby.rb [hunk 1]
```diff
@@ -16,7 +16,7 @@ def file_included?(commit_file)
 
     def perform_file_review(commit_file)
       FileReview.create!(filename: commit_file.filename) do |file_review|
-        team.inspect_file(parsed_source(commit_file)).each do |violation|
+        enabled_rubocop_offenses(parsed_source(commit_file)).each do |violation|
```
teoljungberg: `enabled` sounds a bit to vague to me. Sadly, I don't have any other naming suggestions.

spec/models/linter/ruby_spec.rb
```diff
@@ -580,6 +580,27 @@ def foo
...
 
+    context "with inline configuration" do
+      context "disabling a cop" do
+        it "does not return a violation" do
+          config = stub_ruby_config(
+            "StringLiterals" => {
+              "EnforcedStyle" => "double_quotes",
+            },
+          )
+
```
teoljungberg: I would remove this newline to follow The Four Phase Test

app/models/linter/ruby.rb
thorncp: Yeah, it's the best I could come up with :/

benniemosher: What about `#permitted_rubocop_offenses`? To me that would say that these offenses are the only offenses that are permitted to be displayed.

teoljungberg: I like `permitted_*`. I'll update the code to reflect that.

PR #1001, opened 2015-12-01
Don't fail on violations

thorncp:
This was enabled on Hound itself in order to vet the feature before
releasing it to the world. The Hound team treats violations as
suggestions and sometimes decides to ignore them. In these cases,
failing the PR does not make sense. The feature has been working well
for quite awhile now, so it can be disabled here.

Am I correct in thinking no one else wants this enabled? @thoughtbot/hound

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,4 +1,2 @@
-fail_on_violations: true
-
 ruby:
   config_file: .ruby-style.yml


gylaz:
I have a somewhat biased opinion that we should keep this on because we should drink our own cool-aid. But I would like to be swayed otherwise.

thorncp:
I remember everyone being a little hesitant before adding the feature. If that's changed we can definitely leave it enabled.

teoljungberg:
I enjoy having it enabled. And haven't felt a pain in having it so.

thorncp:
Ok, let's keep it!

PR #1006, opened 2015-12-03
Send email to repo collaborators upon activation

sikachu:
This email notifies user that someone turned on Hound, so that they don't be surprised when Hound starting to comment on issues.

Note that since there is no reliable way of obtaining collaborator's email address using user's token, we currently going to try our best to figure out their email address.

* If the collaborator is on Hound (has authenticated with GitHub), we will use email that Hound stored in the database upon user signed up.
* if the collaborator is not on Hound:
  * If the collaborator has a public email address listed on GitHub profile, we will use that.
  * If there is no email on collaborator's profile, we just silently fails.

https://trello.com/c/o3pUWoNq


app/mailers/mailer.rb
```diff
@@ -0,0 +1,14 @@
...
+  def repo_activation_notification(repo, username, email)
+    @repo = repo
+    @subscriber = repo.subscriber
+    @username = username
+
+    mail(
+      from: "hound@thoughtbot.com",
+      to: email,
+      subject:
+        "[Hound] Hound is now enabled on #{repo.full_github_name} repository"
```
gylaz: Is it common to stick `[business name]` at the beginning of the subject. I know Gmail tags add that, maybe we should leave it off?

app/mailers/mailer.rb
```diff
@@ -0,0 +1,14 @@
+class Mailer < ActionMailer::Base
+  def repo_activation_notification(repo, username, email)
+    @repo = repo
+    @subscriber = repo.subscriber
+    @username = username
+
+    mail(
+      from: "hound@thoughtbot.com",
```
gylaz: What about using a `default from: ...` helper?

app/services/repo_activator.rb
```diff
@@ -10,9 +10,12 @@ def initialize(github_token:, repo:)
   def activate
     change_repository_state_quietly do
       if repo.private?
-        add_hound_to_repo && create_webhook && repo.activate
+        add_hound_to_repo &&
+          create_webhook &&
+          notify_collaborators &&
```
gylaz: I would make this be that last call. We don't want anything failing in `notify_collaborators` to keep the repo from being active.

app/views/mailer/repo_activation_notification.text.erb
```diff
@@ -0,0 +1,12 @@
+Hi <%= @username %>,
+
+Hound is now enabled on https://github.com/<%= @repo.full_github_name %>. When it finds Ruby style guide violations, it will comment inline on pull requests.
+
+You can configure which guidelines Hound should consider a violation. Find out more about configuring Hound at https://houndci.com/configure.
+<% if @subscriber %>
+
+@<%= @subscriber.github_username %> purchased the subscription for <%= @repo.full_github_name %> with a billing address of <%= @subscriber.email_address %>. The card they used will be charged <%= number_to_currency @repo.subscription_price, precision: 2 %> each month that Hound is enabled on the repo.
```
gylaz: Thoughts about leaving this out? Is there any value in including it? I might not be a bad idea to advertise to all the collaborators the amount that is being billed.

app/views/mailer/repo_activation_notification.text.erb
```diff
@@ -0,0 +1,12 @@
...
+<% end %>
+
+Thanks,
+Hound
```
gylaz: The Hound team?

spec/mailers/mailer_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "rails_helper"
+
+RSpec.describe Mailer do
+  describe "#repo_activation_notification" do
+    context "private repository" do
```
gylaz: We usually star off `context` body with "when", "with", or "given". Like `context "give a private repository" do`.

spec/models/repo_spec.rb
```diff
@@ -154,4 +154,14 @@
       end
     end
   end
+
+  describe "#subscriber" do
+    it "returns user on subscription" do
```
gylaz: How about `returns user associated with subscription`? `user on subscription` sounds odd.

spec/services/repo_activator_spec.rb
```diff
@@ -24,6 +24,17 @@
...
 
+        it "sends notifications to collaborators" do
+          repo = create(:repo, private: false, in_organization: true)
+          api = stub_github_api
+          activator = build_activator(repo: repo)
+
+          activator.activate
+
+          expect(api).
+            to have_received(:repo_collaborators).with(repo.full_github_name)
```
gylaz: This expectation doesn't really match the doc string. Should we mock `api. repo_collaborators` and have it return a few emails, then verify that `Mailer` receives `repo_activation_notification` with expected emails?

spec/services/repo_activator_spec.rb
```diff
@@ -56,6 +67,17 @@
           expect(repo.reload).to be_active
         end
 
+        it "sends notifications to collaborators" do
+          repo = create(:repo, private: false, in_organization: true)
```
gylaz: Is it supposed to be `in_organization: false`?

spec/services/repo_activator_spec.rb
```diff
@@ -167,6 +189,68 @@
...
+        let(:repo) { build(:repo) }
+        let(:api) { stub_github_api }
+        let(:activator) { build_activator(repo: repo) }
+
+        before do
+          allow(api).to receive(:repo_collaborators).
+            and_return([{ login: "salbertson" }])
+          allow(Mailer).to receive(:repo_activation_notification).
+            and_return(double(deliver_later: true))
+        end
```
gylaz: This is the first time `before` is used in a spec in this project. Can we continue not using `before`s?

app/models/repo.rb
```diff
@@ -60,6 +60,10 @@ def total_violations
       count
   end
 
+  def subscriber
+    subscription.try(:user)
```
teoljungberg: Thoughts on using a `NullSubscription`?

app/services/repo_activator.rb
```diff
@@ -57,6 +60,28 @@ def create_webhook
     end
   end
 
+  def notify_collaborators
+    github.repo_collaborators(repo.full_github_name).each do |collaborator|
```
teoljungberg: Could this be extracted to a service of it's own?

app/views/mailer/repo_activation_notification.text.erb
```diff
@@ -0,0 +1,12 @@
+Hi <%= @username %>,
```
teoljungberg: Thanks for adding a plain-text email

spec/features/repo_list_spec.rb
```diff
@@ -84,12 +84,19 @@
...
     repo = create(:repo, private: false)
     repo.users << user
     hook_url = "http://#{ENV["HOST"]}/builds"
     stub_repo_request(repo.full_github_name, token)
     stub_add_collaborator_request(username, repo.full_github_name, token)
     stub_hook_creation_request(repo.full_github_name, hook_url, token)
+    stub_repo_collaborators_request(
+      repo.full_github_name,
+      token,
+      collaborator_username
```
teoljungberg: I wonder why the trailing comma violation didn't trigger here...

spec/models/repo_spec.rb
teoljungberg: `returns the subscribed user` ?

spec/services/repo_activator_spec.rb
```diff
@@ -167,6 +189,71 @@
           )
         end
       end
+
+      context "activation notification" do
+        let(:repo) { build(:repo) }
```
teoljungberg: thoughts on inlining the `let`'s?

spec/services/repo_activator_spec.rb
```diff
@@ -167,6 +189,71 @@
...
+        let(:api) { stub_github_api }
+        let(:activator) { build_activator(repo: repo) }
+        let(:mail) { double(deliver_later: true) }
+
+        before do
+          allow(api).to receive(:repo_collaborators).
```
teoljungberg: And these stubs

app/mailers/mailer.rb
sikachu: It's pretty common. I think it helps user to filter out their email, but I don't mind leaving it out.

app/models/repo.rb
sikachu: YAGNI. It's just a delegation with a easy-to-understand name. If there's no subscription, there's no subscriber — that's it.

I also don't think we do some complex weird stuff that warrants that null object yet as well.

app/views/mailer/repo_activation_notification.text.erb
sikachu: I don't mind either way. I think it was in original suggestion. Any thoughts on this @croaky?

spec/features/repo_list_spec.rb
sikachu: Looks like I found a bug? :D

spec/mailers/mailer_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        mail = Mailer.repo_activation_notification(
+          repo,
+          "github_username",
+          "user@example.com"
+        )
+
+        expect(mail.from).to eq ["hound@thoughtbot.com"]
+        expect(mail.to).to eq ["user@example.com"]
+        expect(mail.subject).to eq(
+          "[Hound] Hound is now enabled on #{repo.full_github_name} repository"
```
sikachu: N/A. This is done because it's going over 80 columns. :/

sikachu: N/A.

teoljungberg: It is something consistent throughout the codebase, so I would prefer having it in.

app/models/repo.rb
teoljungberg: I'm OK with it

spec/mailers/mailer_spec.rb
sikachu: Sure. I'm not a fan of this rule though. Make it looks dumb when you already know that method arity is finite.

spec/services/repo_activator_spec.rb
sikachu: This one in this context is correct.

app/services/repo_activator.rb
sikachu: Yep, rewriting this.

spec/services/repo_activator_spec.rb
sikachu: Yep, that has been done.

app/views/mailer/repo_activation_notification.text.erb
croaky: I think I did that because it seems common in our own usage of various software that it can be hard to track down who purchased something, and therefore might have admin rights. People leave organizations. If the team can search their email for "Hound billing" and find this, I thought it might be helpful.

Don't feel strongly, though.

sikachu: I see. I think maybe leaving just the first sentence (`@<%= @subscriber.github_username %> purchased the subscription for <%= @repo.full_github_name %> with a billing address of <%= @subscriber.email_address %>. `) might be good enough in that case. That way, the contributor knows who activated it.

gylaz: Maybe it's worth to just take out the price and leave everything else -- the last four of card might be useful information too.

Also, `with a billing address of` to with a billing email address of`?

sikachu: >the last four of card might be useful information too.

We currently don't have the last four stored on our side. I think it is helpful, but I think it probably should be build on top of this PR given this PR is already big enough as it is ...

I'll make that wording change, thanks!

app/services/repo_activator.rb
```diff
@@ -57,6 +60,34 @@ def create_webhook
     end
   end
 
+  def notify_collaborators
```
sikachu: This has been fixed.

sikachu:
I've responded to all feedbacks. This is ready for another look.

app/services/collaborator_notifier.rb
```diff
@@ -0,0 +1,50 @@
+class CollaboratorNotifier
+  def initialize(repo:, github_token:)
```
teoljungberg: Services commonly respond to `.call` in Hound.

What do you think of adding that for a consistent behavior?

app/views/mailer/repo_activation_notification.text.erb
```diff
@@ -0,0 +1,12 @@
...
+
+Hound is now enabled on https://github.com/<%= @repo.full_github_name %>. When it finds Ruby style guide violations, it will comment inline on pull requests.
```
teoljungberg: Should we wrap the lines at 80 characters in the email?

spec/lib/github_api_spec.rb
```diff
@@ -308,6 +308,29 @@
...
+
+  describe "#user" do
+    it "returns user information" do
+      username = "houndci"
+      api = GithubApi.new(token)
+      request = stub_user_request(username, token)
+
+      api.user(username)
+
+      expect(request).to have_been_requested
```
teoljungberg: Does this assertion match the test description? We're never asserting on the return value from `#user`

spec/lib/github_api_spec.rb
```diff
@@ -308,6 +308,29 @@
...
   end
 
+  describe "#collaborators" do
+    it "returns a list of collaborators" do
+      api = GithubApi.new(token)
+      request = stub_repo_collaborators_request(full_repo_name, token)
+
+      api.repo_collaborators(full_repo_name)
+
+      expect(request).to have_been_requested
```
teoljungberg: Same question as below

app/views/mailer/repo_activation_notification.text.erb
sikachu: It doesn't make sense to do that with interpolations. The line would either look too short or way too long on the client.

spec/models/repo_spec.rb
```diff
@@ -154,4 +154,14 @@
       end
     end
   end
+
+  describe "#subscriber" do
+    it "return the subscribed user" do
```
teoljungberg: We never test the case when there is not subscription. Should we add a test case for that?

spec/services/collaborator_notifier_spec.rb
```diff
@@ -0,0 +1,67 @@
+require "rails_helper"
+
+describe CollaboratorNotifier do
+  describe "#notify" do
+    let(:repo) { build(:repo) }
```
teoljungberg: I'm not a fan of these `let`'s. To keep the code consistent, what do you think about inlining them?

app/services/collaborator_notifier.rb
sikachu:
```
$ ag call app/services | wc -l
       0
```

Are you sure that's the common interface?

teoljungberg: Wrong repository. Hound uses `.run`.

```
hound master % ag run app/services | wc -l
      12
```

spec/lib/github_api_spec.rb
sikachu: Yep, I just changed the test description wording to `makes a request to GitHub`.

spec/lib/github_api_spec.rb
sikachu: ditto

app/services/collaborator_notifier.rb
sikachu: Sounds good.

spec/models/repo_spec.rb
sikachu: test added

sikachu:
@teoljungberg I think I responded to all feedbacks. Ready for another look.

app/services/collaborator_notifier.rb
```diff
@@ -0,0 +1,47 @@
+class CollaboratorNotifier
+  pattr_initialize [:repo!, :github_token!]
+
+  def run(collaborator)
```
teoljungberg: Services commonly have a `.run` method to correspond to `#run`. So you can call `Collaborator.run(arguments)` instead of calling `.new(argumenst).run`

sikachu: But that means we'd have to create the Notifier for _every_ collaborator of the repo. With the fact that initializer and execution are separated, it means that we can reuse this notifier for multiple collaborators.

https://github.com/thoughtbot/hound/pull/1006/files#diff-aaa1fe70069044ed5ccfceeaf7a87c8cR64

Do you think that `.run` method convention outweighs the benefit of lower memory usage?

teoljungberg: Oh, you're right.

Let's leave this as is.

teoljungberg:
This looks great, atleast to me. Would love to hear what the rest of @thoughtbot/hound thinks.

gylaz:
While this code looks good (thanks @sikachu!), I'm weary of adding more code and logic to Hound for features that aren't priority. I'm not actually sure that we **need** this feature, and I'm not :100: on merging this.

gylaz:
Thanks @sikachu but we've decided to close this PR for now. We might come back to this idea in the future, but right now we don't want to introduce this much code.

PR #1009, opened 2015-12-05
Display message for repos that can't be activated

gylaz:
Why:

Users should be able to see the repos they have access to on GitHub,
but cannot activate due to not having enough access (they are not an
admin). We should communicate in the UI that the repo cannot be
activated.

How:

* Store a flag on membership whether a user is an admin of the repo
* Display all of user's repos on the repos list
* Implement `RepoSerializer#admin?` that uses `current_user`.
* Add a message to each repo that cannot be activated stating why.

https://trello.com/c/XVDAV6uf
screenshot 2015-12-04 17 59 18

### DIFF
--- a/app/controllers/repos_controller.rb
+++ b/app/controllers/repos_controller.rb
@@ -8,10 +8,7 @@ class ReposController < ApplicationController
           current_user.repos.clear
         end
 
-        repos = current_user.
-          repos.
-          order(active: :desc, full_github_name: :asc).
-          includes(:subscription)
+        repos = current_user.repos_by_activation_ability.includes(:subscription)
 
         render json: repos
       end
--- a/app/serializers/repo_serializer.rb
+++ b/app/serializers/repo_serializer.rb
@@ -1,5 +1,6 @@
 class RepoSerializer < ActiveModel::Serializer
   attributes(
+    :admin,
     :active,
     :full_github_name,
     :full_plan_name,
@@ -17,5 +18,9 @@ class RepoSerializer < ActiveModel::Serializer
 
   def full_plan_name
     "#{object.plan_type} repo".titleize
+  end
+
+  def admin
+    object.memberships.find_by(user_id: current_user.id).admin?
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -57,10 +57,11 @@ ActiveRecord::Schema.define(version: 20151216235118) do
   add_index "file_reviews", ["build_id"], name: "index_file_reviews_on_build_id", using: :btree
 
   create_table "memberships", force: :cascade do |t|
-    t.integer  "user_id",    null: false
-    t.integer  "repo_id",    null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.integer  "user_id",                    null: false
+    t.integer  "repo_id",                    null: false
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+    t.boolean  "admin",      default: false, null: false
   end
 
   add_index "memberships", ["repo_id"], name: "index_memberships_on_repo_id", using: :btree
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -22,8 +22,7 @@ class GithubApi
   end
 
   def repos
-    all_repos = client.repos(nil, accept: PREVIEW_MEDIA_TYPE)
-    authorized_repos(all_repos)
+    client.repos(accept: PREVIEW_MEDIA_TYPE)
   end
 
   def repo(repo_name)
--- a/spec/lib/github_api_spec.rb
+++ b/spec/lib/github_api_spec.rb
@@ -1,4 +1,5 @@
 require "spec_helper"
+require "config/initializers/constants"
 require "attr_extras"
 require "lib/github_api"
 require "json"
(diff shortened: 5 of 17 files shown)


app/controllers/repos_controller.rb
```diff
@@ -8,8 +8,8 @@ def index
           current_user.repos.clear
         end
 
-        repos = current_user.
-          repos.
+        repos = current_user.repos.
+          order("memberships.admin DESC").
```
teoljungberg: Could this be extracted into a scope to keep the controller free from DB knowledge

app/serializers/repo_serializer.rb
```diff
@@ -18,4 +19,8 @@ def price_in_cents
   def full_plan_name
     "#{object.plan_type} repo".titleize
   end
+
+  def admin
+    object.memberships.find_by(user_id: current_user.id).admin?
```
teoljungberg: Could the membership ever be nil?

db/schema.rb
```diff
@@ -56,10 +56,11 @@
...
   create_table "memberships", force: :cascade do |t|
-    t.integer  "user_id",    null: false
-    t.integer  "repo_id",    null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.integer  "user_id",                    null: false
+    t.integer  "repo_id",                    null: false
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+    t.boolean  "admin",      default: false
```
teoljungberg: Should we add "null: false"?

lib/github_api.rb
```diff
@@ -22,8 +22,7 @@ def scopes
   end
 
   def repos
-    all_repos = client.repos(nil, accept: PREVIEW_MEDIA_TYPE)
-    authorized_repos(all_repos)
+    client.repos(nil, accept: PREVIEW_MEDIA_TYPE)
```
teoljungberg: What is the meaning of "nil" here? Could it br extracted into a constant to clarify its meaning?

spec/lib/github_api_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "spec_helper"
+require "config/initializers/constants"
```
teoljungberg: This is dragged in by the spec_helper already, if memory serves me right

spec/services/repo_synchronization_spec.rb
```diff
@@ -30,7 +30,43 @@
 
       synchronization.start
 
-      expect(user.repos.first).to be_in_organization
+      expect(user.reload.repos.first).to be_in_organization
+    end
+
+    context "when the user is not a repo admin" do
```
teoljungberg: What do you think of "when the user is an admin of the repository"?

teoljungberg: Should we order the tests to have the happy path first?

app/serializers/repo_serializer.rb
gylaz: I don't think so. It will return an empty AR Relation when there are no memberships.

spec/lib/github_api_spec.rb
gylaz: It's not...

spec/services/repo_synchronization_spec.rb
gylaz: To the first comment: Did you mean to say "when the user is **not** an admin of the repository"? This tests the **not** case. I also don't quite think there is much difference between the word `repository` and `repo`. We use `repo` across the app.

teoljungberg: You are absolutely right. My bad.

---
Teo Ljungberg

spec/lib/github_api_spec.rb
teoljungberg: What's not working with this: https://github.com/thoughtbot/hound/blob/master/spec/support/constants.rb ?

---
Teo Ljungberg

gylaz: `Hound` const is not defined (https://github.com/thoughtbot/hound/blob/master/spec/support/constants.rb#L5), thus it's falling into the `rescue`.

teoljungberg: I see.

I'm fine with this as is for now. I would like to fix the helper in another PR, but if you decide to fix it in this PR. I'm definitely OK with that too

gylaz:
@thorncp @teoljungberg Anything else?

spec/serializers/repo_serializer_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    context "when current user is an admin of the repo" do
+      it "returns true" do
+        membership = create(:membership, admin: true)
+        serializer = RepoSerializer.new(
+          membership.repo,
+          scope: membership.user,
+          scope_name: :current_user,
+        )
+
+        expect(serializer.admin).to eq true
```
teoljungberg: since this returns `true/false`. Should the method be named `admin?` ?

teoljungberg:
One suggestion about renaming `Membership#admin`. Besides that, looks good to me.

spec/serializers/repo_serializer_spec.rb
gylaz: Since this is a serializer and it will represent the JSON sent to the front end, I prefer to name this the same as what it will become in JSON. `admin?` technically works, but gets turned into `admin` key in JSON.

PR #1011, opened 2015-12-09
Fix default Python config.

georgebrock:
The Python linter we're using (Flake8) expects a non-JSON configuration format, and passing `{}` to it causes an exception.

Since constants in Ruby aren't inherited in the same way as methods, we need to move the default to a method in order to override it.

### DIFF
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -14,8 +14,6 @@ module Config
 
     private
 
-    DEFAULT_CONTENT = "{}"
-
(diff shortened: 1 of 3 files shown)


thorncp:
LGTM! :package:

georgebrock:
Thanks!

gylaz:
It might be worth also making a similar change in https://github.com/thoughtbot/hound/blob/master/app/models/config/go.rb

PR #1014, opened 2015-12-17
Cache number of violations per build

gylaz:
We display the number of violations per repo on the account page (for
paid repos). That was causing timeouts, since repos can have many builds
and sql `IN` uses a loop for large arrays.
To speed things up, we can cache the number of violations (messages) per
build.

### DIFF
--- a/db/migrate/20151216235118_add_violations_count_to_builds.rb
+++ b/db/migrate/20151216235118_add_violations_count_to_builds.rb
@@ -1,0 +1,5 @@
+class AddViolationsCountToBuilds < ActiveRecord::Migration
+  def change
+    add_column :builds, :violations_count, :integer
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151019123908) do
+ActiveRecord::Schema.define(version: 20151216235118) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -27,6 +27,7 @@ ActiveRecord::Schema.define(version: 20151019123908) do
     t.string   "commit_sha",          limit: 255
     t.text     "payload"
     t.integer  "user_id"
+    t.integer  "violations_count"
   end
 
   add_index "builds", ["repo_id"], name: "index_builds_on_repo_id", using: :btree
--- a/spec/models/violation_spec.rb
+++ b/spec/models/violation_spec.rb
@@ -34,4 +34,15 @@ describe Violation do
       expect(violation.messages_count).to eq 2
     end
   end
+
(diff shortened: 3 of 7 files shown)


db/migrate/20151216235118_add_violations_count_to_builds.rb
```diff
@@ -0,0 +1,5 @@
+class AddViolationsCountToBuilds < ActiveRecord::Migration
+  def change
+    add_column :builds, :violations_count, :integer
```
teoljungberg: Should this default to `0`? Or does it to that by default?

spec/models/violation_spec.rb
```diff
@@ -34,4 +34,15 @@
       expect(violation.messages_count).to eq 2
     end
   end
+
+  describe "#after_create" do
```
teoljungberg: This isn't necessarily a method we're testing, but a flow through the lifecycle of the object.

Would `context "after save callbacks"` be a better fit?

db/schema.rb
```diff
@@ -56,10 +57,11 @@
...
   create_table "memberships", force: :cascade do |t|
-    t.integer  "user_id",    null: false
-    t.integer  "repo_id",    null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.integer  "user_id",                    null: false
+    t.integer  "repo_id",                    null: false
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+    t.boolean  "admin",      default: false
```
thorncp: What is this for? Seems unrelated to this PR.

db/migrate/20151216235118_add_violations_count_to_builds.rb
gylaz: Will adding a default here set all the existing columns to `0`? I wasn't sure and I wanted to avoid that, so that we know if it's `nil` it means it hasn't been calculated.

spec/models/violation_spec.rb
gylaz: I like that idea, but maybe `describe` is a better fit, since `context` is better suited for documenting the special situation/condition.

db/schema.rb
gylaz: Nice catch! This snuck in from a different branch/feature.

db/migrate/20151216235118_add_violations_count_to_builds.rb
teoljungberg: Fair reasoning, I can stand behind that

gylaz: When do you sleep!?

thorncp:
:+1:

db/migrate/20151216235118_add_violations_count_to_builds.rb
teoljungberg: Timezones are finicky.

PR #1016, opened 2015-12-18
View each failure in specs with multiple expects

gylaz:
When we write tests with multiple expects, only the first failure is
reported when the test fails. `aggregate_failures` allows us to see
any or all the failures that happened in a specific test.

https://www.relishapp.com/rspec/rspec-core/docs/expectation-framework-integration/aggregating-failures


teoljungberg:
Seems like a reasonable thing to try out

PR #1017, opened 2015-12-19
Allow users to opt into review based on linter

salbertson:
This change disabled all linters by default unless the linter is enabled in `.hound.yml` or configured using `config_file`.

This, along with a migration path for existing users, should make Hound configuration more explicit and avoid the case where team members are surprised by what Hound comments on after enabling the service.


app/models/hound_config.rb
```diff
@@ -1,18 +1,15 @@
...
-    eslint
-    jscs
-    jshint
-    mdast
-    python
-    swift
-  )
-  LANGUAGES = %w(
+  # Only used in Config::Ruby
+  LINTERS = %w(
```
salbertson: Do we need an array of supported languages anymore? It's only used in one remaining place, to check for "legacy" config files.

teoljungberg: I don't think we need it, so I agree.

teoljungberg:
This is pretty sweet, and I'm a fan of this.

What can we do to move this forward @salbertson ?

gylaz:
Sounds like we aren't making progress on this, and we don't really have a way to support our existing clients (since they expect linters to be on by default). I'm going to close this for now and we can re-visit when we are prepared.

PR #1020, opened 2015-12-30
Update Coffeelint gem

gylaz:
Why:

* We are a few versions behind.
* There is potential bug that was reported by our customer.


teoljungberg:
Is there any change that needs to be made to our default config files for this update?

gylaz:
No. The gem changes are mostly bug fixes.

teoljungberg:
LGTM in that case

PR #1023, opened 2015-12-31
Set admin flag on existing memberships to true

gylaz:
All the existing memberships should have admin status, since we only
added memberships if users were admins of the repos.
Thus, we should preserve that state, otherwise all of the current repos
would say that they require admin permissions, until users re-sync.


thorncp:
Your logic seems sound to me. Any concerns with the migration taking long enough to lock the table while someone is enabling/disabling a repo? `Membership.count # => 537029`.

LGTM

gylaz:
Good point, @thorncp. We'll probably want to put it in maintenance mode when running this migration.

PR #1025, opened 2016-01-07
Report invalid configuration files per linter

teoljungberg:
Since the services does not have any concept of a configuration files,
but only the contents of them. It makes more sense to report invalid
configuration files per linter basis instead of per file basis.

Instead of Hound reporting the invalid configuration file by saying:

    Error parsing config file: config/rubocop.yml...

Hound will instead report:

    Error parsing config file for: ruby...

Which means that the linters does not need to know which file they are
parsing.

https://trello.com/c/42U5K0Bs

### DIFF
--- a/app/models/config/parser_error.rb
+++ b/app/models/config/parser_error.rb
@@ -1,10 +1,10 @@
 module Config
   class ParserError < StandardError
-    attr_reader :filename
+    attr_reader :linter_name
 
-    def initialize(message, filename:)
+    def initialize(message, linter_name:)
       super(message)
-      @filename = filename
+      @linter_name = linter_name
     end
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -13,7 +13,7 @@ en:
     one: "1 violation found."
     other: "%{count} violations found."
   config_error_status:
-    "Error parsing config file: %{filename}. Click \"details\" for
+    "Error parsing config for: %{linter_name}. Click \"details\" for
     assistance."
   hound_error_status: "We've encountered an error while reviewing your code."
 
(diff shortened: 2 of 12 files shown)


config/locales/en.yml
```diff
@@ -13,7 +13,7 @@ en:
     one: "1 violation found."
     other: "%{count} violations found."
   config_error_status:
-    "Error parsing config file: %{filename}. Click \"details\" for
+    "Error parsing config file for: %{linter_name}. Click \"details\" for
```
geoffharcourt: :+1:

geoffharcourt:
@teoljungberg LGTM. This is a great enhancement.

config/locales/en.yml
gylaz: I know that the space provided in the GitHub status is limited. Will the `details` link still be visible? If not, we could remove the word `file` here.

app/models/config/parser_error.rb
```diff
@@ -1,10 +1,10 @@
 module Config
   class ParserError < StandardError
-    attr_reader :filename
+    attr_reader :name
 
-    def initialize(message, filename:)
+    def initialize(message, name:)
```
gylaz: Is it clear that `name` actually means linter name?

gylaz:
:+1:

app/models/config/parser_error.rb
teoljungberg: Miss from my side - fixed in https://github.com/bbatsov/rubocop/issues/2551

config/locales/en.yml
teoljungberg: I think we can skip file regardless. Thanks

teoljungberg:
Thanks for the review @gylaz and @geoffharcourt

PR #1028, opened 2016-01-09
Add endpoint fetch merged RuboCop config

thorncp:
Adds `/config/:owner/:repo/:linter` route for providing users with their
fully merged config file. Currently only supports RuboCop via the `ruby`
linter value, e.g. `/config/thoughtbot/hound/ruby.yml`.

https://trello.com/c/3ZMngMl6

### DIFF
--- a/app/controllers/linter_configs_controller.rb
+++ b/app/controllers/linter_configs_controller.rb
@@ -1,0 +1,39 @@
+class LinterConfigsController < ApplicationController
+  def show
+    case params[:linter]
+    when "ruby"
+      ruby_config = Config::Ruby.new(hound_config, "ruby")
+      config = RubyConfigBuilder.new(ruby_config.content, params[:owner]).config
+      render body: config.to_yaml, content_type: "text/yaml"
+    else
+      head 404
+    end
+  end
+
+  protected
+
+  def repo
+    Repo.find_by(full_github_name: full_github_name)
+  end
+
+  def full_github_name
+    "#{params[:owner]}/#{params[:repo]}"
+  end
+
+  def hound_config
+    HoundConfig.new(head_commit)
+  end
+
+  def head_commit
+    Commit.new(full_github_name, default_branch, github_api)
+  end
+
+  def default_branch
+    repo = github_api.repo(full_github_name)
+    repo["default_branch"]
+  end
+
+  def github_api
+    GithubApi.new(current_user.token)
+  end
+end
--- a/app/controllers/linter_configs_controller.rb
+++ b/app/controllers/linter_configs_controller.rb
@@ -1,39 +1,21 @@
 class LinterConfigsController < ApplicationController
   def show
-    case params[:linter]
-    when "ruby"
-      ruby_config = Config::Ruby.new(hound_config, "ruby")
-      config = RubyConfigBuilder.new(ruby_config.content, params[:owner]).config
-      render body: config.to_yaml, content_type: "text/yaml"
+    config_builder = RepoConfigBuilder.new(
+      repo: repo,
+      token: current_user.token,
+    )
+    config_file = config_builder.config_for(params[:linter])
+
+    if config_file.present?
+      render body: config_file.content, content_type: config_file.format
     else
       head 404
     end
   end
 
-  protected
+  private
 
   def repo
-    Repo.find_by(full_github_name: full_github_name)
-  end
-
-  def full_github_name
-    "#{params[:owner]}/#{params[:repo]}"
-  end
-
-  def hound_config
-    HoundConfig.new(head_commit)
-  end
-
-  def head_commit
-    Commit.new(full_github_name, default_branch, github_api)
-  end
-
-  def default_branch
-    repo = github_api.repo(full_github_name)
-    repo["default_branch"]
-  end
-
-  def github_api
-    GithubApi.new(current_user.token)
+    Repo.find_by(full_github_name: "#{params[:owner]}/#{params[:repo]}")
   end
 end
--- a/app/services/repo_config_builder.rb
+++ b/app/services/repo_config_builder.rb
@@ -1,0 +1,38 @@
+class RepoConfigBuilder
+  def initialize(repo:, token:)
+    @repo = repo
+    @token = token
+  end
+
+  def config_for(linter_name)
+    case linter_name
+    when "ruby"
+      build_ruby_config
+    end
+  end
+
+  private
+
+  def build_ruby_config
+    ruby_config = Config::Ruby.new(hound_config, "ruby")
+    config = RubyConfigBuilder.new(ruby_config.content, @repo.owner.name).config
+    ConfigFile.new(content: config.to_yaml, format: :yaml)
+  end
+
+  def hound_config
+    HoundConfig.new(head_commit)
+  end
+
+  def head_commit
+    Commit.new(@repo.full_github_name, default_branch, github_api)
+  end
+
+  def default_branch
+    repo = github_api.repo(@repo.full_github_name)
+    repo["default_branch"]
+  end
+
+  def github_api
+    @_github_api ||= GithubApi.new(@token)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -20,6 +20,8 @@ Houndapp::Application.routes.draw do
   get "/configuration", to: "pages#configuration"
   get "/faq", to: "pages#show", id: "faq"
 
+  get "/config/:owner/:repo/:linter", to: "linter_configs#show"
+
   resource :account, only: [:show, :update]
   resources :builds, only: [:create]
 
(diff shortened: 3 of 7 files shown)


spec/controllers/linter_configs_controller_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      config = YAML.load(response.body)
+      expect(config["Style/StringLiterals"]).to match(hash_including(
+        "EnforcedStyle" => "double_quotes",
+        "Enabled" => true,
+      ))
+      expect(config["Style/OptionHash"]).to match(hash_including(
+        "Enabled" => true,
+      ))
+      expect(config["Metrics/LineLength"]).to match(hash_including(
+        "Max" => 80,
```
thorncp: This seems wrong.

app/controllers/linter_configs_controller.rb
```diff
@@ -0,0 +1,39 @@
+class LinterConfigsController < ApplicationController
+  def show
+    case params[:linter]
+    when "ruby"
+      ruby_config = Config::Ruby.new(hound_config, "ruby")
+      config = RubyConfigBuilder.new(ruby_config.content, params[:owner]).config
+      render body: config.to_yaml, content_type: "text/yaml"
+    else
+      head 404
```
thorncp: Is there a better action to take here?

teoljungberg: We are merging more configurations than just `ruby` right? Should these also be handled here?

teoljungberg: Nevermind, I did not read the description of the PR.

404 seems reasonable

spec/controllers/linter_configs_controller_spec.rb
teoljungberg: This could probably be resolved by putting `hash_including` on its own line

app/controllers/linter_configs_controller.rb
gylaz: What if we extracted the case statement out of the controller? Maybe we could have a service class that takes in a linter name and returns the merged content. We could 404 if it returned `nil` otherwise we render, or something?

app/controllers/linter_configs_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def default_branch
+    repo = github_api.repo(full_github_name)
+    repo["default_branch"]
+  end
+
+  def github_api
+    GithubApi.new(current_user.token)
+  end
```
gylaz: A lot of this looks like candidate to be pulled out of the controller.

config/routes.rb
```diff
@@ -20,6 +20,8 @@
   get "/configuration", to: "pages#configuration"
   get "/faq", to: "pages#show", id: "faq"
 
+  get "/config/:owner/:repo/:linter", to: "linter_configs#show"
```
gylaz: What do you think about `/:owner/:repo/config/:linter` format? I imagine down the road we'll have similar urls for `/:owner/:repo/settings` and the sorts.

thorncp: That makes sense to me. Any concerns with `:owner` values conflicting with our own routes? E.g., https://github.com/queue.

gylaz: That is an excellent point. Can we use a safe prefix in that case like `/owner`? On the other hand, we could say that the paths are inverted to my proposal -- `/config/thoughtbot/hound`, `/settings/thoughtbot/hound`, `/billing/thoughtbot/hound`, etc.

thorncp: I like that better because it means I don't have to change this PR :dancer:

thorncp: Chatted in-person and we want to go the route of `/owners/:owner/:repo`.

app/controllers/linter_configs_controller.rb
thorncp: Did some refactoring in b0bbdf9b7349c512b71f5c85d05c798871012982. I'm not too keen on the names of all those things, though, and would welcome any suggestions.

config/routes.rb
thorncp: Changed in 5a01875f2e885a49db052942dcffa7986e53ac1c

app/services/repo_config_builder.rb
```diff
@@ -0,0 +1,38 @@
+class RepoConfigBuilder
```
gylaz: Is this really a service or a model?

app/controllers/linter_configs_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      render body: config_file.content, content_type: config_file.format
+    else
+      head 404
+    end
+  end
+
+  private
+
+  def repo
+    Repo.find_by(full_github_name: "#{params[:owner]}/#{params[:repo]}")
```
gylaz: `find_by!` to avoid `nil` problems?

app/services/repo_config_builder.rb
```diff
@@ -0,0 +1,38 @@
...
+  def initialize(repo:, token:)
+    @repo = repo
+    @token = token
+  end
+
+  def config_for(linter_name)
+    case linter_name
+    when "ruby"
```
gylaz: Is the idea to have this case here to lay the framework for additional linters?

app/services/repo_config_builder.rb
```diff
@@ -0,0 +1,38 @@
...
+      build_ruby_config
+    end
+  end
+
+  private
+
+  def build_ruby_config
+    ruby_config = Config::Ruby.new(hound_config, "ruby")
```
gylaz: Unrelated, but I found myself wondering why we pass`"ruby"` to a `Config::Ruby` class.

spec/controllers/linter_configs_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
+require "rails_helper"
+
+describe LinterConfigsController do
+  context "requesting ruby" do
+    it "returns the merged RuboCop config" do
+      user = create(:user)
+      owner = create(:owner, name: "thoughtbot")
+      repo = create(:repo, full_github_name: "thoughtbot/hound", owner: owner)
+      stub_sign_in(user)
+      our_config = {
```
gylaz: What is our? Hound or customer?

spec/controllers/linter_configs_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+        body: hound_yaml,
+        token: user.token,
+      )
+      stub_encoded_contents_request(
+        sha: "master",
+        repo_name: repo.name,
+        file: "ruby.yml",
+        body: YAML.dump(our_config),
+        token: user.token,
+      )
```
gylaz: That setup :eyes:

spec/controllers/linter_configs_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
+require "rails_helper"
+
+describe LinterConfigsController do
+  context "requesting ruby" do
```
gylaz: `when requesting ruby`? or maybe `for ruby`?

spec/controllers/linter_configs_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      )
+      expect(config["Metrics/LineLength"]).to match(
+        hash_including(
+          "Max" => 80,
+        ),
+      )
+    end
+  end
+
+  context "for an unsupported endpoint" do
```
gylaz: `... unsupported linter`?

spec/routing/routes_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+RSpec.describe "application routing" do
+  it "routes the linter configs controller" do
+    expect(get: "/config/foo/bar/baz").to route_to(
+      controller: "linter_configs",
+      action: "show",
+      owner: "foo",
+      repo: "bar",
+      linter: "baz",
+    )
```
gylaz: Noice!

spec/controllers/linter_configs_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      "https://api.github.com/repos/#{repo_name}/contents/#{file}?ref=#{sha}",
+    ).with(
+      headers: { "Authorization" => "token #{token}" },
+    ).to_return(
+      status: 200,
+      body: { content: content }.to_json,
+      headers: { "Content-Type" => "application/json; charset=utf-8" },
+    )
+  end
+end
```
gylaz: What do you think about this being a request spec?

thorncp:
Our current plan is to introduce organization level configuration, so users can specify default linter options for all repositories in their organization. When we introduce this, we plan to default everyone's config to the _current_ Hound config. This will allows us to remove Hound's opinions without disrupting existing user's experiences. If they want to deviate from those defaults, they can do so.

With this plan, the endpoint that provides a user with raw config content (this PR) is not a necessity. It could still provide some use, e.g. showing users what their current config is, but for now it feels like an unnecessary maintenance burden, so I'm going to close this. We can revisit this later if it makes sense.

PR #1029, opened 2016-01-11
Extract service from `CompletedFileReviewJob`

teoljungberg:
To follow the pattern of a job class having a corresponding service
object.

### DIFF
--- a/app/jobs/completed_file_review_job.rb
+++ b/app/jobs/completed_file_review_job.rb
@@ -1,43 +1,15 @@
 class CompletedFileReviewJob
   @queue = :high
 
+  # The following parameters are required for this job to run.
+  # filename
+  # commit_sha
+  # pull_request_number
+  # patch
+  # violations
+  #   [{ line: 123, message: "WAT" }]
   def self.perform(attributes)
-    # filename
-    # commit_sha
-    # pull_request_number
-    # patch
-    # violations
-    #   [{ line: 123, message: "WAT" }]
-
-    build = Build.find_by!(
-      pull_request_number: attributes.fetch("pull_request_number"),
-      commit_sha: attributes.fetch("commit_sha")
-    )
-    file_review = build.file_reviews.find_by!(
-      filename: attributes.fetch("filename")
-    )
-    commit_file = CommitFile.new(
-      patch: attributes.fetch("patch"),
-      filename: nil,
-      commit: nil,
-    )
-
-    attributes.fetch("violations").each do |violation|
-      line = commit_file.line_at(violation.fetch("line"))
-      file_review.build_violation(line, violation.fetch("message"))
-    end
-
-    file_review.complete
-    file_review.save!
-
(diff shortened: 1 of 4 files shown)


app/services/completed_file_review.rb
```diff
@@ -0,0 +1,61 @@
+class CompletedFileReview
```
gylaz: What do you think about using an active form for naming services (i.e. verbs instead of nouns)? Name the services after what it's doing `CompleteFileReview` rather than after what it is?

app/services/completed_file_review.rb
```diff
@@ -0,0 +1,61 @@
...
+      line = commit_file.line_at(violation.fetch("line"))
+      file_review.build_violation(line, violation.fetch("message"))
+    end
+
+    file_review.complete
+    file_review.save!
+  end
+
+  def pull_request
+    @pull_request ||= PullRequest.new(payload, Hound::GITHUB_TOKEN)
```
gylaz: Do we need this memoization?

app/services/completed_file_review.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def payload
+    @payload ||= Payload.new(build.payload)
```
gylaz: Or this memoization?

spec/services/completed_file_review_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        expect(BuildReport).to have_received(:run).with(
+          pull_request: pull_request,
+          build: correct_build,
+          token: Hound::GITHUB_TOKEN,
+        )
+      end
+    end
+  end
+
+  let(:attributes) do
```
gylaz: Generally, I've seen `let`s always be defined at the top. What were your thoughts on putting it here?

app/services/completed_file_review.rb
teoljungberg: I know we've had this discussion before, and I agree with you. But this extracts the logic from the job to a service class of its own. And I think the symetry between having a service object and a job class sharing a similar name helps with readability and reasoning.

I'm opting leaving this as is now, and doing a follow up PR renaming the jobs and service objects to follow your suggested naming convention to help with compatibility during the migration period.

What do you think of that plan?

---
Teo Ljungberg

gylaz: :+1:

spec/services/completed_file_review_spec.rb
teoljungberg: I commonly define `let`s at the bottom so you see the tests at the top, as you open the file.

app/services/completed_file_review.rb
teoljungberg: Thinking more about this - I'll go ahead and rename the service now and followup with renaming the job later.

teoljungberg:
Thanks for reviewing @gylaz

PR #1034, opened 2016-01-16
Upgrade RuboCop to 0.36.0

gylaz:
Fix tests and code for upgrade:
* Set the Ruby Parser Version to `2.3`
* Remove `Rails/DefaultScope` rule since it was removed from RuboCop
* Fix default config test for Ruby's trailing comma

  The default config enforced no comma at the end of the list. There was
  a bug in previous RuboCop version where it wasn't detected properly.
  This test was wrong, and RuboCop improperly did not catch the
  violation. In the new version it properly detects this violation.
* Split `Style/TrailingComma` into `Style/TrailingCommaInLiteral` and

 `Style/TrailingCommaInArguments` per new RuboCop rules.
  Disable them in Hound defaults, to preserve backwards compatibility
  since RuboCop enables them by default.
  Enable them explicitly for thoughbot, to align with the styleguide.
* Use new `Style/MultilineMethodCallIndentation` rule

  `Style/MultilineMethodCallIndentation` takes over checking alignment
  of methods instead of `Style/MultilineOperationIndentation`.
* Extra spaces around `=` result in two violations
* Fix spec that was passing wrong config

  RuboCop expects configuration as hash, but we were passing in a String
  and it was raising an error in the new version.


app/models/linter/ruby.rb
```diff
@@ -39,8 +40,7 @@ def team
     end
 
     def parsed_source(commit_file)
-      absolute_filepath = File.expand_path(commit_file.filename)
-      RuboCop::ProcessedSource.new(commit_file.content, absolute_filepath)
+      RuboCop::ProcessedSource.new(commit_file.content, RUBY_PARSER_VERSION)
```
teoljungberg: I was surprised that we could not pass in `RUBY_VERSION` to `ProcessedSource`. I made a PR allowing that format here https://github.com/bbatsov/rubocop/pull/2655.

All though - this is not a blocker for this PR.

teoljungberg:
This looks great, thanks for picking this up @gylaz

ericboehs:
I agree. Top notch.  Are there any blockers from merging this?

PR #1035, opened 2016-01-18
Allow a linter to respond to many names

teoljungberg:
This allows us to configure that a given linter should respond to
multiple names in the `.hound.yml` configuration file.

https://trello.com/c/oHuaCx6b

### DIFF
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -67,7 +67,11 @@ module Config
     end
 
     def linter_config
-      hound_config.content[linter_name]
+      hound_config.content.slice(*linter_names).values.first
+    end
+
+    def linter_names
+      [linter_name]
     end
 
     def commit
--- a/app/models/config/coffee_script.rb
+++ b/app/models/config/coffee_script.rb
@@ -8,12 +8,11 @@ module Config
       ensure_correct_type(result)
     end
 
-    def linter_config
-      super || hound_config.content[alternate_linter_name]
-    end
-
-    def alternate_linter_name
-      linter_name.sub("_", "")
+    def linter_names
+      [
+        linter_name,
+        linter_name.sub("_", ""),
+      ]
     end
   end
 end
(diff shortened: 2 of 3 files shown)


app/models/config/base.rb
```diff
@@ -67,7 +67,11 @@ def file_path
     end
 
     def linter_config
-      hound_config.content[linter_name]
+      hound_config.content.slice(*linter_names).values.first
```
mike-burns: I see that, elsewhere in the project, `#linter_config` is assumed not to produce `nil`.

Would it be useful to raise here if the config is `nil`, so that this assumption is guaranteed?

```ruby
def linter_config
  find_linter_config.tap do |config|
    if config.nil?
      raise "nil config"
    end
  end
end

def find_linter_config
  hound_config.content.slice(*linter_names).values.first
end
```

teoljungberg: I think it would, all though - we have been talking on how to gather data on such occurrences and haven't come to a conclusion yet.

I'll leave this as is, but update the card to discuss this further.

gylaz: I believe this private `linter_config` method is only used on above (on line 66) and it checks for `nil`.

mike-burns: It is also used on all subclasses of `Base`, e.g.
https://github.com/thoughtbot/hound/blob/9d8d31dfa670bbdc308a750e6810a5ec91d83fd4/app/models/linter/ruby.rb#L12
https://github.com/thoughtbot/hound/blob/fec130bfe8f411c22681cfe74f7cfd133c3b5ca4/app/models/linter/haml.rb#L28
etc.

teoljungberg:
Thanks for reviewing, @mike-burns - I made a card for your suggestion of tracking invalid configuration files(https://trello.com/c/iNhzwxDF)

app/models/config/base.rb
gylaz: @mike-burns That is a different method (with the same name). This is private method on the `Config` namespace, not available to `Linter` namespace classes.

PR #1039, opened 2016-01-21
Add absolute filename to get file exclusion working

teoljungberg:
Rubocop expects the absolute path to a filename for it to be excluded

### DIFF
--- a/app/models/linter/ruby.rb
+++ b/app/models/linter/ruby.rb
@@ -40,7 +40,11 @@ module Linter
     end
 
     def parsed_source(commit_file)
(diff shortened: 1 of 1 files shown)


calleerlandsson:
It seems like we're now passing an extra argument to `RuboCop::ProcessedSource.new`. I don't understand exactly what file is excluded or how. Would it be good to explain this in the commit message or is this obvious to someone with familiarity with the `RubyCop` API?

The code looks good to me!

teoljungberg:
@calleerlandsson Good idea. The third argument is the full path to the file we are processing, which is something we have not passed to Rubocop before.
Which might explain why we have seem comments like `Rename have_comment to have_comment?.` in feature specs. Even though `spec` is ignored by the configuration.

PR #1047, opened 2016-01-29
Update for the mdast -> remark rename

thorncp:
The hound service is being updated to use `remark`, which is `mdast`'s new name, in https://github.com/thoughtbot/hound-mdast/pull/7. This change brings _that_ name change into Hound so everything is consistent.

https://trello.com/c/vFLTg8IQ

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -136,7 +136,7 @@ being used and the default configs for each linter.
  * [hound-go](https://github.com/thoughtbot/hound-go)
 
 1. Markdown (beta)
- * [hound-mdast](https://github.com/thoughtbot/hound-mdast)
+ * [hound-remark](https://github.com/houndci/hound-remark)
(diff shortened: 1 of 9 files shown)


gylaz:
Kinda makes me miss when we supported one linter per language. That way code didn't need to change if we switched linters or it was rename. In reality, I think JS is the only language so far that has two different leading linters.

Looks good though.

teoljungberg:
I think @gylaz brings up a great point, for languages that are only backed by one linter - should we expose the linter or the language in the config level?

thorncp:
My vote would be no, because it would create more of a maintenance burden when that linter is no longer the only one or we decide to switch to another. If the key is the linter name, there's no migration that has to happen: the old stays around for a bit as legacy and we point people to the new one.

gylaz:
> If the key is the linter name, there's no migration that has to happen...

That's a good point as far as configuration is concerned.

gylaz:
:+1:

PR #1050, opened 2016-01-31
Merge RuboCop overrides first

### LINKED ISSUE #1049: Rubocop's `DisabledByDefault: true` doesn't seem to work
Hi! Given the hound config below, for some reason I'm still getting Hound comments in Ruby files, which I was assuming I should not. Am I missing something? 

thanks!

.hound.yml
```
fail_on_violations: false

ruby:
  config_file: .ruby-style.yml
scss:
  enabled: false
javascript:
  enabled: false
coffeescript:
  enabled: false
jscs:
  enabled: false
haml:
  enabled: false
```

.ruby-style.yml
```
AllCops:
  DisabledByDefault: true
```

gylaz:
There is some important code that is run in
`RuboCop::ConfigLoader.merge_with_defaults` -- it checks for presence of
`DisabledByDefault` and disables all of the rules but the overrides.

Because we were calling `RuboCop::ConfigLoader.configuration_from_file`
first (and it calls `merge_with_defaults` in turn) and then merging the
custom config, we were not preserving the `DisabledByDefault` logic.

Fixes #1049

### DIFF
--- a/spec/services/ruby_config_builder_spec.rb
+++ b/spec/services/ruby_config_builder_spec.rb
@@ -4,12 +4,24 @@ require "app/services/ruby_config_builder"
 
 RSpec.describe RubyConfigBuilder do
   context "when there is no custom configuration" do
-    subject(:builder) { described_class.new }
+    it "returns the Hound defaults" do
+      builder = RubyConfigBuilder.new
 
-    it "returns the Hound defaults", aggregate_failures: true do
       config = builder.config
 
+      expect(config["Rails/ActionFilter"]).to match enabled_rule
       expect(config["Style/StringLiterals"]).to match hound_override_rule
+      expect(config["Style/VariableName"]).to match rubocop_default_rule
+    end
+  end
+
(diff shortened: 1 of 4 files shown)


spec/services/ruby_config_builder_spec.rb
```diff
@@ -4,19 +4,34 @@
 
 RSpec.describe RubyConfigBuilder do
   context "when there is no custom configuration" do
-    subject(:builder) { described_class.new }
-
     it "returns the Hound defaults", aggregate_failures: true do
+      builder = described_class.new
```
teoljungberg: We are consistently not using `described_class` in the rest of the codebase, what do you think about adhearing to that in this test?

teoljungberg:
This looks great to me! I really like the cleanup you squeezed in too

gumatias:
Hi guys! just checking (no rush at all) Is there anything I can do to help get this PR merged? thanks!

PR #1052, opened 2016-02-01
Load `config/initializers/constants.rb` once

teoljungberg:
It was loaded multiple times for a test that booted up Rails:
- Once by `rspec` requiring `spec_helper` in `.rspec`.
- Once when Rails itself booted.


.rspec
```diff
@@ -1,2 +1 @@
 --color
---require spec_helper
```
gylaz: It was nice having this -- and is the thoughtbot way. Is there a particular reason this was causing an issue?

teoljungberg: Where is it the thoughtbot way? It's not in suspenders, and was added in https://github.com/thoughtbot/hound/commit/0a6a4bd72a22af9d8465c3e65bf055ca5e0c9448

teoljungberg: The issue occurs because `spec_helper` is loaded before Rails has been loaded for tests that load `rails_helper`. Which makes that the constants initializer was executed twice, once in the `spec_helper`, and once when Rails boots.

teoljungberg: I updated the approach in 2655cdc to be a bit more resilient, with less changes.

gylaz: When I generate a new project with Suspenders, the `.rspec` file contains:
```
--color
--require spec_helper
```

config/initializers/constants.rb
```diff
@@ -1,13 +1,17 @@
...
-  HOST ||= ENV.fetch("HOST")
-  GITHUB_TOKEN ||= ENV.fetch("HOUND_GITHUB_TOKEN")
-  GITHUB_USERNAME ||= ENV.fetch("HOUND_GITHUB_USERNAME")
-  INTERCOM_API_SECRET ||= ENV.fetch("INTERCOM_API_SECRET")
-  MAX_COMMENTS ||= ENV.fetch("MAX_COMMENTS").to_i
-  SEGMENT_KEY ||= ENV.fetch("SEGMENT_KEY")
-  STRIPE_PUBLISHABLE_KEY ||= ENV.fetch("STRIPE_PUBLISHABLE_KEY")
+Hound = Module.new unless defined?(Hound)
+
+if Hound.constants.none?
```
gylaz: This feels like over guarding. In theory this constant should be `Houndapp` because that is what we originally called the app, and that's the constant used throughout the code. (Though ideally, I would love to rename that to `Hound`).

teoljungberg: Do you have any other suggestions than guarding like this?

gylaz: What about not loading it in the `spec_helper`? It's not ideal that `spec_helper` loads `spec/support/**/*.rb` since a lot of those files are for Railsy tests. What if we move that to `rails_helper` and load constants in the tests that need it (and not in the `spec_helper`)?

teoljungberg: We want to load it in the spec_helper, but we don't want to load it in the `rails_helper`. Since rails already preloads it for us.

gylaz: We don't have to load it in the `spec_helper`, right? We can load it in those files that need the constants. Or would that be too cumbersome?

teoljungberg: Good call. As it turns out, there aren't many specs that require the constants to be loaded.

I'm changing this PR to load dotenv in both `spec_helper` and `rails_helper`. Which is much cleaner.

teoljungberg:
I didn't mean to merge that. Re-opening

PR #1054, opened 2016-02-03
Make the test-suite warning free

teoljungberg:
By stubbing `Hound::GITHUB_TOKEN` in `spec/lib/github_api_spec.rb`.

- Remove `spec/support/constants.rb` in favor of stubbing the constants
  in the test suite.


teoljungberg:
References #1052

spec/support/constants.rb
```diff
@@ -1,10 +1,10 @@
 require "dotenv"
 Dotenv.load
 
-begin
+module Hound; end
+
+def load_constants
```
teoljungberg: This was the cleanest solution I could come up with for this to work with both running the entire test suite and individual tests.

If each test that requires the constants to be set, requires the initializer - the whole test suite warns.

/cc @gylaz

spec/lib/github_api_spec.rb
```diff
@@ -1,5 +1,5 @@
 require "spec_helper"
-require "config/initializers/constants"
+load_constants
```
gylaz: Since this is the only test that loads the constants outside of Rails, what if we do:
```rb
unless defined?(Hound)
  require "config/initializers/constants"
end
```
right in this file. We'd need to add `Dotenv` too (maybe directly in `spec_helper`?).

gylaz:
It looks like the only reason we are loading constants in `spec/lib/github_api_spec.rb` is for `Hound::GITHUB_TOKEN`. Can we just fake/stub that token}

teoljungberg:
I feel like there is such a low maintenance cost of having a `load_constants` "macro", that it's worth having it.

teoljungberg:
But, maybe stubbing the constants in each test-case is better.

gylaz:
> I feel like there is such a low maintenance cost of having a load_constants "macro"

Having it is fine, but it's unusual to have a custom method call amongst `require`s.

PR #1055, opened 2016-02-04
Add `ConfigBuilder` to build configuration objects

teoljungberg:


app/models/config_builder.rb
```diff
@@ -0,0 +1,12 @@
+class ConfigBuilder
+  def self.for(hound_config, name)
+    config_class =
+      begin
+        "Config::#{name.classify}".constantize
+      rescue
+        Config::Unsupported
+      end
```
gylaz: How about extracting this to a method to avoid this weird indentation?

spec/models/linter/base_spec.rb
```diff
@@ -1,5 +1,4 @@
-require "spec_helper"
-require "app/models/linter/base"
+require "rails_helper"
```
gylaz: Why this change?

gylaz:
:+1:

spec/models/linter/base_spec.rb
teoljungberg: Since now we are requiring the ConfigBuilder, which in turn builds any of the `Config::*` objects. And I didn't want to require variations of that in this file.

Do you think we should?

teoljungberg: I changed this to require the needed files instead.

PR #1056, opened 2016-02-04
Extract `build_config` from `Config::Base` test

teoljungberg:


spec/models/config/base_spec.rb
```diff
@@ -131,9 +122,28 @@ def parse(file_content)
...
   describe "#linter_names" do
     it "returns a list of names the linter is accessible under" do
-      config = Config::Test.new(double, "test")
+      config = build_config(linter_name: "test")
 
       expect(config.linter_names).to eq ["test"]
     end
   end
+
+  def build_config(options = {})
```
gylaz: What do you think about using kwargs here since `hound_config` and `linter_name` are the only options?

gylaz:
:+1:

spec/models/config/base_spec.rb
teoljungberg: Good call!

PR #1057, opened 2016-02-04
Parse all Config's using their corresponding parser

teoljungberg:
- The Yaml parsers can return `nil`, so guard against such a case by
  returning an empty hash which will work cleanly further up.
- Replace `shared_example "a service based linter"`.


app/models/config/coffee_script.rb
```diff
@@ -10,7 +10,7 @@ def linter_names
     private
 
     def parse(file_content)
-      result = Parser.json(file_content)
+      result = Config::Parser.json(file_content)
```
gylaz: Why is it necessary to specify the namespace?

app/models/config/jscs.rb
```diff
@@ -1,9 +1,21 @@
 module Config
   class Jscs < Base
+    def content
+      @content ||= super || default_content
```
gylaz: Any reason to memoize here but on in `Eslint#content`?

app/models/config/python.rb
```diff
@@ -1,9 +1,17 @@
 module Config
   class Python < Base
+    def content
+      @content ||= super.presence || default_content
```
gylaz: Why use `.presence` here but in other `content` methods?

gylaz:
:+1:

app/models/config/python.rb
teoljungberg: Parsing an empty string using the INI parser returns an empty hash, which is why I added this.

But, re-thinking it. I'll move it to the parser.

app/models/config/jscs.rb
teoljungberg: Good call, I think all of them should be memoized.

app/models/config/coffee_script.rb
teoljungberg: I experienced intermittent failures without this.

teoljungberg:
I've tested this for

JSCS: https://github.com/teoljungberg/houndradish/pull/58
JSHint: https://github.com/teoljungberg/houndradish/pull/60
MDast: https://github.com/teoljungberg/houndradish/pull/61
Ruby: https://github.com/teoljungberg/houndradish/pull/62
Scss: https://github.com/teoljungberg/houndradish/pull/59
Python: https://github.com/teoljungberg/houndradish/pull/63

Which uses the different file formats that could have been effected. But none were.

app/models/config/coffee_script.rb
gylaz: Are you using Spring? I'd love to understand why we need to reference a namespace for code that is inside of that namespace. In theory, we shouldn't need to do that.

app/models/config/eslint.rb
```diff
@@ -1,9 +1,17 @@
...
   class Eslint < Base
+    def serialize(data = content)
+      Config::Serializer.json(data)
+    end
+
     private
 
     def parse(file_content)
-      Parser.raw(file_content)
+      Config::Parser.yaml(file_content)
```
thorncp: Should this be `json`? Or is this the Hound YAML config?

teoljungberg: This is the config file for ESLint, so an `.eslintrc`. Which actually isn't valid JSON, but it is valid YAML. And YAML is a superset of JSON, which I had to learn the hard way this week.

app/models/config/coffee_script.rb
teoljungberg: I'm not. I'll see if I can reach some clarity to why I got a constant missmatch. I'll look into it

teoljungberg: It seems to have been a constant-finding issue, requiring the parser class we want resolved it. 
Which now is fixed.

PR #1059, opened 2016-02-04
ignore NPM packages from JavaScript linting

afeld:
`node_modules/` is equivalent to a `vendor/` folder, so it should be ignored by default when running JSHint.

### DIFF
--- a/config/style_guides/.jshintignore
+++ b/config/style_guides/.jshintignore
@@ -1,1 +1,2 @@
+node_modules/*
 vendor/*


teoljungberg:
I haven't seen a project with `node_modules` checked in to `git`, since that is likely the only use case for ignoring `node_modules`.

Could you link me to such a project?

afeld:
Sorry for the slow response—some people check their `node_modules/` into version control, just like they would with Ruby gems or other dependencies.

* http://www.letscodejavascript.com/v3/blog/2014/03/the_npm_debacle
* http://www.redotheweb.com/2013/09/12/should-you-commit-dependencies.html

I actually can't even remember what my use case was...maybe a GitHub Pages site that used NPM to install dependencies? Had the check them in since Pages doesn't run `npm install`.

gylaz:
Thanks for the PR @afeld!

PR #1062, opened 2016-02-04
Add swift and python language icons for home page

smharley:
image

### DIFF
--- a/app/views/home/index.haml
+++ b/app/views/home/index.haml
@@ -44,6 +44,10 @@
             .language-logo
               = image_tag "languages/sass.svg"
             %h5 Scss
+          %li.language
+            .language-logo
+              = image_tag "languages/swift.svg"
+            %h5 Swift
 
 %section.importance
   .section-content
--- a/app/assets/images/languages/python.svg
+++ b/app/assets/images/languages/python.svg
@@ -1,7 +1,4 @@
 <svg width="100" height="100" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <title>
-        python
-    </title>
     <defs>
         <ellipse id="a" cx="50" cy="50" rx="50" ry="50"/>
         <linearGradient x1="12.959%" y1="11.7%" x2="79.639%" y2="78.453%" id="d">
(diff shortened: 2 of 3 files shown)


croaky:
Beautiful! Would it be a weird flow to make each logo clickable, and go to the documentation for that language?

smharley:
I think the only downside to that @croaky would be taking people to the documentation rather than putting the CTA to sign up/in right in front of them

teoljungberg:
Have we considered both Python And Swift to be stable?

gylaz:
Swift is stable and we're promoting it in #1065 

@smharley Let's not show the Python icon on the homepage just yet, since it's in beta.

croaky:
> Let's not show the Python icon on the homepage just yet, since it's in beta.

It's been in beta for 6 months. How about putting "Python (beta)" on the home page if there's a beta distinction we want to make? In order to get feedback on the beta, we need people to be aware that it exists.

croaky:
@gylaz thoughts on above?

gylaz:
I think having a beta overlay (or something) for the languages that are in beta would be great.

In regards to Python, it could be that it's been that long, however we haven't gotten much indication about how it's been working. Typically, we'll have a language be in "alpha" -- vet it internally to see how it's working. I think it's been enabled on a few Django projects, but the feedback hasn't been very strong. I will reach out to Python devs again.

croaky:
I don't think we have enough volume internally to depend on pinging people in Slack, Trello, etc. would be good to move each new linter faster to publicly-discoverable, marked as beta in every instance.

gylaz:
We can try that, though we don't have the bandwidth to handle much of the customer support in a timely fashion. Since it's beta, I suppose customers should expect that.

smharley:
I can work on a beta label, I'll update this branch/PR

croaky:
Let's merge this!

gylaz:
I think we should take out Python from this PR. Python has not been working (from the feedback) and the Hound team lacks expertise to debug this.

app/views/home/index.haml
```diff
@@ -38,12 +38,20 @@
             %h5 JavaScript
           %li.language
             .language-logo
+              = image_tag "languages/python.svg"
+            %h5 Python
+          %li.language
+            .language-logo
```
gylaz: @smharley Can we remove the Python icon from the landing page? (it's in beta and not working very well).

smharley:
Updated screenshot, without Python
image

smharley:
My plan was to leave the Python logo file in, so we can easily add it when the time comes

gylaz:
👍 Good to merge.

PR #1063, opened 2016-02-05
Extract `LinterEnabledChecker`

teoljungberg:
To see which linters the config has enabled.

Changes:
- Extract `build_linter`.

### DIFF
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -1,6 +1,6 @@
 module Config
   class Base
-    pattr_initialize :hound_config, :linter_name
+    attr_reader_initialize :hound_config, :linter_name
 
     def content
       @content ||= parse(load_content)
--- a/app/models/linter/base.rb
+++ b/app/models/linter/base.rb
@@ -23,9 +23,7 @@ module Linter
     end
 
     def enabled?
-      config.linter_names.any? do |linter_name|
-        hound_config.enabled_for?(linter_name)
-      end
+      CheckEnabledLinter.run(config)
     end
 
     def file_included?(*)
--- a/app/services/check_enabled_linter.rb
+++ b/app/services/check_enabled_linter.rb
@@ -1,0 +1,27 @@
+class CheckEnabledLinter
+  def self.run(*configs)
+    new(*configs).enabled?
+  end
+
+  def initialize(*configs)
+    @configs = configs
+  end
+
+  def enabled?
+    linter_names.any? do |linter_name|
+      hound_configs.any? do |hound_config|
+        hound_config.enabled_for?(linter_name)
+      end
+    end
+  end
+
+  private
+
+  def linter_names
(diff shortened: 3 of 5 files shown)


app/models/config/base.rb
```diff
@@ -1,6 +1,6 @@
 module Config
   class Base
-    pattr_initialize :hound_config, :linter_name
+    attr_reader_initialize :hound_config, :linter_name
```
gylaz: Nice! I like the expanded versions rather than `rattr_initialize`. Should we do this for all `attr_extras` macros going forth?

app/models/linter/base.rb
```diff
@@ -23,9 +23,7 @@ def file_review(commit_file)
     end
 
     def enabled?
-      config.linter_names.any? do |linter_name|
-        hound_config.enabled_for?(linter_name)
-      end
+      LinterEnabledChecker.for(config)
```
gylaz: What do you think about some ideas on naming this class something like `IsLinterEnabled`, or turning it into a service, like `CheckEnabledLinter.run(config)`?

app/models/linter_enabled_checker.rb
```diff
@@ -0,0 +1,27 @@
...
+  def initialize(*configs)
+    @configs = configs
+  end
+
+  def enabled?
+    linter_names.any? do |linter_name|
+      hound_configs.any? do |hound_config|
+        hound_config.enabled_for?(linter_name)
+      end
+    end
```
gylaz: I'm trying to grasp the behavior of this -- and the fact that it's checking multiple linters. Why multiple? What is the use case for that?

app/models/config/base.rb
teoljungberg: I think that we should

app/models/linter/base.rb
teoljungberg: I like the name `CheckEnabledLinter`

app/models/linter_enabled_checker.rb
teoljungberg: This was extracted from my branch of adding a master configuration for an owner, and then there are multiple hound configs in play. And I think it would scale nicely, and keep the diff as small as possible, to let the class receive an array of `HoundConfig`'s.

gylaz: Just so understand a potential scenario, could you describe why it's relevant to check if any of the linters are enabled? Is this a case of `javascript`, `java_script`, `jshint` situation?

teoljungberg: That's exactly right, it is a case of a linter having multiple names. Which also is the implementation we have right now, this only adds so we can check for multiple HoundConfig's

gylaz: Got it. Thanks.

PR #1064, opened 2016-02-05
Reduce spec dependencies for ConfigBuilder

gylaz:
This test/code doesn't need the whole Rails framework.
Shave off precious time on focused spec runs :)


thorncp:
:+1: :racehorse:

PR #1066, opened 2016-02-05
Report invalid HAML format as violation

gylaz:
HamlLint includes the line number in the exception when it encounters an
invalid Haml format. However, the line number is of the line before the
error, thus we need to add one to comment under the erroneous line in
the pull request.

### DIFF
--- a/app/models/linter/haml.rb
+++ b/app/models/linter/haml.rb
@@ -2,6 +2,7 @@ module Linter
   class Haml < Base
     DEFAULT_CONFIG_FILENAME = "haml.yml"
     FILE_REGEXP = /.+\.haml\z/
+    HamlViolation = Struct.new(:line, :message)
 
     def file_review(commit_file)
       @commit_file = commit_file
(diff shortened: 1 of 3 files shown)


app/models/linter/haml.rb
```diff
@@ -43,6 +44,8 @@ def run_linters
         linter.run(content)
         results + linter.lints
       end
+    rescue HamlLint::Exceptions::ParseError => haml_error
+      [HamlViolation.new(haml_error.line + 1, haml_error.message)]
```
thorncp: Do we know it will always be off by one?

thorncp:
LGTM :+1:

app/models/linter/haml.rb
gylaz: We did a bit of digging, this is what we found:

Normally, HamlLint's lines start at 1. However, this exception is raised
by the Haml gem when parsing the file, and HamlLint wraps it.
Haml reports the first line as 0 (zero based line numbering).
Thus, we need to add 1 to the line number when reporting this exception
as violation.

Updated the commit message with this.

teoljungberg: I think this line should be extracted into a private method, but we'll get to that another time.

PR #1067, opened 2016-02-05
Update HamlLint to fix a parse issue

gylaz:
HamlLint had a bug where it was erroring out on invalid attribute tags.

Example:
```haml
$div
  %p{ data-attribute: "foo" }
```

Since the above is not a valid symbol syntax, HamlLint was dying on it
with a message of:
```
undefined method `children' for false:FalseClass
```

The linter has fixed the issue in v0.16.0 version.


thorncp:
:+1:

teoljungberg:
Nicely found!

PR #1068, opened 2016-02-05
Don't pass filename to HamlLint

gylaz:
When we pass a filename to HamlLint and RuboCop linter is enabled then
HamlLint will try to create a Tempfile in the directory that it gets
from the file's path. It won't exist on Heroku, so it throws an
exception.

HamlLint still works without a filename, so let's not pass it in.

Fixes the following issue we've been seeing:
```
No such file or directory @ rb_sysopen - app/views/articles/_form.html.haml.haml_lint.tmp20160205-6932-alnw3e
```

### DIFF
--- a/spec/models/linter/haml_spec.rb
+++ b/spec/models/linter/haml_spec.rb
@@ -115,6 +115,26 @@ describe Linter::Haml do
         end
       end
     end
+
+    context "when RuboCop linter is enabled" do
+      it "does not raise missing dir for Tempfile error" do
+        content = <<-EOS.strip_heredoc
+          %div
+        EOS
+        commit_file = build_commit_file(
+          filename: "/does/not/exist/foo.haml",
+          content: content,
+        )
+        config = {
+          "linters" => {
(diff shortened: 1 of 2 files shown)


thorncp:
Weird. LGTM

spec/models/linter/haml_spec.rb
```diff
@@ -115,6 +115,26 @@
         end
       end
     end
+
+    context "when RuboCop linter is enabled" do
```
teoljungberg: I find it to be slightly confusing that RuboCop is messing with a HAML file. I wonder if we can clarify the context to why that is.

gylaz: I tried to describe it in the commit message -- HamlLint creates a Tempfile in the directory of the file (e.g. `app/views/users` for `app/views/users/show.html.haml`) when RuboCop is enabled. I imagine it's because HamlLint uses the cli version of RuboCop to lint the Ruby code.

PR #1069, opened 2016-02-06
Extract GitHub token finding into own class

gylaz:
* Get that concern out of BuildRunner.
* The plan is to use the new `GithubRepo` class in other places.
* We should never encounter `Octokit::Unauthorized` during builds,
because this is the error that is only thrown when someone tries to
authenticate.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -1,6 +1,4 @@
 class BuildRunner
-  ExpiredToken = Class.new(StandardError)
-
   pattr_initialize :payload
 
   def run
@@ -9,17 +7,8 @@ class BuildRunner
     end
   rescue Config::ParserError => exception
     report_config_file_as_invalid(exception)
-  rescue Octokit::Unauthorized
-    if users_with_token.any?
-      reset_token
-      raise ExpiredToken
-    else
-      raise
-    end
   rescue Octokit::NotFound
-    if token != Hound::GITHUB_TOKEN
-      remove_repo_from_user
-    end
+    remove_current_user_membership
     raise
   end
 
@@ -35,7 +24,7 @@ class BuildRunner
     upsert_owner
     build = create_build
     review_files(build)
-    BuildReport.run(pull_request: pull_request, build: build, token: token)
+    report_on_build(build)
   end
 
   def relevant_pull_request?
@@ -51,42 +40,32 @@ class BuildRunner
       pull_request_number: payload.pull_request_number,
       commit_sha: payload.head_sha,
       payload: payload.build_data.to_json,
-      user: current_user_with_token,
+      user: user_token.user,
     )
   end
 
+  def remove_current_user_membership
+    repo.remove_membership(user_token.user)
+  end
+
   def pull_request
-    @pull_request ||= PullRequest.new(payload, token)
+    @pull_request ||= PullRequest.new(payload, user_token.token)
   end
 
-  def token
-    @token ||= current_user_with_token.try(:token) || Hound::GITHUB_TOKEN
-  end
-
-  def current_user_with_token
-    @current_user_with_token ||= users_with_token.sample
-  end
-
-  def users_with_token
-    repo.users.where.not(token: nil)
-  end
-
-  def last_token_user
-    repo.users.detect { |user| user.token == token }
+  def user_token
+    @user_token ||= UserToken.new(repo)
   end
 
   def repo
     @repo ||= Repo.active.find_by(github_id: payload.github_repo_id)
   end
 
-  def reset_token
-    last_token_user.update_columns(token: nil)
-    @token = nil
-  end
-
-  def remove_repo_from_user
-    last_token_user.repos.destroy(repo)
-    @token = nil
+  def report_on_build(build)
+    BuildReport.run(
+      pull_request: pull_request,
+      build: build,
+      token: user_token.token,
+    )
   end
 
   def track_subscribed_build_started
@@ -110,7 +89,7 @@ class BuildRunner
     @commit_status ||= CommitStatus.new(
       repo_name: payload.full_repo_name,
       sha: payload.head_sha,
-      token: token,
+      token: user_token.token,
     )
   end
 
--- a/spec/models/user_token_spec.rb
+++ b/spec/models/user_token_spec.rb
@@ -1,0 +1,37 @@
+require "app/models/user_token"
+
+describe UserToken do
+  describe "#token" do
+    context "when user with a token is found" do
+      it "returns user's token" do
+        token = "foo_bar_token"
+        user = instance_double("User", token: token)
+        repo = instance_double("Repo", users_with_token: [user])
+        user_token = UserToken.new(repo)
+
+        expect(user_token.token).to eq token
+      end
+    end
+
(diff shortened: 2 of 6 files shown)


app/models/github_repo.rb
```diff
@@ -0,0 +1,27 @@
+class GithubRepo
+  USER_WITH_DEFAULT_TOKEN = User.new(token: Hound::GITHUB_TOKEN)
```
teoljungberg: This doesn't feel like a constant to me, but more like a private method. What do you think about moving it into one?

app/models/github_repo.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(repo)
```
teoljungberg: Should we use `attr_extras` here?

app/services/build_runner.rb
```diff
@@ -35,7 +24,7 @@ def review_pull_request
     upsert_owner
     build = create_build
     review_files(build)
-    BuildReport.run(pull_request: pull_request, build: build, token: token)
+    make_comments(build)
```
teoljungberg: `BuildReport` does more than just commenting on the violations. So what do you think about renaming this to `report_on_build` or `run_build_report`?

spec/models/github_repo_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "spec_helper"
+require "config/initializers/constants"
```
teoljungberg: This is going to cause warnings, as laid out in #1054. I would like to see that PR being merged before merging this one

spec/models/github_repo_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+require "app/models/github_repo"
+
+describe GithubRepo do
+  describe "#token" do
+    context "when no users with token are found" do
```
teoljungberg: Should this be `tokens`?

spec/models/github_repo_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      it "returns hound's token" do
```
teoljungberg: Should we capitalize Hound?

spec/models/github_repo_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        repo = instance_double("Repo", users_with_token: [])
```
teoljungberg: We seem to use more `double`'s than `instance_double`'s in the codebase, which is something I have been thinking about.

Should we use `instance_double`'s for models going forward?

app/models/github_repo.rb
gylaz: I'm not opposed to that idea, but I would love to get more of your thoughts on this. Ideally, I'd like to understand why one way might be better than the other.
The benefit that I see of using a constant, is that it will get evaluated once -- there is only ever one `USER_WITH_DEFAULT_TOKEN`.

app/services/build_runner.rb
gylaz: Sure, I like those names.

app/models/github_repo.rb
gylaz: :+1:

spec/models/github_repo_spec.rb
gylaz: We get more benefit from `instance_double`. There are a lot of `double`s in the codes base, because `instance_double` didn't exist for a long period of Hound's life.

app/models/github_repo.rb
teoljungberg: I don't have a clear idea to why it rubs me the wrong way, maybe it's because the constant is holding an instance of some class.

If the benefit we are looking for is that it only get's evaluated once, we could go with a memoized private method.

spec/models/github_repo_spec.rb
teoljungberg: Valid point. I agree.

app/models/github_repo.rb
gylaz: > If the benefit we are looking for is that it only get's evaluated once, we could go with a memoized private method.

That wouldn't provide the same benefit, unless it was a calls method.

My attempt here is get some concrete ideas of why we'd prefer one approach over the other. If there is some disadvantage to having a constant be a class instance, that would be a good reason not to do it.

thorncp: >  If there is some disadvantage to having a constant be a class instance, that would be a good reason not to do it.

The only disadvantage I can think of is if the object is being mutated. That doesn't seem to be the case here, and we could even `.freeze` it.

gylaz: I ended up moving it to an instance method, because stubbing the constants was more difficult in the test, since those constants have to exist before the file is required.

gylaz:
@teoljungberg @thorncp Ready for another look.

spec/models/user_token_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "app/models/user_token"
+
+describe UserToken do
+  describe "#token" do
+    context "when no users for repo have tokens" do
+      it "returns Hound's token" do
+        stub_const("User", OpenStruct)
```
teoljungberg: Thoughts on using a `double` instead of an `OpenStruct`?

spec/models/user_token_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "app/models/user_token"
+
+describe UserToken do
+  describe "#token" do
+    context "when no users for repo have tokens" do
```
teoljungberg: What do you think about swapping the two contexts here? I prefer having the happy path over the sad path.

spec/models/user_token_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        repo = instance_double("Repo", users_with_token: [user])
+        user_token = UserToken.new(repo)
+
+        expect(user_token.token).to eq token
+      end
+    end
+  end
+
+  describe "#user" do
+    it "returns the user that bears the current token" do
```
teoljungberg: I thoroughly enjoy the wording in this test description, `bears the current token` makes me think the user bears a sword or something similar.

spec/services/build_runner_spec.rb
```diff
@@ -251,12 +234,8 @@
         force_fail_build_creation
         allow(Resque).to receive(:enqueue)
 
-        begin
-          build_runner.run
-        rescue ActiveRecord::StatementInvalid
```
teoljungberg: Interesting to have this in a test.

teoljungberg:
This looks good to me!

spec/models/user_token_spec.rb
gylaz: I'm stubbing a class constant. I'm not sure how I can achieve this with `double`.

teoljungberg: The type of the stubbed class shouldn't matter.

gylaz: I still don't follow. Can you provide some code examples?
What I am trying to do is to not rely on `User` class being loaded -- for the purpose of my test, it can behave just like an `OpenStruct`. Otherwise, a "constant not found" error is raised.

thorncp:
LGTM :crown:

PR #1076, opened 2016-02-11
Update Rubocop to version 0.37.2

monfresh:
**Why**:
The current version used by Hound (0.37.0) contains a bug with the RedundantMerge cop that causes a false positive when calling `update` on ActiveRecord objects. See https://github.com/bbatsov/rubocop/issues/2781

**How**:
Update to 0.37.1 or higher, which fixes the issue.


dzader:
Would love to see this merged sooner than later!

gylaz:
Thanks @monfresh. Merged as efacf3957afd339f5ad2b8a191f0908a5511c0e4.

PR #1080, opened 2016-02-13
Allow users to cancel orphaned, private repos

thorncp:
If a user has activated a private repo then loses access to it on
GitHub, they cannot deactivate the repo in Hound and Hound continues to
charge their credit card. This change includes repos in this state in
the user's repo list, so they can deactivate it.

* Adds the `OrphanedAndMembershipReposQuery` query object, which
  returns repos that a user has a membership and/or subscription for.
* Updates the repo list to include repos that have a subscription but not
  a membership.
* Updates `RepoSubscriber#unsubscribe` to only remove the Hound user
  and webhook if the user has a membership for that repo (meaning they
  have access (probably)).

https://trello.com/c/cb3EUNVC

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -33,7 +33,8 @@ class SubscriptionsController < ApplicationController
   end
 
   def repo
-    @repo ||= current_user.repos.find(params.fetch(:repo_id))
+    @repo ||= current_user.repos.find_by(id: params.fetch(:repo_id)) ||
+      current_user.subscribed_repos.find(params.fetch(:repo_id))
   end
 
   def github_token
--- a/app/serializers/repo_serializer.rb
+++ b/app/serializers/repo_serializer.rb
@@ -21,6 +21,20 @@ class RepoSerializer < ActiveModel::Serializer
   end
 
   def admin
-    object.memberships.find_by(user_id: current_user.id).admin?
+    has_admin_membership? || has_subscription?
+  end
+
+  private
+
+  def membership
+    @membership ||= object.memberships.find_by(user_id: current_user.id)
+  end
+
+  def has_admin_membership?
+    membership.present? && membership.admin?
+  end
+
+  def has_subscription?
+    current_user.subscribed_repos.include?(object)
   end
 end
--- a/app/services/repo_activator.rb
+++ b/app/services/repo_activator.rb
@@ -18,12 +18,10 @@ class RepoActivator
   end
 
   def deactivate
-    change_repository_state_quietly do
-      if repo.private?
-        remove_hound_from_repo
-      end
-
-      delete_webhook && repo.deactivate
+    if skip_github?
+      repo.deactivate
+    else
+      deactivate_with_github
     end
   end
 
@@ -37,6 +35,24 @@ class RepoActivator
     add_error(error)
     Raven.capture_exception(error)
     false
+  end
+
+  def skip_github?
+    repo.subscription.present? && missing_membership?
+  end
+
+  def missing_membership?
+    Membership.where(repo: repo, user: repo.subscription.user).empty?
+  end
+
+  def deactivate_with_github
+    change_repository_state_quietly do
+      if repo.private?
+        remove_hound_from_repo
+      end
+
+      delete_webhook && repo.deactivate
+    end
   end
 
   def remove_hound_from_repo
--- a/spec/features/user_deactivates_a_repo_spec.rb
+++ b/spec/features/user_deactivates_a_repo_spec.rb
@@ -1,0 +1,34 @@
+require "rails_helper"
+
+feature "user deactivates a repo", js: true do
+  context "when the user has a subscription" do
+    context "when the user does not have a membership" do
+      scenario "successfully deactiavtes the repo" do
+        token = "letmein"
+        user = create(:user, token_scopes: "public_repo,user:email")
+        repo = create(:repo, :active, private: true)
+        create(:subscription, user: user, repo: repo)
+        gateway_subscription = instance_double(
+          PaymentGatewaySubscription,
+          unsubscribe: true,
+        )
+        payment_gateway_customer = instance_double(
+          PaymentGatewayCustomer,
+          retrieve_subscription: gateway_subscription,
+        )
+        allow(PaymentGatewayCustomer).to receive(:new).and_return(
+          payment_gateway_customer
+        )
+
+        sign_in_as(user, token)
+        find(".repos .toggle").click
+
+        expect(page).not_to have_css(".active")
+
+        visit current_path
+
+        expect(page).not_to have_selector(".repo")
+      end
+    end
+  end
+end
--- a/spec/serializers/repo_serializer_spec.rb
+++ b/spec/serializers/repo_serializer_spec.rb
@@ -27,5 +27,36 @@ describe RepoSerializer do
         expect(serializer.admin).to eq false
       end
     end
+
+    context "when the current user is not a member of the repo" do
+      context "and does not have a subscription for the repo" do
+        it "returns false" do
+          user = create(:user)
+          repo = create(:repo)
+          serializer = RepoSerializer.new(
+            repo,
+            scope: user,
+            scope_name: :current_user,
+          )
+
+          expect(serializer.admin).to eq false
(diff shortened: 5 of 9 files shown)


app/queries/orphaned_and_membership_repos_query.rb
```diff
@@ -0,0 +1,10 @@
+class OrphanedAndMembershipReposQuery
```
gylaz: `ReposWithMembershipOrSubscriptionQuery`?

app/serializers/repo_serializer.rb
```diff
@@ -21,6 +21,13 @@ def full_plan_name
   end
 
   def admin
-    object.memberships.find_by(user_id: current_user.id).admin?
+    (membership.present? && membership.admin?) ||
```
gylaz: Worth extracting to `has_admin_membership?`

app/services/repo_activator.rb
```diff
@@ -39,6 +37,22 @@ def change_repository_state_quietly
     false
   end
 
+  def skip_github?
+    repo.private? &&
```
gylaz: Do we need this check if we are checking for subscription?

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+feature "user deactivates a repo", js: true do
+  context "when the user has a subscription" do
+    context "when the user does not have a membership" do
+      scenario "successfully deactiavtes the repo" do
+        token = "letmein"
+        user = create(:user, token_scopes: "public_repo,user:email")
+        repo = create(:repo, :active, private: true)
+        create(:subscription, user: user, repo: repo)
+        gateway_subscription = double(unsubscribe: true)
+        payment_gateway_customer = double(
```
gylaz: Worth making this an `instance_double`?

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        )
+        allow(PaymentGatewayCustomer).to receive(:new).and_return(
+          payment_gateway_customer
+        )
+
+        sign_in_as(user, token)
+        find(".repos .toggle").click
+
+        expect(page).not_to have_css(".active")
+        expect(user.repos.active.count).to eq(0)
```
gylaz: Does line 22 already assure there isn't an active repo?

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        expect(gateway_subscription).to have_received(:unsubscribe)
```
gylaz: Worth removing this since it's not a UI test -- a user has no way of verifying this?

app/controllers/subscriptions_controller.rb
```diff
@@ -33,7 +33,8 @@ def activator
   end
 
   def repo
-    @repo ||= current_user.repos.find(params.fetch(:repo_id))
+    @repo ||= current_user.repos.find_by(id: params.fetch(:repo_id)) ||
+      current_user.subscribed_repos.find(params.fetch(:repo_id))
```
gylaz: I suppose this could be optimized with raw SQL -- but that seems fairly complex. So, :+1:

gylaz:
I like it!

app/queries/orphaned_and_membership_repos_query.rb
teoljungberg: Could this be a database view instead, backed by [scenic](/thoughtbot/scenic)?

thorncp: Yes, it could be. @hktouw and I went back and forth about this, but in the end felt like it was a lot of overhead for the benefits it added. Now, I'm having that debate with myself...

I think the biggest hurdle is how records are ordered. Currently, repos where the user is an admin are bubbled to the top. The admin flag is stored in `memberships`, but the new repos in the set don't have a membership record. So, we'd have to normalize the data to determine the admin-ness in the new sql view. With the new "column," we're no longer dealing with `Repo` objects, so we'd have to touch a few more places to account for that (primarily the json serializer). We'd also have to add a few more associations, e.g. `current_user.repos_with_membership_or_subscription`.

Benefits would be:

* Having a single query to get the data set
* Having an `ActiveRecord::Relation`, which could be `order`ed elsewhere (instead of being in the query)

Downsides would be:

* Normalizing the data will touch many files and (probably) make for a big diff
* Adding AR associations for a hyper specific problem

None of this is hard, but it seems like a lot of work to account for a problem that very few people have.

Thoughts?

app/queries/orphaned_and_membership_repos_query.rb
```diff
@@ -0,0 +1,10 @@
...
+  def initialize(user)
+    @user = user
+  end
+
+  def run
+    @user.subscribed_repos.includes(:subscription) |
```
thorncp: Is this union surprising?

Alternatively, this could be a single query, as seen in the [first commit](https://github.com/houndci/hound/commit/0f6d0a6ed6bfb8a0664eae7ae2eb817030aa4049#diff-4676c008b11a5480d73d4a6de01e45b9)

app/serializers/repo_serializer.rb
thorncp: Good idea. I also realized this didn't have a test. Added in 59cb5aff1a80a689dee1fb9b59613723cc8b2338

app/services/repo_activator.rb
thorncp: We don't! Also refactored a bit and added some specs: 319670f7bc9887c8a41406df4f870debc3853190

spec/features/user_deactivates_a_repo_spec.rb
thorncp: Good idea! a1a6e0e61e0db08ee25805f5c33dffe1ebeb7cb4

spec/serializers/repo_serializer_spec.rb
```diff
@@ -27,5 +27,36 @@
         expect(serializer.admin).to eq false
       end
     end
+
+    context "when the current user is not a member of the repo" do
```
hktouw: I don't know if this is possible, but can a user have a subscription and a non-admin membership? I think the code would handle that just fine, but I don't think there is a test for that flow since the above context just checks the case where a user doesn't have a membership as opposed to the case where the user does not have an admin membership (which is what the code is doing). Probably fine though because this context is the actual use case we're trying to solve for? And its probably just mincing details at this point :)

app/queries/orphaned_and_membership_repos_query.rb
gylaz: Not surprising, but a query would be more optimal.

app/services/repo_activator.rb
```diff
@@ -39,6 +37,24 @@ def change_repository_state_quietly
     false
   end
 
+  def skip_github?
+    repo.subscription.present? && missing_membership?
+  end
+
+  def missing_membership?
+    !Membership.where(repo: repo, user: repo.subscription.user).any?
```
gylaz: `!...any?` -> `.empty?`

spec/serializers/repo_serializer_spec.rb
thorncp: I think that is a possible situation to get in, but it does seem unlikely. I imagine the scenario would be:

1. User is a repo admin and enables the repo on houndci.com.
2. User's admin rights are removed from that repo.
3. User tries to disable the repo on houndci.com and fails (they see an error message about not being an admin).

At this point, they can reach out to an admin to disable it, or contact support. I think I'm ok with that. Anyone else feel differently?

app/services/repo_activator.rb
thorncp: Yes. 19860989030eb6db989587a571fefff1152fc02e

app/queries/orphaned_and_membership_repos_query.rb
thorncp: Renamed to `ReposWithMembershipOrSubscriptionQuery ` in 652dbab9b06d19ededec4ee5937c7d4809f36d82

gylaz:
:shipit:

PR #1084, opened 2016-02-19
Update configuration.haml for #scss exclude example

beaktor:
I tried the existing example for #scss exclusion, and it doesn't seem to work. If I remove the `scss:` line in the `.scss-style.yml` file, it does work. Cross-confirmed with an example in the linter documentation:

https://github.com/brigade/scss-lint#configuration

Discussion and :+1: in this thread: https://github.com/houndci/hound/issues/1018#issuecomment-186394304

Please merge accordingly if things look good. Cheers!

### DIFF
--- a/app/views/pages/configuration.haml
+++ b/app/views/pages/configuration.haml
@@ -302,9 +302,8 @@
 
         %code.code-block
           :preserve
-            scss:
-              exclude:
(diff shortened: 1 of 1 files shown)


gylaz:
Thank you @beaktor!

PR #1085, opened 2016-02-26
Rescue parse error of inherited RuboCop configs

gylaz:
We were getting errors because we weren't handling potential errors from
parsing the `inherit_from` files.

Restructure code to make it harder to make this mistake in the future.

### DIFF
--- a/app/models/config/coffee_script.rb
+++ b/app/models/config/coffee_script.rb
@@ -10,9 +10,7 @@ module Config
     private
 
     def parse(file_content)
-      result = Parser.json(file_content)
-
-      ensure_correct_type(result)
+      Parser.json(file_content)
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/models/config/coffee_script.rb
```diff
@@ -10,9 +10,7 @@ def linter_names
     private
 
     def parse(file_content)
-      result = Parser.json(file_content)
-
-      ensure_correct_type(result)
+      Parser.json(file_content)
```
thorncp: I like these got cleaned up!

spec/models/config/ruby_spec.rb
```diff
@@ -103,21 +101,31 @@
...
-            !
-          EOS
-        )
+        commit = stubbed_commit("config/rubocop.yml" => "[]")
         config = build_config(commit)
 
         expect { config.content }.to raise_error(
           Config::ParserError,
-          %r(`config/rubocop.yml` must be a Hash),
+          %r("config/rubocop\.yml" must be a Hash),
```
thorncp: Good :eyes: on that dot

thorncp:
LGTM :gem:

PR #1091, opened 2016-03-04
Add service to run a rebuild

thorncp:
* Add `Rebuilder` to trigger a new build from an existing build for a
  repo and pull request number.

https://trello.com/c/hVwHmtpP

### DIFF
--- a/app/services/rebuild_pull_request.rb
+++ b/app/services/rebuild_pull_request.rb
@@ -1,0 +1,21 @@
+class RebuildPullRequest
+  def initialize(repo:, pull_request_number:)
+    @repo = repo
+    @pull_request_number = pull_request_number
+  end
+
+  def call
+    if latest_build.present?
+      SmallBuildJob.perform_later(latest_build.payload)
+    end
+  end
+
+  private
+
+  def latest_build
+    Build.
+      where(repo: @repo, pull_request_number: @pull_request_number).
+      order(created_at: :desc).
+      first
+  end
+end
--- a/spec/services/rebuild_pull_request_spec.rb
+++ b/spec/services/rebuild_pull_request_spec.rb
@@ -1,0 +1,43 @@
+require "rails_helper"
+
+describe RebuildPullRequest do
+  describe "#call" do
+    context "when a latest build exists" do
+      it "schedules a small build job" do
+        repo = create(:repo)
+        payload = "{}"
+        create(
+          :build,
+          repo: repo,
+          payload: payload,
+          pull_request_number: 42,
+        )
+        rebuilder = RebuildPullRequest.new(repo: repo, pull_request_number: 42)
+        allow(SmallBuildJob).to receive(:perform_later)
+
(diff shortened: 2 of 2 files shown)


spec/services/rebuilder_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+describe Rebuilder do
+  describe "#rebuild" do
+    context "when a latest build exists" do
+      it "schedules a small build job" do
+        repo = create(:repo)
+        payload = stubbed_payload
+        create(
+          :build,
+          repo: repo,
+          payload: payload,
```
gylaz: Could this just be `{}` or `"foo"` for the purpose of this test?

thorncp: Yep! d9dacb8a2ce970df9309eaee4869487295ae50d5

app/services/rebuilder.rb
```diff
@@ -0,0 +1,21 @@
+class Rebuilder
```
gylaz: Thoughts on starting to move to action based named for services. For example, `RerunBuild` or `RebuildPullRequest`? This way we can call the only public method `call` or `perform` and avoid the awkward redundancy (e.g `Rebuilder.rebuild`). Related discussion on our Guides repo: https://github.com/thoughtbot/guides/pull/402

gylaz:
:+1:

app/services/rebuilder.rb
thorncp: Good idea. Renamed to `RebuildPullRequest` in the squash.

app/services/rebuild_pull_request.rb
```diff
@@ -0,0 +1,21 @@
+class RebuildPullRequest
+  def initialize(repo:, pull_request_number:)
+    @repo = repo
+    @pull_request_number = pull_request_number
+  end
+
+  def call
```
teoljungberg: Normally, our services expose class level `.run` method.
Should we do so here, and - should it be named `run` for consistency with the other service objects?

I personally prefer `.call`, but I find consistency to be the most important.

app/services/rebuild_pull_request.rb
```diff
@@ -0,0 +1,21 @@
...
+    if latest_build.present?
+      SmallBuildJob.perform_later(latest_build.payload)
+    end
+  end
+
+  private
+
+  def latest_build
+    Build.
+      where(repo: @repo, pull_request_number: @pull_request_number).
```
teoljungberg: I prefer keeping everything related to sql in the model, should we make a class level method for this?

app/services/rebuild_pull_request.rb
gylaz: I also am leaning towards `call` (because of parity with lambas). We had a card to rename our services to be action/verb based, maybe we should also rename the main method to `call`?

teoljungberg: Sounds good

PR #1092, opened 2016-03-04
Display failed Resque jobs in the admin section

gylaz:
There is a need to view similar Resque failures,
and the Resque backend UI isn't ideal for this.

Group similar errors by error message and show their ids for ease
of resolving them.

* Integrate with Administrate for admin-only acess.
* Add `JobFailure` to wrap around Resque specific API.
* Point Redis to a non-default database in test environment,
  so it doesn't interfere with development data.
* Wrap test runs that touch Redis into a block that cleans up after
itself to avoid leaving around data from previous tests.

https://trello.com/c/SUydnPxN

screen shot 2016-03-04 at 12 38 46 pm


app/models/job_failure.rb
```diff
@@ -0,0 +1,10 @@
+class JobFailure < OpenStruct
+  extend ActiveModel::Naming
+
+  def self.all
+    failures = Resque::Failure.all(0, Resque::Failure.count)
```
thorncp: What does `0` represent? Worth extracting it to a local variable with a descriptive name?

app/models/job_failure.rb
```diff
@@ -0,0 +1,10 @@
...
+    [failures].flatten.map.with_index do |failure, index|
```
thorncp: What do you think about `failures.map...`?

app/models/job_failure.rb
thorncp: Any memory concerns around loading the whole queue into memory?

gylaz: Could be, but probably not in our given situation. Even if there's 10,000 of them -- they are just hashes and it should be that much in the grand scheme of things. However, I could be wrong and we should watch this.

app/models/job_failure.rb
gylaz: I'm not sure I follow you suggestion here. `Resque::Failure.all` either returns an array of hashes or one hash (if there's just one record).

thorncp:
:+1:

PR #1094, opened 2016-03-04
Add script to show what isn't on production

thorncp:
The goal is to get a bird's eye view of what needs to be deployed to
production.

The script assumes all of the hound related repositories are in the same
directory as the main one (this one). It loops over them and attempts to
compare production/master to origin/master, showing the results.

production-status

### DIFF
--- a/bin/production-status
+++ b/bin/production-status
@@ -1,0 +1,34 @@
+#!/usr/bin/env ruby --disable-gems
+
+require "pathname"
+
+parent_directory = File.expand_path("../../..", __FILE__)
+Dir.chdir(parent_directory)
+child_directories = `ls -d */`.split
+git_directories = child_directories.select do |child_directory|
+  Dir.exists?(File.join(child_directory, ".git"))
+end
+
+git_directories.each do |git_directory|
+  puts git_directory
+
+  Dir.chdir(File.join(parent_directory, git_directory))
+  remotes = `git remote`.split
+
+  if remotes.include?("origin") && remotes.include?("production")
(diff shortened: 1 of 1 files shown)


bin/production-status
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env ruby --disable-gems
+
+require "pathname"
+
+parent_directory = File.expand_path("../../..", __FILE__)
+Dir.chdir(parent_directory)
```
gylaz: Worth using the block version of [`Dir.chdir`](http://ruby-doc.org/core-2.2.3/Dir.html#method-c-chdir) to be returned where we started?

bin/production-status
```diff
@@ -0,0 +1,34 @@
...
+  puts git_directory
+
+  Dir.chdir(File.join(parent_directory, git_directory))
+  remotes = `git remote`.split
+
+  if remotes.include?("origin") && remotes.include?("production")
+    `git fetch origin > /dev/null`
+    `git fetch production > /dev/null`
+
+    commits_not_on_production = `git log --graph --pretty='format:%C(yellow)%h %C(green)%ar %C(cyan)%an %C(reset)%s' production/master..origin/master`
```
gylaz: Noice!

gylaz:
I like!

bin/production-status
thorncp: The script doesn't do anything at the end that requires being in the original directory, so I think I'm inclined to avoid the extra level of nesting.

PR #1095, opened 2016-03-04
Overthrow Big Brother

thorncp:
These log messages were useful when we were having issues with token
usage, but I don't think they're needed anymore. Additionally, it
noticeably decrease the size of our logs.


gylaz:
Very nice commit title, and still descriptive. :+1:

gylaz:
FWIW, I relied on this logging as I debugged an issue today.

We've discussed in the past improving our logging to give us a better view into what is going on during a build -- so we can debug more efficiently.

PR #1097, opened 2016-03-05
Configure operator indentation for thoughtbot

gylaz:
Splitting continuous line at the operator, causes Hound to find a
violation. RuboCop's default is to align the operators, but we want the
next line to be indented one extra level.


thorncp:
:+1:

PR #1098, opened 2016-03-07
Don't rescue Octokit::NotFound when setting status

gylaz:
Introduced in e9836eea64baefb0699952c5f35fd65a9ca2df41, the logic was in
regard to adding Hound user to teams. Since, we are not doing that
anymore this is not longer relevant to the original problem.

Currently, there is an issue where we choose a user token to handle the
repo, since a repo could be public, the user's token will work for
processing the build/review.
But, it will fail silently when it tries to set the status. This is the
cause of the GitHub status sometimes not being set, a token that was
picked does not have status setting privileges.
If we raise, the pending status will fail and the job will retry using a
new token.

### DIFF
--- a/spec/lib/github_api_spec.rb
+++ b/spec/lib/github_api_spec.rb
@@ -231,23 +231,6 @@ describe GithubApi do
 
       expect(request).to have_been_requested
     end
-
-    describe "when setting the status returns 404" do
-      it "does not crash" do
-        sha = "abc"
-        api = GithubApi.new(Hound::GITHUB_TOKEN)
-        stub_failed_status_creation_request(
-          full_repo_name,
-          sha,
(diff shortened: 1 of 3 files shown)


spec/lib/github_api_spec.rb
```diff
@@ -233,19 +233,18 @@
     end
 
     describe "when setting the status returns 404" do
```
gylaz: Does this test have any value anymore? We stub the response then assert that the error is raised. I'm thinking it could be removed along with the `stub_failed_status_creation_request` helper.

thorncp: :scissors: :it:

thorncp:
LGTM

PR #1102, opened 2016-03-15
Add file_reviews.linter_name

jbusser:
[Related Github issue](https://github.com/houndci/hound/issues/1101)

Generally speaking, for any given language, Hound supports only one linter. For
example, Rubocop will handle all `.rb` files; SCSS-lint will lint `.scss` files
and so on.

And then there's JavaScript.

For JavaScript, Hound supports three linters: JShint, JSCS, and ESlint,
each of which can be individually enabled or disabled. If both JShint and
JSCS are enabled, however, Hound code reviews will never complete for a given
build, and this is because `CompleteFileReivew#file_review` will only look up
a `FileReview` by its `filename` attribute, and when multiple linters are
working on a single file, that is not enough information. The second through nth
`FileReivew` for that file never gets found, and its `completed_at` never gets
updated. When that happens, `Build#complete?` always returns false.

This change creates a migration that adds a new column to the `file_reviews`
table, which will store the name of the linter assigned to a given file review.
The corresponding `FileReview` model will acquire a model-level validation to
ensure the presence of this new linter-name attribute.

It then updates `Linter::Base#enqueue_job` to add the linter_name attribute to
the hash of review-job attributes so that linters can pick up this new
attribute. When their reviews are complete, the linters can be configured to
pass the linter name back to Hound, along with any errors and the filename, just
as before.

The last part of this solution involves updating
`CompleteFileReview#file_review` to anticipate an optional `linter_name`
attribute in its dequeued job attributes, which it will use to perform a more
precise search of `FileReview` objects in the relevant build. In order to avoid
a large coordinated upgrade effort, this method will fall back to the previous
locator logic, which only operates on the job's `filename` attribute.

### DIFF
--- a/app/models/linter/coffee_script.rb
+++ b/app/models/linter/coffee_script.rb
@@ -6,7 +6,10 @@ module Linter
     FILE_REGEXP = /.+\.coffee(\.js)?(\.erb)?\z/
 
     def file_review(commit_file)
-      FileReview.create!(filename: commit_file.filename) do |file_review|
+      FileReview.create!(
+        filename: commit_file.filename,
+        linter_name: name,
+      ) do |file_review|
         content = content_for_file(commit_file)
 
         lint(content).each do |violation|
--- a/app/models/linter/ruby.rb
+++ b/app/models/linter/ruby.rb
@@ -15,7 +15,10 @@ module Linter
     private
 
     def perform_file_review(commit_file)
-      FileReview.create!(filename: commit_file.filename) do |file_review|
+      FileReview.create!(
+        filename: commit_file.filename,
+        linter_name: name,
+      ) do |file_review|
         permitted_rubocop_offenses(commit_file).each do |violation|
           line = commit_file.line_at(violation.line)
           file_review.build_violation(line, violation.message)
--- a/app/services/complete_file_review.rb
+++ b/app/services/complete_file_review.rb
@@ -47,8 +47,21 @@ class CompleteFileReview
   end
 
   def file_review
-    @file_review ||= build.file_reviews.
-      find_by!(filename: attributes.fetch("filename"))
+    @file_review ||= build.file_reviews.find_by!(file_review_properties)
+  end
+
+  def file_review_properties
+    if attributes.has_key?("linter_name")
+      legacy_file_review_search_properties.merge(
+        linter_name: attributes.fetch("linter_name"),
+      )
+    else
+      legacy_file_review_search_properties
+    end
+  end
+
+  def legacy_file_review_search_properties
+    { filename: attributes.fetch("filename") }
   end
 
   def commit_file
--- a/db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
+++ b/db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
@@ -1,0 +1,9 @@
+class AddLinterNameToFileReviews < ActiveRecord::Migration
+  def up
+    add_column :file_reviews, :linter_name, :string
+  end
+
+  def down
+    remove_column :file_reviews, :linter_name
+  end
+end
--- a/db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
+++ b/db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
@@ -1,0 +1,35 @@
+class BackfillLinterNamesAddNullConstraint < ActiveRecord::Migration
+  def up
+    sql = <<-SQL
+      UPDATE file_reviews
+      SET linter_name = CASE
+        WHEN filename ~* '\\.coffee(\\.js)?(\\.erb)?\\Z' THEN 'coffee_script'
+        WHEN filename ~* '\\.go\\Z' THEN 'go'
+        WHEN filename ~* '\\.haml\\Z' THEN 'haml'
+        WHEN filename ~* '(\\.es6|\\.es6\\.js)\\Z' THEN 'eslint'
+        WHEN filename ~* '\\.js\\Z' THEN 'jshint'
+        WHEN filename ~* '(\\.md|\\.markdown)\\Z' THEN 'remark'
+        WHEN filename ~* '\\.py\\Z' THEN 'python'
+        WHEN filename ~* '\\.rb\\Z' THEN 'ruby'
+        WHEN filename ~* '\\.scss\\Z' THEN 'scss'
+        WHEN filename ~* '\\.swift\\Z' THEN 'swift'
+        ELSE 'unsupported'
+        END;
+    SQL
+
+    execute(sql)
+
+    change_column_null :file_reviews, :linter_name, false
+  end
+
+  def down
+    change_column_null :file_reviews, :linter_name, true
+
+    sql = <<-SQL
+      UPDATE file_reviews
+      SET linter_name = null;
+    SQL
+
+    execute(sql)
+  end
+end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -15,6 +15,7 @@ FactoryGirl.define do
     build
 
     filename "the_thing.rb"
+    linter_name "ruby"
   end
 
   factory :repo do
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,16 +150,18 @@ Resque queue.
 
 Linter jobs are created with the following arguments:
 
+* `commit_sha` - The git commit SHA of the code snippet. This should be passed
+  through to the outbound queue.
 * `config` - The configuration for the linter. This will be linter specific.
 * `content` - The source code to check for violations.
 * `filename` - The name of the source file for the code snippet. This should be
   passed through to the outbound queue.
-* `commit_sha` - The git commit SHA of the code snippet. This should be passed
-  through to the outbound queue.
-* `pull_request_number` - The GitHub pull request number. This should be passed
-  through to the outbound queue.
+* `linter_name` - A string that identifies which linter is assigned to this
+  linter job. This must be passed through to the outbound queue unmodified.
 * `patch` - The patch content from GitHub for the file being reviewed. This
   should be passed through to the outbound queue.
+* `pull_request_number` - The GitHub pull request number. This should be passed
+  through to the outbound queue.
 
 Once linting is complete, resulting violations should be posted to the outbound
 "CompletedFileReviewJob" queue:
@@ -173,8 +175,8 @@ Once linting is complete, resulting violations should be posted to the outbound
   provided by the inbound queue.
 * `commit_sha` - The git commit SHA of the code snippet. This is provided by the
   inbound queue.
-* `pull_request_number` - The GitHub pull request number. This is provided by the
-  inbound queue.
+* `linter_name` - A string that identifies which linter is assigned to this
+  linter job. This is provided by the inbound queue.
 * `patch` - The patch content from GitHub for the file being reviewed. This is
   provided by the inbound queue.
 
--- a/app/models/linter/base.rb
+++ b/app/models/linter/base.rb
@@ -15,6 +15,7 @@ module Linter
       file_review = FileReview.create!(
         build: build,
         filename: commit_file.filename,
+        linter_name: name,
       )
 
       enqueue_job(attributes)
@@ -44,6 +45,7 @@ module Linter
         config: config.serialize,
         content: commit_file.content,
         filename: commit_file.filename,
+        linter_name: name,
         patch: commit_file.patch,
         pull_request_number: build.pull_request_number,
       }
--- a/app/models/linter/haml.rb
+++ b/app/models/linter/haml.rb
@@ -7,7 +7,10 @@ module Linter
     def file_review(commit_file)
       @commit_file = commit_file
 
-      FileReview.create!(filename: commit_file.filename) do |file_review|
+      FileReview.create!(
+        filename: commit_file.filename,
+        linter_name: name,
+      ) do |file_review|
         run_linters.map do |violation|
           line = commit_file.line_at(violation.line)
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -52,6 +52,7 @@ ActiveRecord::Schema.define(version: 20160513002940) do
     t.datetime "created_at",   null: false
     t.datetime "updated_at",   null: false
     t.string   "filename",     null: false
+    t.string   "linter_name",  null: false
   end
 
   add_index "file_reviews", ["build_id"], name: "index_file_reviews_on_build_id", using: :btree
--- a/spec/models/linter/eslint_spec.rb
+++ b/spec/models/linter/eslint_spec.rb
@@ -59,6 +59,7 @@ describe Linter::Eslint do
         EslintReviewJob,
         filename: commit_file.filename,
         commit_sha: build.commit_sha,
+        linter_name: "eslint",
         pull_request_number: build.pull_request_number,
         patch: commit_file.patch,
         content: commit_file.content,
--- a/spec/models/linter/go_spec.rb
+++ b/spec/models/linter/go_spec.rb
@@ -42,6 +42,7 @@ describe Linter::Go do
         GoReviewJob,
         filename: commit_file.filename,
         commit_sha: build.commit_sha,
+        linter_name: "go",
         pull_request_number: build.pull_request_number,
         patch: commit_file.patch,
         content: commit_file.content,
--- a/spec/models/linter/jscs_spec.rb
+++ b/spec/models/linter/jscs_spec.rb
@@ -59,6 +59,7 @@ describe Linter::Jscs do
         JscsReviewJob,
         filename: commit_file.filename,
         commit_sha: build.commit_sha,
+        linter_name: "jscs",
         pull_request_number: build.pull_request_number,
         patch: commit_file.patch,
         content: commit_file.content,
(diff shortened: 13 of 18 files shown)


app/models/linter/jscs.rb
```diff
@@ -1,5 +1,7 @@
 module Linter
   class Jscs < Base
+    include JobWithLinterName
```
teoljungberg: Could we create a class out of this instead? To use composition instead of inheritance?

app/models/linter/ruby.rb
```diff
@@ -15,7 +15,9 @@ def file_included?(commit_file)
     private
 
     def perform_file_review(commit_file)
-      FileReview.create!(filename: commit_file.filename) do |file_review|
+      FileReview.create!(
+        filename: commit_file.filename,
+        linter_name: name) do |file_review|
```
teoljungberg: This formatting is somewhat against how we have done things in Hound.

We usually format these expressions as:

```ruby
FileReview.create!(
  filename: commit_file.filename,
  linter_name: name,
) do |file_review|
# ...
end

```

app/services/complete_file_review.rb
```diff
@@ -47,8 +47,10 @@ def build
   end
 
   def file_review
-    @file_review ||= build.file_reviews.
-      find_by!(filename: attributes.fetch("filename"))
+    @file_review ||= (build.file_reviews.find_by(
+      filename: attributes.fetch("filename"),
+      linter_name: attributes["linter_name"]) ||
```
teoljungberg: Same as above.

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
```diff
@@ -0,0 +1,5 @@
+class AddLinterNameToFileReviews < ActiveRecord::Migration
+  def change
+    add_column :file_reviews, :linter_name, :string
```
teoljungberg: Should this have a default value, and `null: false`?

jbusser: Sure. I could do that. I will remove the model-level validation then as well.

app/models/linter/jscs.rb
jbusser: Embarrassed to say I don't think I'm following you here. I thought my including a module was following composition over inheritance. What do you have in mind?

app/models/linter/ruby.rb
jbusser: No problem at all.

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
jbusser: Do you have a preference for a default value? I know @gylaz supported the approach implemented in this PR because knowing the linter responsible for each file review had reporting value. For all file reviews created before this PR gets merged, we won't necessarily have that information. With reporting in mind should the default value be something like "unknown" or "default?"

app/models/linter/jscs.rb
teoljungberg: Including modules is basically the same as inheriting from a class.

I would imagine a class along the lines off to `JobAttributesFor.call(commit_file)`.

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
teoljungberg: For the default value, I don't have any preference. Thoughts @gylaz ?

app/models/linter/jscs.rb
jbusser: Rather than doing that, what if I went ahead and changed `Linter::Base#build_review_job_attributes` to supply the new linter_name attribute? The change is additive and therefore won't affect any linters, but it will be present in the payload if/when those linters ever need it.

Changing `Linter::Base#build_review_job_attributes` was my original approach but then saw it caused every linter's test to fail because their Resque payloads had changed, but perhaps it's worth the trouble? Or do you think a separate class is still a better design?

teoljungberg: I would add it to `Linter::Base#build_review_job_attributes` and leave it at that.

We'll extract a new class when we need one.

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
gylaz: Ideally, we would figure out what linter it should be. What if the migration set the linter name based on the file extension?

app/models/linter/jscs.rb
gylaz: :+1: on adding it to the `Linter::Base#build_review_job_attributes` if it doesn't break any of the services that don't support that argument.

However, say you wanted to only override it for a few linters, you could do:
```
module Linter
  class Jscs < Base
    def build_review_job_attributes(commit_file)
      super.merge(linter_name: "foo")
    end
  end
end
```

app/models/file_review.rb
```diff
@@ -2,6 +2,8 @@ class FileReview < ActiveRecord::Base
   belongs_to :build
   has_many :violations
 
+  validates_presence_of :linter_name
```
gylaz: IMO, adding model-level validations is only useful when a user/client (UI or API) will consume them. If there is no such user, then DB constraints should be the gospel.

app/models/linter/coffee_script.rb
```diff
@@ -6,7 +6,9 @@ class CoffeeScript < Base
     FILE_REGEXP = /.+\.coffee(\.js)?(\.erb)?\z/
 
     def file_review(commit_file)
-      FileReview.create!(filename: commit_file.filename) do |file_review|
+      FileReview.create!(
+        filename: commit_file.filename,
+        linter_name: name) do |file_review|
```
gylaz: `) do |file_review|` should be on its own line according to our guides.

app/services/complete_file_review.rb
gylaz: Can we extract this first part into a method to clean this up?

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
jbusser: I could do that, but we would still need to decide on a rule for .js files, wouldn't we? If I'm given only a filename from a file review, I don't think it's possible to see what Hound configuration was in effect for a given repo during a given build, is it? I wouldn't necessarily know whether JShint, JSCS, or ESlint was active, right?

gylaz: You are correct. I think at the time of the migration we could make a best guess -- say all the `*.js` are `jshint` since both `jscs` and `eslint` support were added recently and if we wrongly guess it's not a big deal. Alternatively, we can set `*.js` reviews to be `javascript` (and treat it as legacy before we made the switch to multi JS linters).

jbusser:
Hi, guys --

I've fixed all the style errors; added the linter_name param to all Resque payloads; made a best effort at backfilling linter names; and I cleaned up that kind of gross implementation of `CompleteFileReview#file_review`.

Thank you for taking the time to review this.

jbusser:
Could I get a status update on this pull? I would like to get this committed since my company pays for Hound to lint some private repos. This change addresses a real problem we experience.

The CI error is the result of a failed security audit and does not appear to be related to this change. However, I'm happy to update the Administrate gem in a separate pull if you would like.

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
```diff
@@ -0,0 +1,42 @@
...
+    backfill_linter_names
+    change_column :file_reviews, :linter_name, :string, null: false
+  end
+
+  def down
+    remove_column :file_reviews, :linter_name
+  end
+
+  def backfill_linter_names
+    FileReview.reset_column_information
```
gylaz: Why is this line needed?

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
```diff
@@ -0,0 +1,42 @@
...
+      Linter::Haml,
+      Linter::Jshint,
+      Linter::Eslint,
+      Linter::Remark,
+      Linter::Python,
+      Linter::Ruby,
+      Linter::Scss,
+      Linter::Swift,
+      Linter::Unsupported,
+    ]
```
gylaz: When one of these linters is removed, this migration will fail for future contributors. Not sure the best way to handle this situation, maybe a one off task is a better fit for this?

gylaz:
@jbusser The JSCS support is in beta, and now that it is merging with ESLint we are likely to drop support for it soon.

@teoljungberg @thorncp Since this situation won't arise if we drop support for JSCS, do we still think it's a good to do this? (analytics, future linters, etc.)

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
teoljungberg: Couldn't this use the list of linters already available in `Linter::Collection::LINTERS`?

teoljungberg:
I can still see that there's value in keeping track which linter is responsible for which file review, eventhough `JSCS` is being moved out of the picture in the near future.

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
```diff
@@ -0,0 +1,42 @@
...
+      file_review.update linter_name: linter_name_for(file_review.filename)
+    end
+  end
+
+  def linter_name_for(filename)
+    linters.
+      detect { |linter| linter.can_lint?(filename) }.
+      name.
+      demodulize.
+      underscore
```
gylaz: Can we use the `Linter::Base#name` method here instead?

spec/factories.rb
```diff
@@ -15,6 +15,7 @@
     build
 
     filename "the_thing.rb"
+    linter_name "base"
```
gylaz: Maybe this should be a real linter, like `ruby`? I don't think it can ever be `base`.

gylaz:
@jbusser A couple more things, then we can move forward with merging this. Thanks for pushing on this!

jbusser:
Thank you, @gylaz and @teoljungberg. I'll try to get those changes pushed up tonight.

db/migrate/20160314122828_add_linter_name_to_file_reviews.rb
jbusser: @teoljungberg I certainly could. I took this collection from `Linter::Collection::LINTERS`. The only difference was removing the entry for `Linter::JSCS`. I chose to do that for data-accuracy. The regexp for matching the JSHint linter is similar to JSCS', but JSCS appears before JSHint in the array. That means every file with a `.js` suffix would be retroactively assigned a JSCS linter. Since jshint support predated JSCS support, I thought that would provide a more accurate history.

teoljungberg: The issue that both @gylaz and I've pointed out is that this migration will have to be updated every time the list of linters changes, which is sub-optimal.

jbusser: Yep. I get it. The migration depends too heavily on the state of the codebase. Related question: what is your opinion on referring AR models in migrations?

In the weeks since I pushed this code, I encountered a problem in my day job were we made a material change to a model that resulted in breaking earlier migrations. To correct this, we followed Justin Searls' advice outlined here: http://blog.testdouble.com/posts/2014-11-04-healthy-migration-habits.html

I still feel a migration is the only way out of this problem. I can't see how to ensure a not-null integrity constraint on file_reviews.linter_name with a combination of migration and one-off rake task.

So what if I:
* Defined a `MigrationFileReview` model here, breaking any dependency on `FileReivew`
* Defined a map of regexps and linter names, breaking the dependency on anything in `Linter::Collection`

gylaz: > The only difference was removing the entry for Linter::JSCS.

Once #1119 lands, it won't be an issue anymore?

> what is your opinion on referring AR models in migrations?

We typically avoid using AR models in migrations, preferring SQL queries.
However, sometimes it will make sense to use an AR model, so what we woud do is:
```ruby
class AddLinterNameToFileReviews < ActiveRecord::Migration
  class FileReview < ActiveRecord::Base
  ...
end
```
Then `FileReview.find_each` will always work (well, until AR decides to rename/remove that method).

I think that in this case we can write a SQL query, with a `CASE` statement that does a regex match on the filename. It's more work, but it does not ties us to any additional dependencies.

jbusser: :+1:

jbusser:
I just wanted to let you guys know that this is ready for another review. The Ci failure is (still) the result of the out-of-date administrate gem and not necessarily related to this change. I'm still happy to fix that in separate pull if that would help.

teoljungberg:
@jbusser Try rebasing against master to remove the CI failure.

db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
```diff
@@ -0,0 +1,35 @@
...
+        WHEN filename ~* '(\\.md|\\.markdown)\\Z' THEN 'remark'
+        WHEN filename ~* '\\.py\\Z' THEN 'python'
+        WHEN filename ~* '\\.rb\\Z' THEN 'ruby'
+        WHEN filename ~* '\\.scss\\Z' THEN 'scss'
+        WHEN filename ~* '\\.swift\\Z' THEN 'swift'
+        ELSE 'unsupported'
+        END;
+    SQL
+
+    ActiveRecord::Base.connection.execute(sql)
```
gylaz: This could just be `execute(sql)`.

jbusser: Would you prefer I change it? I used examples from `20150213132248_backfill_timestamps_add_null_constraints.rb` and `20141219224840_inactivate_repos_without_privacy_or_org_info.rb`. It's no issue for me either way.

db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
```diff
@@ -0,0 +1,35 @@
+class BackfillLinterNamesAddNullConstraint < ActiveRecord::Migration
+  def up
+    sql = <<-SQL
+      UPDATE file_reviews
+      SET linter_name = CASE
+        WHEN filename ~* '\\.coffee(\\.js)?(\\.erb)?\\Z' THEN 'coffee_script'
+        WHEN filename ~* '\\.go\\Z' THEN 'go'
+        WHEN filename ~* '\\.haml\\Z' THEN 'haml'
+        WHEN filename ~* '(\\.es6|\\.es6\\.js)\\Z' THEN 'eslint'
```
gylaz: Should thre be an `|` after `.es6`?

thorncp:
This is great, thanks @jbusser! :+1: :dancers:

db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
jbusser: I believe we do want the `|` because we want to duplicate the regexp from `Linter::Eslint`:
```
    FILE_REGEXP = /.+(\.js|\.es6|\.es6\.js)\z/
```
I am deliberately skipping the `.js` pattern because I want that to fall through to match the jshint linter in the next line.

db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
```diff
@@ -0,0 +1,35 @@
...
+        WHEN filename ~* '\\.js\\Z' THEN 'jshint'
+        WHEN filename ~* '(\\.md|\\.markdown)\\Z' THEN 'remark'
+        WHEN filename ~* '\\.py\\Z' THEN 'python'
+        WHEN filename ~* '\\.rb\\Z' THEN 'ruby'
+        WHEN filename ~* '\\.scss\\Z' THEN 'scss'
+        WHEN filename ~* '\\.swift\\Z' THEN 'swift'
```
gylaz: I believe all of these need to be written with a single `\` as an escaping character.

jbusser: I thought so, too, because of the heredoc, but it didn't work when I tested the migration. I needed the double escapes.

db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
gylaz: I see.

Should we move the `jshint` case above this? Right now, `jshint` will pickup `.es6.js` files (unless it's explicitly disabled). Or, are you making a best guess that if the file extension is `.es6.js` then it probably was also linted with ESLint explicitly?

gylaz:
Thanks for your help and patience @jbusser!

db/migrate/20160430171015_backfill_linter_names_add_null_constraint.rb
jbusser: Yes, the latter. It was an attempt at trying to make the historical record of linters and jobs associated with them line up with the order in which Hound supported them. Jshint support ESlint, so I wanted to give it the win, if you will, in the event of any ties.

jbusser:
This was my first significant open-source contribution. Thanks for making it a pleasant experience.

croaky:
Thank YOU, @jbusser!

PR #1103, opened 2016-03-15
Set JSCS default config as string, not hash

jbusser:
Given a basic JSCS configuration of

```
jscs:
  enabled: true
```

Instances of `SmallBuildJob` will log an exception to Resque's `failed` queue.
The exception reads "no implicit conversion of Hash into String."

The root cause turned out to be `Config::Jscs#default_content`. Instead of being
a string, it was a Ruby hash, and when a user provided no configuration file in
his `.hound.yml`, `Config::Jscs` attempted to parse that hash via `YAML.parse`,
resulting in the exception.

This change removes `Config::Jscs#default_content`, delegating to
`Config::Base#default_content`, which is already a String. It also changes the
parser used from YAML to JSON, which more closely aligns with the implementation
of `Config::Jshint` and the format of JSCS's configuration file.

### DIFF
--- a/app/models/config/jscs.rb
+++ b/app/models/config/jscs.rb
@@ -9,9 +9,5 @@ module Config
     def parse(file_content)
       Parser.yaml(file_content)
     end
-
-    def default_content
-      {}
-    end
   end
 end
--- a/spec/models/config/jscs_spec.rb
+++ b/spec/models/config/jscs_spec.rb
@@ -18,7 +18,7 @@ describe Config::Jscs do
   end
 
   describe "#serialize" do
-    it "serializes the parsed content into YAML" do
+    it "serializes the parsed content into JSON" do
       raw_config = <<-EOS.strip_heredoc
         { "disallowKeywordsInComments": true }
       EOS


teoljungberg:
Here's the rub. JSCS config file isn't actually JSON, since it allows trailing commas, comments. Which are not a part of the JSON spec.

But, it is valid YAML - which is a superset of JSON.

jbusser:
TIL.

I can restore the YAML.parse, but what do you think about the default_config issue?

teoljungberg:
Where does the exception occur? And why?

`Linter::*#serialize` should serialize `#content` from a hash to a string.

jbusser:
I think the problem is in `Config::Base#load_content`. The `file_path` method is falsy when jscs is enabled in `.hound.yml` and no `config_file` option is specified. It then evaluates `default_content`. The value returned from `default_content` needs to be a string because that gets passed to `safe_parse`. That method, in turn, returns the hash that `#serialize` expects.

gylaz:
I've ran into this same situation with ESLint (and fixed it). :+1:

gylaz:
Thanks @jbusser.

Issue #1106, opened 2016-03-16
ESLint: Support for javascript file


brgaulin:
ESLint supports Javascript, JSON, and YAML.
http://eslint.org/docs/user-guide/configuring#configuration-file-formats

So today our project has something like this
```javascript
module.exports = {
    extends: "airbnb"
}
```

Which can't be read by hound. I may work on a PR for this later today, unfortunately, JS ones can't be verified I think by Ruby code(ExecJS doesn't support module exports)

marcbachmann:
I suppose that might be more complicated on the service side because houndci probably doesn't want to execute javascript code because of security issues.
This is only a guess. I have no idea whether they already execute third party code on their server.

teoljungberg:
@marcbachmann You're right about that.

This is sadly not feasable in the current landscape of Hound.

PR #1107, opened 2016-03-25
Support .eslint.js

brgaulin:
ESLint is a bit complicated, and one of its complications, is how it handles config files.

In hound currently, we pass the config file to the queue and assume it is JSON.

This change, simply passes through file content as well as the file name so that when eslint goes to run it will know what to do.

This enables support for .eslint.js and .eslint.yaml, as well as .eslint.json and .eslint

In the future, we should probably find a better way to pass around configs? but this should work for now.

### DIFF
--- a/app/models/config/eslint.rb
+++ b/app/models/config/eslint.rb
@@ -1,15 +1,17 @@
 module Config
   class Eslint < Base
-    def serialize(data = content)
-      Serializer.json(data)
+    def serialize(_data = nil)
+      Serializer.json(
+        raw_content: load_content,
+        file_name: file_path,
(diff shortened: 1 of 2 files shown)


brgaulin:
This change is linked to https://github.com/houndci/eslint/pull/22 The changes to hound are not backwards compatible, but the changes to hound-eslint are, so we would want to deploy that one first, then the changes to hound

thorncp:
Thank you for your efforts, @brgaulin.

As mentioned in #1106, executing user supplied JavaScript is a security issue that we aren't currently set up to handle, and it's not a priority for us at this time.

PR #1108, opened 2016-03-25
Updating Rubocop Rule Names

brgaulin:
Looks like the names of the rubocops have changed when hound recently updated, this caused failing PRs.


gylaz:
Thank you @brgaulin.

PR #1112, opened 2016-04-13
Check for presence of line in HAML exception

gylaz:
Sometimes the HAML exception does not contain a line number.

### DIFF
--- a/app/models/linter/haml.rb
+++ b/app/models/linter/haml.rb
@@ -44,7 +44,11 @@ module Linter
         results + linter.lints
       end
     rescue HamlLint::Exceptions::ParseError => haml_error
-      [HamlViolation.new(haml_error.line + 1, haml_error.message)]
+      if haml_error.line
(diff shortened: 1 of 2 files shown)


app/models/linter/haml.rb
```diff
@@ -41,7 +41,9 @@ def run_linters
         results + linter.lints
       end
     rescue HamlLint::Exceptions::ParseError => haml_error
-      [HamlViolation.new(haml_error.line + 1, haml_error.message)]
+      if haml_error.line
```
gylaz: This should check if `haml_error` is not `nil`.

gylaz: Actually, this is fine. The `haml_error` is present, it just doesn't contain a line.

gylaz:
I figured out one situation that will cause an error without a line and wrote a test.

gylaz:
Closing, since Haml linter has been extracted into https://github.com/houndci/linters.

PR #1118, opened 2016-04-28
Lookup ignore files using the linter's name

gylaz:
There was a bug where `JsIgnore` looked up the ignore file only for the
`javascript` linter. It now takes in a linter name and uses that as the
key to look up in `.hound.yml`.


teoljungberg:
Looks good!

PR #1119, opened 2016-04-28
Remove JSCS linter support

gylaz:
JSCS is merging into ESLint: http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint


gylaz:
Closing this for now. We can revisit when we actually decide to stop supporting JSCS.

PR #1124, opened 2016-05-06
Remove Code Climate badge

thorncp:
The badge has said 4.0 for many months (years?) and the Hound team does
not seem to be getting any value from Code Climate's analysis.

Anyone feel that this badge should stay?


gylaz:
👍

PR #1127, opened 2016-05-06
Lint HAML via service

thorncp:
Removes in-line HAML linting in favor of scheduling jobs to be picked up
by a service monitoring the `haml_review` queue.

Depends on https://github.com/houndci/linters/pull/28

https://trello.com/c/8BkUFmhA


app/jobs/haml_review_job.rb
```diff
@@ -0,0 +1,3 @@
+class HamlReviewJob
+  @queue = :haml_review
```
gylaz: Thoughts about all jobs going to the same queue? I'm thinking, we wouldn't have to spin up multiple workers that sit idle for languages/linters that see less action.

Alternatively, we can schedule to multiple queues, but have a worker that listens on all/most queues, and see how that feels?

app/models/linter/haml.rb
```diff
@@ -1,71 +1,5 @@
...
-    def default_config
-      HamlLint::ConfigurationLoader.load_file(default_config_file)
-    end
-
-    def default_config_file
-      DefaultConfigFile.new(
-        DEFAULT_CONFIG_FILENAME,
-        repository_owner_name
-      ).path
-    end
```
gylaz: Do we need to preserve this merging of config (Hound/thoughtbot defaults + user overrides)?

spec/models/style_checker_spec.rb
```diff
@@ -214,30 +214,13 @@
...
+        pull_request = stub_pull_request(commit_files: [commit_file])
 
-          violation_messages = pull_request_violations(pull_request)
+        violation_messages = pull_request_violations(pull_request)
 
-          expect(violation_messages).not_to include(
-            "`%div.foo` can be written as `.foo` since `%div` is implicit",
-          )
-        end
+        expect(violation_messages).to be_empty
```
gylaz: This doesn't seem to fit the heredoc.

app/jobs/haml_review_job.rb
thorncp: Not having idle workers is a good point for that. I do think using the queue name is a convenient way to handle what linters should be invoked, though I don't know if that outweighs the downside of idle workers. It would require more changes on the linters end, so maybe worth investigating in another PR?

spec/models/style_checker_spec.rb
thorncp: Ha, you're right. I copied it [from above](https://github.com/houndci/hound/blob/ct-haml-service/spec/models/style_checker_spec.rb#L195), so it seem all of these are wrong.

app/jobs/haml_review_job.rb
gylaz: I was thinking along the lines of using meta-programming to determine what needs to run, something like `"#{job_args['linter']}Service".constantize.new(...).run` on `linters` side. Another PR sounds good.

app/models/linter/haml.rb
thorncp: I just added a commit to https://github.com/houndci/linters/pull/28 that does this.

spec/models/style_checker_spec.rb
thorncp: I don't see how these specs are providing any value, so I deleted them all in 2452ebe9a53e64456f902faf58416e64a060a420. Am I missing something?

gylaz: I think these originally served as sort of integration tests. They seem useless now, but... is there no way to verify that the right thing is happening when `StyleChecker.new(pull_request, build).review_files` is called? Scheduled job? Something else?

PR #1128, opened 2016-05-13
Ensure uniqueness of repos by github_id

gylaz:
Previously, we were ensuring repo uniqueness by name, but since we are
tightly coupled to GitHub, we should be using their ids.

There was a recent bug that led to this:

* On GitHub, `thoughtbot/constable-api` repo was renamed to `thoughtbot/constable`
* On GitHub, `thoughtbot/constable` repo was renamed to `thoughtbot/constable-react`
* When we were syncing repos, we would find the `thoughtbot/constable`
  and create a membership for it
* Then we would see the `thoughtbot/constable-react`. Nnot find it
  by name, then look it up by `github_id` and pull up the same
  `thoughtbot/constable` record and try to create a membership,
  which would fail.

If we relied on `github_id` for uniquness we would eventually update the
names to match what GitHub has.

### DIFF
--- a/db/migrate/20160513002940_add_github_id_uniqueness_on_repos.rb
+++ b/db/migrate/20160513002940_add_github_id_uniqueness_on_repos.rb
@@ -1,0 +1,23 @@
+class AddGithubIdUniquenessOnRepos < ActiveRecord::Migration
+  def up
+    remove_uniqueness(:full_github_name)
+    add_uniqueness(:github_id)
+  end
+
+  def down
+    remove_uniqueness(:github_id)
+    add_uniqueness(:full_github_name)
+  end
+
+  private
+
+  def remove_uniqueness(column)
+    remove_index :repos, column
+    add_index :repos, column
+  end
+
+  def add_uniqueness(column)
+    remove_index :repos, column
+    add_index :repos, column, unique: true
+  end
+end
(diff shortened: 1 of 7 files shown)


gylaz:
My only concern with this approach is that we can have multiple repos with the same name in our db. For example:

* Repo `thoughtbot/test` is created on GitHub (gets a unique id).
* We sync it and persist it.
* Repo `thoughtbot/test` is deleted on GitHub (we know nothing about this).
* A new repo `thoughtbot/test` is created on GitHub (gets a new unique id).
* We sync the new repo and now have two repos by the name of `thoughtbot/test`.

db/migrate/20160513002940_add_github_id_uniqueness_on_repos.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def down
+    remove_uniqueness(:github_id)
+    add_uniqueness(:full_github_name)
+  end
+
+  private
+
+  def remove_uniqueness(column)
```
teoljungberg: Nice use of helpers

thorncp:
I think this is an improvement! 👍 :sandal:

> My only concern with this approach is that we can have multiple repos with the same name in our db.

I don't think we currently have any real insight into how often this happens, but it seems infrequent enough to deal with manually as it occurs?

gylaz:
Looks good on staging. Merging.

PR #1129, opened 2016-05-13
Remove JSCS documenation

gylaz:
In preparation for dropping support of JSCS,
we're removing the documenation so users don't discover it.


teoljungberg:
Looks good!

thorncp:
:+1: :cloud:

PR #1130, opened 2016-05-13
Add index for builds pr number and sha

thorncp:
When completing a file review job, we find the associated build record
by pull request number and commit sha. With about 1.6 million records
in the table, this query is averaging 1820ms to execute.

This adds an index on the pull_request_number/commit_sha pair to improve
the performance of that query.

### DIFF
--- a/db/migrate/20160513205551_add_index_for_builds_pull_request_number_and_commit_sha.rb
+++ b/db/migrate/20160513205551_add_index_for_builds_pull_request_number_and_commit_sha.rb
@@ -1,0 +1,5 @@
+class AddIndexForBuildsPullRequestNumberAndCommitSha < ActiveRecord::Migration
+  def change
+    add_index :builds, [:commit_sha, :pull_request_number]
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160513002940) do
+ActiveRecord::Schema.define(version: 20160513205551) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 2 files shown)


db/migrate/20160513205551_add_index_for_builds_pull_request_number_and_commit_sha.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexForBuildsPullRequestNumberAndCommitSha < ActiveRecord::Migration
+  def change
+    add_index :builds, [:pull_request_number, :commit_sha]
```
gylaz: I could see this making sense as being a unique index.

gylaz:
👍

db/migrate/20160513205551_add_index_for_builds_pull_request_number_and_commit_sha.rb
thorncp: I thought that too, but we have been toying with the idea of re-triggering a build, so it would prevent that from being a possibility.

thorncp: I wonder if it makes sense to have `commit_sha` first in the index, so a query searching for just that value can take advantage of the index. That seems way more likely than search by just `pull_request_number`.

thorncp: Updated in 45b0d70c344e3ecc4f6746ef685dec83193ebb30

gylaz: Good idea.

thorncp:
For posterity, here's the explain plans for the following query:

```sql
SELECT "builds".* FROM "builds" WHERE "builds"."pull_request_number" = 42 AND "builds"."commit_sha" = 'foo' LIMIT 1
```

### Before

```text
 Limit  (cost=0.00..259667.72 rows=1 width=753)
   ->  Seq Scan on builds  (cost=0.00..259667.72 rows=1 width=753)
         Filter: ((pull_request_number = 42) AND ((commit_sha)::text = 'foo'::text))
```

### After

```text
 Limit  (cost=0.55..8.57 rows=1 width=732)
   ->  Index Scan using index_builds_on_commit_sha_and_pull_request_number on builds  (cost=0.55..8.57 rows=1 width=732)
         Index Cond: (((commit_sha)::text = 'foo'::text) AND (pull_request_number = 42))
```

PR #1131, opened 2016-05-13
Upgrade ActiveModelSerializers

### LINKED ISSUE #1090: Updating ActiveModelSerializers
So, I'm a maintainer of AMS and was wondering if you had any thoughts of updating to the [latest RC](https://rubygems.org/gems/active_model_serializers/versions) or any blockers if you've already discussed it :)

There doesn't appear to be any discussion in the repo here, and the version was hard-coded in https://github.com/houndci/hound/pull/536

gylaz:
Closes #1090

### DIFF
--- a/app/serializers/repo_serializer.rb
+++ b/app/serializers/repo_serializer.rb
@@ -27,7 +27,7 @@ class RepoSerializer < ActiveModel::Serializer
   private
 
   def membership
-    @membership ||= object.memberships.find_by(user_id: current_user.id)
+    @membership ||= object.memberships.find_by(user_id: scope.id)
   end
 
   def has_admin_membership?
@@ -35,6 +35,6 @@ class RepoSerializer < ActiveModel::Serializer
   end
 
   def has_subscription?
-    current_user.subscribed_repos.include?(object)
+    scope.subscribed_repos.include?(object)
   end
 end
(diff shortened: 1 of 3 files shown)


app/serializers/repo_serializer.rb
```diff
@@ -27,14 +27,14 @@ def admin
   private
 
   def membership
-    @membership ||= object.memberships.find_by(user_id: current_user.id)
+    @membership ||= object.memberships.find_by(user_id: scope.id)
```
thorncp: I hadn't seen `scope` before. The name is really confusing!

Posteriy: https://github.com/rails-api/active_model_serializers/tree/0-9-stable#customizing-scope

thorncp:
:+1: :tophat:

app/serializers/repo_serializer.rb
gylaz: Worth wrapping that in a `current_user` method or alias?

thorncp: It seems like that's going to be the AMS API going forward and it might be confusing later if we deviate from it. So, probably not?

gylaz: 👌

bf4:
Did you do any performance testing on this?

I'm surprised the change was so minimal.  Awesome!  Please leave any feedback in the AMS repo if you think the docs could be improved in any way

PR #1133, opened 2016-05-16
Lint JSX file extension with ESLint

gylaz:


teoljungberg:
Looks good!

I still wonder if we should allow users to configure which additional extensions they want to lint in their `.hound.yml`...

nneal:
Really appreciate this fix!! 👍

BreahnaDavid:
:+1: Thank you!

PR #1136, opened 2016-05-20
Renames config keys to be linter names

jackvnimble:
This makes it a bit more clear in a config file which linter will
actually used. This change nests the JSHint and ESLint linters
under JavaScript header on the configuration page

https://trello.com/c/CRtnokdz


app/models/config/base.rb
```diff
@@ -79,7 +79,7 @@ def file_path
     end
 
     def linter_config
-      hound_config.content.slice(*linter_names).values.first
+      hound_config.content.values_at(*linter_names).compact.first
```
jackvnimble: @gylaz not sure if you still prefer `values_at` when if requires compact (non_extant keys return nil)

jackvnimble:
The aformentioned nesting of ESLint and JSHint under Javascript

image
image

jackvnimble:
Should we update the default config to reflect these changes? Where does that live?

Also, https://raw.githubusercontent.com/houndci/hound/master/config/style_guides/haml.yml is a broken link.

thorncp:
> Should we update the default config to reflect these changes? Where does that live?

I think it makes sense to do that, in an attempt to encourage people to use linter names.

Good catch on the broken link! It should be https://raw.githubusercontent.com/houndci/linters/master/config/haml.yml

app/models/config/base.rb
gylaz: I can go either way. Your call.

thorncp:
Oh, I didn't answer your question. There isn't a default .hound.yml file that gets used. The config/linter objects themselves know if they're enabled by default or not. But a place to update to the linter name keys is on the config page, where it talks about .hound.yml - https://houndci.com/configuration#configuration

app/models/config/coffee_script.rb
```diff
@@ -2,6 +2,7 @@ module Config
   class CoffeeScript < Base
```
gylaz: What do you think about renaming these files and classes after their linters...

app/models/config/coffee_script.rb
```diff
@@ -2,6 +2,7 @@ module Config
...
     def linter_names
       [
+        "coffeelint",
```
gylaz: ... then here we use the old key (`coffeescript`)?

app/models/config/swift.rb
```diff
@@ -1,5 +1,13 @@
 module Config
   class Swift < Base
+    def linter_names
+      [
+        "swiftlint",
+        "swift_lint",
```
gylaz: Do we need to support this one?

app/views/pages/configuration.haml
```diff
@@ -10,9 +10,7 @@
       %li
         = link_to "Coffeescript", "#coffeescript"
```
gylaz: Worth camel-casing this while we're touching this fiel?

jackvnimble:
Ah, cool, I missed that one. Just changed it 👍

app/views/pages/configuration.haml
```diff
@@ -141,92 +139,98 @@
...
+          If you need to change the way Hound is configured, simply copy the
+          #{link_to "default config", javascript_config_url, target: :blank}.
+          into your project, make changes and reference the file in your
+          %em.code .hound.yml
+
+        %p
+          %code.code-block
+            :preserve
+              jshint:
+                config_file: .javascript-style.json
```
gylaz: How about renaming this to `.jshintrc` which is what JSHint uses by default?

app/models/config/coffee_script.rb
jackvnimble: I think that's a good idea

app/views/pages/configuration.haml
```diff
@@ -141,92 +139,98 @@
...
+        %p
+          If you would like to ignore certain JavaScript files, simply copy the
+          #{link_to "default config", javascript_ignore_url, target: :blank}
+          into your project, make changes and reference the file in your
+          %em.code .hound.yml
+
+          %code.code-block
+            :preserve
+              jshint:
+                ignore_file: .javascript_ignore
```
gylaz: `.jshintignore`?

app/models/config/swift.rb
jackvnimble: Probably not, I was mimicing the java_script one.

app/views/pages/configuration.haml
jackvnimble: Definitely

app/views/pages/configuration.haml
```diff
@@ -246,7 +250,7 @@
       %p
         %code.code-block
           :preserve
-            scss:
+            scss-lint:
```
gylaz: Is it strange that `coffeelint` doesn't have a hyphen but this does? Worth standardizing on one way?

app/views/pages/configuration.haml
jackvnimble: Good idea.

app/models/config/base.rb
jackvnimble: I think I'll keep it. `values_at` is much more clear than slice

app/views/pages/configuration.haml
thorncp: The CLI commands are `coffeelint` and `scss-lint`. Makes sense to me for our config keys to be consistent with that.

app/views/pages/configuration.haml
```diff
@@ -299,7 +303,7 @@
       %p
         %code.code-block
           :preserve
-            haml:
+            haml_lint:
```
gylaz: This one is `_`, but the one above is `-`.

app/views/pages/configuration.haml
gylaz: Maybe we should support both? If someone configures `coffeelint` first, they are likely to assume SCSS will be `scsslint`, and vice versa. It's just one extra entry in the `linter_names` array. Or am I being paranoid?

app/views/pages/configuration.haml
jackvnimble: Good catch, should be `-`

app/views/pages/configuration.haml
thorncp: I like supporting both. More flexibility will probably make for a better user experience.

jackvnimble: So add `coffee-lint` and `scsslint`?

gylaz: Yes to `scsslint`.

`coffee-lint`... maybe?

spec/models/linter/rubocop_spec.rb
```diff
@@ -697,19 +697,19 @@ def violations_with_config(config = stub_ruby_config)
 
     def build_linter(
       hound_config: build_hound_config,
-      config: stub_ruby_config,
+      config: stub_rubocop_config,
```
jackvnimble: Yeah... this took me a while to figure out :(

teoljungberg: Yeah, I'm not happy on how this turned out either - saying this as the guy who wrote it.

gylaz:
Closing this due to inactivity. We still want to do this, but probably handle this one linter at a time.

PR #1137, opened 2016-05-20
Improve formatting of failed jobs admin page

gylaz:
Make it so things are visible on the screen, rather than scrolling
horizontally.

Uses features from the latest Administrate release, to provide custom
CSS styles.

### DIFF
--- a/app/assets/stylesheets/admin/administrate.scss
+++ b/app/assets/stylesheets/admin/administrate.scss
@@ -1,0 +1,11 @@
+.collection-data {
+  th {
+    td {
+      white-space: nowrap;
+    }
+  }
+
+  td {
+    white-space: normal;
+  }
+}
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/admin/administrate.scss
```diff
@@ -0,0 +1,11 @@
+.collection-data {
+  th {
+    td {
```
gylaz: I think I'm okay with this.

config/initializers/administrate.rb
```diff
@@ -0,0 +1,2 @@
+Rails.application.config.assets.precompile += %w( admin/administrate.css )
```
thorncp: Is this a requirement for the administrate 0.2.0?

thorncp:
:+1: :trumpet:

config/initializers/administrate.rb
gylaz: Yes... ?

Without this I get the following error:
screenshot 2016-05-27 10 30 04

PR #1140, opened 2016-05-24
Separate explanation of dev setup from ngrok setup

jcmorrow:
If it isn't necessary for most developers to use ngrok then we should make that clear in the contributing guide because ngrok provides a substantial barrier to developer setup and could easily stop people from
contributing.

 Also, if people are using ngrok our current directions are incorrect, as the command for running ngrok has changed slightly and our recommended setup (with a custom subdomain) is no longer provided to
free accounts.

 I'm still not sure this is 100% the correct structure, but I think it will get people up and running faster. It would be nice to include a little bit about what type of feature development would require ngrok integration, as someone new to the codebase will almost certainly not know whether or not they need it. As someone who is new myself, I'm still not sure which parts of the codebase require it, so I can't include it in this PR.


CONTRIBUTING.md
```diff
@@ -25,19 +25,40 @@ Here are a few technical guidelines to follow:
...
+
+1. Run `foreman start`. Foreman will start the web server and the resque
+   background job queue. **NOTE**: `rails server` will not load the appropriate
+environment variables and you'll get a "Missing `secret_key_base` for
+'development' environment" error. Similarly, `heroku local` and `forego start`
+will fail to properly load `.env.local`.
+
+1. Open `localhost:5000` in a browser.
+
+## Setup Ngrok to Allow Webhooks
```
jcmorrow: It would be nice to start this section off with something like `If you want to contribute to X, Y, and Z features you should use ngrok to connect your development machine to github's webhooks`.

gylaz: Ngrok is necessary to receive pushes from GitHub via our web hook. If you'd like to test GitHub sending a pull request notification into the app, then GitHub must access your app from the outside. In such cases, Ngrok or similar, is required.

Feel free to reword ^.

CONTRIBUTING.md
```diff
@@ -25,19 +25,41 @@ Here are a few technical guidelines to follow:
...
+
+Ngrok allows GitHub to make requests via webhook to start a build. To get
+started with ngrok, sign up for an [ngrok] account and create a `~/.ngrok` file
+with the following:
 
     `auth_token: <your-token>`
 
-1. Launch ngrok with a custom subdomain on port 5000.
+1. Launch ngrok with a custom subdomain (note: this requires a paid ngrok
```
gylaz: I don't think a paid account is **required**, but it's certainly more convenient. With a free Ngrok account, a subdomain will be generated each time you run Ngrok, so you'll have to go and update your GitHub app settings.

gylaz:
Couple of thoughts. This looks good.

CONTRIBUTING.md
thorncp: > Ngrok is necessary to receive pushes from GitHub via _a_ web hook.

CONTRIBUTING.md
```diff
@@ -25,19 +25,41 @@ Here are a few technical guidelines to follow:
...
+   background job queue. **NOTE**: `rails server` will not load the appropriate
+environment variables and you'll get a "Missing `secret_key_base` for
+'development' environment" error. Similarly, `heroku local` and `forego start`
+will fail to properly load `.env.local`.
+
+1. Open `localhost:5000` in a browser.
+
+## Setup Ngrok to Allow Webhooks
+
+Ngrok allows GitHub to make requests via webhook to start a build. To get
```
thorncp: What do you think about making `To get started...` the first item in the list?

thorncp:
The new version of ngrok uses a different domain, ngrok.io, so I think all instances of `<your-initials>-hound.ngrok.com` should be changed to `<your-initials>-hound.ngrok.io`.

jcmorrow:
good catch @thorncp!

jcmorrow:
Maybe one last look at the final language (to make sure my rewording still makes sense)? @gylaz @thorncp

gylaz:
> The new version of ngrok uses a different domain, ngrok.io

Might be a good idea to update `.env` too.

jcmorrow:
Good call @gylaz, addressed in 5fa7d6b98654828936178b8a7afde9e3b8db3fa6

CONTRIBUTING.md
```diff
@@ -25,22 +25,56 @@ Here are a few technical guidelines to follow:
...
+   background job queue. **NOTE**: `rails server` will not load the appropriate
+environment variables and you'll get a "Missing `secret_key_base` for
+'development' environment" error. Similarly, `heroku local` and `forego start`
+will fail to properly load `.env.local`.
+
+1. Open `localhost:5000` in a browser.
+
+## Setup Ngrok to Allow Webhooks
+
+ Ngrok allows hound to receive webhooks from GitHub. If you'd like to develop or
```
gylaz: Is there a space before `Ngrok`?

Also, capitalize Hound?

CONTRIBUTING.md
```diff
@@ -25,22 +25,56 @@ Here are a few technical guidelines to follow:
...
+test a feature involving GitHub sending a pull request notification to your
+local hound server you'll need to have ngrok or something similar set up.
 
-    `auth_token: <your-token>`
+ To get started with ngrok, sign up for an [ngrok] account and configure ngrok
```
gylaz: Same here? 👀

jcmorrow: Yeah, this is a bad markdown habit of mine. Good catch.

CONTRIBUTING.md
```diff
@@ -25,22 +25,56 @@ Here are a few technical guidelines to follow:
...
+locally by installing ngrok and running:
 
-1. Launch ngrok with a custom subdomain on port 5000.
+    `ngrok authtoken <your-token>`
 
-    `ngrok -subdomain=<your-initials>-hound 5000`
+1. Launch ngrok on port 5000 (we recommend running ngrok with a custom subdomain
+   for easy and persistent configuration, but this requires a paid ngrok
```
gylaz: Why is this line indented but not the next?

CONTRIBUTING.md
```diff
@@ -25,22 +25,56 @@ Here are a few technical guidelines to follow:
...
+account. You can still run hound with a free ngrok account, but it will require
```
gylaz: Capitalize to `Hound` in the whole file?

gylaz:
LGTM.

CONTRIBUTING.md
jcmorrow: When I tell vim to reflow my bulleted list this is always what it does: it indents the second line only. I'm not sure why.

.env
```diff
@@ -6,7 +6,7 @@ DB_REAPING_FREQUENCY=10
 ENABLE_HTTPS=no
 GITHUB_CLIENT_ID=github_client_id
 GITHUB_CLIENT_SECRET=github_client_secret
-HOST=your-initials-hound.ngrok.com
+HOST=your-subdomain.ngrok.io
```
thorncp: :100:

CONTRIBUTING.md
thorncp: I recall having the same issue (I do think it's a bug), but I don't know what I did to fix it. I may be willing to do some sleuthing through my history to try to find out, if you're interested.

Do you mind manually indenting these in the meantime?

thorncp:
:+1: :tropical_fish:

CONTRIBUTING.md
jcmorrow: I did a bunch of googling towards the end of yesterday and didn't come up with a fix yet, but yeah in the meantime I'll take care of this.

PR #1141, opened 2016-05-24
Add ESLint to lists of linters

jcmorrow:
There are two places in the docs where we list our linters but leave out ESLint. I've added in references in the styles of the other linters.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,9 +120,13 @@ and services being used, and the default configuration for each linter.
  * [CoffeeLint](https://github.com/clutchski/coffeelint)
  * [config](https://raw.githubusercontent.com/houndci/hound/master/config/style_guides/coffeescript.json)
 
-1. JavaScript (JSHint)
- * [houndci/jshint](https://github.com/houndci/jshint)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -124,6 +124,10 @@ and services being used, and the default configuration for each linter.
  * [houndci/jshint](https://github.com/houndci/jshint)
  * [config](https://raw.githubusercontent.com/houndci/jshint/master/config/.jshintrc)
 
+1. ES6/JSX(ESLint)
```
gylaz: I'm not a fan of this heading, since ESLint can lint plain JavaScript, as well as JSHint can lint ES6. How about combining the tho linters under the JavaScript heading?

gylaz:
👍

CONTRIBUTING.md
jcmorrow: Yeah, that sounds right. I see we've been dealing with this in a couple of other places as well, and I agree that a JavaScript heading is useful.

jcmorrow: Addressed in 737493a541058acdd150584d262db38790d734ac

PR #1142, opened 2016-05-24
Track repo activation on the client side

jcmorrow:
In order to get correct statistics for our paid marketing efforts we
need conversion events to be tracked on the client side so that segment
can drop in the appropriate conversion pixels.

https://trello.com/c/sublWZZi

### DIFF
--- a/app/assets/javascripts/directives/repo.js.coffee.erb
+++ b/app/assets/javascripts/directives/repo.js.coffee.erb
@@ -6,6 +6,7 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription, 
       scope.processing = true
 
       scope.repo.$activate()
+        .then((repo) -> track_repo_activated(repo))
         .catch((response) -> alert(errorMessage(response.data)))
         .finally(-> scope.processing = false)
 
@@ -54,6 +55,7 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription, 
 
     saveSubscription = (subscription) ->
       subscription.$save().then((response) ->
+        track_repo_activated(response)
         scope.repo.active = true
         scope.repo.stripe_subscription_id = response.stripe_subscription_id
       ).catch(->
@@ -84,4 +86,14 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription, 
           deactivateRepo()
       else
         createSubscription()
+
+    track_repo_activated = (repo) ->
+      window.analytics.track(
+        "Repo Activated",
+        properties: {
+          name: repo.full_github_name,
+          private: repo.private,
+          revenue: repo.price_in_dollars
+        }
+      )
 ]
--- a/app/controllers/activations_controller.rb
+++ b/app/controllers/activations_controller.rb
@@ -6,7 +6,6 @@ class ActivationsController < ApplicationController
 
   def create
     if activator.activate
-      analytics.track_repo_activated(repo)
       render json: repo, status: :created
     else
       analytics.track_repo_activation_failed(repo)
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -6,8 +6,6 @@ class SubscriptionsController < ApplicationController
 
   def create
     if activator.activate && create_subscription
-      analytics.track_repo_activated(repo)
-
       render json: repo, status: :created
     else
       activator.deactivate
--- a/app/models/analytics.rb
+++ b/app/models/analytics.rb
@@ -20,17 +20,6 @@ class Analytics
     )
   end
 
-  def track_repo_activated(repo)
-    track(
-      event: "Repo Activated",
-      properties: {
-        name: repo.full_github_name,
-        private: repo.private,
-        revenue: repo.plan_price,
-      }
-    )
-  end
-
   def track_repo_deactivated(repo)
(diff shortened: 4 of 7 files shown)


app/assets/javascripts/directives/repo.js.coffee.erb
```diff
@@ -5,7 +5,7 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription,
     activateRepo = ->
       scope.processing = true
 
-      scope.repo.$activate()
+      scope.repo.$activate().then((repo) -> track_repo_activated(repo))
         .catch((response) -> alert(errorMessage(response.data)))
```
jcmorrow: Coffeescript's implicit return interacts weirdly with promise chaining sometimes. It's hard for me to tell if this `catch` will still work appropriately in all cases.

jcmorrow:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;sublWZZi&#x2F;760-move-the-hound-payment-analytics-event-client-side"> Move the Hound payment analytics event client-side</a>

app/assets/javascripts/directives/repo.js.coffee.erb
gylaz: Yes, `catch` should fire for all failures (even those that can happen in `then`'s success).

What do you think about bringing `.then` on a new line, like `catch`?

gylaz:
What is the value of moving this event tracking from server-side to client-side? /cc @croaky

app/assets/javascripts/directives/repo.js.coffee.erb
```diff
@@ -84,4 +85,14 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription,
...
       else
         createSubscription()
+
+    track_repo_activated = (repo) ->
+      window.analytics.track(
+        "Repo Activated",
+        properties: {
+          name: repo.full_github_name,
+          private: repo.private,
+          revenue: repo.plan_price
```
thorncp: Is `plan_price` available here? It looks like the API renders this as [`price_in_cents`](https://github.com/houndci/hound/blob/78684b7d4ae0285cb168bc5f0cb80e64b1acf9b3/app/serializers/repo_serializer.rb#L10).

croaky:
> What is the value of moving this event tracking from server-side to client-side? /cc @croaky

In the current state, we can't tell which channels (Twitter, Facebook, Google) or specific campaigns are performing well or poorly. This change fixes that, which enables us to iterate on paid marketing experiments, which should drive new users getting value from Hound.

croaky:
https://help.segment.com/hc/en-us/articles/205511905-Client-side-vs-server-side-event-tracking

app/assets/javascripts/directives/repo.js.coffee.erb
jcmorrow: I like the newline idea. I had it like that originally and moved it  to the same line to be consistent with line 56, but I'm going to change it back.

app/assets/javascripts/directives/repo.js.coffee.erb
jcmorrow: Good catch! I've addressed this in10dfa330495102ea33ae91ff351d73703ac10851

app/assets/javascripts/directives/repo.js.coffee.erb
```diff
@@ -84,4 +86,15 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription,
           deactivateRepo()
       else
         createSubscription()
+
+    track_repo_activated = (repo) ->
+      console.log(repo);
```
croaky: Would be cool if Hound could catch leftover debugging statements

jcmorrow: I support this idea!

app/assets/javascripts/directives/repo.js.coffee.erb
```diff
@@ -84,4 +86,14 @@ App.directive 'repo', ['Subscription', 'StripeCheckout', 'User', (Subscription,
...
       else
         createSubscription()
+
+    track_repo_activated = (repo) ->
+      window.analytics.track(
+        "Repo Activated",
+        properties: {
+          name: repo.full_github_name,
+          private: repo.private,
+          revenue: (repo.price_in_cents / 100).toFixed(2)
```
croaky: Might be nice to have a `repo.price_in_dollars` method but I don't feel strongly.

croaky:
I don't know much about Angular so if you wanted a second pair of eyes there, feel free. Code looks good to me!

app/assets/javascripts/directives/repo.js.coffee.erb
gylaz: Hmm, I have a vague recollection of CoffeeLint doing that. If so, I wonder why it didn't catch this.

jcmorrow:
I really wanted to add a JS spec for this, but it feels like a lot of work given how little I know of angular/jasmine. I did verify that the events are working via segment. What do you think @gylaz?

gylaz:
> I really wanted to add a JS spec for this, but it feels like a lot of work given how little I know of angular/jasmine. I did verify that the events are working via segment.
👍 We only have 1 or 2 JS tests anyways 😞

gylaz:
:shipit:

PR #1143, opened 2016-05-26
Add `disabled_for?` to `HoundConfig` and check if a config is disabled before checking whether it is enabled by default

jcmorrow:
Previously, `hound_config#enabled_for?` would return `true` for any
value that was not:

 - Explicitly set to false in the supplied config file
 - A language marked as beta

This didn't account for aliases and could result in someone disabling an alias but still running the actual linter (e.g. if someone disabled 'javascript', `jshint` would still run).

This adds an additional method `disabled_for?` and modifies both of these methods to **only** check their respective states (in other words, `enabled_for?` doesn't care if something is disabled, and
`disabled_for?` doesn't care if something is enabled. This allows `CheckEnabledLinter` to first check whether a particular config is disabled before it checks whether the config is explicity enabled or default.

Future work: 

 - Get `Linter::Collection::LINTER_NAMES` to return aliases as well
 - Allow linters to exclude other linters (e.g. if you enable `eslint` disable `jslint` automatically)


app/models/hound_config.rb
```diff
@@ -1,22 +1,26 @@
 class HoundConfig
   CONFIG_FILE = ".hound.yml"
-  BETA_LANGUAGES = %w(
+  BETA_LINTERS = %w(
```
jcmorrow: Addressed in 06b7f9405bec25df21a3fedaf84c1cd4b778e740

gylaz: We could add `# frozen_string_literal: true` to the top of the file.

app/models/hound_config.rb
```diff
@@ -55,11 +59,11 @@ def options_for(name)
...
 
   def default_options_for(name)
-    { "enabled" => !beta?(name) }
+    { "enabled" => default?(name) }
   end
 
-  def beta?(name)
-    BETA_LANGUAGES.include?(name)
+  def default?(name)
+    DEFAULT_LINTERS.include? name
```
gylaz: Rather than having another array, can we leverage the `Linter::Collection::LINTERS` array? This would also allow us to use the aliases, and check those.

gylaz: To be clear, I'm proposing this instead of `CONFIGURABLE_LINTERS`.

jcmorrow:
There's an additional question around things that are explicitly disabled via aliases (e.g. the currently failing spec). The way the spec if written indicates that `hound_config` is responsible for understanding aliases, but actually it doesn't contain any alias logic. Instead all of that logic is in the linter configs themselves (the only linters with aliases right now are `config/jshint` and `config/coffee_script`. It seems like the specific config file should actually be responsible for checking on whether an alias is explicitly disabled. Thoughts?

spec/models/hound_config_spec.rb
```diff
@@ -54,13 +64,13 @@
...
         commit = stubbed_commit(
           ".hound.yml" => <<-EOS.strip_heredoc
-            javascript:
+            jshint:
               enabled: false
           EOS
         )
         hound_config = HoundConfig.new(commit)
 
-        expect(hound_config).not_to be_enabled_for("jshint")
```
gylaz: Is it worth asserting this too, to be sure that `jshint` is also disabled?

app/models/hound_config.rb
jcmorrow: I really like this idea! I'll try it out.

spec/models/hound_config_spec.rb
jcmorrow: Oh good catch, this change is in the wrong place. This is testing for aliases when it should just be testing what is explicitly set.

app/models/hound_config.rb
jcmorrow: It came out alright, but I've left out the aliases part, partially because I'm concerned about how complex that makes composing this constant and also because of my comment below.

app/models/linter/collection.rb
```diff
@@ -13,6 +13,9 @@ class Collection
       Linter::Scss,
       Linter::Swift,
     ].freeze
+    LINTER_NAMES = Linter::Collection::LINTERS.map do |linter_class|
+      linter_class.name.demodulize.downcase
+    end
```
jcmorrow: I moved this into here because it's logically related, but it's not used in this class, so that's a downside. I considered a class method on `HoundConfig` but that felt bad because it was a class method that just reached out to another class's constants. Thoughts?

app/models/hound_config.rb
jcmorrow: Addressed in d938474144bc80dd0f4e15332a0db40d60112b16.

app/services/check_enabled_linter.rb
```diff
@@ -8,6 +8,10 @@ def initialize(*configs)
   end
 
   def enabled?
+    if disabled?
+      return false
```
thorncp: What do you think about refactoring this to not require an early return?

An idea could be to extract the `any?` loop below to its own method, and having this method be something like `none_disabled? && any_enabled?`

jcmorrow: I like it. 👍

jcmorrow: Done in 0b3ce43a96c6c4e3adcbc02dc719bcaa3f3e2c15

app/models/hound_config.rb
```diff
@@ -38,28 +38,20 @@ def parse(file_content)
...
-
   def enabled?(config)
     config["enabled"] || config["Enabled"]
   end
 
-  def options_for(name)
-    config = content[name] || {}
-    default_options_for(name).merge(config)
+  def disabled?(config)
+    config["enabled"] == false || config["Enabled"] == false
```
thorncp: Could this be `!config["enabled"] || !config["Enabled"]` ?

thorncp: Oh, this checking if the keys are present _and_ they're `false`? If so, it might be a good move to be more explicit with that.

app/models/linter/collection.rb
```diff
@@ -13,6 +13,9 @@ class Collection
       Linter::Scss,
       Linter::Swift,
     ].freeze
+    LINTER_NAMES = Linter::Collection::LINTERS.map do |linter_class|
```
thorncp: Since `LINTERS` is defined above, we don't need the namespace, e.g.  `LINTER_NAMES = LINTERS.map ...`

app/models/hound_config.rb
jcmorrow: Indeed. If they are null that's not good enough, they have to be actively set to false. I can add a presence check to make that more clear.

app/models/linter/collection.rb
thorncp: I don't think I have any qualms with it here.

spec/models/config/ruby_spec.rb
```diff
@@ -1,4 +1,17 @@
 require "spec_helper"
+require "app/models/linter/base"
+require "app/models/linter/coffee_script"
```
thorncp: Are all of these required? Since this is the spec for `Config::Ruby`, I'd expect the other linters wouldn't need to be loaded.

jcmorrow: Yeah, it because they're required for `Linter::Collection` and we're using that one constant from there. I really feel like that constant doesn't go there but I don't know where it **does** go. 😖

app/models/linter/collection.rb
jcmorrow: Done in d29bc08d54ce78e9bf2ad1eac5bb4fd8dac574bc

app/models/hound_config.rb
jcmorrow: Addressed in d29bc08d54ce78e9bf2ad1eac5bb4fd8dac574bc

app/models/hound_config.rb
```diff
@@ -1,31 +1,31 @@
...
+
+    disabled?(config)
+  end
+
   def enabled_for?(name)
-    configured?(name)
+    key = normalize_key(name)
+    config = options_for(key)
+
+    enabled?(config) || default?(key)
```
jackvnimble: I may be looking at this wrong but wouldn't this always be true if a linter is non-beta? I guess it's technically correct since you check both this and `disabled_for?` using `&&g in the `CheckEnabledLinter` object but it does seem odd. 

What if `enabled_for` represented something like `!disabled || (beta && enabled?)` then could we get rid of `disabled_for?` and rely only on `enabled_for?`?

spec/models/config/ruby_spec.rb
jackvnimble: Or we could put this in another file and require it? Maybe both? Idk

jcmorrow: I like that `require` better at the very least.

app/models/hound_config.rb
jcmorrow: The way I see it we always need `enabled_for?` and `disabled_for?` because if we have just `enabled_for?` we can't tell if something is explicitly disabled or just an alias. So for example if you disabled jshint by putting `javascript: enabled: false` in your config there is no difference in the way the API responds (because javascript isn't a default, it returns false from `enabled_for?` whether it's explicitly disabled or just not in the config). Hopefully that makes sense, it's hard to explain via interwebz 🎹

spec/models/config/ruby_spec.rb
thorncp: At this point my preference would be to load Rails to avoid the burden of maintaining all those `require`s.

app/models/hound_config.rb
jackvnimble: Haha definitely hard to explain over the interwebz (and not for that matter). Think I'm following you now after looking more closely at `CheckEnabledLinter`.

spec/models/config/ruby_spec.rb
teoljungberg: Sounds like maybe a `app/models/linter.rb` is needed that requires _all_ the linters?

app/models/hound_config.rb
```diff
@@ -38,28 +38,28 @@ def parse(file_content)
...
-  def enabled?(config)
-    config["enabled"] || config["Enabled"]
+  def config_set?(config)
+    !config["enabled"].nil? || !config["Enabled"].nil?
   end
 
-  def options_for(name)
-    config = content[name] || {}
-    default_options_for(name).merge(config)
+  def config_false?(config)
```
gylaz: Ugh, really not liking this name. I'm also getting very confused in the sea of `enabled?`, `enabled_for?` and `config_false?`. This points to a smell.

How is this method different from `enabled?`? Can we combine the two methods? Perhaps using `fetch`?

app/models/hound_config.rb
```diff
@@ -38,28 +38,28 @@ def parse(file_content)
...
+    config["enabled"] == false || config["Enabled"] == false
   end
 
-  def default_options_for(name)
-    { "enabled" => !beta?(name) }
+  def options_for(name)
+    content[name] || {}
```
gylaz: `content.fetch(name, {})`?

app/models/hound_config.rb
gylaz: What if we flipped this logic to `default?(key) && !disabled?(config)`? Would it make code easier to reason about?

app/models/hound_config.rb
```diff
@@ -38,28 +38,28 @@ def parse(file_content)
     Config::Parser.yaml(file_content) || {}
   end
 
-  def configured?(name)
-    key = normalize_key(name)
-    config = options_for(key)
+  def enabled?(config)
```
gylaz: Perhaps, we can call this method `enabled_key_present?`?

app/models/hound_config.rb
```diff
@@ -38,28 +38,28 @@ def parse(file_content)
...
 
-    enabled?(config)
+  def disabled?(config)
+    config_set?(config) && config_false?(config)
   end
 
-  def enabled?(config)
-    config["enabled"] || config["Enabled"]
+  def config_set?(config)
+    !config["enabled"].nil? || !config["Enabled"].nil?
```
gylaz: Once again, this method seems to be doing something very similar to `enabled?`.

app/models/hound_config.rb
jcmorrow: Yeah, naming is the hard thing here. I'll take another shot.

app/services/check_enabled_linter.rb
```diff
@@ -8,14 +8,26 @@ def initialize(*configs)
   end
 
   def enabled?
+    !any_disabled? && any_enabled?
+  end
+
+  private
+
+  def any_disabled?
```
gylaz: Thoughts on renaming this to `any_alias_disabled?`?

app/services/check_enabled_linter.rb
```diff
@@ -8,14 +8,26 @@ def initialize(*configs)
   end
 
   def enabled?
+    !any_disabled? && any_enabled?
```
gylaz: This is likely a dead horse and I'm beating it, but why can't we just ask whether any linter alias is enabled, then do the appropriate checking logic like "both default and not disabled, or enabled": `(default? && !disabled?) || enabled?`

jcmorrow: Hmmm... I do think if we move the `default?` logic into this service object things will read better. I'll give it a shot.

jcmorrow:
@gylaz I'm really not sure I made it any better 😖

app/models/hound_config.rb
```diff
@@ -1,31 +1,24 @@
...
 
   def content
     @content ||= parse(commit.file_content(CONFIG_FILE))
   end
 
+  def disabled_for?(name)
+    key = normalize_key(name)
+    config = options_for(key)
+
+    config_enabled(config) === false
```
gylaz: Why triple equals?

app/models/hound_config.rb
```diff
@@ -1,31 +1,24 @@
...
+  end
+
   def enabled_for?(name)
-    configured?(name)
+    key = normalize_key(name)
+    config = options_for(key)
+
+    config_enabled(config)
```
gylaz: Should we use a `!!` prefix to ensure this returns a boolean?

app/models/linter/collection.rb
```diff
@@ -13,6 +13,9 @@ class Collection
       Linter::Scss,
       Linter::Swift,
     ].freeze
+    LINTER_NAMES = LINTERS.map do |linter_class|
+      linter_class.name.demodulize.downcase
+    end
```
gylaz: Does it make sense to move this into `app/services/check_enabled_linter.rb`? Does anyone else use it?

app/services/check_enabled_linter.rb
```diff
@@ -8,14 +16,32 @@ def initialize(*configs)
   end
 
   def enabled?
+    !any_alias_disabled? && (any_alias_enabled? || any_alias_default?)
```
gylaz: I'm actually liking this better!

app/services/check_enabled_linter.rb
```diff
@@ -8,14 +16,32 @@ def initialize(*configs)
...
+  end
+
+  private
+
+  def any_alias_disabled?
+    linter_names.any? do |linter_name|
+      hound_configs.any? do |hound_config|
+        hound_config.disabled_for?(linter_name)
```
gylaz: I'm still a bit confused on why this needs a separate method. Why can't we use `enabled_for?`?

app/models/hound_config.rb
```diff
@@ -1,31 +1,24 @@
...
   end
 
+  def disabled_for?(name)
+    key = normalize_key(name)
+    config = options_for(key)
+
+    config_enabled(config) === false
+  end
+
   def enabled_for?(name)
```
gylaz: What if this method was called `linter_alias_enabled?`, to make it more obvious?

app/models/linter/collection.rb
jcmorrow: It's also used to check if a config is legacy in `app/models/config/ruby.rb`

app/services/check_enabled_linter.rb
jcmorrow: Because we're using a boolean to represent three states:

1. Something can be explicitly enabled
1. Something can be explicitly disabled
1. Something can be not explicitly set (use the defaults)

If we only have one method (`enabled_for?`) then we can't distinguish between the second two states.

app/models/hound_config.rb
jcmorrow: Yeah, that's a good idea here!

app/models/hound_config.rb
jcmorrow: Because nil is falsey, and in this case, it's only disabled if it is **exactly** false.

app/models/hound_config.rb
jcmorrow: yeah, there's a lot of options here: `linter_enabled?`, `linter_name_enabled?`, `linter_alias_disabled?`. The last sounds fine to me, and I think all of them are more descriptive than `enabled_for?`.

app/models/hound_config.rb
teoljungberg: That sounds wrong to me - we should handle `nil` and `false` equally.

app/models/hound_config.rb
teoljungberg: I like `linter_alias_{enabled,disabled}?`

spec/models/config/ruby_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "spec_helper"
+require "spec/support/linters"
```
teoljungberg: I like this abstraction.

jcmorrow:
I'm going to close this in favor of: https://github.com/houndci/hound/pull/1149

PR #1144, opened 2016-05-26
Calculate config names from their class names

jcmorrow:
Previously you would create a `Ruby` config by doing:

    Config::Ruby.new(hound_config, "ruby")

This seems unnecessary since the information that it is a ruby linter config is right in the class name. It also implies that you might want to do something like:

    Config::Ruby.new(hound_config, "python")

But this just changes the way the object calculates `linter_names`, it doesn't actually provide additional  functionality that the app is using.

Besides making the API for creating configs simpler, this change allows us to more easily fetch aliases from a type of config, since `linter_names` can become a class method.

### DIFF
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -1,6 +1,6 @@
 module Config
   class Base
-    attr_reader_initialize :hound_config, :linter_name
+    attr_reader_initialize :hound_config
 
     def content
       @content ||= ensure_correct_type(safe_parse(load_content))
@@ -21,6 +21,10 @@ module Config
     private
 
     attr_implement :parse, [:file_content]
+
+    def linter_name
+      self.class.name.demodulize.underscore
+    end
 
     def safe_parse(content)
       parse(content)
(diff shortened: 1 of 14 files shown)


app/models/config/base.rb
```diff
@@ -22,6 +22,10 @@ def serialize(data = content)
 
     attr_implement :parse, [:file_content]
 
+    def name
```
gylaz: How about still keeping this as `linter_name`? Benefits? Less diff noise, and `config.linter_name` is a bit more self-explanatory.

spec/models/config/base_spec.rb
```diff
@@ -20,7 +20,12 @@ def parse(file_content)
   describe "#content" do
     context "when there is no config content for the given linter" do
       it "does not raise" do
-        config = build_config(linter_name: "unconfigured_linter")
+        hound_config = double(
+          "HoundConfig",
+          commit: double("Commit", file_content: ""),
```
gylaz: How about `instance_double` here and line 23?

gylaz:
Nice cleanup! 🎂

app/models/config/base.rb
jcmorrow: I like that idea.

teoljungberg:
For future reference, this was merged as https://github.com/houndci/hound/commit/66a336a4e83b4864dd830dfb9647edf8298dbcf6

teoljungberg:
Nice cleanup Josh!

PR #1147, opened 2016-06-01
RubyConfigParser should return a Config, not a Hash, on error

jcmorrow:
When hound rescues from a `TypeError` right now it is returning a `Hash`, but if the operation had succeeded it would have returned a `RuboCop::Config`.

Change: return the default `RuboCop::Config` in `rescue` block.

See: https://github.com/houndci/hound/issues/1146 and a reproduced failure: 
failure

### DIFF
--- a/app/services/ruby_config_builder.rb
+++ b/app/services/ruby_config_builder.rb
@@ -21,7 +21,7 @@ class RubyConfigBuilder
   def combined_overrides
     RuboCop::ConfigLoader.merge(normalized_hound_config, normalized_overrides)
   rescue TypeError
-    hound_config
+    normalized_hound_config
   end
 
   def normalized_overrides
--- a/spec/services/ruby_config_builder_spec.rb
+++ b/spec/services/ruby_config_builder_spec.rb
@@ -1,6 +1,7 @@
 require "rubocop"
 require "app/models/default_config_file"
 require "app/services/ruby_config_builder"
+require "app/models/config/parser"
 
 RSpec.describe RubyConfigBuilder do
   context "when there is no custom configuration" do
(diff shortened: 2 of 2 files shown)


app/models/config/ruby.rb
```diff
@@ -23,7 +23,13 @@ def parse_inherit_from(config)
         result.merge(ancestor_config)
       end
 
-      inherited_config.merge(config.except("inherit_from"))
+      merged_config = inherited_config.merge(config.except("inherit_from"))
+
+      if(merged_config.has_key?("inherit_from"))
```
teoljungberg: This isn't javascript, josh!

spec/models/config/ruby_spec.rb
```diff
@@ -115,6 +115,43 @@
         expect { config.content }.to raise_error(Config::ParserError)
       end
     end
+    context "with nested inherit_froms" do
```
teoljungberg: Could we add a newline above here?

spec/models/config/ruby_spec.rb
```diff
@@ -115,6 +115,43 @@
...
+      it "returns the merged configuration using `inherit_from`" do
+        rubocop = <<-EOS.strip_heredoc
```
teoljungberg: Instead of using `strip_heredoc` -  what do you think of using the squiggly heredoc instead `<<~EOS` to achieve the same thing.

app/models/config/ruby.rb
jcmorrow: Don't tell me when I can and can't write javascript!

spec/models/config/ruby_spec.rb
jcmorrow: I like it! I also forget that it exists. Kind of a lot.

teoljungberg:
Solid!

jcmorrow:
@teoljungberg sorry to crush your earlier comments with a force push, but I realized I was fixing two different things in the same PR and wanted to just get this merged before fixing the other issue (https://github.com/houndci/hound/pull/1148)

gylaz:
👍

PR #1148, opened 2016-06-02
WIP: Rubocop config with nested inherit_froms

jcmorrow:
Hound can only follow one `inherit_from` configuration. [Rubocop supports nested `inherit_from`s](http://rubocop.readthedocs.io/en/latest/configuration/#inheritance), and we can do the same by recursively parsing the inherited file until we no longer can find any `inherited_from` statement in the config. 

Still to do: get this to work with relative file paths and possibly even URL's

### DIFF
--- a/app/models/config/ruby.rb
+++ b/app/models/config/ruby.rb
@@ -23,7 +23,13 @@ module Config
         result.merge(ancestor_config)
       end
 
-      inherited_config.merge(config.except("inherit_from"))
+      merged_config = inherited_config.merge(config.except("inherit_from"))
+
+      if merged_config.has_key?("inherit_from")
+        parse_inherit_from(merged_config)
+      else
+        merged_config
+      end
     end
 
     def legacy?
--- a/app/models/config/ruby.rb
+++ b/app/models/config/ruby.rb
@@ -14,19 +14,24 @@ module Config
       Parser.yaml(file_content)
     end
 
-    def parse_inherit_from(config)
+    def parse_inherit_from(config, parsed_files = [])
       inherit_from = Array(config.fetch("inherit_from", []))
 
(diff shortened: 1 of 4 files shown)


app/models/config/ruby.rb
```diff
@@ -23,7 +23,13 @@ def parse_inherit_from(config)
         result.merge(ancestor_config)
       end
 
-      inherited_config.merge(config.except("inherit_from"))
+      merged_config = inherited_config.merge(config.except("inherit_from"))
+
+      if merged_config.has_key?("inherit_from")
```
gylaz: Can that inherited file have more `inherit_from`? Should we loop? Do we say: "Nay, we won't support more than one nested level?"

jcmorrow: Ideally, we support what rubocop supports. So, if something works locally running rubocop, it would be nice if it worked on hound as well. This should work **almost** exactly like rubocop, except that it doesn't follow relative file paths. I have a plan for that, but first I want to investigate how rubocop does it and see if I can essentially export this job to them at least somewhat.

gylaz: My theory is that you can infinitely nest `inherit_from` in RuboCop, but I did not test this theory.

jcmorrow: Yeah, and that should also be possible with this code, with the exception of the relative filenames piece.

thorncp: Are we doing anything to detect circular dependencies?

jcmorrow: No 😨

thorncp: 🌀

jcmorrow:
I'm not sure what the right way forward on this. Closing for now.

PR #1149, opened 2016-06-02
HoundConfig merges default and user configurations

jcmorrow:
Previously, `hound_config#enabled_for?` would return `true` for any value that was not:

 - Explicitly set to false in the supplied config file
 - A language marked as beta

Additionally, HoundConfig didn't account for aliases. This could result in someone disabling an alias (e.g. `javascript`) but still running the actual linter (e.g. `jshint`). Instead of trying to perform boolean logic about what is and is not activated, this strategy just starts with a default configuration and merges the user-specified configuration into it using a series of transforms.

Thanks to @jackvnimble for the idea!

### DIFF
--- a/app/models/hound_config.rb
+++ b/app/models/hound_config.rb
@@ -12,48 +12,45 @@ class HoundConfig
   attr_reader_initialize :commit
 
   def content
-    @content ||= parse(commit.file_content(CONFIG_FILE))
+    @_content ||= default_config.deep_merge(resolved_aliases_config)
   end
 
   def linter_enabled?(name)
-    configured?(name)
+    key = name.downcase
+    config = options_for(key)
+
+    !!config["enabled"]
   end
 
   def fail_on_violations?
-    !!(content["fail_on_violations"])
+    !!content["fail_on_violations"]
   end
 
   private
+
+  def default_config
+    Linter::Collection.linter_names.each.with_object({}) do |name, config|
+      config[name] = { "enabled" => !BETA_LINTERS.include?(name) }
+    end
+  end
+
+  def resolved_aliases_config
+    ResolveConfigAliases.run(normalized_config)
+  end
+
+  def normalized_config
+    NormalizeConfig.run(parsed_config)
+  end
+
+  def parsed_config
+    parse(commit.file_content(CONFIG_FILE))
+  end
 
   def parse(file_content)
     Config::Parser.yaml(file_content) || {}
   end
 
-  def configured?(name)
-    key = normalize_key(name)
-    config = options_for(key)
-
-    enabled?(config)
-  end
-
-  def enabled?(config)
-    config["enabled"] || config["Enabled"]
-  end
-
   def options_for(name)
-    config = content[name] || {}
-    default_options_for(name).merge(config)
-  end
-
-  def default_options_for(name)
-    { "enabled" => !beta?(name) }
-  end
-
-  def beta?(name)
-    BETA_LINTERS.include?(name)
-  end
-
-  def normalize_key(key)
-    key.downcase.sub("_", "")
+    content.fetch(name, {})
   end
 end
--- a/app/services/normalize_config.rb
+++ b/app/services/normalize_config.rb
@@ -1,0 +1,27 @@
+class NormalizeConfig
+  def self.run(config)
+    new(config).run
+  end
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.reduce({}) do |normalized_config, (key, value)|
+      normalized_key = normalize_key(key)
+      if value.is_a? Hash
+        normalized_config[normalized_key] = NormalizeConfig.run(value)
+      else
+        normalized_config[normalized_key] = value
+      end
+      normalized_config
+    end
+  end
+
+  private
+
+  def normalize_key(key)
+    key.downcase
+  end
+end
--- a/app/services/resolve_config_aliases.rb
+++ b/app/services/resolve_config_aliases.rb
@@ -1,0 +1,26 @@
+class ResolveConfigAliases
+  ALIASES = {
+    "javascript" => "jshint",
+    "java_script" => "jshint",
+    "coffeescript" => "coffee_script",
+  }.freeze
+
+  def self.run(config)
+    new(config).run
+  end
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.reduce({}) do |resolved_config, (key, value)|
+      if ALIASES.keys.include? key
+        resolved_config[ALIASES[key]] = value
+      else
+        resolved_config[key] = value
+      end
+      resolved_config
+    end
+  end
+end
--- a/spec/models/hound_config_spec.rb
+++ b/spec/models/hound_config_spec.rb
@@ -1,25 +1,23 @@
 require "spec_helper"
-require "spec/support/linters"
 require "app/models/config/parser"
+require "app/services/normalize_config"
+require "app/services/resolve_config_aliases"
 require "app/models/hound_config"
 
 describe HoundConfig do
   describe "#content" do
-    it "returns the content of the .hound.yml file" do
+    it "returns the contents of .hound.yml, merged with the default config" do
       commit = stubbed_commit(
         ".hound.yml" => <<-EOS.strip_heredoc
           ruby:
-            enabled: true
             config_file: config/rubocop.yml
         EOS
       )
       hound_config = HoundConfig.new(commit)
 
-      expect(hound_config.content).to eq(
-        "ruby" => {
-          "enabled" => true,
-          "config_file" => "config/rubocop.yml",
-        },
+      expect(hound_config.content["ruby"]).to eq(
+        "enabled" => true,
+        "config_file" => "config/rubocop.yml",
       )
     end
   end
@@ -42,13 +40,27 @@ describe HoundConfig do
       it "returns false" do
         commit = stubbed_commit(
           ".hound.yml" => <<-EOS.strip_heredoc
-            scss:
+            ruby:
               enabled: false
           EOS
         )
         hound_config = HoundConfig.new(commit)
 
-        expect(hound_config).not_to be_enabled_for("scss")
+        expect(hound_config).not_to be_enabled_for("ruby")
+      end
+    end
+
+    context "when the given language is enabled" do
+      it "returns true" do
+        commit = stubbed_commit(
+          ".hound.yml" => <<-EOS.strip_heredoc
+            remark:
+              enabled: true
+          EOS
+        )
+        hound_config = HoundConfig.new(commit)
+
+        expect(hound_config).to be_enabled_for("remark")
       end
     end
 
@@ -56,7 +68,7 @@ describe HoundConfig do
       it "returns false" do
         commit = stubbed_commit(
           ".hound.yml" => <<-EOS.strip_heredoc
-            javascript:
+            java_script:
               enabled: false
           EOS
         )
@@ -66,75 +78,41 @@ describe HoundConfig do
       end
     end
 
-    context "when the given language is supported but unconfigured" do
+    context "given an unsupported language" do
+      it "returns false" do
+        commit = stubbed_commit(".hound.yml" => "")
+        hound_config = HoundConfig.new(commit)
+        unsupported_linter = "some_random_linter"
+
+        expect(hound_config).not_to be_enabled_for(unsupported_linter)
+      end
+    end
+
+    context "when the enabled key is capitalized" do
       it "returns true" do
         commit = stubbed_commit(
           ".hound.yml" => <<-EOS.strip_heredoc
-            scss:
-              config_file: config/.scss_lint.yml
+            python:
+              Enabled: true
           EOS
         )
         hound_config = HoundConfig.new(commit)
 
-        expect(hound_config).to be_enabled_for("scss")
+        expect(hound_config).to be_enabled_for("python")
       end
     end
 
-    context "given a language in beta" do
-      context "when the given language is enabled" do
-        it "returns true" do
-          commit = stubbed_commit(
-            ".hound.yml" => <<-EOS.strip_heredoc
-              python:
-                enabled: true
-            EOS
-          )
-          hound_config = HoundConfig.new(commit)
+    context "when the given language has underscores in it" do
+      it "converts them and returns true" do
+        commit = stubbed_commit(
+          ".hound.yml" => <<-EOS.strip_heredoc
+            coffeescript:
+              enabled: true
+          EOS
+        )
+        hound_config = HoundConfig.new(commit)
 
-          expect(hound_config).to be_enabled_for("python")
-        end
-      end
-
-      context "when the enabled key is capitalized" do
-        it "returns true" do
-          commit = stubbed_commit(
-            ".hound.yml" => <<-EOS.strip_heredoc
-              python:
-                Enabled: true
-            EOS
-          )
-          hound_config = HoundConfig.new(commit)
-
-          expect(hound_config).to be_enabled_for("python")
-        end
-      end
-
-      context "when the given language has underscores in it" do
-        it "converts them and returns true" do
-          commit = stubbed_commit(
-            ".hound.yml" => <<-EOS.strip_heredoc
-              coffeescript:
-                enabled: true
-            EOS
-          )
-          hound_config = HoundConfig.new(commit)
-
-          expect(hound_config).to be_enabled_for("coffee_script")
-        end
-      end
-
-      context "when the given language is disabled" do
-        it "returns false" do
-          commit = stubbed_commit(
-            ".hound.yml" => <<-EOS.strip_heredoc
-              python:
-                enabled: false
-            EOS
-          )
-          hound_config = HoundConfig.new(commit)
-
-          expect(hound_config).not_to be_enabled_for("python")
-        end
+        expect(hound_config).to be_enabled_for("coffee_script")
       end
     end
   end
--- a/spec/services/normalize_config_spec.rb
+++ b/spec/services/normalize_config_spec.rb
@@ -1,0 +1,15 @@
+require "app/services/normalize_config"
+
+describe NormalizeConfig do
+  describe "#run" do
+    context "given a hash with keys containing capital letters" do
+      it "downcases the keys" do
+        config = { "Ruby" => { "Enabled" => true } }
+
+        expect(NormalizeConfig.run(config)).to eq(
+          "ruby" => { "enabled" => true },
+        )
+      end
+    end
+  end
+end
--- a/spec/services/resolve_config_aliases_spec.rb
+++ b/spec/services/resolve_config_aliases_spec.rb
@@ -1,0 +1,17 @@
+require "app/services/resolve_config_aliases"
+
+describe ResolveConfigAliases do
+  describe "#run" do
+    context "when the config contains aliases" do
+      it "renames them to the appropriate linter" do
+        config = {
+          "javascript" => { "enabled" => false },
+          "ruby" => { "enabled" => false },
+        }
+
+        expect(ResolveConfigAliases.run(config).keys).
+          to match_array(["jshint", "ruby"])
+      end
+    end
+  end
+end
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -10,21 +10,17 @@ module Config
       []
     end
 
-    def linter_names
-      [linter_name]
-    end
-
     def serialize(data = content)
       data
+    end
+
+    def linter_name
+      self.class.name.demodulize.underscore
     end
 
     private
 
     attr_implement :parse, [:file_content]
-
-    def linter_name
-      self.class.name.demodulize.underscore
-    end
 
     def safe_parse(content)
       parse(content)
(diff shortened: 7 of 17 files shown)


app/services/config_alias_resolver.rb
```diff
@@ -0,0 +1,31 @@
...
+  ALIASES = {
+    jshint: [:javascript, :java_script],
+    coffeescript: [:coffee_script],
+  }
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
```
jcmorrow: I agree hound! 🐶

app/services/config_alias_resolver.rb
```diff
@@ -0,0 +1,25 @@
+class ConfigAliasResolver
+  ALIASES = {
+    "javascript" => "jshint",
+    "java_script" => "jshint",
+    "coffee_script" => "coffeescript"
```
gylaz: Listen to Da Hound!

app/models/hound_config.rb
```diff
@@ -1,31 +1,30 @@
...
+    "go" => { "enabled" => true },
+    "haml" => { "enabled" => true },
+    "jscs" => { "enabled" => false },
+    "jshint" => { "enabled" => true },
+    "remark" => { "enabled" => false },
+    "python" => { "enabled" => false },
+    "ruby" => { "enabled" => true },
+    "scss" => { "enabled" => true },
+    "swift" => { "enabled" => true },
+  }.freeze
```
gylaz: Here's an idea: Since the linter names can change (yet again). How about generating a default config from the list of `LINTERS` and setting enabled to `true` (or `false` if in `BETA`)?

app/models/hound_config.rb
```diff
@@ -1,31 +1,30 @@
...
 
   attr_reader_initialize :commit
 
   def content
-    @content ||= parse(commit.file_content(CONFIG_FILE))
+    merged_config
```
gylaz: Any value in having this method call another rather than just inlining the logic?

app/models/hound_config.rb
```diff
@@ -34,32 +33,38 @@ def fail_on_violations?
 
   private
 
-  def parse(file_content)
-    Config::Parser.yaml(file_content) || {}
+  def resolved_aliases_config
+    @resolved_aliases_config ||= ConfigAliasResolver.new(normalized_config).run
```
gylaz: Probably don't need to memoize this, since we memoize right after this is called.

app/models/hound_config.rb
```diff
@@ -34,32 +33,38 @@ def fail_on_violations?
...
   end
 
-  def configured?(name)
-    key = normalize_key(name)
-    config = options_for(key)
+  def normalized_config
+    @normalized_config ||= ConfigNormalizer.new(parsed_config).run
```
gylaz: Same here.

app/models/hound_config.rb
jcmorrow: Good call, I was moving things around and definitely got carried away with the memoizing.

app/models/hound_config.rb
jcmorrow: Nope, it used to do more and now it doesn't, inlining

app/models/hound_config.rb
```diff
@@ -34,32 +33,38 @@ def fail_on_violations?
...
-  def options_for(name)
-    config = content[name] || {}
-    default_options_for(name).merge(config)
+  def merge(base_hash, derived_hash)
+    result = base_hash.merge(derived_hash)
+    keys_appearing_in_both = base_hash.keys & derived_hash.keys
+    keys_appearing_in_both.each do |key|
+      next unless base_hash[key].is_a?(Hash)
+      result[key] = merge(base_hash[key], derived_hash[key])
+    end
```
gylaz: I'm having a hard time parsing this (and why we need this). How about extracting this into an intention revealing method?

app/services/config_alias_resolver.rb
```diff
@@ -0,0 +1,25 @@
+class ConfigAliasResolver
```
gylaz: How about naming this service after what it does `ResolveConfigAliases`}

app/services/config_alias_resolver.rb
```diff
@@ -0,0 +1,25 @@
...
+  }.freeze
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    config_with_aliases_resolved = {}
+
+    @config.each_pair do |config_key, config_value|
```
gylaz: Using `each` is more idiomatic: `@config.each do |key, value|`

gylaz: Even better would be to use `reduce` or `each_with_object`.

app/services/config_normalizer.rb
```diff
@@ -0,0 +1,27 @@
+class ConfigNormalizer
```
gylaz: `NormalizeConfig`?

jcmorrow: Oh yeah, I guess we name them like tasks rather than nouns don't we? Sounds good to me.

app/services/config_alias_resolver.rb
jcmorrow: Yeah I just switched something else from `each` to `inject` and had the same thought, these things are basically supposed to be functional so those techniques make more sense here.

app/models/hound_config.rb
jcmorrow: yeah, it should be called like `deep_merge` or something like that. I'm not sure if it can get much more broken up due to the recursive nature, but I can take a shot.

app/services/resolve_config_aliases.rb
```diff
@@ -0,0 +1,22 @@
...
+    "java_script" => "jshint",
+    "coffee_script" => "coffeescript",
+  }.freeze
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.reduce({}) do |config_without_aliases, (config_key, config_value)|
```
gylaz: This is actually wrong, according to thoughtbot styleguide. I didn't realize this but since we moved this under `houndci` org, we no longer get the thoughtbot rules.

app/models/hound_config.rb
```diff
@@ -34,32 +31,34 @@ def fail_on_violations?
...
   end
 
-  def options_for(name)
-    config = content[name] || {}
-    default_options_for(name).merge(config)
+  def merge(base_hash, derived_hash)
+    result = base_hash.merge(derived_hash)
+    keys_appearing_in_both = base_hash.keys & derived_hash.keys
+    keys_appearing_in_both.each do |key|
+      next unless base_hash[key].is_a?(Hash)
```
teoljungberg: `next` inside of an `each` loop is not a ruby convention - try using a `select` or `reject` first to limit the data set.

teoljungberg:
@jcmorrow This PR seems to use work that was done in #1143 - which makes the diff quite confusing to read.

app/models/hound_config.rb
teoljungberg: This could also be refactored into a `reduce`.

app/models/hound_config.rb
```diff
@@ -34,32 +31,34 @@ def fail_on_violations?
 
   private
 
-  def parse(file_content)
-    Config::Parser.yaml(file_content) || {}
+  def resolved_aliases_config
+    ResolveConfigAliases.new(normalized_config).run
```
teoljungberg: What do you think about using a top level `.run` method? Which is a convention in this project.

app/models/hound_config.rb
```diff
@@ -34,32 +31,34 @@ def fail_on_violations?
...
   end
 
-  def configured?(name)
-    key = normalize_key(name)
-    config = options_for(key)
-
-    enabled?(config)
+  def normalized_config
+    NormalizeConfig.new(parsed_config).run
```
teoljungberg: Same as above.

app/services/normalize_config.rb
```diff
@@ -0,0 +1,23 @@
+class NormalizeConfig
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.inject({}) do |normalized_config, (key, value)|
```
teoljungberg: We prefer `reduce` over `inject` - https://github.com/thoughtbot/guides/blob/5deeaaa2006a7f895f697cc99668eadaa4abe693/style/ruby/README.md.

Also - I'm unsure why this isn't being warned by Hound.

spec/models/hound_config_spec.rb
```diff
@@ -1,138 +1,125 @@
...
+          commit = stubbed_commit(".hound.yml" => "")
+          hound_config = HoundConfig.new(commit)
+
+          expect(hound_config).not_to be_enabled_for("remark")
         end
       end
     end
 
-    context "when the given language is disabled" do
+    context "given a language that is disabled in the config file" do
```
teoljungberg: I don't think this wording needed to change.

teoljungberg: Nor the test bellow.

spec/models/hound_config_spec.rb
```diff
@@ -173,4 +160,8 @@
       end
     end
   end
+
+  def be_enabled_for(linter_name)
```
teoljungberg: Nice refactoring!

spec/services/config_normalizer_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "app/services/normalize_config"
+
+describe NormalizeConfig do
+  context "given a hash with keys containing capital letters" do
```
teoljungberg: Should we have a `describe "#run"` here?

spec/services/config_alias_resolver_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "app/services/resolve_config_aliases"
+
+describe ResolveConfigAliases do
+  context "when the content contains configured aliases" do
```
teoljungberg: Should we have a `describe "#run"` here?

teoljungberg: I would replace `content` with `config` in this context, especially since the class name has `config` in it.

teoljungberg: Also, the spec name doesn't match the class name.

spec/services/config_normalizer_spec.rb
teoljungberg: The spec name does not match the class name.

app/services/normalize_config.rb
jcmorrow: Hound warned me the other way, hence the change 😕

spec/services/config_alias_resolver_spec.rb
jcmorrow: Good catches all!

spec/models/hound_config_spec.rb
jcmorrow: I kind of prefer the contexts to be consistent in their wording, but it's not a necessary change.

jcmorrow:
@teoljungberg I probably should have been more clear, this is an alternative to https://github.com/houndci/hound/pull/1143. I wanted to be able to see how both solutions worked out without smashing older work with the newer approach.

app/models/hound_config.rb
jcmorrow: TIL, Rails has a `deep_merge` method. http://api.rubyonrails.org/v2.3/classes/ActiveSupport/CoreExtensions/Hash/DeepMerge.html

app/services/normalize_config.rb
```diff
@@ -0,0 +1,19 @@
+class NormalizeConfig
+  def self.run(config)
+    config.reduce({}) do |normalized_config, (key, value)|
```
jcmorrow: @teoljungberg ❓

teoljungberg: Beats me...

gylaz: I've made this comment else where in this PR:
> This is actually wrong, according to thoughtbot styleguide. I didn't realize this but since we moved this under houndci org, we no longer get the thoughtbot rules.

So, we should add this rule back in to our `.ruby-style.yml`.

app/models/hound_config.rb
```diff
@@ -1,31 +1,28 @@
...
-    coffeescript
-    go
-    haml
-    javascript
-    python
-    ruby
-    scss
-    swift
-  )
+  DEFAULT_CONFIG = Linter::Collection::LINTER_NAMES.inject({}) do |config, name|
```
gylaz: If you use `each_with_object` you can avoid needing `config` on line 12.

gylaz: Constant feels like a wrong tool for this, since we are doing some logic. Thoughts on using a class method for this instead?

app/services/normalize_config.rb
gylaz:
```yaml
Style/CollectionMethods:
  Enabled: true
  PreferredMethods:
    find: detect
    inject: reduce
    collect: map
    find_all: select
```

app/models/hound_config.rb
jcmorrow: That's a good call.

app/models/hound_config.rb
```diff
@@ -1,31 +1,27 @@
...
-    coffeescript
-    go
-    haml
-    javascript
-    python
-    ruby
-    scss
-    swift
-  )
+  DEFAULT_CONFIG =
```
teoljungberg: This also seems like it should be a method.

app/services/normalize_config.rb
```diff
@@ -0,0 +1,27 @@
+class NormalizeConfig
+  def self.run(config)
+    new(config).normalize
```
teoljungberg: We usually name the instance method the same as its static facade - so `.run` -> `#run`.

app/services/resolve_config_aliases.rb
```diff
@@ -0,0 +1,26 @@
+class ResolveConfigAliases
```
teoljungberg: I wonder about the use case of this service object - we already have a way of resolving aliases for a config using `Config::Base#linter_names`.

Couldn't we use that approach and have fewer moving parts?

spec/models/hound_config_spec.rb
```diff
@@ -1,37 +1,44 @@
 require "spec_helper"
+require "app/models/config/parser"
+require "app/services/normalize_config"
+require "app/services/resolve_config_aliases"
 require "app/models/hound_config"
 
 describe HoundConfig do
   describe "#content" do
-    it "returns the content of the .hound.yml file" do
+    it "returns the contents of the .hound.yml file merged with the defaults" do
```
teoljungberg: I would add a comma after `file`

teoljungberg:
I'm somewhat confused over the approach in this PR, as in we've built new functionality to replace something that we already have. So some questions off the bat:

- Doesn't the existing configuration setup work with aliases? If not - we should fix that instead of introducing new concepts. See `Config::Base#linter_names`.
- There's a bit too much going on in this PR, and it seems like they are isolated enough that it can be extracted into multiple PR's.

jcmorrow:
@teoljungberg Our current setup doesn't really work. It **sort of** works, but it misses a lot of corner cases. The biggest thing it misses is:

Currently if I disable an alias of a default linter, it doesn't disable the linter. So if your `hound.yml` is:

    coffee_script: 
      enabled: false

coffeescript will still be run. The only reason this works:

    javascript: 
      enabled: false

is because `jshint` is under `BETA_LANGUAGES`, while `javascript` is under `LANGUAGES`, which doesn't make any sense since they are aliases for the same linter. So doing this:


    jshint:
      enabled: false

Actually wouldn't disable the jshint linter.

----

I agree that it would be nice to just make the current approach work, but it involves making the changes I made in https://github.com/houndci/hound/pull/1143 and those were much more fragile and harder to grok than the approach taken here. I'm open to breaking this up into smaller PR's. I'll see if I can start that today.

app/services/resolve_config_aliases.rb
jcmorrow: The problem with this is the problem of disabled aliases, as I mentioned in the main discussion thread. Might be worth talking over on a hangout?

spec/services/resolve_config_aliases_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+describe ResolveConfigAliases do
+  describe "#run" do
+    context "when the config contains aliases" do
+      it "renames them to the appropriate linter" do
+        config = {
+          "javascript" => { "enabled" => false },
+          "ruby" => { "enabled" => false },
+        }
+
+        expect(ResolveConfigAliases.run(config).keys).to eq(["jshint", "ruby"])
```
thorncp: Might be good to use `match_array` here so we don't rely on the ordering of the keys in the hash.

spec/services/normalize_config_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "app/services/normalize_config"
+
+describe NormalizeConfig do
+  describe "#run" do
+    context "given a hash with keys containing capital letters" do
+      it "downcases the keys" do
+        config = { "Ruby" => { "Enabled" => true } }
+
+        expect(NormalizeConfig.run(config).keys).to eq ["ruby"]
```
thorncp: What do you think about `expect(NormalizeConfig.run(config)).to have_key("ruby")`?

app/services/resolve_config_aliases.rb
thorncp: Another option could be `each_with_object`. A benefit would be being able to delete L23, since the method keeps its own accumulator.

app/models/hound_config.rb
```diff
@@ -25,35 +28,29 @@ def fail_on_violations?
...
-    Config::Parser.yaml(file_content) || {}
-  end
-
-  def configured?(name)
-    key = normalize_key(name)
-    config = options_for(key)
-
-    enabled?(config)
+  def default_config
+    Linter::Collection.linter_names.each.with_object({}) do |name, config|
```
thorncp: Would it be beneficial for this generated config to be run through the normalizer and alias resolver? `Linter::Collection.linter_names` is including underscores, e.g. `coffee_script`, but config coming from the user is being resolved to `coffeescript`. Is the difference being resolved somewhere else?

jcmorrow: I don't think I understand the question. If the user provides a key `coffeescript` it's resolved as an alias for the true linter name: `coffee_script`, which is enabled by default. `coffee_script` is being passed in here and returns true as expected.

thorncp: Maybe I'm confused? Maybe a script would be a more clear representation:

```ruby
require "./config/environment"

class FakeCommit
  def file_content(*)
    <<~YAML
      ---
      coffee_script:
        enabled: false
    YAML
  end
end

commit = FakeCommit.new
hound_config = HoundConfig.new(commit)

puts hound_config.linter_enabled?("coffee_script") # => true
```

I think I'd expect it to return `false`. Is that correct?

jcmorrow: 😑

app/models/hound_config.rb
```diff
@@ -12,11 +12,14 @@ class HoundConfig
   attr_reader_initialize :commit
 
   def content
-    @content ||= parse(commit.file_content(CONFIG_FILE))
+    @merged_config ||= default_config.deep_merge(resolved_aliases_config)
```
gylaz: We tend to name memoized variables after the method. Better yet, to add `_`, so `@_content`.

app/services/normalize_config.rb
```diff
@@ -0,0 +1,26 @@
...
+    new(config).run
+  end
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.each_with_object({}) do |(key, value), normalized_config|
+      normalized_key = normalize_key key
```
gylaz: How about parens around `key`?

app/models/hound_config.rb
thorncp: It seems to be because the order in `ResolveConfigAliases` is `"coffee_script" => "coffeescript"` (I haven't tried). Would it make sense to flip those?

jcmorrow: I think the aliases were just set up wrong:

`coffee_script` was aliasing to `coffeescript`, but it should be the other way around.

thorncp: :+1:

jcmorrow:
@teoljungberg We're still feeling some pain from this and I'd like to merge it, but I don't want to do so if you feel uncomfortable with the approach. What do you think?

teoljungberg:
@jcmorrow The thing I'm bothered by is the duplication between the various linter names and their aliases. Which are present in `ResolveConfigAliases` and `Linter::Base#linter_names`. But since they aren't connected at this point - I'm unsure it can be avoided.

I did make some improvements and refactorings to address the feedback and some of my own. But besides that - I'm fine merging this and moving on.

app/services/normalize_config.rb
```diff
@@ -0,0 +1,23 @@
+class NormalizeConfig
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.reduce({}) do |normalized_config, (key, value)|
```
gylaz: @teoljungberg  Why did you change this from `each_with_object` to `reduce`? It seems like `each_with_object` is a better fit, since line 18 can be removed.

jcmorrow:
@teoljungberg I totally understand your concern about duplicating the linter aliases. I've removed them from the linter classes as they are no longer necessary 👍

app/services/normalize_config.rb
teoljungberg: I find `reduce` easier to reason / read than `each_with_object` - if I'm alone in that camp I'm happy changing it back.

PR #1151, opened 2016-06-06
Remove users with bad tokens from repo

gylaz:
GitHub returns `401 - Bad credentials` status when a user with
expired/invalid token tries to set it. We should handle that case and
remove that token from the repo's memberships.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -7,7 +7,7 @@ class BuildRunner
     end
   rescue Config::ParserError => exception
     report_config_file_as_invalid(exception)
-  rescue Octokit::NotFound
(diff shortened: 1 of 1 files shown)


jcmorrow:
@gylaz is this usually because something has actually expired or is it usually because Hound is removed from the repo?

gylaz:
My best guess is that this happens when a user has been removed from the repo. Hound will first use a token of any user that is associated with the repo -- it assumes they have access -- if it fails to get it, it disassociates that user (and in the end falls back to Hound's own token).

jcmorrow:
Makes sense, LGTM! 🌂

PR #1153, opened 2016-06-07
Move linter names to Linter::Collection

jcmorrow:
Move linter names to Linter::Collection and rename `BETA_LANGUAGES` to `BETA_LINTERS`. This is a change isolated from https://github.com/houndci/hound/pull/1149


app/models/linter/collection.rb
```diff
@@ -14,6 +14,10 @@ class Collection
       Linter::Swift,
     ].freeze
 
+    def self.linter_names
+      LINTERS.map { |klass| klass.name.demodulize.downcase }
```
gylaz: It's `self.class.name.demodulize.underscore` in `app/models/linter/base.rb`. I'm not sure which one we *should* use, but it seems they should be consistent.

app/models/hound_config.rb
```diff
@@ -1,22 +1,12 @@
...
-    go
-    haml
-    javascript
-    python
-    ruby
-    scss
-    swift
-  )
+  ).freeze
+  CONFIG_FILE = ".hound.yml".freeze
```
gylaz: How about using `# frozen_string_literal: true` a the top of the file?

gylaz:
Two questions, then LGTM.

app/models/linter/collection.rb
jcmorrow: Yeah `underscore` is superior here.

app/models/hound_config.rb
```diff
@@ -1,22 +1,13 @@
+# frozen_string_literal: true
```
jcmorrow: @gylaz does it go on the top line outside of the class or inside of the class?

gylaz: This is correct. First line in the file

PR #1154, opened 2016-06-07
Rename `enabled_for?` to `linter_enabled?`

jcmorrow:
This is a change isolated from https://github.com/houndci/hound/pull/1149. The new naming is clearer and makes `HoundConfig` more readable.

### DIFF
--- a/spec/models/hound_config_spec.rb
+++ b/spec/models/hound_config_spec.rb
@@ -22,7 +22,7 @@ describe HoundConfig do
     end
   end
 
-  describe "#enabled_for?" do
+  describe "#linter_enabled?" do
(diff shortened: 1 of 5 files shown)


spec/models/hound_config_spec.rb
```diff
@@ -173,4 +173,8 @@
       end
     end
   end
+
+  def be_enabled_for(linter_name)
+    be_linter_enabled linter_name
```
gylaz: I think we've been fans of parens in Ruby, expect specific DSLs.

gylaz:
👍

teoljungberg:
I _love_ the size of these PRs!

PR #1156, opened 2016-06-10
Owner level configuration

jcmorrow:
As a user, I'd like to be able to specify a repository which acts as the default configuration for all of a particular owner's repos.

**Changes**

 - `config_repo`, and `config_enabled` columns added to `owners` table
 - all configs get a `merge` method which takes another config and returns the result of combining them
 - `OwnerHoundConfigBuilder` takes a repo and a `HoundConfig` and returns either the repo owner's `HoundConfig` or, failing that, the `HoundConfig` it was passed.
 - stub `OwnerHoundConfigBuilder` in a lot of tests (I don't love this but I don't see a great way around it either)

**Still working on**

 - higher-level specs for opening pull requests where owner level configs should be used and verifying that they are generated correctly.

### DIFF
--- a/app/models/linter/base.rb
+++ b/app/models/linter/base.rb
@@ -42,7 +42,7 @@ module Linter
     def build_review_job_attributes(commit_file)
       {
         commit_sha: build.commit_sha,
-        config: config.serialize,
+        config: merged_config,
         content: commit_file.content,
         filename: commit_file.filename,
         linter_name: name,
@@ -57,6 +57,18 @@ module Linter
 
     def job_class
       "#{name.classify}ReviewJob".constantize
+    end
+
+    def merged_config
+      owner_config.merge(config)
+    end
+
+    def owner_config
+      @owner_config ||= ConfigBuilder.for(owner_hound_config, name)
+    end
+
+    def owner_hound_config
+      OwnerHoundConfigBuilder.run(build.repo, hound_config)
     end
 
     def config
--- a/app/services/owner_hound_config_builder.rb
+++ b/app/services/owner_hound_config_builder.rb
@@ -1,0 +1,25 @@
+class OwnerHoundConfigBuilder
+  HEAD = "HEAD".freeze
+  def self.run(repo, default)
+    new(repo, default).run
+  end
+
+  def initialize(repo, default)
+    @repo = repo
+    @default = default
+  end
+
+  def run
+    if repo.owner.has_config_repo?
+      github = GithubApi.new(Hound::GITHUB_TOKEN)
+      commit = Commit.new(repo.owner.config_repo, HEAD, github)
+      HoundConfig.new(commit)
+    else
+      default
+    end
+  end
+
+  private
+
+  attr_reader :repo, :default
+end
--- a/spec/models/config/base_spec.rb
+++ b/spec/models/config/base_spec.rb
@@ -135,6 +135,26 @@ describe Config::Base do
     end
   end
 
+  describe "#merge" do
+    context "when the override contains keys not in the base config" do
+      it "returns a hash contains all keys and values" do
+        hound_config = instance_double(
+          "HoundConfig",
+          content: {
+            "test" => { "config_file" => "config-file.txt" }
+           },
+           commit: stubbed_commit("config-file.txt" => "some key: some value"),
+        )
+        base_config = build_config(hound_config: hound_config)
+        override_config = build_config(hound_config: hound_config)
+
+        merged_config = base_config.merge(override_config)
+
+        expect(merged_config).to include("some key" => "some value")
+      end
+    end
+  end
+
   def build_config(hound_config: build_hound_config)
     Config::Test.new(hound_config)
   end
--- a/spec/models/linter/scss_spec.rb
+++ b/spec/models/linter/scss_spec.rb
@@ -50,6 +50,40 @@ describe Linter::Scss do
         config: "{}",
       )
     end
+
+    context "when the owner has a config repo set" do
+      it "schedules a review job with the owner's config" do
+        owner = build(
+          :owner,
+          config_enabled: true,
+          config_repo: "organization/style",
+        )
+        repo = build(:repo, owner: owner)
+        build = build(
+          :build,
+          commit_sha: "foo",
+          pull_request_number: 123,
+          repo: repo,
+        )
+        linter = build_linter(build)
+        stub_scss_config({})
+        commit_file = build_commit_file(filename: "lib/a.scss")
+        allow(Resque).to receive(:enqueue)
+
+        linter.file_review(commit_file)
+
+        expect(Resque).to have_received(:enqueue).with(
+          ScssReviewJob,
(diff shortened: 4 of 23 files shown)


app/models/linter/base.rb
```diff
@@ -59,6 +59,18 @@ def job_class
...
+    def merged_config
+      owner_config.merge(config)
+    end
+
+    def owner_config
+      @owner_config ||= ConfigBuilder.for(owner_hound_config, name)
+    end
+
+    def owner_hound_config
+      OwnerHoundConfigBuilder.run(build.repo, hound_config)
```
gylaz: `BuildOwnerHoundConfig`? Or how about just `BuildHoundConfig`?

app/services/owner_hound_config_builder.rb
```diff
@@ -0,0 +1,25 @@
+class OwnerHoundConfigBuilder
+  HEAD = "HEAD".freeze
```
gylaz: `# frozen_string_literal: true`?

app/services/owner_hound_config_builder.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def initialize(repo, default)
+    @repo = repo
+    @default = default
+  end
+
+  def run
+    if repo.owner.has_config_repo?
+      github = GithubApi.new(Hound::GITHUB_TOKEN)
```
gylaz: Can you use `UserToken` here to get the right token?

spec/models/config/base_spec.rb
```diff
@@ -135,6 +133,46 @@ def parse(file_content)
...
+          "HoundConfig",
+          content: { "test" => { "config_file" => "config-file.txt" } },
+          commit: stubbed_commit("config-file.txt" => "some key: base "),
+        )
+        override_hound_config = instance_double(
+          "HoundConfig",
+          content: { "test" => { "config_file" => "config-file.txt" } },
+          commit: stubbed_commit("config-file.txt" => "some key: override"),
+        )
+
```
gylaz: Do we need this newline?

spec/models/config/base_spec.rb
```diff
@@ -135,6 +133,46 @@ def parse(file_content)
...
+        base_hound_config = instance_double(
+          "HoundConfig",
+          content: { "test" => { "config_file" => "config-file.txt" } },
+          commit: stubbed_commit("config-file.txt" => "some key: base "),
+        )
+        override_hound_config = instance_double(
+          "HoundConfig",
+          content: { "test" => { "config_file" => "config-file.txt" } },
+          commit: stubbed_commit("config-file.txt" => "some key: override"),
+        )
```
gylaz: Could probably extract this into a method called `build_hound_config` and pass whatever is differring.

spec/support/helpers/linter_helper.rb
```diff
@@ -64,6 +64,12 @@ def stubbed_commit(configuration)
 
     commit
   end
+
+  def stub_owner_hound_config
+    allow(OwnerHoundConfigBuilder).to receive(:run) do |_repo, default|
+      default
+    end
```
gylaz: What does this block do? Is it necessary?

jcmorrow:
@teoljungberg I based a lot of this off of your branch, so :eyes: would be very appreciated 😃

app/models/linter/base.rb
jcmorrow: I like `BuildOwnerHoundConfig`. I think the `Owner` in the name is important, because it doesn't build a hound config for the repo you give it, it extracts the owner from the repo and builds a hound config for that.

app/services/owner_hound_config_builder.rb
jcmorrow: I didn't know that was an object, but yeah, probably.

spec/models/config/base_spec.rb
jcmorrow: Nope.

spec/support/helpers/linter_helper.rb
jcmorrow: Yeah, so this cuts out all of the logic around fetching the config from the owner's config repo, and instead just says "pretend you went and got that hound config and then pass back the one you were given originally". I'm not sure this is the best way to stub it out but it's important for some of these tests that we don't override the hound config they set by returning a static stub all the time.

gylaz: Ah, so `_repo` and `default` are the arguments that get passed in. Got it.

app/services/owner_hound_config_builder.rb
thorncp: Is there a more descriptive name for this constant that we can use? I'm not sure what `"HEAD"` means in this context, because we're not mentioning a branch name anywhere.

spec/models/linter/scss_spec.rb
```diff
@@ -50,13 +52,48 @@
...
+        stub_scss_config({})
+        commit_file = build_commit_file(filename: "lib/a.scss")
+        allow(Resque).to receive(:enqueue)
+
+        linter.file_review(commit_file)
+
+        expect(Resque).to have_received(:enqueue).with(
+          ScssReviewJob,
+          commit_sha: build.commit_sha,
+          config: "{some_key: some_value}",
```
thorncp: Where is this coming from? Would it be possible to see it in the spec set up?

jcmorrow: I'm not sure how I want to make this work, so right now it's nowhere, this spec doesn't pass 😞

blackjid:
hey @jcmorrow 
how the linters in https://github.com/houndci/linters will know about the correct configuration???

jcmorrow:
@blackjid the config values are part of the job that is pushed onto the resque queue by hound and when the correct linter pulls it off it parses the config specified in the job. At least, that's my understanding of it after only a couple of weeks working on Hound 😄

blackjid:
ah, thanks!. that makes sense. that way the jobs are self-contained. I'm looking forward for this to be working!!! Now I run a fork with my own configurations... :)

spec/models/linter/scss_spec.rb
```diff
@@ -50,16 +52,80 @@
         config: "{}",
       )
     end
+
+    context "when the owner has a config repo set" do
+      it "schedules a review job with the owner's config" do
```
jcmorrow: @thorncp can you take another look at this spec? I really don't like how long it is, but it *is* testing the right thing. What do you think?

thorncp: This looks good to me. I think the length is an indicator of how complex it is and there's not much you can do about it without doing a larger refactor. You could move the set up steps to helper methods, but I prefer seeing them inline in the spec itself.

jcmorrow: Yeah, I tried to extract as much set up as I could while still making the set up steps clear. I will add a spec for this scenario to each linter and then I'll feel a lot more comfortable about this feature.

PR #1157, opened 2016-06-15
Add BuildOwnerHoundConfig service

jcmorrow:
I just realized I can split these owner level config changes up and not change anything on the user facing side, so here's the first of several smaller PR's.

Owners can now have a config repository configured. When given an
`owner` `BuildOwnerHoundConfig` will fetch the HEAD commit of the repo
and give back a `HoundConfig` built from it. If the owner does not have
a config repo or it is disabled `BuildOwnerHoundConfig` will return
false.

### DIFF
--- a/app/services/build_owner_hound_config.rb
+++ b/app/services/build_owner_hound_config.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+class BuildOwnerHoundConfig
+  LATEST_SHA = "HEAD"
+
+  def self.run(*args)
+    new(*args).run
+  end
+
+  def initialize(owner)
+    @owner = owner
+  end
+
+  def run
+    if owner.has_config_repo?
+      github = GithubApi.new(user_token)
+      commit = Commit.new(config_repo.full_github_name, LATEST_SHA, github)
+      HoundConfig.new(commit)
+    end
+  end
+
+  private
+
+  attr_reader :owner
+
+  def user_token
+    UserToken.new(config_repo).token
+  end
+
+  def config_repo
+    Repo.find_or_create_with(full_github_name: owner.config_repo)
+  end
+end
--- a/db/migrate/20160608183021_add_config_to_owner.rb
+++ b/db/migrate/20160608183021_add_config_to_owner.rb
@@ -1,0 +1,24 @@
+class AddConfigToOwner < ActiveRecord::Migration
+  def up
+    add_column :owners, :config_enabled, :boolean, null: false, default: false
+    add_column :owners, :config_repo, :string
+
+    execute <<~SQL
+      UPDATE
+        owners
+      SET
+        config_repo = 'houndci/legacy-config',
+        config_enabled = true
+      FROM
+        repos
+      WHERE
+        repos.owner_id = owners.id AND repos.active = true;
+    SQL
+
+  end
+
+  def down
+    remove_column :owners, :config_enabled
+    remove_column :owners, :config_repo
+  end
+end
--- a/spec/models/owner_spec.rb
+++ b/spec/models/owner_spec.rb
@@ -36,4 +36,30 @@ describe Owner do
       end
     end
   end
+
+  describe "#has_config_repo?" do
+    context "when the owner has a config repo set and enabled" do
+      it "returns true" do
+        owner = create(:owner, config_repo: "org/style", config_enabled: true)
+
+        expect(owner).to have_config_repo
+      end
+    end
+
+    context "when the owner has a config repo set and disabled" do
+      it "returns false" do
+        owner = create(:owner, config_repo: "org/style", config_enabled: false)
+
+        expect(owner).not_to have_config_repo
+      end
+    end
+
+    context "when the owner does not have a config repo set" do
+      it "returns false" do
+        owner = create(:owner, config_repo: "")
+
+        expect(owner).not_to have_config_repo
+      end
+    end
+  end
 end
--- a/spec/services/build_owner_hound_config_spec.rb
+++ b/spec/services/build_owner_hound_config_spec.rb
@@ -1,0 +1,62 @@
+require "rails_helper"
+
+describe BuildOwnerHoundConfig do
+  describe "#run" do
+    context "when the owner has a configuration set" do
+      it "returns the configuration of that repo" do
+        owner = instance_double(
+          "Owner",
+          has_config_repo?: true,
+          config_repo: "thoughtbot/guides",
+        )
+        commit = stubbed_commit(
+          ".hound.yml" => <<~EOS
+            ruby:
+              enabled: true
+          EOS
+        )
+        allow(Commit).to receive(:new).and_return(commit)
+
+        owner_config = BuildOwnerHoundConfig.run(owner)
+
+        expect(owner_config.content).to eq("ruby" => { "enabled" => true })
+      end
+    end
+
+    context "when the owner does not have a configuration set" do
+      it "returns nil" do
+        owner = instance_double(
+          "Owner",
+          has_config_repo?: false,
+          config_repo: "thoughtbot/guides",
+        )
+
+        owner_config = BuildOwnerHoundConfig.run(owner)
+
+        expect(owner_config).to be_nil
+      end
+    end
+
+    context "when the owner's configuration is unreachable" do
+      it "returns an empty hound config" do
+        owner = instance_double(
+          "Owner",
+          has_config_repo?: true,
+          config_repo: "organization/private_style_guides",
+        )
+        stub_failure_on_repo(repo: "organization/private_style_guides")
+
+        owner_config = BuildOwnerHoundConfig.run(owner)
+
+        expect(owner_config.content).to eq({})
+      end
+    end
+  end
+
+  def stub_failure_on_repo(repo:)
+    stub_request(
+      :get,
+      %r"https://api.github.com/repos/#{repo}/contents/*",
+    ).to_return(status: 404, headers: {})
+  end
+end
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -8,4 +8,8 @@ class Owner < ActiveRecord::Base
     owner.save!
     owner
   end
+
+  def has_config_repo?
+    config_enabled? && config_repo.present?
+  end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160513205551) do
+ActiveRecord::Schema.define(version: 20160608183021) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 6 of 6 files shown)


spec/models/owner_spec.rb
```diff
@@ -36,4 +36,30 @@
       end
     end
   end
+
+  describe "#has_config_repo?" do
+    context "when the owner has a config repo set and enabled" do
+      it "returns true" do
+        owner = create(:owner, config_repo: "org/style", config_enabled: true)
+
+        expect(owner.has_config_repo?).to be true
```
thorncp: `expect(owner).to have_config_repo` ?

app/services/build_owner_hound_config.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def initialize(owner)
+    @owner = owner
+  end
+
+  def run
+    if owner.has_config_repo?
+      github = GithubApi.new(Hound::GITHUB_TOKEN)
+      commit = Commit.new(owner.config_repo, HEAD, github)
```
thorncp: What does HEAD mean in this context? It is eventually passed to the GitHub API to fetch a file, but I wouldn't expect GitHub to have a HEAD, because it doesn't have a working tree. Does it related to GitHUb's concept of a default branch?

spec/services/build_owner_hound_config_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        allow(Commit).to receive(:new).and_return(commit)
+
+        owner_config = BuildOwnerHoundConfig.run(owner)
+
+        expect(owner_config.content).to eq("ruby" => { "enabled" => true })
+      end
+    end
+
+    context "when the owner does not have a configuration set" do
+      it "returns the default it was passed" do
```
thorncp: This differs from the expectation. Should it be `"returns false"` ?

jcmorrow: Good catch. Changed this from a previous version.

app/services/build_owner_hound_config.rb
teoljungberg: I know I sketched this out this way, when I did my spike on this months ago, and decided on `HEAD`. Didn't test the various edge cases out when doing so, so the time is ripe for it now.

jcmorrow: So, it's not really necessary, if you pass in no `nil`, `"HEAD"`, or `""` as sha it just gives you the file contents at the tip of master.

jcmorrow: But to answer your question yes it does allow GitHub to default to a branch.

jcmorrow: Actually, now that I think about it, `"master"` is probably more clear.

teoljungberg: An edge case would be if a users changes their hound config in a branch that's not `master` we wouldn't pick the changes.

jcmorrow: OK, digged into it and yes, if we pass in `"HEAD"` (or `nil`) as the ref and the repo has a branch other than master set as the default branch, then we will still get the file back as it is in the default branch.

jcmorrow: Are there any other edge case github setups you can think of @teoljungberg?

app/services/build_owner_hound_config.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+class BuildOwnerHoundConfig
+  HEAD = "HEAD"
```
gylaz: Does `HEAD` default to latest on master? We could pass `master` here too, and it will fetch the latest master.

Also, what do you think about calling this something a bit more descriptive, like `LATEST_SHA` or `COMMIT_SHA` or something?

app/services/build_owner_hound_config.rb
```diff
@@ -0,0 +1,26 @@
...
+    new(*args).run
+  end
+
+  def initialize(owner)
+    @owner = owner
+  end
+
+  def run
+    if owner.has_config_repo?
+      github = GithubApi.new(Hound::GITHUB_TOKEN)
```
gylaz: Can we use `UserToken` to get a token rather than `Hound::GITHUB_TOKEN`?

app/services/build_owner_hound_config.rb
gylaz: I think we should be assuming we're going to get the latest from master, and not know anything about stuff on branches.

app/services/build_owner_hound_config.rb
```diff
@@ -0,0 +1,26 @@
...
+      commit = Commit.new(owner.config_repo, HEAD, github)
+      HoundConfig.new(commit)
+    else
+      false
```
gylaz: Why does this `run` method return `false`?

db/migrate/20160608183021_add_config_to_owner.rb
```diff
@@ -0,0 +1,13 @@
+class AddConfigToOwner < ActiveRecord::Migration
+  def up
+    add_column :owners, :config_enabled, :boolean, null: false, default: false
+    add_column :owners, :config_repo, :string
+
+    Owner.update_all(config_enabled: true, config_repo: "houndci/legacy-config")
```
gylaz: What do you think about making this a bit more sophisticated, and setting this to `true` only if there are any active repos? We could probably do this with a join.

Also, rather than relying on `Owner` AR model, why not just use sql query?

app/services/build_owner_hound_config.rb
jcmorrow: So if someone has something other than master as the default you think we should still point to master?

app/services/build_owner_hound_config.rb
jcmorrow: It will default to the lastest commit on the default branch, which almost always master, but is occasionally not master. I think `LATEST_SHA` is quite descriptive.

db/migrate/20160608183021_add_config_to_owner.rb
jcmorrow: Both of those ideas sound good. 👍

app/services/build_owner_hound_config.rb
jcmorrow: It returns false if the owner doesn't have a config repo set up. I see the other alternative as passing back the default hound config (basically `HoundConfig.new(nil)`, which is probably also fine. I don't know which is better.

app/services/build_owner_hound_config.rb
gylaz: Do you mean if they set a different default branch in GitHub? I didn't realize people could do that. I guess then it should be what ever the default is. Does `""` always grab from the default branch?

app/services/build_owner_hound_config.rb
gylaz: I'm not sure that it's expected of `run` to return a boolean. It's not a predicate method. How about returning nothing (implicit `nil`)?

app/services/build_owner_hound_config.rb
jcmorrow: It looks to me like only `nil` or `HEAD` honor the default branch, but it doesn't seem to be explicitly covered in the docs.

app/services/build_owner_hound_config.rb
jcmorrow: I'm also fine with that.

app/services/build_owner_hound_config.rb
```diff
@@ -0,0 +1,32 @@
...
+  private
+
+  attr_reader :owner
+
+  def user_token
+    @user_token ||= UserToken.new(config_repo).token
+  end
+
+  def config_repo
+    @repo ||= Repo.find_or_create_with(full_github_name: owner.config_repo)
```
gylaz: By convention we name the memoizing ivar the same as the method name.

app/services/build_owner_hound_config.rb
```diff
@@ -0,0 +1,32 @@
...
+      HoundConfig.new(commit)
+    end
+  end
+
+  private
+
+  attr_reader :owner
+
+  def user_token
+    @user_token ||= UserToken.new(config_repo).token
```
gylaz: I don't think we need to memoize this.

In general, we try to avoid memoizing unless necessary.

db/migrate/20160608183021_add_config_to_owner.rb
```diff
@@ -0,0 +1,25 @@
+class AddConfigToOwner < ActiveRecord::Migration
+  def up
+    add_column :owners, :config_enabled, :boolean, null: false, default: false
+    add_column :owners, :config_repo, :string
+
+    ActiveRecord::Base.connection.execute(<<~SQL
```
gylaz: Can just be `execute <<~SQL`.

db/migrate/20160608183021_add_config_to_owner.rb
```diff
@@ -0,0 +1,25 @@
...
+          owners
+        SET
+          config_repo = 'houndci/legacy-config',
+          config_enabled = true
+        FROM
+          repos
+        WHERE
+          repos.owner_id = owners.id AND repos.active = true;
+      SQL
+    )
```
gylaz: The parens can be dropped.

app/services/build_owner_hound_config.rb
jcmorrow: Oh yeah, I had the same thought here and then didn't remove it. You're right, the performance gain is pretty much non-existent.

db/migrate/20160608183021_add_config_to_owner.rb
jcmorrow: cool!

db/migrate/20160608183021_add_config_to_owner.rb
```diff
@@ -0,0 +1,25 @@
...
+    ActiveRecord::Base.connection.execute(<<~SQL
+        UPDATE
+          owners
+        SET
+          config_repo = 'houndci/legacy-config',
+          config_enabled = true
+        FROM
+          repos
+        WHERE
+          repos.owner_id = owners.id AND repos.active = true;
```
gylaz: How does this work? Does it look for any repo that matches this criteria?

gylaz:
One concern about the update in the migration. Did you test it and does it seem to do what we want?

gylaz:
Other than that, looks good to merge.

jcmorrow:
Yeah it turns out join statements look very strange when done in update statements, but I did test it against my local database and it performed what we wanted.

db/migrate/20160608183021_add_config_to_owner.rb
jcmorrow: You can use a traditional `FROM` statement in updates, but per postgres it can't include the name of the table being updated unless you're performing a self join.

>from_list
>A list of table expressions, allowing columns from other tables to appear in the WHERE condition and the update expressions. This is similar to the list of tables that can be specified in the FROM Clause of a SELECT statement. **Note that the target table must not appear in the from_list, unless you intend a self-join (in which case it must appear with an alias in the from_list).**

And so instead you use the syntax above without a join statement.

PR #1159, opened 2016-06-15
Owner Level Ruby Configuration

jcmorrow:
`RubyConfigBuilder` no longer knows anything about hound defaults or thoughtbot
 defaults, it just acts as a bridge between hashes (what our
 `Config::Ruby` objects produce) and `RuboCop::Config`s.

`Linter::Ruby` configures cops with the results of merging the repo's
configuration into the owner's configuration. Only some of the specs
have been rewritten, there's still more to rewrite on that end, but it does work, I just want to perfect the setup steps and make sure we're testing all scenarios.

### DIFF
--- a/app/models/linter/ruby.rb
+++ b/app/models/linter/ruby.rb
@@ -9,7 +9,7 @@ module Linter
     end
 
     def file_included?(commit_file)
-      !linter_config.file_to_exclude?(commit_file.filename)
+      !merged_config.file_to_exclude?(commit_file.filename)
     end
 
     private
@@ -36,7 +36,7 @@ module Linter
     def team
       RuboCop::Cop::Team.new(
         RuboCop::Cop::Cop.all,
-        linter_config,
+        merged_config,
         rubocop_options,
       )
     end
@@ -49,19 +49,41 @@ module Linter
       )
     end
 
-    def linter_config
-      @linter_config ||= config_builder.config
+    def merged_config
+      @merged_config ||=
+        if build.repo.owner.has_config_repo?
+          rubocop_config_builder(owner_config.content).
+            merge(repo_config.content)
+        else
+          rubocop_config_builder(repo_config.content).config
+        end
     end
 
-    def config_builder
-      RubyConfigBuilder.new(config.content, repository_owner_name)
+    def rubocop_config_builder(content)
+      RubyConfigBuilder.new(content)
+    end
+
+    def owner_config
+      ConfigBuilder.for(owner_hound_config, "ruby")
+    end
+
+    def owner_hound_config
+      BuildOwnerHoundConfig.run(build.repo.owner)
+    end
+
+    def repo_config
+      ConfigBuilder.for(repo_hound_config, "ruby")
+    end
+
+    def repo_hound_config
+      hound_config
     end
 
     # This is deprecated in favor of RuboCop's DisplayCopNames option.
     # Let's track how often we see this and remove it if we see fit.
     def rubocop_options
-      if linter_config.delete("ShowCopNames")
-        Analytics.new(repository_owner_name).track_show_cop_names
+      if merged_config.delete("ShowCopNames")
+        Analytics.new(build.full_github_name).track_show_cop_names
         { debug: true }
       end
     end
--- a/app/services/ruby_config_builder.rb
+++ b/app/services/ruby_config_builder.rb
@@ -2,42 +2,29 @@ class RubyConfigBuilder
   HOUND_DEFAULTS_FILENAME = "ruby.yml".freeze
   VIRTUAL_FILENAME = "".freeze
 
-  def initialize(overrides = {}, repository_owner_name = nil)
-    @overrides = overrides
-    @repository_owner_name = repository_owner_name
+  def initialize(content = {})
+    @content = content
   end
 
   def config
     RuboCop::ConfigLoader.merge_with_default(
-      combined_overrides,
+      normalized_content,
       VIRTUAL_FILENAME,
     )
   end
 
+  def merge(overrides)
+    RuboCop::ConfigLoader.merge(content, normalize_config(overrides))
+  rescue NoMethodError, TypeError
+    config
+  end
+
   private
 
-  attr_reader :overrides, :repository_owner_name
+  attr_reader :content
 
-  def combined_overrides
-    RuboCop::ConfigLoader.merge(normalized_hound_config, normalized_overrides)
-  rescue TypeError
-    normalized_hound_config
-  end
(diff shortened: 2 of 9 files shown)


app/models/linter/ruby.rb
```diff
@@ -49,18 +49,31 @@ def parsed_source(commit_file)
...
+    def owner_config
+      ConfigBuilder.for(owner_hound_config, "ruby")
+    end
+
+    def owner_hound_config
+      BuildOwnerHoundConfig.run(build.repo.owner)
+    end
+
+    def repo_config
+      ConfigBuilder.for(hound_config, "ruby")
```
gylaz: Worth calling this `repo_hound_config`?

spec/models/linter/ruby_spec.rb
```diff
@@ -66,7 +66,17 @@
...
+            Style/IfUnlessModifier:
+              Description: Favor modifier if/unless usage when you have a single-line body.
+              StyleGuide: https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier
+              Enabled: false
+              MaxLineLength: 80
+          CODE
+
+          build = build_with_stubbed_owner_config(config)
+
+          expect(violations_in(<<-CODE.strip_heredoc, build: build, config: nil)).to eq []
```
gylaz: Can remove the `strip_heredoc` and use `<~` to make this fit.

app/models/linter/ruby.rb
jcmorrow: It comes from `Linter::Base` which I'm trying to avoid changing so that we can introduce changes on a per-linter basis, but yeah I might alias it to that just to be consistent.

app/models/linter/ruby.rb
```diff
@@ -4,12 +4,18 @@ class Ruby < Base
     FILE_REGEXP = /.+\.rb\z/
     RUBY_PARSER_VERSION = 2.3
 
+    def initialize(hound_config:, build:, repository_owner_name: "")
```
gylaz: Could you make this change in `Base`? Given that `repository_owner_name` is always passed for other linters, it might be a very low risk.

jcmorrow: I hadn't thought about it that way but yeah, it is probably very safe.

spec/models/linter/ruby_spec.rb
```diff
@@ -191,7 +120,7 @@ def is_something?
           violations = ["Space inside parentheses detected."]
 
           expect(violations_in(<<-CODE.strip_heredoc)).to eq violations
-            logger( "test")
+            logger( 'test')
```
gylaz: Why did you choose to convert this to single quotes?

spec/models/linter/ruby_spec.rb
```diff
@@ -307,6 +224,18 @@ def two
...
 
+      describe "for is_* method name" do
+        it "returns violations" do
+          violations = ["Rename `is_something?` to `something?`."]
+
+          expect(violations_in(<<~CODE)).to eq violations
+            def is_something?
+              'something'
+            end
+          CODE
```
gylaz: This is a bit nitpicky, but I find that kind of usage of heredoc to be hard to read. What do you think about defining a variable like `code` or `content`, then passing it to `violations_in`?

spec/models/linter/ruby_spec.rb
```diff
@@ -555,44 +396,133 @@ def foo
...
+    context "with legacy hound configuration set on owner" do
+      describe "for single line conditional" do
+        it "returns no violations" do
+          expect(legacy_violations_in(<<~CODE)).to eq []
+            if signed_in? then redirect_to dashboard_path end
+          CODE
+        end
+      end
+
+      describe "for has_* method name" do
```
gylaz: Should this be `context`?

spec/models/linter/ruby_spec.rb
```diff
@@ -555,44 +396,133 @@ def foo
...
+        it "returns no violations" do
+          expect(legacy_violations_in(<<~CODE)).to eq []
+            def has_something?
+              "something"
+            end
+          CODE
+        end
+      end
+
+      describe "when using find" do
```
gylaz: Should this be `context`?

spec/models/linter/ruby_spec.rb
```diff
@@ -555,44 +396,133 @@ def foo
...
-    context "with inline configuration" do
-      context "disabling a cop" do
-        it "does not return a violation" do
-          config = stub_ruby_config(
-            "StringLiterals" => {
-              "EnforcedStyle" => "double_quotes",
-            },
-          )
+    context "with legacy hound configuration set on owner" do
+      describe "for single line conditional" do
```
gylaz: Should this be `context`?

Might be copy/pasta. So feel free to fix or ignore.

spec/models/linter/ruby_spec.rb
jcmorrow: I agree. I tried to follow the style of what was there already, but I really like this form:

    expected_violations = []
    code = "blah"

    violations = violations_in(code)

    expect(violations).to eq(expected_violations)

which is used in a few specs but not in most. I'll see if I can write a macro for switching it.

spec/models/linter/ruby_spec.rb
jcmorrow: it is just copied, but I can still go through and fix them.

spec/models/linter/ruby_spec.rb
jcmorrow: Because now hound doesn't expect double quotes unless you specifically configure it that way 😄

app/services/ruby_config_builder.rb
```diff
@@ -2,50 +2,39 @@ class RubyConfigBuilder
...
+    normalize_config(content)
   end
 
   def normalize_config(config)
     RuboCop::Config.new(config, VIRTUAL_FILENAME).tap do |new_config|
       new_config.add_missing_namespaces
       new_config.make_excludes_absolute
     end
-  rescue NoMethodError
+  rescue NoMethodError, TypeError
```
thorncp: What scenario causes the `TypeError`?

thorncp:
This looks good to me (note: I did not look at the spcs 🙈 ) :+1:

app/services/ruby_config_builder.rb
jcmorrow: It happens when calling `make_excludes_absolute` on things that contain arrays instead of just nested hashes. It tries to access via `[]` and the array tries to cast the `String` into an `Integer`.

PR #1160, opened 2016-06-20
Return empty HoundConfig when owner has no config

jcmorrow:
We've been getting some errors on staging when I set non-sensical values for config repos (things that either can't be parsed by the GitHub API or don't exist). This changes the way `BuildOwnerHoundConfig` works to allow for inputs like that. It also changes all behavior where `BuildOwnerHoundConfig` would previously return nil and instead returns a default `HoundConfig` (this allows us to merge into it without worrying about whether or not it's the correct type). Lastly, instead of trying to save config repos (which were getting validation errors because we don't have their GitHub ID) it just discards the config repo after grabbing the default Hound GitHub token from it.

### DIFF
--- a/app/services/build_owner_hound_config.rb
+++ b/app/services/build_owner_hound_config.rb
@@ -11,10 +11,11 @@ class BuildOwnerHoundConfig
   end
 
   def run
-    if owner.has_config_repo?
-      github = GithubApi.new(user_token)
+    if owner.has_config_repo? && config_repo_reachable?
       commit = Commit.new(config_repo.full_github_name, LATEST_SHA, github)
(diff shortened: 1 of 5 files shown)


app/services/build_owner_hound_config.rb
```diff
@@ -11,22 +11,34 @@ def initialize(owner)
...
+      HoundConfig.new(EmptyCommit.new)
     end
   end
 
   private
 
   attr_reader :owner
 
+  def github
+    @github ||= GithubApi.new(user_token)
```
teoljungberg: Underscore?

gylaz:
LGTM.

PR #1161, opened 2016-06-24
Thoughtbot should use thoughtbot config

jcmorrow:
Migrate thoughtbot owner to use the guides as its owner level config.

### DIFF
--- a/db/migrate/20160624165658_update_thoughtbot_owner_configs.rb
+++ b/db/migrate/20160624165658_update_thoughtbot_owner_configs.rb
@@ -1,0 +1,23 @@
+class UpdateThoughtbotOwnerConfigs < ActiveRecord::Migration
+  def up
+    execute <<~SQL
+      UPDATE
+        owners
+      SET
+        config_repo = 'thoughtbot/guides'
+      WHERE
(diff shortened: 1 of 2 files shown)


db/migrate/20160624165658_update_thoughtbot_owner_configs.rb
```diff
@@ -0,0 +1,23 @@
+class UpdateThoughtbotOwnerConfigs < ActiveRecord::Migration
+  def up
+    execute <<~SQL
+      UPDATE
+        owners
+      SET
+        config_repo = 'thoughtbot/guides'
+      WHERE
+        owners.name = 'thoughtbot'
```
thorncp: `owners.name in ('thoughtbot', 'houndci')` ?

gylaz: Good call on `houndci`. We could test it right away.

thorncp:
LGTM :bank:

db/migrate/20160624165658_update_thoughtbot_owner_configs.rb
jcmorrow: Ah yeah, good call.

PR #1162, opened 2016-06-24
Owner-level CoffeeScript configuration

jcmorrow:
- Merge the owner's coffeescript config (defaults to default
   coffeescript config) with the repo level overrides.
 - Change documentation to reflect owner-level config


app/models/config/coffee_script.rb
```diff
@@ -1,9 +1,13 @@
...
   class CoffeeScript < Base
+    def merge(config)
+      content.deep_merge(config.content)
+    end
+
     private
 
     def parse(file_content)
-      Parser.json(file_content)
+      Config::Parser.json(file_content)
```
gylaz: Why is adding this namespace necessary? This code is in the `Config` namespace, so it should resolve correctly.

app/models/linter/coffee_script.rb
```diff
@@ -40,20 +39,24 @@ def erb?(file)
...
-      default_config.merge(config.content)
+    def merged_config
+      owner_config.merge(repo_config)
     end
 
-    def default_config
-      config = File.read(default_config_file)
-      Config::Parser.json(config)
+    def owner_config
+      ConfigBuilder.for(owner_hound_config, "coffee_script")
```
gylaz: Should this be `coffeescript`?

app/views/pages/configuration.haml
```diff
@@ -89,14 +89,11 @@
       %p
         Hound uses
         #{link_to "RuboCop", "https://github.com/bbatsov/rubocop", target: :blank}
-        with this
-        #{link_to "config", ruby_config_url, target: :blank}
-        by default.
+        the default RuboCopConfig.
```
gylaz: `RuboCop configuration`? Might be valuable linking that to https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml

gylaz:
Looking good!

app/models/config/coffee_script.rb
jcmorrow: Huh, at one point I was getting a namespace conflict when I ran this spec by itself, but it's not happening anymore! I'll remove it.

app/views/pages/configuration.haml
jcmorrow: Good call.

thorncp:
> +100 −796

Yaaaaaasssssssss ✂️

PR #1170, opened 2016-06-27
Clear working workers count in Resque

gylaz:
Resque will show something like "2 of 356 Workers Working" which is
clearly false. This will clear that count, back to sane numbers.


thorncp:
:+1:

PR #1173, opened 2016-07-08
Use Rubocop service

thorncp:
Updates `Linter::Ruby` to remove all the in-line linting code and
instead fall-back to the job based approach provided by `Linter::Base`.

We're moving to job names based on linters, so this introduces
`RubocopReviewJob`, which is scheduled by `Linter::Ruby`. I tried
renaming `Linter::Ruby` and `Config::Ruby` to `Rubocop` and the config
key from `ruby` to `rubocop` (with an alias), but so much spec setup and
assumptions rely on the Ruby name that it wasn't worth including as a
part of this change.

https://trello.com/c/jG5McIv8


thorncp:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;jG5McIv8&#x2F;772-extract-rubocop-linter-to-linters"> Extract RuboCop linter to `linters`</a>

app/models/linter/ruby.rb
```diff
@@ -1,86 +1,11 @@
...
-
-    # This is deprecated in favor of RuboCop's DisplayCopNames option.
-    # Let's track how often we see this and remove it if we see fit.
-    def rubocop_options
-      if merged_config.delete("ShowCopNames")
-        Analytics.new(build.full_github_name).track_show_cop_names
-        { debug: true }
-      end
+    def job_name
+      "RubocopReviewJob"
```
teoljungberg: Why did you opt for this instead of `RubyReviewJob`?

teoljungberg: Reading the PR description explained it.

app/models/linter/ruby.rb
```diff
@@ -1,86 +1,11 @@
...
-
-    def parsed_source(commit_file)
-      RuboCop::ProcessedSource.new(
-        commit_file.content,
-        RUBY_PARSER_VERSION,
-        File.join(Rails.root, commit_file.filename),
-      )
-    end
-
-    def merged_config
```
jcmorrow: So where will we do this now? In the linters repo?

gylaz: Linters can't do this, as they don't have a way of fetching remote config files. We still need to do this in Hound proper.

spec/features/builds_spec.rb
```diff
@@ -41,10 +41,11 @@
     stub_pull_request_comments_request(repo_name, pr_number)
     comment_request = stub_simple_comment_request
     stub_status_requests(repo_name, pr_sha)
+    stub_review_job(RubocopReviewJob)
 
     page.driver.post builds_path, payload: payload
 
-    expect(comment_request).to have_been_requested
+    expect(comment_request).to have_been_requested.at_least_once
```
gylaz: Is it getting called more than once? If so, then why?

spec/models/linter/ruby_spec.rb
```diff
@@ -20,843 +20,48 @@
...
-  )
-    Linter::Ruby.new(
-      hound_config: hound_config,
-      build: build,
-    )
-  end
-
   def stub_ruby_config(config = {})
-    stubbed_ruby_config = double("RubyConfig", content: config)
+    stubbed_ruby_config = double(
```
gylaz: Worth making this an `instance_double` now?

spec/models/style_checker_spec.rb
```diff
@@ -138,6 +99,17 @@
     end
   end
 
+  def pull_request_file_reviews(pull_request)
+    build = build(
+      :build,
+      repo: build(:repo, owner: build(:owner, config_enabled: false)),
```
gylaz: What if we pull out `build(:repo...` into a variable, it might make this fit on one line.

app/models/linter/ruby.rb
thorncp: Made an attempt to bring it back in 15f0fb3bffd767750c6b97075f57ae1f8ecbb736. It was a little janky to get it working well with `Linters::Base` so I moved it into `Config::Ruby`. It feels like a bandaid but I'm not sure how else to do it short of refactoring all the config stuff 😕

app/models/config/ruby.rb
```diff
@@ -1,10 +1,15 @@
...
+      super(hound_config)
+      @owner = owner
+    end
+
     def content
       if legacy?
         hound_config.content
       else
-        parse_inherit_from(super)
+        owner_config_content.merge(parse_inherit_from(super))
```
thorncp: I guess this should do a deep merge???

spec/features/builds_spec.rb
thorncp: It's because `CompleteFileReview` makes comments before `BuildReport.run` finishes, so it sees the build as [complete](https://github.com/houndci/hound/blob/15f0fb3bffd767750c6b97075f57ae1f8ecbb736/app/services/build_report.rb#L13) and comments again. I don't think it's a problem in production (anymore than the existing services), just kind of a pain in the spec here. Once all the linters are moved out of the Rails app, we should be able to remove the commenting parts of `BuildReport`.

app/models/linter/ruby.rb
jcmorrow: I think this approach is fine for now. We can maybe do some other refactoring as we move owner-level configuration out to other linters.

app/models/config/ruby.rb
```diff
@@ -14,6 +19,18 @@ def parse(file_content)
       Parser.yaml(file_content)
     end
 
+    def owner_config_content
+      if @owner.present?
```
gylaz: Do we need this check? In the last year we've had `0` build from repos that don't have an owner.

app/services/ruby_config_builder.rb
```diff
@@ -1,39 +0,0 @@
...
-
-  def normalize_config(config)
-    RuboCop::Config.new(config, VIRTUAL_FILENAME).tap do |new_config|
-      new_config.add_missing_namespaces
-      new_config.make_excludes_absolute
-    end
-  rescue NoMethodError, TypeError
-    RuboCop::Config.new
-  end
-end
```
gylaz: :cry: Goodbye good ol' config merging code.

gylaz:
What's the next step here? Test on staging?

app/models/config/ruby.rb
thorncp: This is a little weird and could probably be cleaned up well with some refactoring. For every repo that we build a config for, an owner will be present, but then we "recurse" and build the owner's config (the line below this one). The owner's config does not have an owner, and is `nil`.

PR #1177, opened 2016-07-17
Extract a method to read fixture file content

abunashir:
We are building response body from the fixtures file in a couple of places and each time we are doing the same thing to read a fixture file content. This commit extracts it to a method which will deal with reading the fixture file content.


spec/support/helpers/github_api_helper.rb
```diff
@@ -257,6 +258,10 @@ def status_request_body(description, state, target_url)
     }
   end
 
+  def response_with_fixture_file(filename)
```
jcmorrow: I like this idea! What do you think about calling the method something like `read_fixture`? When I read `response_with_fixture_file` I wasn't sure how the file was being used.

spec/support/helpers/github_api_helper.rb
```diff
@@ -257,6 +258,10 @@ def status_request_body(description, state, target_url)
...
+    File.read("spec/support/fixtures/#{filename}")
```
jcmorrow: What about using `File.join("spec", "support", "fixtures", filename)` here?

spec/support/helpers/github_api_helper.rb
abunashir: Yes, you are right. I'll update it soon.

abunashir:
@jcmorrow : I updated the code. Any thoughts?

gylaz:
Looks good. I've merged this. Thanks @abunashir!

PR #1179, opened 2016-07-22
Fix Style/CollectionMethods to be consistent with linked style guide

bjeanes:
`#reduce` is preferred over `#inject` as it's a more common name in languages other than SmallTalk.

### DIFF
--- a/config/style_guides/ruby.yml
+++ b/config/style_guides/ruby.yml
@@ -12,7 +12,7 @@ Style/CollectionMethods:
     collect!: map!
     find: detect
     find_all: select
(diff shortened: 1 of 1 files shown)


jcmorrow:
Thanks for the suggestion @bjeanes! Hound is actually no longer using `config/style_guides/ruby.yml` (it will be removed soon as part of  https://github.com/houndci/hound/pull/1162). We will instead be starting off new users with the default rubocop config (which does indeed prefer `reduce` over `inject`).

bjeanes:
Ah cool no worries then.

PR #1180, opened 2016-07-22
Check for conflicting linters + resolve conflicts

jcmorrow:
We now have multiple linters which can lint the same file (e.g. `jshint`
and `eslint`). It probably only makes sense to use one of these at a
time, so Hound now checks to see if there are any conflicts between
enabled linters (specifically if eslint is enabled and javascript is not
explicitly disabled) and disables the conflicts explicitly.

Other changes:

 - Turn on `jshint` by default again (because I believe this is the
   expected behavior)

### DIFF
--- a/app/services/resolve_config_conflicts.rb
+++ b/app/services/resolve_config_conflicts.rb
@@ -1,0 +1,22 @@
+class ResolveConfigConflicts
+  CONFLICTS = { "eslint" => "jshint" }.freeze
+
+  def self.run(config)
+    new(config).run
+  end
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.reduce({}) do |resolved_config, (linter, options)|
+      if CONFLICTS.has_key?(linter) && options["enabled"] == true
+        resolved_config[CONFLICTS[linter]] = { "enabled" => false }
+      else
+        resolved_config[linter] = options
+      end
+      resolved_config
+    end
+  end
+end
--- a/spec/models/hound_config_spec.rb
+++ b/spec/models/hound_config_spec.rb
@@ -2,6 +2,7 @@ require "spec_helper"
 require "app/models/config/parser"
 require "app/services/normalize_config"
 require "app/services/resolve_config_aliases"
+require "app/services/resolve_config_conflicts"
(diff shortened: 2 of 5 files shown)


app/services/resolve_config_conflicts.rb
```diff
@@ -0,0 +1,22 @@
...
+    new(config).run
+  end
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.reduce({}) do |resolved_config, (key, value)|
+      if CONFLICTS.keys.include?(key) && value["enabled"] == true
```
thorncp: `Hash` implements `include?`, so `.keys` isn't necessary.

app/services/resolve_config_conflicts.rb
```diff
@@ -0,0 +1,22 @@
...
+  def self.run(config)
+    new(config).run
+  end
+
+  def initialize(config)
+    @config = config
+  end
+
+  def run
+    @config.reduce({}) do |resolved_config, (key, value)|
```
thorncp: What do you think about naming the block params after what they represent? Something like `(linter, options)`?

spec/models/hound_config_spec.rb
```diff
@@ -78,6 +79,20 @@
       end
     end
 
+    context "when the given language conflicts with an enabled linter" do
+      it "returns false" do
+        commit = stubbed_commit(
```
thorncp: Worth stubbing the `ResolveConfigConflicts::CONFLICTS` constant here so the setup is explicit?

thorncp:
Few questions/suggestions, but LGTM :part_alternation_mark:

app/services/resolve_config_conflicts.rb
jcmorrow: Yeah, I spent about 2 minutes staring at this trying to think of a better name and gave up. `(linter, options)` is not perfect, but it's better than `(key, value)` 😄

spec/models/hound_config_spec.rb
jcmorrow: yes.

thorncp:
🚢

PR #1181, opened 2016-07-29
Include original config options after transform

jcmorrow:
The ResolveConfigConflicts object was working in the sense that it
disables linters which conflict with configured and enabled linters.
However, it was not correctly including the config options for the
configured and enabled linter. An example:

A .hound.yml of:

    eslint:
        enabled: true

would accurately disable `jshint`. However, it would not enable `eslint`.
This change allows it to do both.


thorncp:
LGTM :camel:

PR #1182, opened 2016-08-11
Go, Don't Go

davidsawyer:
The section about Go in the docs ought to read:

## Go

To disable Go style checking, add the following to your `.hound.yml`

```
go:
  enabled: false
```


gylaz:
Thanks @davidsawyer!

PR #1186, opened 2016-08-19
Silence 404 errors when setting status

gylaz:
Sometimes users will change `houndci-bot`'s permissions so that it
cannot set the status of the GitHub PR. This generates a lot of
failure noise in our job queues.

This actually doesn't affect the functionality of Hound that much,
since it will still review the PR and make comments when violations are
found. The PR status just won't be set/updated.

### DIFF
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -139,5 +139,7 @@ class GithubApi
       description: description,
       target_url: target_url
     )
+  rescue Octokit::NotFound
+    # noop
   end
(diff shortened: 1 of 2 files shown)


jcmorrow:
LGTM 👍 I definitely appreciate less noise around failures like this. I'd like to ignore sentry less.

gylaz:
Going to close this, in favor of better grouping on `/admin/job_failures` page.

PR #1191, opened 2016-08-26
Initialize Jshint config with content

jcmorrow:
# Context

Most of our configs are initialized by passing in a `hound_config`. In a
lot of cases this makes sense, but it makes it difficult to do things
like merge configs, which is a crucial part of enabling owner level
configurations.

# Changes

- Make Config::Jshint initialize with a string of content
- Allow `parse` to take other arguments, but default to the`raw_content` that the object was initialized with (similar to how `serialize` works already)

# Future Development

Right now the `JshintConfigBuilder` is overly specialized. I'd like to start expanding this to other linters, slowly making `JshintConfigBuilder` more generic, and then ultimately remove what is the `ConfigBuilder` right now with the generic version of `JshintConfigBuilder`

### DIFF
--- a/app/models/config/jshint.rb
+++ b/app/models/config/jshint.rb
@@ -1,12 +1,24 @@
 module Config
   class Jshint < Base
+    pattr_initialize :raw_content
+
     def serialize(data = content)
       Serializer.json(data)
     end
 
+    def content
+      @_content ||= ensure_correct_type(safe_parse(raw_content))
+    end
+
+    def merge(raw_overrides)
+      parsed_overrides = parse(raw_overrides)
+      merged_content = content.deep_merge(parsed_overrides)
+      Config::Jshint.new(serialize(merged_content))
+    end
+
     private
 
-    def parse(file_content)
+    def parse(file_content = raw_content)
       Parser.json(file_content)
     end
   end
--- a/app/models/jshint_config_builder.rb
+++ b/app/models/jshint_config_builder.rb
@@ -1,0 +1,51 @@
+class JshintConfigBuilder
+  pattr_initialize :hound_config
+
+  def self.for(hound_config)
+    new(hound_config).config
+  end
+
+  def config
+    Config::Jshint.new(load_content)
+  end
+
+  private
+
+  def load_content
+    if file_path
+      if url?
+        fetch_url
+      else
+        commit.file_content(file_path)
+      end
+    else
+      "{}"
+    end
+  end
+
+  def fetch_url
+    response = Faraday.new.get(file_path)
+
+    if response.success?
+      response.body
+    else
+      raise_parse_error("#{response.status} #{response.body}")
+    end
+  end
+
+  def url?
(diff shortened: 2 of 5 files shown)


app/models/jshint_config_builder.rb
```diff
@@ -0,0 +1,61 @@
...
+    Config::Jshint
+  end
+
+  def linter_name
+    "jshint"
+  end
+
+  def load_content
+    if file_path
+      if url?
```
jcmorrow: @gylaz we had talked about removing this feature altogether, do you have any more thoughts on that?

app/models/jshint_config_builder.rb
```diff
@@ -0,0 +1,61 @@
+class JshintConfigBuilder < ConfigBuilder
+  def self.for(hound_config)
+    new(hound_config).config
+  end
+
+  def initialize(hound_config)
```
thorncp: Use `pattr_initialize` for consistency?

thorncp:
This looks good to me :hotsprings:

app/models/jshint_config_builder.rb
jcmorrow: Ah yeah, originally I was doing additional things here so I took it out but this is a good call.

app/models/jshint_config_builder.rb
teoljungberg: I for one really like the feature, and I'll gladly keep it around until we've finished all of "organizational level configuration"

app/models/config/jshint.rb
```diff
@@ -1,12 +1,24 @@
 module Config
   class Jshint < Base
+    pattr_initialize :raw_content
+
     def serialize(data = content)
       Serializer.json(data)
     end
 
+    def content
+      @content ||= ensure_correct_type(safe_parse(raw_content))
```
teoljungberg: Thoughts on using `@_content` here?

app/models/jshint_config_builder.rb
```diff
@@ -0,0 +1,59 @@
+class JshintConfigBuilder < ConfigBuilder
+  pattr_initialize :hound_config
+
+  def self.for(hound_config)
+    new(hound_config).config
+  end
+
+  def config
+    config_class.new(load_content)
```
teoljungberg: Since this method is only used once, what do you think about inlining it?

app/models/jshint_config_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  private
+
+  def config_class
+    Config::Jshint
+  end
+
+  def linter_name
```
teoljungberg: Thoughts on extracting this to a constant?

app/models/jshint_config_builder.rb
jcmorrow: That sounds like a reasonable time to reassess.

teoljungberg:
As discussed in person, I think this is a good first step towards owner level configuration. 

LGTM

PR #1197, opened 2016-09-02
Add static Builds page so users can rerun a build

smharley:
I didn't commit the changes to the `routes` that would be needed to view this page, figured we'd want to keep it hidden

* Refactor font-weight
* Refactor transitions
* Add spacing variables
* Build Builds page with static content

https://trello.com/c/FLrhuBGo

### DIFF
--- a/app/assets/stylesheets/pages/_builds.scss
+++ b/app/assets/stylesheets/pages/_builds.scss
@@ -1,0 +1,73 @@
+.build-button {
+  background-color: $base-background-color;
+  border: 2px solid $base-border-color;
+  border-bottom-left-radius: $base-border-radius;
+  border-bottom-right-radius: $base-border-radius;
+  color: $light-font-color;
+  font-weight: $font-weight-normal;
+  margin: 0 -2px -2px;
+  padding: $small-spacing 0;
+  text-align: center;
+  transition: $base-transition;
+
+  &:hover {
+    background-color: $purple;
+    border-color: $purple;
+    color: $white;
+    cursor: pointer;
+  }
+}
+
+.build-list {
+  margin: $large-spacing auto;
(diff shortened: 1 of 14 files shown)


app/assets/stylesheets/pages/_builds.scss
```diff
@@ -0,0 +1,72 @@
+.build-button {
+  background-color: $base-background-color;
+  border: 2px solid $base-border-color;
+  border-radius: 0 0 $base-border-radius $base-border-radius;
```
smharley: Hmm I'm not sure `border-radius` shorthands work like that Hound.. I'll 👀 into it

smharley:
image

app/assets/stylesheets/pages/_builds.scss
smharley: Are we using the default thoughtbot SCSS style? Shorthands of length `4` should be allowed:

https://github.com/thoughtbot/guides/blob/master/style/sass/.scss-lint.yml#L169

```
  Shorthand:
    enabled: true
    allowed_shorthands: [1, 2, 3, 4]
```

thorncp: We aren't using the file you linked yet. There's a Hound copy at https://github.com/houndci/linters/blob/master/config/scss.yml that has probably gotten out of sync.

smharley: Ah, looks like that's the case:

https://github.com/houndci/linters/blob/master/config/scss.yml#L169

```
  Shorthand:
    enabled: true
    allowed_shorthands: [1, 2, 3]
```

thorncp:
Looks good to me! :mount_fuji:

PR #1198, opened 2016-09-02
Scss Owner Level Config

jcmorrow:
Turns out once it's done for one linter service it gets easier to replicate.


spec/models/linter/scss_spec.rb
```diff
@@ -20,36 +20,91 @@
...
 
   describe "#file_review" do
+    context "when the owner does not have a configuration set up" do
+      it "enqueues a file review with the local config" do
+        build = build(:build, commit_sha: "foo", pull_request_number: 123)
+        config_content = <<~CFG
+          linters:
+            Indentation:
+              width: 2
+          CFG
```
gylaz: Could Hound be freaking bout because this is indented one extra?

spec/models/linter/scss_spec.rb
```diff
@@ -20,36 +20,91 @@
...
+    end
+
+    context "when the owner has a configuration set up" do
+      it "enqueues a file review with the owner config merged with the local" do
+        stubbed_owner_config = stubbed_commit(
+          ".hound.yml" => <<~YML,
+            scss:
+              config_file: .scss.yml
+          YML
+          ".scss.yml" => <<~YML
```
gylaz: Let's try to avoid `<<~` with `=>`, and capture to a var instead. It doesn't look the best either.

app/models/mergeable_config_builder.rb
```diff
@@ -1,16 +1,22 @@
-class JshintConfigBuilder
-  pattr_initialize :hound_config
+class MergeableConfigBuilder
```
gylaz: Does this make more sense as a service? `MergeConfigs.call` or something?

jcmorrow: This is literally just a parallel hierarchy to `ConfigBuilder` until I can remove it completely. I can easily change the name to something else but once everything is ported over to use this it should just become `ConfigBuilder`.

app/models/config/scss.rb
```diff
@@ -1,12 +1,24 @@
...
+    def content
+      @_content ||= ensure_correct_type(safe_parse(raw_content))
+    end
+
+    def merge(raw_overrides)
+      parsed_overrides = parse(raw_overrides)
+      merged_content = content.deep_merge(parsed_overrides)
+      Config::Scss.new(serialize(merged_content))
+    end
+
```
gylaz: I'm officially confused by our Config classes. Who and when calls `content`, who calls `merge`? `serialize`? 😴

app/models/mergeable_config_builder.rb
thorncp: :+1: for a verb'd service.

spec/models/config/scss_spec.rb
```diff
@@ -14,13 +14,40 @@
...
-      config = build_config(commit)
+      config = Config::Scss.new(raw_config)
 
       expect(config.content).to eq Config::Parser.yaml(raw_config)
     end
   end
 
+  describe "#merge" do
+    it "combines the existing configuration with the overrides" do
+      raw_config = <<-EOS.strip_heredoc
```
thorncp: Use squiggly here for consistency?

spec/models/config/scss_spec.rb
```diff
@@ -14,13 +14,40 @@
...
+        linters:
+          BangFormat:
+            enabled: true
+            space_before_bang: true
+            space_after_bang: false
+      EOS
+      raw_overrides = <<-EOS.strip_heredoc
```
thorncp: Use squiggly here for consistency?

app/models/config/scss.rb
jcmorrow: Here's the config interface I'm using:

- It's initialized with a string, which it parses.
- It has a public method called `content` which produces a hash that it parsed from the string
- It has a public method called `serialize` which is the opposite of parse, it turns the content hash into a string
- It has a public method called `merge` which takes a valid config string, merges it into its own content, and then produces a new config with the result of the merge.

spec/models/config/scss_spec.rb
jcmorrow: I will if I figure out how to make hound stop barking 😢

app/models/mergeable_config_builder.rb
jcmorrow: I agree it should be BuildConfig, not sure why it wasn't that in the beginning, but while I'm mirroring the other hierarchy I'd prefer to either not change them or change both to make it clear that this is a temporary parallel.

app/models/linter/scss.rb
```diff
@@ -1,5 +1,21 @@
 module Linter
   class Scss < Base
     FILE_REGEXP = /.+\.scss\z/
+
+    def config
+      owner_config.merge(local_config.serialize)
```
gylaz: Can we standardize on one method linters call on the config? Either `content` or `merge`? Otherwise, it's starting to get harder to keep track of these.

app/models/mergeable_config_builder.rb
thorncp: Did see @jcmorrow 's comment until afterwards. If it's temporary I don't think there's much of a need to change it now. Though, in the end it'd be nice to verbify `ConfigBuilder`.

spec/models/config/scss_spec.rb
thorncp: I'm okay with us ignoring Hound's bark if it means we have consistency .

app/models/config/scss.rb
gylaz: Is anyone using the public `content` method? Maybe `merge` should be private, and `content` could call it?

teoljungberg: I agree with @gylaz, I think `#merge` should live in its respective config class.

spec/models/config/scss_spec.rb
teoljungberg: I agree, consistency above all.

app/models/config/scss.rb
teoljungberg: I'd make `merge` private

app/models/mergeable_config_builder.rb
teoljungberg: I still see this can be moved to a service in the meanwhile and then deprecate `ConfigBuilder` for the newly created service.

spec/models/linter/scss_spec.rb
```diff
@@ -20,36 +20,95 @@
   end
 
   describe "#file_review" do
+    context "when the owner does not have a configuration set up" do
+      it "enqueues a file review with the local config" do
+        build = build(:build, commit_sha: "foo", pull_request_number: 123)
+        config_content = <<~CFG
```
teoljungberg: All heredocs in this file have different annotations. What do you think about using `<<~EOS` for all of them for simplicity.

spec/models/mergeable_config_builder_spec.rb
```diff
@@ -21,7 +21,7 @@
           content: { "jshint" => { "config_file" => ".jshintrc" } },
         )
 
-        config = JshintConfigBuilder.for(hound_config)
+        config = described_class.for(hound_config, "jshint")
```
teoljungberg: We aren't using `described_class` anywhere else in the codebase, thoughts on naming the class by it's name for consistency?

app/models/config/scss.rb
jcmorrow: @gylaz @teoljungberg can you clarify this suggestion? How could you make merge private if you have to take in some content to merge with? It seems to me like if it's private you can only merge with your own content, which is useless.

spec/models/mergeable_config_builder_spec.rb
jcmorrow: Yup, good call.

spec/models/linter/scss_spec.rb
jcmorrow: Good suggestion!

app/models/linter/scss.rb
jcmorrow: So, `merge` and `content` serve quite different purposes, I believe all linters call both.

app/models/config/scss.rb
gylaz: I was thinking modifying the `content` method to take in optional args. Then this particular content could become:
```rb
def content(raw_overrides)
  @_content ||= merge(ensure_correct_type(safe_parse(raw_content)), raw_overrides)
end
```
or something along those lines.

app/models/mergeable_config_builder.rb
teoljungberg: After revisiting this, could we extract the purpose of `MergableConfigBuilder` into its own PR?

app/models/mergeable_config_builder.rb
```diff
@@ -1,16 +1,22 @@
...
 
   def config
-    Config::Jshint.new(load_content)
+    config_class.new(load_content)
   end
 
   private
 
+  def config_class
+    "Config::#{name.classify}".constantize
```
teoljungberg: I think this was the original purpose of having the `ConfigBuilder`. I for one like that abstraction.

spec/models/config/scss_spec.rb
teoljungberg: Since we resolved the yelling from Hound, I think we can rewrite these into squiggly heredocs now

app/models/mergeable_config_builder.rb
jcmorrow: I'd like to get these changes in so that we can officially be using owner level configurations for all of the major linters. I 100% believe that there will be improvements left to make once that process is done, and I'd prefer to make them with more perspective.

teoljungberg:
Is there anything I can do to help you get this merged and going @jcmorrow ?

jcmorrow:
@teoljungberg sorry I totally missed your review. I'm making some changes as per your and @gylaz's suggestions and I'll be trying to get it merged today.

app/models/config/scss.rb
jcmorrow: @gylaz if you make this change then calling

`scss_config.content({})` just returns a copy of `scss_config`, which feels wrong. I suppose we don't actually need the `content` method, so we could rename it to merge? Or maybe everything should just call `serialize` and `serialize can optionally take in some args. I think that might be the neatest way.

jcmorrow:
@teoljungberg I *am* going to pull these out into separate PR's. Closing in favor of that.

teoljungberg:
Let me know if I can be off any assistance, @jcmorrow.

PR #1199, opened 2016-09-02
Store violations count when file review completes

gylaz:
Rather than updating `build.violations_count` after each violation,
do it once after the file review is complete.

This should alleviate the DEADLOCK error, like:
```
PG::TRDeadlockDetected: ERROR: deadlock detected DETAIL: Process 2333
waits for ShareLock on transaction 19601816; blocked by process 2430.
Process 2430 waits for ExclusiveLock on tuple (223360,8) of relation
16407 of database 16385; blocked by process 2333. HINT: See server log
for query details. : UPDATE "builds" SET "violations_count" = $1,
"updated_at" = $2 WHERE "builds"."id" = $3
```
Since we'll be updating that model less frequently during the build.

### DIFF
--- a/app/models/linter/coffee_script.rb
+++ b/app/models/linter/coffee_script.rb
@@ -5,10 +5,19 @@ module Linter
     FILE_REGEXP = /.+\.coffee(\.js)?(\.erb)?\z/
 
     def file_review(commit_file)
-      FileReview.create!(
-        filename: commit_file.filename,
-        linter_name: name,
-      ) do |file_review|
+      create_file_review(commit_file).tap do |file_review|
+        file_review.build.increment!(
+          :violations_count,
+          file_review.violations.map(&:messages_count).sum,
+        )
+      end
+    end
+
+    private
+
+    def create_file_review(commit_file)
+      filename = commit_file.filename
+      FileReview.create!(filename: filename, linter_name: name) do |file_review|
         content = content_for_file(commit_file)
 
         lint(content).each do |violation|
@@ -20,8 +29,6 @@ module Linter
         file_review.complete
       end
     end
-
-    private
 
     def lint(content)
       Coffeelint.lint(content, merged_config)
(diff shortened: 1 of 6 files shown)


app/models/linter/coffee_script.rb
```diff
@@ -21,8 +30,6 @@ def file_review(commit_file)
       end
     end
 
-    private
```
thorncp: Was this intentional?

spec/features/account_spec.rb
```diff
@@ -77,27 +77,21 @@
...
     create(:subscription, repo: paid_repo, user: user)
 
     sign_in_as(user)
 
     visit account_path
 
-    expect(find('td.reviews-given')).to have_text("3");
-    expect(find('td.violations-caught')).to have_text("2");
+    expect(find("td.reviews-given")).to have_text("3");
+    expect(find("td.violations-caught")).to have_text("5");
```
thorncp: Ha, where'd these come from

app/models/linter/coffee_script.rb
gylaz: `private` has moved up to line 16.

spec/features/account_spec.rb
gylaz: haha, 😂

app/models/linter/coffee_script.rb
thorncp: Aha! 👴

thorncp:
Looks good to me. I'm guessing we won't have this problem at all when coffeescript becomes a service.

spec/features/account_spec.rb
gylaz: Oh that @frechg https://github.com/houndci/hound/commit/44e271252cf6efcfd737237c4a7519e78c0364ed#diff-81550665e424c64c3425af0488c548cfR58

PR #1200, opened 2016-09-03
Delay CompleteFileReviewJob on Resque::TermException

gylaz:
This could be causing our double commenting issue.

If we encounter an error, we likely want to delay for a bit.


teoljungberg:
Worth giving a shot! But, the test failure is concerning.

gylaz:
I'm going to close this. I think the issue might be resolved now that we've moved out all the linters from the repo, and not calling `BuildReport` twice.

PR #1201, opened 2016-09-03
Simplify high-level linting logic

gylaz:
The collaboration between `StyleChecker` and `Linter::Collection`
introduced additional complexity.

The new implementation feels easier to understand and is less code.


teoljungberg:
Some background - the idea of `Linter::Collection` was to act as a facade for all of the linters we were going to use. But I think it's fair `StyleChecker` has that responsibility.

app/models/hound_config.rb
```diff
@@ -9,6 +21,10 @@ class HoundConfig
 
   attr_reader_initialize :commit
 
+  def self.linter_names
```
teoljungberg: Could this be a constant instead?

app/models/hound_config.rb
```diff
@@ -27,7 +43,7 @@ def fail_on_violations?
   private
 
   def default_config
-    Linter::Collection.linter_names.each.with_object({}) do |name, config|
+    self.class.linter_names.each.with_object({}) do |name, config|
```
teoljungberg: I prefer writing the class name instead of `self.class`. Your preference.

teoljungberg:
I really like this refactoring!

app/models/hound_config.rb
gylaz: Good idea. Were you thinking:
```rb
LINTER_NAMES = LINTERS.map { |klass| klass.name.demodulize.underscore }
```
or listing them out like `BETA_LINTERS`?

teoljungberg: I thought of the former, which gives us a nice bit of flexiblity.

teoljungberg:
This LGTM, but I'd be happy to hear what @thorncp and @jcmorrow thinks

PR #1205, opened 2016-09-16
Simplify enabled linter checking

thorncp:
Removes `CheckEnabledLinter` in favor of performing the check inline.
`CheckEnabledLinter` was capable of supporting many config objects, each
with their own `HoundConfig`. Over time the actual usage of it has been
reduced to checking a single linter name given a single `HoundConfig`,
which simplifies the process so much that having a dedicated class is
overkill.


gylaz:
I think one of the issues `CheckEnabledLinter` was trying to address was legacy language keys. (e.g `javascript` and `jshint`). I think we want to preserve that functionality.

spec/models/linter/base_spec.rb
```diff
@@ -33,21 +33,19 @@ class Test < Base
   describe "#enabled?" do
     context "when the hound config is enabled for the given language" do
       it "returns true" do
-        allow(CheckEnabledLinter).to receive(:run).and_return(true)
-        linter = build_linter
+        hound_config = instance_double(HoundConfig, linter_enabled?: true)
```
gylaz: We can remove the require of `hound_config` if we use the string versions `"HoundConfig"`.

thorncp: If we use the string we can't use `instance_double` to verify the stubbed method exists. Do you think giving that up is worth dropping a require?

gylaz: Well, that's the point of using the string version. In focused tests the `"HoundConfig"` will "just work" (without verifying), then when the entire suite is run and `HoundConfig` is loaded and present, RSpec will verify the API.

thorncp: Huh, I had no idea that was a feature. Is it documented anywhere?

Updated in the rebase: a1ff362f8f500564512a36199a6e9a81ea8983b5

gylaz: A combination of https://relishapp.com/rspec/rspec-mocks/docs/verifying-doubles and example in https://relishapp.com/rspec/rspec-mocks/docs/verifying-doubles/using-an-instance-double ?

PR #1206, opened 2016-09-16
Use Coffeelint service

thorncp:
* Removes coffeelint gem
* Udpates `Linter::CoffeeScript` and `Config::CoffeeScript` to use the
  service infrastructure
* Updates `Config::CoffeeScript` to be aware of owner config
* Adds `CoffeelintReviewJob`

Note: CoffeeScript is in a similar spot as Ruby, in that all of the
naming on Rails app side (this side) uses the name of the language
(CoffeeScript), while the linters side is based on the name of the
linter (Coffeelint).

https://trello.com/c/pj5BdqFU


app/models/config/coffee_script.rb
```diff
@@ -1,13 +1,34 @@
 module Config
   class CoffeeScript < Base
-    def merge(config)
-      content.deep_merge(config.content)
+    def initialize(hound_config, owner: nil)
+      super(hound_config)
+      @owner = owner
```
gylaz: At this point, does it make sense to have `Base` take in and set the `owner`? If so, probably in another PR.

app/models/linter/coffee_script.rb
```diff
@@ -1,69 +1,17 @@
...
-    end
-
-    def repo_config
-      ConfigBuilder.for(repo_hound_config, "coffee_script")
-    end
-
-    def repo_hound_config
-      hound_config
+    def config
+      Config::CoffeeScript.new(hound_config, owner: build.repo.owner)
```
gylaz: Is it worth using the `Base`'s behavior for this?

app/models/config/coffee_script.rb
thorncp: Yes, I think so. Another PR would be good.

app/models/linter/coffee_script.rb
thorncp: `Base` is using `ConfigBuilder`, so I don't really want try to mesh that together with knowledge of owners in this PR.

gylaz: My thinking was that it would be "standard". Just let the parent do the generic thing, which I believe results in the same thing as what you're doing here. I would likely prefer to remove the `ConfigBuilder` at some point. But right now, it might be good to introduce less deviation.

thorncp: I don't see how `ConfigBuilder` handles owner config. What am I missing?

gylaz: Ah, owner config. Gotcha. That's what I was missing.

thorncp:
> Maybe also remove the BuildReport line from BuildRunner?

Done in https://github.com/houndci/hound/pull/1207

PR #1208, opened 2016-09-19
Add ENV variable for worker timeout

thorncp:


lib/tasks/resque.rake
```diff
@@ -9,7 +9,7 @@ namespace :resque do
     Resque.workers.each do |w|
       worker_start_time = w.processing.fetch("run_at", Time.current).to_time
       time_running = Time.current - worker_start_time
-      max_time_running = 10.minutes
+      max_time_running = ENV.fetch("RESQUE_WORKER_TIMEOUT", 10).to_i.minutes
```
teoljungberg: Since this is in minutes, thoughts on incorporating that into the environment variable? Something like `RESQUE_WORKER_TIMEOUT_IN_MINUTES`.

teoljungberg:
I haven't gotten into using the new GitHub Review workflow, but this looks good to me besides the already laid out remarks.

teoljungberg:
Thoughts on getting this merged @thorncp ?

PR #1214, opened 2016-09-23
Add functionality to remove resolved comments

gylaz:
Since GitHub removed the comment collapsing feature, we need a way to
clean-up comments after the underlying style violation has been
resolved.


gylaz:
Still need to test at the high-level, but ready for review.

app/policies/commenting_policy.rb
```diff
@@ -1,27 +1,40 @@
...
-  def allowed_for?(violation)
+  def comment_on?(violation)
     unreported_violation_messages(violation).any?
   end
 
+  def comment_matches_any_violation?(comment, violations)
+    violations.any? do |violation|
+      comment.user.login == Hound::GITHUB_USERNAME &&
+        matches_location?(violation, comment) &&
+        messages_match?(violation, comment)
```
thorncp: What do you think about naming this method `matches_messages?` to be consistent with `matches_location?`?

app/services/build_report.rb
```diff
@@ -7,11 +7,13 @@ def initialize(pull_request:, build:, token:)
     @build = build
     @pull_request = pull_request
     @token = token
+    @commenter = Commenter.new(pull_request)
```
thorncp: Needs an `attr_reader`?

app/models/pull_request.rb
```diff
@@ -14,13 +16,17 @@ def commit_files
...
-      comment: violation.messages.join("<br>"),
+      comment: violation.messages.join(COMMENT_LINE_DELIMITER),
       commit: head_commit,
       filename: violation.filename,
       patch_position: violation.patch_position
     )
   end
 
+  def delete_comment(comment_id)
+    delete_pull_request_comment(full_repo_name, comment_id)
```
thorncp: Is this method available here?

teoljungberg: Doesn't look like it, we have to use either `hound_github` or `user_github`.

app/policies/commenting_policy.rb
teoljungberg: I like this suggestion 👍

app/services/build_report.rb
teoljungberg: 👍

app/services/commenter.rb
```diff
@@ -4,15 +4,23 @@ class Commenter
 
   def comment_on_violations(violations)
     violations.each do |violation|
-      if commenting_policy.allowed_for?(violation)
+      if policy.comment_on?(violation)
```
teoljungberg: I'd keep this as `commenting_policy` to reflect the class name.

app/services/commenter.rb
```diff
@@ -4,15 +4,23 @@ class Commenter
...
         pull_request.comment_on_violation(violation)
       end
     end
   end
 
+  def remove_resolved_violations(violations)
+    pull_request.comments.each do |comment|
+      unless policy.comment_matches_any_violation?(comment, violations)
```
teoljungberg: Instead of doing an `each` with a conditional inside it, what do you think of a `select` or `reject` followed by an `each` block?

spec/policies/commenting_policy_spec.rb
```diff
@@ -83,17 +83,114 @@
...
+
+        expect(result).to eq false
+      end
+    end
+  end
+
   def stub_comment(options = {})
     defaults = {
-      position: nil
+      user: double(login: Hound::GITHUB_USERNAME),
```
teoljungberg: Thoughts on naming this `double`?

spec/policies/commenting_policy_spec.rb
```diff
@@ -83,17 +83,114 @@
...
     }
-    double(:comment, defaults.merge(options))
+    double(:github_comment, defaults.merge(options))
```
teoljungberg: Thoughts on using the string form as we have in the rest of the code base? `double("GithubComment")`.

spec/services/commenter_spec.rb
```diff
@@ -10,7 +10,7 @@
           pull_request = double(:pull_request, comment_on_violation: nil)
           violation = double(:violation)
           commenter = Commenter.new(pull_request)
-          policy = double(:commenting_policy, allowed_for?: true)
+          policy = double(:commenting_policy, comment_on?: true)
```
teoljungberg: Same suggestion / question as above regarding the string form.

spec/services/commenter_spec.rb
```diff
@@ -45,4 +45,47 @@
       end
     end
   end
+
+  describe "#remove_resolved_violations" do
+    it "deletes comments for violations that are no longer present" do
+      existing_comments = [
+        build_comment(id: 1, position: 4, body: "foo bar"),
```
teoljungberg: Does the `id`'s matter?

app/services/commenter.rb
gylaz: Hmm, I don't see any particular benefit to that, other than functional programming street cred. Seems less efficient.

spec/policies/commenting_policy_spec.rb
gylaz: What do you mean here?

gylaz: Oh, i gotcha. 👍

spec/services/commenter_spec.rb
gylaz: It does, because it's used in the expectation on line 68. I'll see if I can improve the readability.

gylaz:
@teoljungberg @thorncp this is ready for another look.

app/services/commenter.rb
```diff
@@ -4,15 +4,23 @@ class Commenter
...
-      if commenting_policy.allowed_for?(violation)
+      if commenting_policy.comment_on?(violation)
         pull_request.comment_on_violation(violation)
       end
     end
   end
 
+  def remove_resolved_violations(violations)
+    pull_request.comments.each do |comment|
+      if commenting_policy.delete_comment?(comment, violations)
```
teoljungberg: I for one like the functional way of `select` + `each`, I'm unsure if there's a performance hit since it's an enumerable. I'd be interested in doing some benchmarks of that at some point.

spec/policies/commenting_policy_spec.rb
```diff
@@ -83,17 +83,111 @@
...
+
+          expect(result).to eq false
         end
       end
     end
   end
 
+  describe "#delete_comment?" do
+    context "when commented by Hound's user" do
+      context "and messages match" do
```
teoljungberg: I really like the symmetry of how this reads!

spec/policies/commenting_policy_spec.rb
```diff
@@ -83,17 +83,111 @@
...
+        pull_request = stub_pull_request(comments: [comment])
+        commenting_policy = CommentingPolicy.new(pull_request)
+
+        result = commenting_policy.delete_comment?(comment, [violation])
+
+        expect(result).to eq false
+      end
+    end
+
+    context "when messages don't match" do
```
teoljungberg: Thoughts on writing out `don't` into `does not`?

app/models/pull_request.rb
```diff
@@ -14,13 +16,17 @@ def commit_files
...
-      comment: violation.messages.join("<br>"),
+      comment: violation.messages.join(COMMENT_LINE_DELIMITER),
       commit: head_commit,
       filename: violation.filename,
       patch_position: violation.patch_position
     )
   end
 
+  def delete_comment(comment)
+    hound_github.delete_pull_request_comment(full_repo_name, comment.id)
```
thorncp: Should we try using the user's token first? Theoretically we'll be making twice as many API requests, 2 for every violation, so we could end up exhausting Hound's quota. I don't suspect it will be a pressing issue immediately, so we deploy without and see how it affects things.

gylaz: Good point. Let's monitor it and then make the call.

PR #1216, opened 2016-09-26
Set status to success when no files to review

gylaz:
When there are not reviews scheduled for a build, we never set the
status back from pending.

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -24,6 +24,9 @@ class BuildRunner
     upsert_owner
     build = create_build
     review_files(build)
+    if build.file_reviews.empty?
+      set_no_violations_status
+    end
   end
 
   def relevant_pull_request?
@@ -90,5 +93,9 @@ class BuildRunner
       commit_sha: payload.head_sha,
       linter_name: exception.linter_name,
     )
+  end
+
+  def set_no_violations_status
+    commit_status.set_success(0)
   end
 end
--- a/spec/services/build_runner_spec.rb
+++ b/spec/services/build_runner_spec.rb
@@ -223,6 +223,23 @@ describe BuildRunner do
       end
     end
 
+    context "when no files need to be reviewed" do
+      it "sets the final status as passing" do
+        repo = create(:repo, :active)
+        build_runner = make_build_runner(repo: repo)
+        github_api = stubbed_github_api
+        stubbed_pull_request_with_file("foo.whatever", "hello world")
+
+        build_runner.run
+
+        expect(github_api).to have_received(:create_success_status).with(
+          repo.name,
+          stubbed_payload.head_sha,
+          I18n.t(:complete_status, count: 0),
+        )
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


app/services/build_runner.rb
```diff
@@ -24,6 +24,9 @@ def review_pull_request
     upsert_owner
     build = create_build
     review_files(build)
+    if build.file_reviews.empty?
+      commit_status.set_success(0)
```
teoljungberg: Could we extract the violations count of `0` into something meaningful? I had to look at the method to see what it meant.

akutz:
@gylaz,

Thank you for such a quick turn-around! I'm not sure how I haven't encountered this until now. We've been using Hound for a while, but recently I've been working a lot with our Makefile, resulting in non-code changes. Thank you again!

spec/services/build_runner_spec.rb
```diff
@@ -223,6 +223,23 @@ def force_fail_build_creation
...
+        repo = create(:repo, :active)
+        build_runner = make_build_runner(repo: repo)
+        github_api = stubbed_github_api
+        stubbed_pull_request_with_file("foo.whatever", "hello world")
+
+        build_runner.run
+
+        expect(github_api).to have_received(:create_success_status).with(
+          repo.name,
+          "somesha",
```
teoljungberg: I've always been bothered by this magic value in this test.

spec/services/build_runner_spec.rb
```diff
@@ -341,10 +358,11 @@ def stubbed_payload(options = {})
   end
 
   def stubbed_github_api
-    github_api = double(
+    github_api = instance_double(
       "GithubApi",
       create_pending_status: nil,
-      create_error_status: nil
+      create_success_status: nil,
+      create_error_status: nil,
```
teoljungberg: Good cleanup~

app/services/build_runner.rb
gylaz: Would it be better to extract `commit_status.set_success(0)` to a private method of `set_no_violations_status`? Or perhaps make the argument to `set_status` be a kwarg, like `commit_status.set_success(violations_count: 0)`?

I just feel like adding a `NO_VIOATIONS_COUNT = 0` constant is a bit redundant.

Thoughts?

teoljungberg: I think the kwarg idea is nice, but that would mean that the changeset would be larger.

I think a `set_no_violations_status` is a good first step to take!

gylaz:
@akutz This was a recent regression as we moved the last linter from this repo into [`linters`](https://github.com/houndci/linters).

spec/services/build_runner_spec.rb
gylaz: How about https://github.com/houndci/hound/pull/1216/files#diff-d483ebbf053cbd4d49cb92d8b163a425R237 ?

teoljungberg: Love it!

PR #1218, opened 2016-09-30
Support rebuilding a PR

thorncp:
Adds /builds index action, which lists the most recent pull requests
with builds that the current user has access to and provides a button to
rebuild them.

There's no link to this page from the UI at the moment.

rebuild


thorncp:
[Ability to rerun a build](https://trello.com/c/FLrhuBGo/725-ability-to-rerun-a-build)

app/controllers/builds_controller.rb
```diff
@@ -3,6 +3,10 @@ class BuildsController < ApplicationController
   skip_before_action :verify_authenticity_token, only: [:create]
   skip_before_action :authenticate, only: [:create]
 
+  def index
+    render locals: { builds: recent_builds_by_repo }
```
gylaz: Yeah! 💯

config/locales/en.yml
```diff
@@ -43,3 +43,7 @@ en:
       "If Hound finds any violations of your style guide it will
       comment inline on the violation in your GitHub pull
       request."
+
+  rebuilds:
+    new: "Rerun Build"
```
gylaz: What do you think about using Rails view structure for this?
```yaml
builds:
  index:
    rebuild: Rerun Build
    success: Rebuild triggered
```
Then in the view `t('.rebuild')` and `t('.success')` in the controller?

spec/features/user_rebuilds_a_pull_request_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+feature "User rebuilds a pull request" do
+  scenario "from the builds page" do
+    user = create(:user, token_scopes: "public_repo,user:email")
+    repo = create(:membership, user: user).repo
+    build = create(:build, repo: repo)
+    allow(RebuildPullRequest).to receive(:call)
```
gylaz: What do you think about spying on the `SmallBuildJob` here? This way `RebuildPullRequest` gets to be tested at the integration level.

config/routes.rb
```diff
@@ -15,16 +15,19 @@
   get "/sign_out", to: "sessions#destroy"
   get "/configuration", to: "pages#configuration"
   get "/faq", to: "pages#show", id: "faq"
+  get "/recent_builds", to: "pages#show", id: "builds"
```
gylaz: Do we need this? Does this clash with `builds#index` below?

spec/queries/recent_builds_by_repo_query_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        repo: repo1,
+        created_at: 1.hour.ago,
+      )
+      _old_build2 = create(
+        :build,
+        pull_request_number: 2,
+        repo: repo2,
+        created_at: 1.hour.ago,
+      )
+      recent_build2 = create(:build, pull_request_number: 2, repo: repo2)
```
gylaz: Worth putting this line before `_old_build2` so it's obvious that is the old build of this one?

config/locales/en.yml
thorncp: Good call: 4df9eb724cd28e01b8120cbf3fdaf17fa9c2f314

spec/features/user_rebuilds_a_pull_request_spec.rb
thorncp: I like it 4df9eb724cd28e01b8120cbf3fdaf17fa9c2f314

config/routes.rb
thorncp: Don't need it. Forgot to remove it. Also removed the high voltage page in 8e3a1ddc7ef630b249c8d2e3dbc56eee49a9f74d

spec/queries/recent_builds_by_repo_query_spec.rb
thorncp: Sure 4df9eb724cd28e01b8120cbf3fdaf17fa9c2f314

PR #1219, opened 2016-09-30
Remove RuboCop gem

gylaz:
RuboCop is now entrirely handled by the `linters` repo.

Also, remove `astrolabe` since it was needed for RuboCop.


thorncp:
:+1:

PR #1220, opened 2016-09-30
Remove old violation_count method

gylaz:


thorncp:
👍 ✂️

teoljungberg:
Do we really want this? Don't we want to update the build status with the violation count if there are any?

thorncp:
I assumed "old violation_count method" meant it wasn't used anymore, but the Circle failure says otherwise. Do we have a replacement for this method?

gylaz:
There is a `violations_count` column on `builds` that replaces this.

teoljungberg:
@gylaz In that case, we'd need to update the reporter to use that method.

app/services/build_report.rb
```diff
@@ -37,14 +37,14 @@ def track_subscribed_build_completed
...
+      commit_status.set_failure(build.violations_count)
     else
-      commit_status.set_success(build.violation_count)
+      commit_status.set_success(build.violations_count)
     end
   end
 
   def fail_build?
-    hound_config.fail_on_violations? && build.violation_count > 0
+    hound_config.fail_on_violations? && build.violations_count > 0
```
gylaz: Me thinks this is a bad rule for us to follow. ✂️ it?

thorncp: I'd be okay removing it :+1:

thorncp:
LGTM :traffic_light:

PR #1222, opened 2016-10-03
Facelift

sardaukar:
Just a Rubocop once-over


teoljungberg:
We have a policy against rewriting existing code just for the sake of bringing it inline withthe current style guide. But we make those changes when we touch that part of the codebase.

With that in mind, what does this PR resolve?

sardaukar:
@teoljungberg just a once-over, as stated - there were a bunch of places with what seemed easy wins (inline guards, and so on) so I thought of automating it with Rubocop. But is seems that Hound and it don't see eye to eye. Nevermind then, my bad :|

PR #1229, opened 2016-10-17
Replace Angular with ReactJS (first try)

sardaukar:
Complete switchover to React - this is the baseline for me and @LkeMitchll work on tiered pricing. The only feature missing is the animation when the repo gets activated/deactivated, since it will probably need a tweak anyway.


app/assets/javascripts/components/account_credit_card_updater.js.jsx.erb
```diff
@@ -0,0 +1,54 @@
...
+    });
+  }
+
+  render = () => {
+    return (
+      <h3>
+        Monthly Billing
+        {
+          this.props.stripe_customer_id_present
+          ?
```
BlakeWilliams: Instead of using a ternary here, thoughts on extracting an `if/else`?

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,77 @@
...
+          <div className="form-group">
+            <label>Email address for receipts</label>
+            <input
+              type="email"
+              placeholder={this.state.emailAddress || this.props.billable_email}
+              onChange={(evt) => this.setState({emailInput: evt.target.value, addressChanged: null})}
+            ></input>
+            {
+              this.state.addressChanged !== null
+              ?
```
BlakeWilliams: I think this is a good sign to extract another component so we can avoid complicated ternaries in JSX.

app/assets/javascripts/components/organization.js.jsx
```diff
@@ -0,0 +1,39 @@
+class Organization extends React.Component {
+  filterBySearchTerm = (repo) => {
```
BlakeWilliams: This reads like it's filtering but returns a boolean, thoughts on naming it something like `canShow`? or `isHidden`?

app/assets/javascripts/components/organization.js.jsx
```diff
@@ -0,0 +1,39 @@
...
+    } = this.props;
+
+    return (
+      <div className="organization">
+        <header className="organization-header">
+          <h2 className="organization-header-title">{data.name}</h2>
+        </header>
+        <section className="repo_listing">
+          <ul className="repos">
+            {repos && repos.filter(this.filterBySearchTerm).map( (repo) => (
```
BlakeWilliams: Extra space after `.map(`

app/assets/javascripts/components/organization.js.jsx
```diff
@@ -0,0 +1,39 @@
...
+              <Repo
+                repo={repo}
+                key={repo.id}
+                onRepoClicked={onRepoClicked}
+                isProcessingId={isProcessingId}
+              />
+            )) || null}
```
BlakeWilliams: Instead of an `|| null`, I think this could be extracted into a component, or delegate the responsibility of rendering null to the `Repo` component.

app/assets/javascripts/components/repo_tools.js.jsx
```diff
@@ -0,0 +1,27 @@
...
+      isSyncing,
+      onRefreshClicked,
+    } = this.props;
+
+    return (
+      <div className="repo-tools">
+        <RepoToolsSearch onSearchInput={onSearchInput} />
+        {
+          showPrivateButton
+          ?
```
BlakeWilliams: Thoughts on putting the ternary on the same line?

app/assets/javascripts/components/repo_tools_refresh.js.jsx.erb
```diff
@@ -0,0 +1,22 @@
+class RepoToolsRefresh extends React.Component {
+  render = () => {
+    const { isSyncing, onRefreshClicked } = this.props;
+    const buttonText = isSyncing
+      ?
```
BlakeWilliams: Could this be an if/else or extracted into a function?

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,278 @@
...
+      url: `/repos/${repo.id}/activation`,
+      type: "POST",
+      dataType: "text"
+    });
+  }
+
+  commitRepoToState = (repo) => {
+    repoIdx = _.findIndex(this.state.repos, {id: repo.id});
+
+    const newRepos = React.addons.update(
```
BlakeWilliams: I think this could be changed to use `...` instead of `update`

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,278 @@
...
+          showPrivateButton={!has_private_access}
+          onSearchInput={this.onSearchInput}
+          onRefreshClicked={this.onRefreshClicked}
+          onPrivateClicked={this.onPrivateClicked}
+          isSyncing={this.state.isSyncing}
+        />
+        {
+          this.state.isSyncing
+          ?
+            <ReposSyncSpinner/>
```
BlakeWilliams: Extract a component instead of using a ternary here?

vendor/assets/javascripts/es5-shim.js
```diff
@@ -0,0 +1,2061 @@
+/*!
```
teoljungberg: This shim could be installed via [Rails Assets] to avoid vendoring it.

[Rails Assets]: https://rails-assets.org/#/components/es5-shim

app/assets/javascripts/components/account_credit_card_updater.js.jsx.erb
```diff
@@ -0,0 +1,50 @@
...
+      token: this.updateCustomerCreditCard
+    }).open({
+      panelLabel: "Update Card",
+      allowRememberMe: false
+    });
+  }
+
+  render() {
+    if (this.props.stripe_customer_id_present) {
+      return (
```
teoljungberg: I'm unsure what is the convention in React, but I find it difficult to read an if / else-statment followed by the markup.

What do you think about extracting the two branches out into private methods?

app/assets/javascripts/classnames.js
```diff
@@ -0,0 +1,48 @@
+/*!
```
teoljungberg: Couldn't this package also be installed via [Rails Assets]?

[Rails Assets]: https://rails-assets.org/#/components/classnames

app/assets/javascripts/components/account_credit_card_updater.js.jsx.erb
```diff
@@ -0,0 +1,50 @@
...
+  updateCustomerCreditCard(stripeToken) {
+    $.ajax({
+      url: `/credit_card.json`,
+      type: "PUT",
+      data: {card_token: stripeToken.id},
+      dataType: "json"
+    });
+  }
+
+  onUpdateCreditCard(evt) {
```
teoljungberg: I commonly like to spell out `event` instead of using shorthands like `e` or `evt`.

app/assets/javascripts/components/account_credit_card_updater.js.jsx.erb
```diff
@@ -0,0 +1,50 @@
...
+    evt.preventDefault();
+
+    StripeCheckout.configure({
+      key: Hound.settings.stripePublishableKey,
+      image: "<%= image_path('icon.svg') %>",
+      email: $("input[type=email]").attr('placeholder'),
```
teoljungberg: Mixed use of quotes? Should we stick to double quotes?

app/assets/javascripts/components/account_credit_card_updater.js.jsx.erb
```diff
@@ -0,0 +1,50 @@
...
+      allowRememberMe: false
+    });
+  }
+
+  render() {
+    if (this.props.stripe_customer_id_present) {
+      return (
+        <h3>
+          Monthly Billing
+          <a href="#" className="update-card" onClick={(evt) => this.onUpdateCreditCard(evt)}>
```
teoljungberg: Where does `evt` come from in this context? Is it a React thing?

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,59 @@
...
+  onUpdateEmail(evt) {
+    evt.preventDefault();
+
+    $.ajax({
+      url: `/account.json`,
+      type: "PUT",
+      data: {billable_email: this.state.emailInput},
+      dataType: "text",
+      success: () => {
+        this.setState({
```
teoljungberg: Aren't the `{` brackets unnecessary in ES6? Or am I mixing up coffeescript with ES6?

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,59 @@
...
+  render() {
+    return (
+      <article className="account-details">
+        <h3>Update account settings</h3>
+        <form>
+          <div className="form-group">
+            <label>Email address for receipts</label>
+            <input
+              type="email"
+              placeholder={this.state.emailAddress || this.props.billable_email}
```
teoljungberg: Again, not knowing React. I'd extract a method to encapsulate branching like this.

app/assets/javascripts/components/repo.js.jsx
```diff
@@ -0,0 +1,35 @@
...
+          {"repo--active": repo.active},
+          {"repo--processing": isProcessingId === repo.id}
+        )}
+      >
+        <div className="repo-name">
+          {repo.full_github_name}
+        </div>
+        <div className={classNames(
+          "repo-activation-toggle",
+          {"repo-activation-toggle--private": !repo.active && repo.price_in_cents > 0}
```
teoljungberg: Again, I'd extract this out to a method to encapsulate this logic. Perhaps put it on `repo`?

app/assets/javascripts/components/repo.js.jsx
```diff
@@ -0,0 +1,35 @@
...
+        )}>
+          <span className="repo-private-label">
+            Private - ${repo.price_in_dollars}
+          </span>
+          <RepoActivationButton
+            repo={repo}
```
teoljungberg: This reads quite nicely!

app/assets/javascripts/components/repo_list.js.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+    const repoName = repo.full_github_name.toLowerCase();
+    return repoName.indexOf(filterTerm.toLowerCase()) !== -1;
+  }
+
+  render() {
+    const { repos, onRepoClicked, isProcessingId } = this.props;
+
+    if (repos.length > 0) {
+      return (
```
teoljungberg: Again, without knowing React. I'd extract the two if branches out of the `render` method.

app/assets/javascripts/components/repo_tools.js.jsx
```diff
@@ -0,0 +1,21 @@
...
+      onSearchInput,
+      showPrivateButton,
+      isSyncing,
+      onRefreshClicked,
+    } = this.props;
+
+    return (
+      <div className="repo-tools">
+        <RepoToolsSearch onSearchInput={(evt) => onSearchInput(evt)} />
+        { showPrivateButton ? <RepoToolsPrivate /> : null }
```
teoljungberg: Could we instead write out the ternary to a proper if statement?

app/assets/javascripts/components/repo_tools_refresh.js.jsx.erb
```diff
@@ -0,0 +1,22 @@
+class RepoToolsRefresh extends React.Component {
+  render() {
+    const { isSyncing, onRefreshClicked } = this.props;
+    const buttonText = isSyncing
+      ?
```
teoljungberg: Could we write out the ternary to a proper if statement?

app/assets/javascripts/components/repo_tools_refresh.js.jsx.erb
```diff
@@ -0,0 +1,22 @@
...
+        "<%= I18n.t('syncing_repos') %>"
+      :
+        "<%= I18n.t('sync_repos') %>";
+
+    return (
+      <div className="repo-tools-refresh">
+        <button
+          className="repo-tools-refresh-button"
+          disabled={isSyncing ? "disabled" : null}
```
teoljungberg: Again, without knowing React. Could we extract a method out of this branching?

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,285 @@
+class ReposContainer extends React.Component {
+  fetchReposAndOrgs() {
+    $.ajax({
+      url: "/repos.json",
+      type: "GET",
+      dataType: "json",
+      success: (data) => {
+        if (data.length == 0) {
```
teoljungberg: Could this success handler be extracted out to a method of its own?

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,285 @@
...
+        repo,
+        (token) => this.createSubscriptionWithNewCard(repo, token)
+      );
+    }
+  }
+
+  createSubscriptionWithExistingCard(repo) {
+    this.createSubscription(
+      { repo_id: repo.id}
+    ).then( (resp) => {
```
teoljungberg: It seems like we don't add a space after `(` in the rest of this code. Should it be removed?

teoljungberg: This `then` handler also seems to be re-used elsewhere. Worth extracting a common method that can be re-used?

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,285 @@
...
+    });
+  }
+
+  render() {
+    const { has_private_access } = this.props;
+
+    return (
+      <div>
+        <RepoTools
+          showPrivateButton={!has_private_access}
```
teoljungberg: Could we have the opposite as a method? `public_access_only`?

app/views/accounts/show.html.erb
```diff
@@ -0,0 +1,99 @@
...
+    <p class="message">
+      <%= t(
+        "account.bulk_pricing",
+        mail_to: mail_to("hound@thoughtbot.com")
+      ).html_safe %>
+    </p>
+  </aside>
+  <div class="account-content">
+    <article class="billing-breakdown">
+      <%= react_component('AccountCreditCardUpdater',
```
teoljungberg: Should these be uniformly double quotes? (goes for the entire file)

app/views/repos/index.html.erb
```diff
@@ -0,0 +1,11 @@
+<% content_for :page_title, 'Your Repos' %>
```
teoljungberg: Which quotes are we sticking with here? (goes for the entre file)

spec/features/repo_list_spec.rb
```diff
@@ -6,18 +6,21 @@
...
+      full_github_name: "#{user.github_username}/inaccessible-repo",
+    )
     activatable_repo = create(:repo, full_github_name: "thoughtbot/my-repo")
     create(:membership, repo: activatable_repo, user: user, admin: true)
     create(:membership, repo: restricted_repo, user: user, admin: false)
 
     sign_in_as(user)
 
-    within ".repo:nth-of-type(1)" do
+    within "[data-org-name=thoughtbot]" do
```
teoljungberg: Where does the value of `thoughtbot` come from? Could it be clarified somehow?

vendor/assets/javascripts/es5-shim.js
sardaukar: Can you control the order with a rails assets gem? It has to be required before React.

sardaukar: TIL - I hadn't used rails-assets before!

app/assets/javascripts/classnames.js
sardaukar: Same as for `es5-shim`

app/assets/javascripts/components/account_email_updater.js.jsx
sardaukar: You are :D

app/assets/javascripts/components/repo.js.jsx
sardaukar: I'm relatively new to JS, but the `repo` objects come directly from a JSON call - is it not an antipattern to enrich them with methods? I've extracted the class' toggler statement to a variable with a name that tries to make sense.

app/assets/javascripts/components/repo_list.js.jsx
sardaukar: Tell me what you think of the change into separate components.

app/assets/javascripts/components/repos_container.js.jsx.erb
sardaukar: Thanks on the handler being re-used! I find it easier to read arrow-functions with the space after `(` to clearly demark the arguments being passed.

app/assets/javascripts/components/repos_container.js.jsx.erb
sardaukar: I find it much easier to read `showPrivateButton={!has_private_access}` as "show private button if user doesn't have private access". Is `!` considered harmful?

app/views/repos/index.html.erb
sardaukar: Thanks for the quote-spotting. I used an online Haml to ERB converter and didn't remember to check.

sardaukar:
@teoljungberg thanks for the feedback! First big React and ES6 codebase, so keep it coming please.

app/assets/javascripts/components/account_credit_card_updater.js.jsx.erb
BlakeWilliams: > I'm unsure what is the convention in React

In general it's keeping the markup in the `render` function rather than private methods so you  can see what the component is doing at a glance.

app/assets/javascripts/components/repo_activation_button.js.jsx
```diff
@@ -0,0 +1,27 @@
+class RepoActivationButton extends React.Component {
+  render() {
+    const {
+      repo,
+      onRepoClicked,
+      isProcessingId,
+    } = this.props;
+    const disabledState = (isProcessingId === repo.id) ? "disabled" : null;
```
BlakeWilliams: I think this could be nicely extracted into a getter.

eg:

```
get isDisabled() {
  return isProcessingId === repo.id
}
```

Then you can access it like a property `this.disabledState` (no parens). I assume it should be returning a boolean?

app/assets/javascripts/components/repo_tools.js.jsx
```diff
@@ -0,0 +1,27 @@
+class RepoTools extends React.Component {
+  render() {
+    const {
+      onSearchInput,
+      showPrivateButton,
+      isSyncing,
+      onRefreshClicked,
+    } = this.props;
+
+    if (showPrivateButton) {
```
BlakeWilliams: I think this could be nice as a ternary in the JSX below.

app/assets/javascripts/components/repo_tools_refresh.js.jsx.erb
```diff
@@ -0,0 +1,24 @@
+class RepoToolsRefresh extends React.Component {
+  render() {
+    const { isSyncing, onRefreshClicked } = this.props;
+
+    if (isSyncing) {
+      var buttonText = "<%= I18n.t('syncing_repos') %>";
+    } else {
+      var buttonText = "<%= I18n.t('sync_repos') %>";
+    }
+    const isDisabled = isSyncing ? "disabled" : null;
```
BlakeWilliams: Should this be a boolean? Thoughts on extracting into a getter like I mentioned above?

app/assets/javascripts/components/repo_tools_refresh.js.jsx.erb
```diff
@@ -0,0 +1,24 @@
+class RepoToolsRefresh extends React.Component {
+  render() {
+    const { isSyncing, onRefreshClicked } = this.props;
+
+    if (isSyncing) {
```
BlakeWilliams: Also a getter, maybe?

app/assets/javascripts/components/repos_container.js.jsx.erb
BlakeWilliams: You can omit the parens in arrow functions when it accepts only one argument making it `resp =>`. Conventionally there is no space between `(` and `(`.

app/assets/javascripts/application.js
```diff
@@ -1,12 +1,9 @@
...
-//= require angular-resource
 //= require lodash
+//= require jquery3
 //= require namespaced
 //= require_self
+//= require es5-shim
+//= require react
+//= require react_ujs
+//= require components
 //= require_tree .
```
gylaz: Will this line require `./components` as well? Maybe we don't need it here. Alternatively, we can remove `app/assets/javascripts/components.js` and just require `classnames` above this line.

app/assets/javascripts/components.js
```diff
@@ -0,0 +1,2 @@
+//= require classnames
+//= require_tree ./components
```
gylaz: I can't comment on the empty `app/assets/javascripts/components/.gitkeep` file, so I'm leaving this here: Can we remove `app/assets/javascripts/components/.gitkeep`?

app/assets/javascripts/components/account_credit_card_updater.js.jsx.erb
```diff
@@ -0,0 +1,50 @@
...
+      headers: {
+        "X-XSRF-Token": this.props.authenticity_token
+      }
+    });
+  }
+
+  updateCustomerCreditCard(stripeToken) {
+    $.ajax({
+      url: `/credit_card.json`,
+      type: "PUT",
```
gylaz: What is the difference between using backticks and double-quotes around strings (this line and the one above it)?

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,72 @@
...
+            <input
+              type="email"
+              placeholder={placeholder}
+              onChange={
+                (event) => this.setState(
+                  {
+                    emailInput: event.target.value,
+                    addressChanged: null
+                  }
+              )}
```
gylaz: I think the JS community, prefers having opening `(` and `{` on the same line. Thoughts on this format?
```js
onChange={
  (event) => this.setState({
    emailInput: event.target.value,
    addressChanged: null
  })
}
```

app/assets/javascripts/components/account_email_updater_message.js.jsx
```diff
@@ -0,0 +1,25 @@
+class AccountEmailUpdaterMessage extends React.Component {
+  render() {
+    if (this.props.addressChanged == null) {
```
gylaz: Isn't this logical branch the same as the `else` on line 15? Am I missing some condition logic here?

app/assets/javascripts/components/empty_repo_list.js.jsx
```diff
@@ -0,0 +1,7 @@
+class EmptyRepoList extends React.Component {
+  render() {
+    return (
+      <ul className="repos"></ul>
+    );
```
gylaz: Can these three lines be one-lined? Or is that not conventional/idiomatic?

app/assets/javascripts/components/organization.js.jsx
```diff
@@ -0,0 +1,27 @@
+class Organization extends React.Component {
+  render() {
+    const {
+      data,
```
gylaz: What is `data` meant to hold in this case? It looks like only `name` is used from it, does it make sense to make it high-level property like the ones below? (sorry, I'm not familiar with React wording, so am making up my own).

app/assets/javascripts/components/organizations_list.js.jsx
```diff
@@ -0,0 +1,42 @@
+class OrganizationsList extends React.Component {
+  reposForOrg(org) {
+    if ( _.has(org, "id") ) {
```
gylaz: White spaces after `(` and before `)` look odd. Thoughts on removing them?

app/assets/javascripts/components/organizations_list.js.jsx
```diff
@@ -0,0 +1,42 @@
...
+      return _.filter(this.props.repos, (repo) => {
+        return repo.owner.id === org.id;
+      });
+    } else {
+      return _.filter(this.props.repos, (repo) => {
+        return this.orgName(repo.full_github_name) === org.name;
+      });
+    }
```
gylaz: What do you think about moving the `if/else` inside the `_.filter`? Should cause less duplication.

app/assets/javascripts/components/organizations_list.js.jsx
```diff
@@ -0,0 +1,42 @@
...
+      repos,
+      onRepoClicked,
+      filterTerm,
+      isProcessingId,
+      organizations,
+    } = this.props;
+
+    return (
+      <ul className="organizations">
+        {organizations.map( (org) => (
```
gylaz: Remove the space after `(`?

app/assets/javascripts/components/organizations_list.js.jsx
```diff
@@ -0,0 +1,42 @@
...
+          <Organization
+            data={org}
+            key={org.id || org.name}
```
gylaz: In what case would `org` not have an `id`?

app/assets/javascripts/components/organizations_list.js.jsx
```diff
@@ -0,0 +1,42 @@
...
+            repos={(repos && this.reposForOrg(org)) || null}
```
gylaz: Do we need the first `repos &&`? Will `repos` ever be `null`?

app/assets/javascripts/components/populated_repo_list.js.jsx
```diff
@@ -0,0 +1,27 @@
+class PopulatedRepoList extends React.Component {
+  canShow(repo) {
+    const { filterTerm } = this.props;
+
+    if (filterTerm == null) { return true; }
```
gylaz: Thoughts on avoiding this early return, and instead wrapping the logic in an `if/else`?

app/assets/javascripts/components/populated_repo_list.js.jsx
```diff
@@ -0,0 +1,27 @@
...
+    const repoName = repo.full_github_name.toLowerCase();
+    return repoName.indexOf(filterTerm.toLowerCase()) !== -1;
+  }
+
+  render() {
+    const { repos, onRepoClicked, isProcessingId } = this.props;
+
+    return (
+      <ul className="repos">
+        {repos.filter(repo => this.canShow(repo)).map( (repo) => (
```
gylaz: Should that space after `(` be there?

app/assets/javascripts/components/repo_activation_button.js.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+    if (repo.admin) {
+      return (
+        <button
+          className="repo-toggle"
+          onClick={() => onRepoClicked(repo.id)}
```
gylaz: Curious if the `() =>` is necessary. Does something along the lines of `onClick=onRepoClicked(repo.id)` work?

app/assets/javascripts/components/repo_tools_refresh.js.jsx.erb
```diff
@@ -0,0 +1,24 @@
...
+      var buttonText = "<%= I18n.t('syncing_repos') %>";
+    } else {
+      var buttonText = "<%= I18n.t('sync_repos') %>";
```
gylaz: So, the problem with ERB'ing this into a JS file is that when the underlying i18n string changes, but this file doesn't, it won't be re-complied. Thus the changes to your i18n text won't propagate here.

The best way to solve this is to use the `Hound.settings` object to pass both of these strings along. (or just don't use i18n here).

app/assets/javascripts/components/repo_tools_search.js.jsx.erb
```diff
@@ -0,0 +1,14 @@
+class RepoToolsSearch extends React.Component {
+  render() {
+    return (
+      <div className="repo-tools-search">
+        <input
+          className="repo-search-tools-input"
+          placeholder="<%= I18n.t('search_placeholder') %>"
```
gylaz: See https://github.com/houndci/hound/pull/1229/files#r84565331

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,295 @@
...
+        organizations: _.uniqWith(organizations, _.isEqual)
+      });
+
+      this.setState({isSyncing: false});
+    }
+  }
+
+  orgName(full_github_name) {
+    return _.split(full_github_name, "/")[0];
+  }
```
gylaz: This looks duplicated from `OrganizationsList`. Can we extract this and share it?

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,295 @@
...
+      );
+
+    }).catch( (error) => {
+      alert("Your subscription could not be activated.");
+    });
+  }
+
+  createSubscriptionWithNewCard(repo, stripeToken) {
+    this.createSubscription(
+      {
```
gylaz: Put this on the same line as `(`? Same for the closing brace.

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,295 @@
...
+          organizations={this.state.organizations}
+          repos={this.state.repos}
+          filterTerm={this.state.filterTerm}
+          onRepoClicked={(event) => this.onRepoClicked(event)}
+          isProcessingId={this.state.isProcessingId}
+         />
+      </div>
+    );
+  }
+}
```
gylaz: An almost 300 line file... Not sure how common this is in React, but I'm guessing there is away to extract/abstract some of this functionality into a class or two?

app/views/accounts/show.html.erb
```diff
@@ -0,0 +1,99 @@
+<% content_for :page_title, "Your Account" %>
```
gylaz: Commenting on this line, but this is in regards to using ERB. As a team, we've been okay with Haml. I don't think we've made a decision to switch off of it, and I'm not sure that this PR is the right place to make this call. Thoughts?

spec/javascripts/controllers/account_controller_spec.coffee
```diff
@@ -1,44 +0,0 @@
...
-        }
-
-        scope.showStripeForm()
-
-        http.expectPUT("/credit_card", card_token: stripeToken.id).respond(422)
-        callback = stripeCheckout.open.calls.argsFor(0)[1]
-        callback(stripeToken)
-        http.flush()
-
-        expect(window.alert).toHaveBeenCalled()
```
gylaz: I haven't seen any React tests... Is that a thing?

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,295 @@
...
+
+  onPrivateClicked(event) {
+    $.post("/auth/github?access=full");
+  }
+
+  onSearchInput(term) {
+    this.setState({filterTerm: term});
+  }
+
+  track_repo_activated(repo) {
```
gylaz: Camelcase?

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,72 @@
...
+        });
+      },
+      error: () => {
+        this.setState({addressChanged: false});
+      }
+    });
+  }
+
+  render() {
+    const { billable_email } = this.props;
```
gylaz: Do we need this as a `const`? Can we inline it into the next line? Camelcase?

app/assets/javascripts/components/repos_container.js.jsx.erb
```diff
@@ -0,0 +1,295 @@
...
+    if (repo.private) {
+      var eventName = "Private Repo Activated";
+      var price = repo.price_in_dollars;
+    } else {
+      var eventName = "Public Repo Activated";
+      var price = 0.0;
+    }
+
+    window.analytics.track(eventName, {
```
gylaz: I've figured out the failure -- the reason specs are failing. It's due to this line raising an error. In order to fix this, we need to set `SEGMENT_KEY` in `.env` file to some string other than blank, like `your_segment_key`. @sardaukar likely you had it set to something non-blank on your machine and that's why it was passing for you.

app/assets/javascripts/components.js
sardaukar: Yes, I'll get to it on my next cleanup round for this PR

app/assets/javascripts/components/account_email_updater.js.jsx
sardaukar: It's a format ¯\_(ツ)_/¯ :D I'll change it

app/assets/javascripts/components/empty_repo_list.js.jsx
sardaukar: I find it less easy to parse at a glance, but I can change it.

app/assets/javascripts/components/organization.js.jsx
sardaukar: Needs a better name, yeah :|

app/assets/javascripts/components/organizations_list.js.jsx
sardaukar: Again, I find it easier to parse. But I'll conform to the standard, if that's the case.

app/assets/javascripts/components/organizations_list.js.jsx
sardaukar: This is tricky - the test data has `org`s without `id`s - I assume the Angular code also did something similar to appease the specs. Ideally, I'd like to either find out if Github ever does this, and update the spec data to match.

app/assets/javascripts/components/repo_activation_button.js.jsx
sardaukar: Without it, the function will be scoped to `window` and not the component instance.

app/assets/javascripts/components/repos_container.js.jsx.erb
sardaukar: See https://github.com/houndci/hound/pull/1229#discussion_r84656710

app/views/accounts/show.html.erb
sardaukar: I kinda had to change to ERB to silence a couple of Hound issues about lines being too long on the `react_component` call outs, and being unable to break it up with `|`. That made me think Haml is the worst, and a quick change to ERB fixed it. I didn't change _all_ of the views, just those that were a problem.

spec/javascripts/controllers/account_controller_spec.coffee
sardaukar: There was a single Angular spec about the account page - I'm not sure how to proceed for testing in React, but it's something on my list for this PR.

app/assets/javascripts/components/organizations_list.js.jsx
gylaz: > Ideally, I'd like to either find out if Github ever does this, and update the spec data to match.

Agreed. Updating specs to match reality seems like the best approach.

app/views/accounts/show.html.erb
gylaz: HAML allows to break up long lines, (nearly) identical to ERB. I haven't ran into situations where I wasn't able to do it.

spec/javascripts/controllers/account_controller_spec.coffee
gylaz: I'm okay with doing this in a follow up PR (this one is getting quite long). Worth making a card on Trello for this?

sardaukar: That would be awesome!

app/assets/javascripts/components/repo_activation_button.js.jsx
sardaukar: It returns `"disabled"` or `null` for the disabled HTML attribute on the button, but yeah a getter seems better.

app/assets/javascripts/components/repo_tools.js.jsx
sardaukar: Yeah, so I had it like that originally :D I agree it looks better that way

app/assets/javascripts/components/account_email_updater.js.jsx
sardaukar: The snake_case is in use because of https://github.com/houndci/hound/blob/master/app/controllers/accounts_controller.rb#L7 not sure if it's worth the change

app/assets/javascripts/components/account_email_updater_message.js.jsx
gylaz: Thoughts on combining this onto a logic tree with `if/esle if/else`? Might be good to make the `true` check explicit to avoid confusion.

app/assets/javascripts/components/account_credit_card_updater.js.jsx
```diff
@@ -0,0 +1,50 @@
+class AccountCreditCardUpdater extends React.Component {
+  componentWillMount() {
+    $.ajaxSetup({
+      headers: {
+        "X-XSRF-Token": this.props.authenticity_token
+      }
+    });
+  }
```
gylaz: I'm wondering if there's a way to do it for all of the components that need it. Is it possible to extend `React.Component` somehow? @BlakeWilliams any thoughts?

app/assets/javascripts/components/account_credit_card_updater.js.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+  updateCustomerCreditCard(stripeToken) {
+    $.ajax({
+      url: "/credit_card.json",
+      type: "PUT",
+      data: {card_token: stripeToken.id},
```
gylaz: Worth adding a space after `{` and before `}`. I see it done in some places but not others. Would be nice to be consistent.

app/assets/javascripts/components/account_email_updater_message.js.jsx
gylaz: So, what I was thinking is:
```js
if (this.props.addressChanged === true) {
  return (...);
} else if (this.props.addressChanged === true) {
  return (...);
} else {
  return null;
}
```

Thoughts?

app/assets/javascripts/components/organizations_list.js.jsx
```diff
@@ -0,0 +1,42 @@
...
+      organizations,
+    } = this.props;
+
+    return (
+      <ul className="organizations">
+        {organizations.map( org => (
+          <Organization
+            name={org.name}
+            key={org.id || org.name}
+            repos={this.reposForOrg(org) || null}
```
gylaz: Won't `reposForOrg` always return something?

app/assets/javascripts/components/populated_repo_list.js.jsx
```diff
@@ -0,0 +1,29 @@
+class PopulatedRepoList extends React.Component {
+  canShow(repo) {
+    const { filterTerm } = this.props;
+
+    if (filterTerm == null) {
```
gylaz: Should this be `===`?

app/assets/javascripts/components/repo_tools.js.jsx
```diff
@@ -0,0 +1,25 @@
...
+      onRefreshClicked,
+    } = this.props;
+
+    return (
+      <div className="repo-tools">
+        <RepoToolsSearch onSearchInput={(event) => onSearchInput(event)} />
+        {
+          showPrivateButton
+          ? <RepoToolsPrivate />
+          : null
```
gylaz: Does this read better on the same line?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+    return _.split(name, "/")[0];
+  }
+
+  state = {
+    isSyncing: false,
+    isProcessingId: null,
+    filterTerm: null,
+    repos: [],
+    organizations: []
+  }
```
gylaz: Worth moving assignments to the top of class?

app/assets/javascripts/components/repo_tools_refresh.js.jsx
```diff
@@ -0,0 +1,31 @@
+class RepoToolsRefresh extends React.Component {
+  get isSyncing() {
+    return this.props.isSyncing;
+  }
+
+  get disabledState() {
```
gylaz: Thoughts on calling this `state()` and returning `"enabled"` instead of `null`?

app/assets/javascripts/components/repo_tools_refresh.js.jsx
```diff
@@ -0,0 +1,31 @@
...
+    return (this.isDisabled ? "disabled" : null);
```
gylaz: Where does `isDisabled` come from?

app/assets/javascripts/components/repo_tools_refresh.js.jsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  render() {
+    const { onRefreshClicked } = this.props;
+
+    if (this.isSyncing) {
+      var buttonText = Hound.settings.syncingButtonText;
```
gylaz: `var` to `let`?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+        alert("Your repos failed to sync.");
+      }
+    });
+  }
+
+  onPrivateClicked(event) {
+    $.post("/auth/github?access=full");
+  }
+
+  onSearchInput(term) {
```
gylaz: I see us pass both `event` and `event.target.value` to this. We should make sure we're consistent.

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  createSubscriptionWithExistingCard(repo) {
+    this.createSubscription({
+      repo_id: repo.id
+    }).then( resp => {
+      this.activateAndTrackRepoSubscription(
+        repo, resp.stripe_subscription_id
+      );
+
```
gylaz: Remove this newline?

app/assets/javascripts/components/repo_tools_search.js.jsx
```diff
@@ -0,0 +1,14 @@
+class RepoToolsSearch extends React.Component {
+  render() {
+    return (
+      <div className="repo-tools-search">
+        <input
+          className="repo-search-tools-input"
+          placeholder={Hound.settings.searchPlaceholder}
+          type="text"
+          onChange={(event) => this.props.onSearchInput(event.target.value)}
```
gylaz: Thoughts on making `onSearchInput` take `event`? Then we can do `onChange={this.onSearchInput}`.

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+  activateRepo(repo) {
+    return $.ajax({
+      url: `/repos/${repo.id}/activation`,
+      type: "POST",
+      dataType: "text"
+    });
+  }
+
+  commitRepoToState(repo) {
+    repoIdx = _.findIndex(this.state.repos, {id: repo.id});
```
gylaz: Or maybe this should be a `const`.

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+      this.trackRepoActivated(repo);
+      this.commitRepoToState(repo);
+    }).catch( () => {
+      alert("Your repo could not be enabled.");
+    });
+  }
+
+  onRepoClicked(id) {
+    this.setState({isProcessingId: id});
+    let repo = _.find(this.state.repos, {id: id});
```
gylaz: `let` to `const`?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  handleSync() {
+    $.ajax({
+      url: "/user.json",
+      type: "GET",
+      dataType: "json",
+      success: data => {
+        if (data.refreshing_repos) {
+          setTimeout(() => { this.handleSync() }, 1000);
```
gylaz: `setTimeout(this.handleSync, 1000)`?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  onRefreshClicked(event) {
+    this.setState({isSyncing: true});
+
+    $.ajax({
+      url: "/repo_syncs.json",
+      type: "POST",
+      dataType: "text", // to trigger success() on 201 and empty response
+      success: () => {
```
gylaz: `success: this.handleSync`?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+    this.setState({filterTerm: term});
+  }
+
+  trackRepoActivated(repo) {
+    if (repo.private) {
+      var eventName = "Private Repo Activated";
+      var price = repo.price_in_dollars;
+    } else {
+      var eventName = "Public Repo Activated";
+      var price = 0.0;
```
gylaz: `var`s to `let`s or `const`s?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+  render() {
+    const { has_private_access } = this.props;
+
+    return (
+      <div>
+        <RepoTools
+          showPrivateButton={!has_private_access}
+          onSearchInput={(event) => this.onSearchInput(event)}
+          onRefreshClicked={(event) => this.onRefreshClicked(event)}
+          onPrivateClicked={(event) => this.onPrivateClicked(event)}
```
gylaz: Pass the function directly on these 3 lines `...={this.onPrivateClicked}`?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+          isSyncing={this.state.isSyncing}
+        />
+        <ReposView
+          isSyncing={this.state.isSyncing}
+          organizations={this.state.organizations}
+          repos={this.state.repos}
+          filterTerm={this.state.filterTerm}
+          onRepoClicked={(event) => this.onRepoClicked(event)}
```
gylaz: `onRepoClicked={this.onRepoClicked}`?

app/assets/javascripts/components/repos_view.js.jsx
```diff
@@ -0,0 +1,28 @@
...
+      return (
+        <ReposSyncSpinner/>
+      );
+    } else {
+      return (
+        <OrganizationsList
+          organizations={organizations}
+          repos={repos}
+          filterTerm={filterTerm}
+          onRepoClicked={(evt) => onRepoClicked(evt)}
```
gylaz: `onRepoClicked={onRepoClicked}`?

app/views/builds/index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <%= form_for :rebuild, url: repo_rebuilds_path(build.repo) do |form| %>
+          <%= form.hidden_field :pull_request_number do %>
+            value: build.pull_request_number
+          <% end %>
+          <%= form.submit t(".new"), class: "build-button" %>
+        <% end %>
+      </div>
+    <% end %>
+  <% end %>
+</ul>
```
gylaz: These Haml to ERB changes are really not related to this work and are making this PR bigger than it needs to be.

app/assets/javascripts/components/account_credit_card_updater.js.jsx
sardaukar: I have a branch up with webpack support that mitigates this - but it would be _even more_ changes in, so that will address this issue at a future time

app/assets/javascripts/components/repo_tools.js.jsx
sardaukar: there are many opinions :D

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
+class ReposContainer extends React.Component {
+  fetchReposAndOrgs() {
+    $.ajax({
+      url: "/repos.json",
+      type: "GET",
+      dataType: "json",
+      success: data => {
```
sardaukar: yeah, it's a holdover I think

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,293 @@
...
+        this.onFetchReposAndOrgsSuccess(data);
+      },
+      error: () => {
+        alert("Your repos failed to load.");
+      }
+    });
+  }
+
+  onFetchReposAndOrgsSuccess(data) {
+    if (data.length == 0) {
```
sardaukar: 👍

app/assets/javascripts/components/repos_container.js.jsx
sardaukar: yes, `const`

app/assets/javascripts/components/account_credit_card_updater.js.jsx
```diff
@@ -0,0 +1,50 @@
...
+    });
+  }
+
+  onUpdateCreditCard(event) {
+    event.preventDefault();
+
+    StripeCheckout.configure({
+      key: Hound.settings.stripePublishableKey,
+      image: Hound.settings.iconPath,
+      email: $("input[type=email]").attr("placeholder"),
```
sardaukar: this is just because the placeholder value is always set to the correct one - on load, the current user email is passed as a prop and the placeholder is set to it, and when the user updates her email, it gets synced to the placeholder as well. Moving it to state would be clearer

app/assets/javascripts/components/populated_repo_list.js.jsx
```diff
@@ -0,0 +1,29 @@
...
+      return repoName.indexOf(filterTerm.toLowerCase()) !== -1;
+    }
+  }
+
+  render() {
+    const { repos, onRepoClicked, isProcessingId } = this.props;
+
+    return (
+      <ul className="repos">
+        {repos.filter(repo => this.canShow(repo)).map( repo => (
```
sardaukar: yeah, going overboard with anon functions :D

app/assets/javascripts/components/repo_activation_button.js.jsx
```diff
@@ -0,0 +1,30 @@
+class RepoActivationButton extends React.Component {
+  get disabledState() {
```
sardaukar: I think calling anything `state` in a React codebase is not a good way to go

app/assets/javascripts/components/repo_tools_private.js.jsx
```diff
@@ -0,0 +1,19 @@
+class RepoToolsPrivate extends React.Component {
+  render() {
+    const { onPrivateClicked } = this.props;
+
+    return (
+      <div className="repo-tools-private">
+        <form className="button_to" method="post" action="/auth/github?access=full">
```
sardaukar: I'll refer to @purinkle for this one...

app/assets/javascripts/components/repo_tools_refresh.js.jsx
```diff
@@ -0,0 +1,31 @@
+class RepoToolsRefresh extends React.Component {
+  get isSyncing() {
+    return this.props.isSyncing;
+  }
```
sardaukar: Why would that be better?

gylaz: Not sure, but that seems to be a pattern here to extract things from `this.props` in that matter. A getter is useful for computed variables, but looks unnecessary here since `isSyncing` isn't going to change during this page's life time... is it?

app/assets/javascripts/components/account_credit_card_updater.js.jsx
purinkle: Resolved in d6c595b

app/assets/javascripts/components/organizations_list.js.jsx
purinkle: Resolved in 268a1a3

app/assets/javascripts/components/populated_repo_list.js.jsx
purinkle: Resolved in 555908a

app/assets/javascripts/components/repo_tools.js.jsx
purinkle: Resolved in 775c4b1

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 68bd6cc

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in c00d279

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 7b37166

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 56a2872

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+    $.ajax({
+      url: "/account.json",
+      type: "PUT",
+      data: {billable_email: this.state.emailInput},
+      dataType: "text",
+      success: () => {
+        this.setState({
+          addressChanged: true,
+          emailAddress: this.state.emailInput
```
purinkle: Resolved in 1963715

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,69 @@
...
+                addressChanged: null
+              })}
+            ></input>
+            <AccountEmailUpdaterMessage
+              addressChanged={this.state.addressChanged}
+            />
+          </div>
+          <div className="form-actions">
+            <button className="button-small" onClick={
+              event => this.onUpdateEmail(event)
```
purinkle: Resolved in 58fa9d0

app/assets/javascripts/components/populated_repo_list.js.jsx
purinkle: Resolved in aae8bcf

app/assets/javascripts/components/populated_repo_list.js.jsx
```diff
@@ -0,0 +1,29 @@
...
+          <Repo
+            repo={repo}
+            key={repo.id}
```
purinkle: [It is recommended](https://facebook.github.io/react/docs/lists-and-keys.html#keys) that keys should be given to the elements inside an array to give the elements a stable identity.

app/assets/javascripts/components/repo_tools_private.js.jsx
purinkle: Resolved in 6dce241

app/assets/javascripts/components/repo_tools_refresh.js.jsx
purinkle: Resolved in 34e39a8

app/assets/javascripts/components/repo_tools_refresh.js.jsx
purinkle: Resolved in ae86ab0

app/assets/javascripts/components/repo_tools_refresh.js.jsx
purinkle: Resolved in ae86ab0

app/assets/javascripts/components/repo_tools_refresh.js.jsx
purinkle: Resolved in f4a397d

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 48035a0

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 24b81e7

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 743b209

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 2833396

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in b893e27

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in b8cf1c5

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,64 @@
+class AccountEmailUpdater extends React.Component {
+  state = {
```
purinkle: I have tried declaring this with both a `let` and a `var`.

React did not seem to like the changes and started to get syntax errors.

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,64 @@
...
+    });
+  }
+
+  onUpdateEmail(event) {
+    event.preventDefault();
+
+    $.ajax({
+      url: "/account.json",
+      type: "PUT",
+      data: {billable_email: this.state.emailInput},
```
purinkle: Resolved in a3ff3b7

app/assets/javascripts/components/account_email_updater.js.jsx
```diff
@@ -0,0 +1,64 @@
...
+      dataType: "text",
+      success: () => {
+        this.setState({ addressChanged: true });
+      },
+      error: () => {
+        this.setState({addressChanged: false});
```
purinkle: Resolved in a3ff3b7

app/views/repos/index.haml
```diff
@@ -6,4 +6,9 @@
   = render 'onboarding'
 
 %section.organization-list{ 'organization-list' => '' }
-= render 'preload_template', template: 'organization_list'
+= react_component( |
```
purinkle: Resolved in 2adfe0a

app/assets/javascripts/components/account_email_updater.js.jsx
sardaukar: @gylaz this is a class property, not a standard variable

Gemfile
```diff
@@ -35,6 +33,9 @@ gem "sass-rails"
 gem "split", require: "split/dashboard"
 gem "stripe"
 gem "uglifier", ">= 2.7.2"
+gem "react-rails"
```
teoljungberg: Thoughts on :abc: sorting these gems?

Gemfile
```diff
@@ -47,14 +48,14 @@ group :development, :test do
   gem "dotenv-rails"
   gem "foreman"
   gem "jasmine-rails"
-  gem "poltergeist"
   gem "rspec-rails", ">= 3.4"
   gem "bundler-audit", require: false
+  gem "quiet_assets"
```
teoljungberg: Thoughts on :abc: sorting these gems?

app/assets/javascripts/components/repo_tools_refresh.js.jsx
```diff
@@ -0,0 +1,25 @@
+class RepoToolsRefresh extends React.Component {
+  render() {
+    const { isSyncing, onRefreshClicked } = this.props;
+
+    let buttonText = null;
+
+    if (isSyncing) {
+      buttonText = Hound.settings.syncingButtonText;
```
teoljungberg: worth extracting a `buttonText` method that we can use instead?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,290 @@
...
+        "X-XSRF-Token": this.props.authenticity_token
+      }
+    });
+    this.setState({isSyncing: true});
+    this.fetchReposAndOrgs();
+  }
+
+  createSubscription(options) {
+    return $.ajax({
+      url: `/repos/${options.repo_id}/subscription.json`,
```
teoljungberg: Could `options` be named `subcriptionOptions` or something along those lines?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -0,0 +1,290 @@
...
+  onPrivateClicked(event) {
+    $.post("/auth/github?access=full");
+  }
+
+  onSearchInput(event) {
+    this.setState({filterTerm: event.target.value});
+  }
+
+  trackRepoActivated(repo) {
+    let eventName = null, price = null
```
teoljungberg: Missing a semicolon?

Gemfile
purinkle: Resolved in cfd1cf8

Gemfile
purinkle: Resolved in cfd1cf8

app/assets/javascripts/components/repo_tools_refresh.js.jsx
purinkle: Resolved in 7a9d975

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 29846c7

app/assets/javascripts/components/repos_container.js.jsx
purinkle: Resolved in 30162ea

app/assets/javascripts/components/account_email_updater_message.js.jsx
teoljungberg: @purinkle This might be worth exploring.

PR #1231, opened 2016-10-20
New pricing plans added to homepage

LkeMitchll:
**Before:**
screen shot 2016-10-20 at 4 41 30 pm

**After:**
screen shot 2016-10-20 at 4 41 48 pm


app/views/shared/_plan.html.erb
```diff
@@ -0,0 +1,16 @@
...
+      Up to
+      <strong><%= allowance %></strong>
+      Repos
+    <% end %>
+  </div>
+  <div class="plan-price">
+    <%= price %>
+    <small>month</small>
+  </div>
+</div>
```
gylaz: I know there's some talk about replacing Haml with Erb, but until we decide that as a team, I think we should stick with Haml. Mixing the two isn't idea.

teoljungberg: I think @sardaukar will replace all HAML with ERB with his work on Reactifying the Angular part of the app.

esparta:
I know it sounds like breaking KISS here, but... How about move this logic far from the view?
Having the plans (with possible changing values) into a yaml:
```yaml
opensource:
  name: Hound
  price: $0
private:
  - name: Chihuahua
    price: $49
    allowance: 4

  - name: Labrador
    price: $99
    allowance: 10

  - name: Great Dane
    price: $249
    allowance: 30
```

Then send the values as object/collection on `HomeController`:
```ruby
class HomeController < ApplicationController
  # ...
  def index
     # This should be moved to a class
     plans = YAML.load(File.read(plans.yml))
     opensource = OpenStruct.new(plans["opensource"].merge(opensource: true))
     privates = plans["private"].map do |plan|
        OpenStruct.new(plan.merge(opensource: false))
     end
     render locals: { opensource: opensource, privates: privates}
  end 
```
On the private block:
```ruby
<%= render partial: "shared/plan", collection: privates %>
```
On the opensource block:
```ruby
<%= render partial: "shared/plan", locals: { repo: opensource } %
``` 
And the partial:
```erb
<div class="plan <%= repo.open_source ? 'plan-open-source' : '' %>">
  <h5 class="plan-title"><%= repo.name %></h5>
  <div class="plan-allowance">
    <% if repo.open_source %>
      Unlimited
    <% else %>
      Up to
      <strong><%= repo.allowance %></strong>
      Repos
    <% end %>
  </div>
  <div class="plan-price">
    <%= repo.price %>
    <small>month</small>
  </div>
</div>
```

With this if you ever change the plans (new prices, new tiers), the changes should be done on the yaml file, not the view.

teoljungberg:
@esparta Valid points, without having been a part of building this out. Here's my take on it:

It makes sense to move the plans into the backend. Starting with keeping them in a class or a YAML file, and moving towards keeping the plans in the DB when they change often enough.
But we don't know that they will change. So to start off with having them in the view, until proven other otherwise, is fine reasoning to me.

app/views/shared/_plan.html.erb
sardaukar: @gylaz @teoljungberg yep, that's been done on the base `ba-reactify` branch that serves as the baseline for this work

sardaukar:
merged into #1237

PR #1234, opened 2016-10-20
WIP: Add pricing tier change page

LkeMitchll:
screen shot 2016-10-20 at 5 14 23 pm

This is mostly styling of static markup (hence WIP). Needs to be integrated with pricing implementation once that's finished, also many need converting to react components.

**Notes:**

- This is currently completely static, so I'm simply mocking this up on a view and a simple route.
- This page will be displayed when the user changes plans, up or down-grade.
- As before plans need to be marked current (see #1232)
- Plans on this screen also need to be marked as 'new', much like the current marker - the `<span class="new-plan">` should be present as should the `.new` class.
- Clicking 'Upgrade' will trigger the Stripe modal to change the price paid.
- Clicking 'cancel' will deactivate the clicked repo and return to the repo list.

**Todo:**
- [ ] Convert from a mockup page and remove temporary routing

**Depends on #1232 and #1233 so rebase from those.**


ilyakatz:
is there a reason why the pricing is not showing up anywhere on the site?

gylaz:
@ilyakatz If you're logged out you can see pricing at the bottom of https://houndci.com

If you are logged in, you can see the pricing by clicking on your user avatar or visiting `https://houndci.com/account`

PR #1235, opened 2016-10-22
Rename a few fields on repos and users

gylaz:
I've been meaning to rename these, as the longer versions are always a
pain to type (when dealing with customer support, and debugging).

There shouldn't be any confusion using these new, shorter names.

* `full_github_name` -> `name`
* `email_address` -> `email`
* `github_username` -> `username`


teoljungberg:
Yes! And thank you.

LGTM

PR #1238, opened 2016-10-28
Make nav on repos index sticky

LkeMitchll:
hound

When the user has many repos they cannot see the allowance indicator at the top of the page when scrolled, this solves by 'sticking' the nav to the screen top on scroll.

### DIFF
--- a/app/assets/javascripts/sticky-header.js
+++ b/app/assets/javascripts/sticky-header.js
@@ -1,0 +1,5 @@
+$( document ).ready(function() {
+  $(window).on("scroll touchmove", function () {
+    $('.global-header.repo-index').toggleClass('shrunk', $(document).scrollTop() > 0);
+  });
+});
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/sticky-header.js
```diff
@@ -0,0 +1,5 @@
+$( document ).ready(function() {
+  $(window).on("scroll touchmove", function () {
+    $('.global-header.repo-index').toggleClass('shrunk', $(document).scrollTop() > 0);
+  });
+});
```
gylaz: Should we make this use React?

app/views/application/_header.html.erb
```diff
@@ -1,56 +1,58 @@
...
-      </li>
-      <li class="auth-container">
-        <%= render "application/auth_button" %>
-      </li>
-    <% end %>
-  </ul>
-</header>
+        <li>
+          <%= link_to account_path, class: "account" do %>
+            <% if current_user.email.present? %>
```
gylaz: We only have 354 users out of 13013 on production without email. Thoughts on not having this `if` check and always showing the avatar (which will be the default Gravatar for those 2.7% of users)?

purinkle:
Included as part of #1244

PR #1239, opened 2016-10-28
Add markup for open source on account page

LkeMitchll:
screen shot 2016-10-28 at 2 43 07 pm

This page previously had no mockup for the open-source plan.

There may be a better way to implement the partials than the clunky looking way I am doing it now. Looking for feedback on that.

### DIFF
--- a/app/views/accounts/show.html.erb
+++ b/app/views/accounts/show.html.erb
@@ -4,9 +4,37 @@
   <aside class="account-nav">
     <h3>Plans</h3>
 
-    <%= render "shared/plan_vertical", new: false, current: true, name: "Chihuahua", allowance: "4", price: "$49" %>
-    <%= render "shared/plan_vertical", new: false,  current: false, name: "Labrador", allowance: "10", price: "$99" %>
-    <%= render "shared/plan_vertical", new: false,  current: false, name: "Great Dane", allowance: "30", price: "$249" %>
+    <%= render "shared/plan_vertical",
+      new: false,
+      current: false,
+      open_source: true,
+      name: "Hound",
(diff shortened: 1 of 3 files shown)


app/views/accounts/show.html.erb
```diff
@@ -4,9 +4,37 @@
...
     <h3>Plans</h3>
 
-    <%= render "shared/plan_vertical", new: false, current: true, name: "Chihuahua", allowance: "4", price: "$49" %>
-    <%= render "shared/plan_vertical", new: false,  current: false, name: "Labrador", allowance: "10", price: "$99" %>
-    <%= render "shared/plan_vertical", new: false,  current: false, name: "Great Dane", allowance: "30", price: "$249" %>
+    <%= render "shared/plan_vertical",
+      new: false,
+      current: false,
+      open_source: true,
+      name: "Hound",
```
gylaz: What about calling this one `Open Source`, so it's obvious that the plan only applies to those repos? Right now it looks like we have a plan that costs `$0` for unlimited amount repos.

app/views/shared/_plan_vertical.html.erb
```diff
@@ -12,9 +12,13 @@
...
     <div class="plan-allowance">
-      Up to
-      <strong><%= allowance %></strong>
-      Repos
+      <% if open_source %>
+        Unlimited
+      <% else %>
+        Up to
+        <strong><%= allowance %></strong>
+        Repos
```
gylaz: Worth clarifying `Private Repos`?

purinkle:
Included as part of #1244

PR #1245, opened 2016-11-04
Remove comment deletion feature

gylaz:
GitHub brought back outdated comment collapsing feature that we relied
on previously.

Comment removal does have an advantage over comment collapsing:
if a violation was resolved by another line in the file, GitHub would
not collapse it since the underlying line didn't change (but our comment
deletion logic would remove it).

However, the disadvantage is making more requests and using more of our
API request limit. We've been hitting API rate limit recently and this
is a much more sensitive issuer for us, since it keeps our customers
from getting reviews.


spec/support/helpers/configuration_helper.rb
```diff
@@ -1,5 +0,0 @@
-module ConfigurationHelper
-  def spy_on_file_read
```
teoljungberg: Is this change also related?

app/models/analytics.rb
```diff
@@ -51,15 +51,6 @@ def track_build_completed(repo)
     )
   end
 
-  def track_show_cop_names
```
gylaz: No, but I was trying to be lazy (and also remove unneeded code). I'll move to another PR.

PR #1247, opened 2016-11-08
Fix hangovers from the removal of Angular

purinkle:
Previously, we had attempted to remove Angular from the application, which had left us with some obsolete code and a few incorrect side effects. Removed all traces of Angular and fixed the side effects.


app/views/accounts/show.haml
```diff
@@ -8,13 +8,10 @@
...
   .account-content
     %article.billing-breakdown
-      %h3
-        Monthly Billing
-        - if current_user.stripe_customer_id.present?
-          %a{ href: '#', class: 'update-card', 'ng-controller' => 'accountController', 'ng-click' => 'showStripeForm()' }
-            %i.fa.fa-credit-card
-            Update Credit Card
-
+      = react_component("AccountCreditCardUpdater",
```
teoljungberg: I normally like to write these as:

```ruby
= react_component(
  "AccountCreditCardUpdater",
  # ...
)
```

gylaz: I believe Haml/Slim won't let you break each argument on its own line, so this is rather common. You can do `= react_compotent(\` however.

teoljungberg: Good to know!

I'm fine leaving this as is.

thorncp:
🎊

PR #1251, opened 2016-11-17
Add a new Stripe migration task

purinkle:
Previously, the Stripe migration task was using the Hound DB as its canonical source, which was causing discrepancies between the Hound DB and the Stripe DB. Update the migration task to use the Stripe DB as the canonical source and bring the Hound DB in-line.


lib/migrate_tiered_subscriptions.rb
```diff
@@ -3,44 +3,61 @@ class NoSuchTierError < StandardError; end
...
   DISCOUNT_COUPON_ID = "tiered_pricing_existing".freeze
 
-  def initialize(user)
-    @user = user
-  end
-
-  def self.migrate!(user)
-    new(user).migrate!
+  def self.migrate!
+    new.migrate!
```
r00k: I think it'd be very useful to retain the ability to run this for just a single user for testing.

lib/migrate_tiered_subscriptions.rb
```diff
@@ -3,44 +3,61 @@ class NoSuchTierError < StandardError; end
...
 
   private
 
-  def customer
-    Stripe::Customer.retrieve(@user.stripe_customer_id)
-  end
+  def update_subscriptions_for(customer)
+    user = User.find_by_stripe_customer_id(customer.id)
+
+    if user
```
r00k: This would be an odd failure, yes? Should we log this somehow? Even to stdout for later investigation?

lib/migrate_tiered_subscriptions.rb
```diff
@@ -3,44 +3,61 @@ class NoSuchTierError < StandardError; end
...
+      subscriptions = customer.subscriptions.all(limit: 1)
```
r00k: All...but limited to 1?

lib/migrate_tiered_subscriptions.rb
```diff
@@ -3,44 +3,61 @@ class NoSuchTierError < StandardError; end
...
+
+      subscriptions.each do |stripe_subscription|
+        user.subscriptions.each_with_index do |subscription, index|
+          if index.zero?
```
r00k: Rather than mutating an existing local subscription like this, what if we made a new one and blew away all the old ones?

lib/migrate_tiered_subscriptions.rb
r00k: If it's just 1, do we need the `subscriptions.each do |stripe_subscription|` below?

lib/migrate_tiered_subscriptions.rb
purinkle: Split this out into two classes so that individual customers can be migrated.

lib/migrate_tiered_subscriptions.rb
purinkle: It would be, but I have seen it on staging.

lib/migrate_tiered_subscriptions.rb
purinkle: After looking into it, we actually need to keep all of the subscriptions. I have amended the class so that we now update each subscription appropriately.

lib/migrate_stripe_subscription.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def current_tier
+    user.current_tier
+  end
+
+  def customer_id
+    customer.id
+  end
+
+  def delete_stripe_subscriptions(subscriptions = stripe_subscriptions)
```
gylaz: Are we ever passing anything to this method? I got a bit confused between User's subscriptions (from db) and subscriptions from Stripe here.

lib/migrate_stripe_subscription.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def update_db(id)
+    subscriptions.each do |subscription|
+      subscription.update_attribute(:stripe_subscription_id, id)
+    end
+  end
+
+  def user
+    User.find_by_stripe_customer_id(customer_id)
```
gylaz: `User.find_by(stripe_customer_id: customer_id)`?

lib/migrate_stripe_subscription.rb
```diff
@@ -0,0 +1,66 @@
...
+      subscriptions.each(&:delete)
+    end
+  end
+
+  def plan_id
+    current_tier.id
+  end
+
+  def stripe_subscriptions
+    customer.subscriptions.all(limit: 1)
```
gylaz: Any reason to fetch 1 at a time?

purinkle: This is to do with the way that the Stripe gem handles pagination. We only really care about one of the existing subscriptions. This grabs the first one and gives us a handler to delete the rest. There was no obvious way of just popping one off the stack and getting rid of the others.

lib/migrate_stripe_customers.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    while customers.has_more
+      run
+    end
+  end
+
+  private
+
+  def all_customers
+    Stripe::Customer.all(limit: 100)
```
purinkle: It looks like we have an old version of the Stripe gem. The `all` method doesn't exist in the version we are using.

lib/migrate_stripe_customers.rb
```diff
@@ -0,0 +1,23 @@
+class MigrateStripeCustomers
+  def self.run
+    new.run
+  end
+
+  def run(customers = all_customers)
+    customers.each { |customer| migrate_subscriptions_for(customer) }
+
+    while customers.has_more
+      run
```
purinkle: This line calls the method we are already in

purinkle:
Merged.

PR #1252, opened 2016-11-18
Loop through Stripe elements

purinkle:
Previously, the Rake task to migrate Stripe customers onto the new payment plans only looped through the first 100 customers, which meant we never updated the other 800+ customers and the script got stuck in an infinite loop. Update the Rake tasks to pagination through the rest of the Stripe elements.


r00k:
👍🏻

purinkle:
Merged.

PR #1255, opened 2016-11-23
Remove references to old pricing structure

LkeMitchll:
Before:
screen shot 2016-11-23 at 11 31 58 am

After:
screen shot 2016-11-23 at 11 32 10 am

The onboarding flash has a reference to the old pricing structure.

Removed the explicit dollar amount, replaced with a link to the account page pricing breakdown

Also removes the price tags next to the 'activate' buttons:

Before:
screen shot 2016-11-23 at 4 55 55 pm

After:
screen shot 2016-11-23 at 4 56 00 pm

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -34,9 +34,9 @@ en:
     step_one_alt_title: "Hound is active on %{count} of your repos!"
     step_one_desc:
       "Hound watches for pull requests on all active repos.
-      Activating public repos costs $0; private repos cost $12."
-    step_two_title: "Open a GitHub pull request on the repo."
-    step_two_alt_title: "Open a pull request on an active repo."
+      Activating public repos is free; private repos are priced by tier."
+    step_one_link: "Details"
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -34,9 +34,9 @@ en:
     step_one_alt_title: "Hound is active on %{count} of your repos!"
     step_one_desc:
       "Hound watches for pull requests on all active repos.
-      Activating public repos costs $0; private repos cost $12."
-    step_two_title: "Open a GitHub pull request on the repo."
-    step_two_alt_title: "Open a pull request on an active repo."
+      Activating public repos is free; private repos are priced by tier."
+    step_one_link: "Details"
+    step_two_title: "Open a GitHub pull request on the active repo."
```
LkeMitchll: I would argue that there was no need for two step-two titles, we should always assume the user has activated a  repo to get to the next step. This is a neater combination.

PR #1256, opened 2016-11-23
Bind `this` to email update event

purinkle:
Previously, `this` was not bound to the email update event, which meant the resulting function was called in the correct context and users could not update their email address. Bound `this` to the email update.

https://trello.com/c/FjeiUk41

### DIFF
--- a/spec/support/user_on_page.rb
+++ b/spec/support/user_on_page.rb
@@ -1,0 +1,20 @@
+require "capybara/dsl"
+
+class UserOnPage
+  include Capybara::DSL
+
+  def update(email_address)
+    fill_in "email_address", with: email_address
+    click_on "Update Email"
+  end
+
+  def updated?
+    flash_element.has_text? "Email address updated!"
+  end
+
+  private
+
+  def flash_element
+    find("[data-role='flash']")
+  end
+end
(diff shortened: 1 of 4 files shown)


purinkle:
[Can't update billing email](https://trello.com/c/FjeiUk41/792-can-t-update-billing-email)

spec/support/user_on_page.rb
```diff
@@ -0,0 +1,20 @@
...
+class UserOnPage
+  include Capybara::DSL
+
+  def update(email_address)
+    fill_in "email_address", with: email_address
+    click_on "Update Email"
+  end
+
+  def updated?
+    flash_element.has_content? "Email address updated!"
```
gylaz: `has_content?` is an alias of `has_text?`. What do you think about using the shorter, non-alias version?

spec/support/user_on_page.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  private
+
+  def flash_element
+    find("[data-role='flash']")
+  end
+end
```
gylaz: Beautiful!

teoljungberg: I really like this extraction!

PR #1257, opened 2016-11-24
[WIP] Using webpack, and Jest for tests

sardaukar:


.gitignore
```diff
@@ -5,3 +5,7 @@
 /log/*.log
 /tmp
 dump.rdb
+
+node_modules/
+app/assets/webpack/app-bundle.js
```
teoljungberg: Is the location of these files configurable? If so, maybe putting them under `tmp` is a better solution?

teoljungberg: Webpack also might be something we can pull from Rails Assets?

sardaukar: Webpack is a whole complex tool that kinda likes things its own way :D The location of the files is configurable, yes - why do you think we should move them to `tmp`?

teoljungberg: Mainly because then we don't have to ignore something in `app`.

teoljungberg: Oh well, it was just a thought.

app/assets/javascripts/components/__tests__/notify_tier_change-test.jsx
```diff
@@ -0,0 +1,52 @@
+import NotifyTierChange from '../notify_tier_change.js';
+
+let repo = {id: 1, name: "Test repo"};
+let plans = [
+  {name: "Chihuahua", price: 49, allowance: 4},
+  {name: "Labrador", price: 99, allowance: 10},
+  {name: "Great Dane", price: 249, allowance: 30},
```
purinkle: Add space inside literal object definitions.

app/assets/javascripts/components/__tests__/organization-test.jsx
```diff
@@ -0,0 +1,23 @@
+import Organization from '../organization.js';
+
+it('renders an organization with ID appropriately (no repo processing)', () => {
+  const org = {
+    id: 1,
+    name: "Test org"
+  }
+  const repos = [
+    {id: 1, name: "Test repo"}
```
purinkle: Add space inside literal object definitions.

app/assets/javascripts/components/__tests__/repos_container-test.jsx
```diff
@@ -0,0 +1,84 @@
...
+//     />
+//   );
+
+//   wrapper.find(".repo-toggle").simulate('click');
+
+//   expect(spy.calledOnce).toBe(true);
+
+//   fetchReposStub.restore();
+//   activateRepoStub.restore();
+// });
```
purinkle: Are you keeping this?

app/assets/javascripts/components/__tests__/update_account_email-test.jsx
```diff
@@ -0,0 +1,50 @@
...
+//     />
+//   );
+
+//   wrapper.find('.button-small').simulate('click');
+
+//   expect(wrapper.find('.fa-check')).to.have.length(1);
+//   expect(wrapper.props().addressChanged).toBe(true);
+
+//   //jQueryStub.restore();
+// });
```
purinkle: And this.

app/assets/javascripts/components/__tests__/repos_container-test.jsx
sardaukar: Sadly, no. I'll leave it up to round 2 of coverage upping :|

app/assets/javascripts/components.jsx
```diff
@@ -0,0 +1,59 @@
...
+app.RepoToolsRefresh = RepoToolsRefresh;
+app.RepoToolsSearch = RepoToolsSearch;
+app.ReposContainer = ReposContainer;
+app.ReposSyncSpinner = ReposSyncSpinner;
+app.ReposView = ReposView;
+app.TierPlan = TierPlan;
+app.UpdateAccountCreditCard = UpdateAccountCreditCard;
+app.UpdateAccountEmail = UpdateAccountEmail;
+app.UpdateAccountEmailMessage = UpdateAccountEmailMessage;
+app.UpgradeSubscriptionLink = UpgradeSubscriptionLink;
```
gylaz: Is this a generated file? Where does it get used?

app/assets/javascripts/components/__tests__/__snapshots__/empty_repo_list-test.jsx.snap
```diff
@@ -0,0 +1,4 @@
+exports[`test renders an empty unordered list 1`] = `
+<ul
+  className="repos" />
+`;
```
gylaz: Is this a snap for tests? When do these get updated? How do we know to update it so it doesn't get out of sync with the actual view?

app/assets/javascripts/components/__tests__/__snapshots__/upgrade_subscription_link-test.jsx.snap
```diff
@@ -0,0 +1,8 @@
+exports[`test renders appropriately 1`] = `
+<a
+  className="repo-toggle tier-change-accept"
+  href="javascript:void(0);"
+  onClick={[Function]}>
+  Upgrade
+</a>
+`;
```
gylaz: Is testing with these snaps actually useful? Could an alternative be unit testing pure JS React functions, then using feature testing to test the UI?

app/assets/javascripts/components/empty_repo_list.js.jsx
```diff
@@ -3,3 +3,5 @@ class EmptyRepoList extends React.Component {
     return (<ul className="repos"></ul>);
   }
 }
+
+module.exports = EmptyRepoList;
```
gylaz: Why do we need to start doing this? Is this for tests?

app/assets/javascripts/components/__tests__/repos_container-test.jsx
```diff
@@ -0,0 +1,42 @@
...
+    />
+  );
+  const textInput = wrapper.find('.repo-search-tools-input');
+  textInput.simulate('change', {target: {value: "new text"}});
+
+  expect(wrapper.state('filterTerm')).toBe('new text');
+});
+
+it('fetches repos and organizations on mount', () => {
+  let stub = sinon.stub(ReposContainer.prototype, 'fetchReposAndOrgs');
```
moimael: Any reason for using sinon instead of jest here ? Jest has awesome mocking support and it you would need one less dependency.
https://facebook.github.io/jest/docs/mock-functions.html

circle.yml
```diff
@@ -1,3 +1,7 @@
+machine:
+  node:
+    version: 7.1.0
```
moimael: Maybe use 7.4.0 here ?

package.json
```diff
@@ -0,0 +1,92 @@
...
+  "scripts": {
+    "build": "webpack --config webpack.config.babel.js",
+    "build:watch": "webpack -w --config webpack.config.babel.js",
+    "build:prod": "webpack -p --config webpack.config.babel.js",
+    "test": "jest app/assets/javascripts/components/__tests__/*-test.jsx",
+    "test:watch": "jest --watch app/assets/javascripts/components/__tests__/*-test.jsx",
+    "test:coverage": "jest --coverage app/assets/javascripts/components/__tests__/*-test.jsx"
+  },
+  "engines": {
+    "node": "7.1",
```
moimael: node 7.4.0 ?

app/assets/javascripts/components.jsx
sardaukar: This places the components in the `app` namespace, so they can be referenced by both Rails and Jest test runner environments.

app/assets/javascripts/components/__tests__/__snapshots__/upgrade_subscription_link-test.jsx.snap
sardaukar: I'm told it's the next best thing after pre-sliced bread :D https://facebook.github.io/jest/blog/2016/07/27/jest-14.html

app/assets/javascripts/components/empty_repo_list.js.jsx
sardaukar: It's needed for the test runner environment, yes.

app/assets/javascripts/components.jsx
```diff
@@ -0,0 +1,59 @@
...
+import UpdateAccountEmail from './components/update_account_email.js';
+import UpdateAccountEmailMessage from './components/update_account_email_message.js';
+import UpgradeSubscriptionLink from './components/upgrade_subscription_link.js';
+
+import Ajax from './lib/ajax.js';
+
+import React from 'react';
+import ReactAddonsUpdate from 'react-addons-update';
+import _ from 'lodash';
+import $ from 'jquery';
```
sardaukar: I'm using it for AJAX calls, to avoid polyfills for `fetch`

app/assets/javascripts/components/__tests__/jestsetup.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+};
+
+// Make Enzyme functions available in all test files without importing
+import { shallow, render, mount } from 'enzyme';
+global.shallow = shallow;
+global.render = render;
+global.mount = mount;
+
+// Skip createElement warnings but fail tests on any other warning
```
sardaukar: Got it from here https://hackernoon.com/testing-react-components-with-jest-and-enzyme-41d592c174f#.t49fmax29

app/assets/javascripts/components/__tests__/notify_tier_change-test.jsx
```diff
@@ -0,0 +1,52 @@
...
+let repo = {id: 1, name: "Test repo"};
+let plans = [
+  { name: "Chihuahua", price: 49, allowance: 4 },
+  { name: "Labrador", price: 99, allowance: 10 },
+  { name: "Great Dane", price: 249, allowance: 30 },
+]
+
+function make_active_plan(target_name) {
+  return _.map(plans, (plan) => {
+    if (plan.name == target_name) {
```
sardaukar: Good one, missed it! About the simplification, I find it harder to parse personally :D

app/assets/javascripts/components/__tests__/repos_container-test.jsx
sardaukar: Stemming from tutorials I got on it. I'll go over and see if Jest (in its new version) can take on these responsabilities now

circle.yml
sardaukar: Sure

app/assets/javascripts/components/__tests__/repos_container-test.jsx
sardaukar: Actually, I think I'll spend some time trying to use Redux, since I think "plain" React gets really messy quickly with its "props down the tree" approach to coherent state on a parent component. In the meantime, I think merging this in would help.

app/assets/javascripts/components.jsx
sardaukar: It's mostly a disdain for them, and the anxiety and sense of dread for everything breaking when I replace it :| And when it inevitably breaks, the fear some polyfill won't be as documented as jQuery itself...

app/assets/javascripts/components/__tests__/repos_container-test.jsx
moimael: Yes I think it's a good idea. I was actually thinking of suggesting it but it seemed out of scope for this pull request.

app/assets/javascripts/components.jsx
moimael: https://github.com/github/fetch

It's kind of the goto polyfill for fetch, it follows the spec very closely so you can just ditch it when every browser support fetch and it should work. It already use the native browser fetch implementation if available. We use it for quiet some time now and it's very stable and well documented. (the api isn't that big anyway) And it's well maintained. Otherwise check axios if the polyfill is really something you don't want. Unminified JQuery is 250KB, Axios 39KB and fetch polyfill 12KB. That's quite a huge diff.

PR #1258, opened 2016-11-24
Update header with tier information

purinkle:
Previously, the header was only updated with the latest tier information when the user refreshed the page, which meant it was not obvious to the user what the current state of their payment plan was. Updated the activation and deactivation logic to update the allowance element in the header appropriately.

https://trello.com/c/EJLrc4J5

### DIFF
--- a/app/assets/javascripts/components/repo_allowance.js.jsx
+++ b/app/assets/javascripts/components/repo_allowance.js.jsx
@@ -1,0 +1,16 @@
+class RepoAllowance extends React.Component {
+  render() {
+    const { subscribedRepoCount, tierAllowance } = this.props;
+
+    return(
+      <div className="allowance">
+        { "Private Repos " }
+        <strong>
+          <span data-role="subscribed-repo-count">{subscribedRepoCount}</span>
+          { " / " }
+          <span data-role="tier-allowance">{tierAllowance}</span>
+        </strong>
+      </div>
+    )
+  }
+}
--- a/app/assets/javascripts/components/repos_container.js.jsx
+++ b/app/assets/javascripts/components/repos_container.js.jsx
@@ -102,6 +102,7 @@ class ReposContainer extends React.Component {
       repo.active = false;
       repo.stripe_subscription_id = null;
       this.commitRepoToState(repo);
+      this.updateSubscribedRepoCount();
     }).catch( () => {
       alert("Your repo could not be disabled.");
     });
@@ -127,12 +128,36 @@ class ReposContainer extends React.Component {
     }
   }
 
+  updateSubscribedRepoCount() {
+    this.getUser().then(user => {
+      const subscribedRepoCount = user.subscribed_repo_count;
+      const tierAllowance = user.tier_allowance;
+
+      if (subscribedRepoCount === 0) {
+        $("[data-role='allowance-container']").remove();
+      } else if (subscribedRepoCount === 1 && $(".allowance").length === 0) {
+        ReactDOM.render(
+          <RepoAllowance
+            subscribedRepoCount={subscribedRepoCount}
+            tierAllowance={tierAllowance}
+          />,
+          $("[data-role='account-actions']")
+            .prepend("<li data-role='allowance-container'></li>")
+            .children()[0]
+        );
+      } else {
+        $("[data-role='subscribed-repo-count']").text(subscribedRepoCount);
+        $("[data-role='tier-allowance']").text(tierAllowance);
+      }
+    });
+  }
+
   activateAndTrackRepoSubscription(repo, stripeSubscriptionId) {
     repo.active = true;
     repo.stripe_subscription_id = stripeSubscriptionId;
     this.trackRepoActivated(repo);
-
     this.commitRepoToState(repo);
+    this.updateSubscribedRepoCount();
   }
 
   onSubscriptionError(repo, error) {
@@ -214,17 +239,20 @@ class ReposContainer extends React.Component {
     });
   }
 
-  handleSync() {
-    $.ajax({
+  getUser() {
+    return $.ajax({
       url: "/user.json",
       type: "GET",
       dataType: "json",
-      success: data => {
-        if (data.refreshing_repos) {
-          setTimeout(this.handleSync.bind(this), 1000);
-        } else {
-          this.fetchReposAndOrgs();
-        }
+    });
+  }
+
+  handleSync() {
+    this.getUser().then(data => {
+      if (data.refreshing_repos) {
+        setTimeout(this.handleSync.bind(this), 1000);
+      } else {
+        this.fetchReposAndOrgs();
       }
     });
   }
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -1,7 +1,22 @@
 class UserSerializer < ActiveModel::Serializer
-  attributes :id, :username, :card_exists, :refreshing_repos
+  attributes(
+    :card_exists,
+    :id,
+    :refreshing_repos,
+    :subscribed_repo_count,
+    :tier_allowance,
+    :username,
+  )
 
   def card_exists
     object.stripe_customer_id.present?
+  end
+
+  def subscribed_repo_count
+    object.subscribed_repos.count
+  end
+
+  def tier_allowance
+    object.current_tier.allowance
   end
 end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -48,6 +48,7 @@ FactoryGirl.define do
       in_organization true
       private true
     end
+    trait(:private) { private true }
 
     sequence(:name) { |n| "user/repo#{n}" }
     github_id
@@ -58,15 +59,20 @@ FactoryGirl.define do
   factory :user do
     username { generate(:github_name) }
 
+    trait(:with_github_scopes) { token_scopes "public_repo,user:email" }
     trait(:stripe) { stripe_customer_id "cus_2e3fqARc1uHtCv" }
   end
 
   factory :membership do
+    trait(:admin) { admin true }
+    trait(:private) { association :repo, :private }
+
     user
     repo
   end
 
   factory :subscription do
+    trait(:active) { association :repo, :active }
     trait(:inactive) { deleted_at { 1.day.ago } }
 
     sequence(:stripe_subscription_id) { |n| "stripesubscription#{n}" }
--- a/spec/features/listing_pricings_spec.rb
+++ b/spec/features/listing_pricings_spec.rb
@@ -66,6 +66,6 @@ feature "Listing Pricings" do
     wait_for_ajax
 
     expect(current_path).to eq repos_path
-    expect(page).to have_content "Private Repos 5/10"
+    expect(page).to have_text "Private Repos 5 / 10"
   end
 end
--- a/spec/features/user_activates_a_repo_spec.rb
+++ b/spec/features/user_activates_a_repo_spec.rb
@@ -1,0 +1,44 @@
+require "rails_helper"
+
+RSpec.feature "User activates a repo", :js do
+  scenario "user upgrades from free tier" do
+    callback_endpoint = "http://#{ENV['HOST']}/builds"
+    user = create(:user, :with_github_scopes, :stripe)
+    membership = create(:membership, :admin, :private, user: user)
+    current_plan = user.current_tier.id
+    upgraded_plan = user.next_tier.id
+    repo = membership.repo
+    token = "letmein"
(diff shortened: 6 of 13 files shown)


purinkle:
[Update allowance on subscription change](https://trello.com/c/EJLrc4J5/793-update-allowance-on-subscription-change)

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -127,12 +128,35 @@ class ReposContainer extends React.Component {
     }
   }
 
+  updateSubscribedRepoCount() {
+    this.getUser().done(user => {
+      let subscribed_repo_count = user.subscribed_repo_count;
+      let tier_allowance = user.tier_allowance;
```
gylaz: What about making these `const`s? Also, camel case?

app/serializers/user_serializer.rb
```diff
@@ -1,7 +1,32 @@
...
+
+  def tier_allowance
+    current_tier.allowance
+  end
+
+  private
+
+  def current_tier
+    object.current_tier
+  end
```
gylaz: Why is this method needed? Can `current_tier` be added to `attributes` list at the top?

app/serializers/user_serializer.rb
```diff
@@ -1,7 +1,32 @@
...
+
+  def subscribed_repos
+    object.subscribed_repos
+  end
```
gylaz: Is this necessary? Isn't it hiding a LOD violation?

spec/features/user_activates_a_repo_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+RSpec.feature "User activates a repo", :js do
+  scenario "user upgrades from free tier" do
+    callback_endpoint = "http://#{ENV['HOST']}/builds"
+    user = create(:user, :github, :stripe)
+    username = Hound::GITHUB_USERNAME
```
gylaz: What is the purpose of storing this constant into a local var? To make the code using the var fit on one line?

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -31,4 +31,62 @@
...
+    tier = Tier.new(user)
+    previous_tier = tier.previous
+    downgraded_plan = previous_tier.id
+    stub_customer_find_request
+    stub_subscription_find_request(first_subscription)
+    stub_subscription_find_request(second_subscription)
+    stub_subscription_update_request(plan: downgraded_plan, repo_ids: "")
+
+    sign_in_as(user, "letmein")
+    expect(page).to have_content first_subscription.repo.name
```
gylaz: How about a newline before this `expect` and one after line 49, to delineate the phases?

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -31,4 +31,62 @@
...
+    expect(page).to have_content second_subscription.repo.name
+    first(".repo--active").find(".repo-toggle").click
+
+    expect(page).to have_content "Private Repos 1 / 4"
```
gylaz: `have_content` -> `have_text`

spec/features/listing_pricings_spec.rb
```diff
@@ -66,6 +66,6 @@
     wait_for_ajax
 
     expect(current_path).to eq repos_path
-    expect(page).to have_content "Private Repos 5/10"
+    expect(page).to have_content "Private Repos 5 / 10"
```
gylaz: `have_content` -> `have_text`

spec/features/user_activates_a_repo_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    stub_add_collaborator_request(username, repo.name, token)
+    stub_customer_find_request
+    stub_hook_creation_request(repo.name, callback_endpoint, token)
+    stub_subscription_create_request(plan: current_plan, repo_ids: repo.id)
+    stub_subscription_update_request(plan: upgraded_plan, repo_ids: repo.id)
+
+    sign_in_as(user, "letmein")
+    click_on "Activate"
+
+    expect(page).to have_content "Private Repos 2 / 4"
```
gylaz: `have_content` -> `have_text`

spec/features/user_deactivates_a_repo_spec.rb
purinkle: It is not really an expectation. This line is required so that the next stage of the current phase of the test will not fire before it is ready to.

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -127,12 +128,34 @@ class ReposContainer extends React.Component {
     }
   }
 
+  updateSubscribedRepoCount() {
+    this.getUser().done(user => {
```
sardaukar: I believe the usage of `done` here might be harmful - a `then`/`catch` logic seems better to make sure we handle errors inbound from the `getUser()` call

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -31,4 +31,62 @@
...
+    downgraded_plan = previous_tier.id
+    stub_customer_find_request
+    stub_subscription_find_request(first_subscription)
+    stub_subscription_find_request(second_subscription)
+    stub_subscription_update_request(plan: downgraded_plan, repo_ids: "")
+
+    sign_in_as(user, "letmein")
+    expect(page).to have_text first_subscription.repo.name
+    expect(page).to have_text second_subscription.repo.name
+    first(".repo--active").find(".repo-toggle").click
```
gylaz: Since you're using `expect` above this to ensure the element is present, what you could do instead is not use `first` (because it doesn't block/wait) and use `find` instead, something like:
```rb
find(".repo--active:nth-of-type(1) .repo-toggle").click
```

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -31,4 +31,62 @@
...
+    stub_subscription_find_request(subscription)
+    tier = Tier.new(user)
+    previous_tier = tier.previous
+    downgraded_plan = previous_tier.id
+    stub_customer_find_request
+    stub_subscription_update_request(plan: downgraded_plan, repo_ids: "")
+
+    sign_in_as(user, "letmein")
+    expect(page).to have_text subscription.repo.name
+    first(".repo--active").find(".repo-toggle").click
```
gylaz: Similar here, instead of `expect`:
```rb
find(".repo--active:nth-of-type(1) .repo-toggle").click
```

spec/features/user_deactivates_a_repo_spec.rb
```diff
@@ -31,4 +31,62 @@
...
+    subscription = create(:subscription, :active, user: user)
+    stub_subscription_find_request(subscription)
+    tier = Tier.new(user)
+    previous_tier = tier.previous
+    downgraded_plan = previous_tier.id
+    stub_customer_find_request
+    stub_subscription_update_request(plan: downgraded_plan, repo_ids: "")
+
+    sign_in_as(user, "letmein")
+    expect(page).to have_text subscription.repo.name
```
gylaz: and here.

app/assets/javascripts/components/repo_activation_button.js.jsx
```diff
@@ -1,8 +1,22 @@
 class RepoActivationButton extends React.Component {
+  get isActive() {
+    return this.repo.active;
+  }
+
+  get buttonText() {
+    if (!this.isActive) {
+      return "Activate";
+    }
+  }
```
gylaz: If this method returns nothing, will there still be a button (with no text)? Or does the `<button>` tag disappear without text?

app/assets/javascripts/components/repo_allowance.js.jsx
```diff
@@ -0,0 +1,19 @@
+class RepoAllowance extends React.Component {
+  render() {
+    const { subscribedRepoCount, tierAllowance } = this.props;
+
+    return(
+      <div className="allowance">
+        Private Repos
+        { " " }
```
gylaz: Is this a white space? Is this best done with CSS padding/margin?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -214,17 +237,20 @@ class ReposContainer extends React.Component {
...
-        if (data.refreshing_repos) {
-          setTimeout(this.handleSync.bind(this), 1000);
-        } else {
-          this.fetchReposAndOrgs();
-        }
+    });
+  }
+
+  handleSync() {
+    this.getUser().done(data => {
```
gylaz: `done` -> `then`?

spec/factories.rb
```diff
@@ -58,15 +59,20 @@
   factory :user do
     username { generate(:github_name) }
 
+    trait(:github) { token_scopes "public_repo,user:email" }
```
gylaz: What do you think about naming this trait `with_scopes` or `with_github_scopes`? I'm not sure I would understand what `create(:user, :github)` would mean -- since all users are mirrors of GitHub.

r00k:
Just wanted to check in on this.

@purinkle, are you able to address these comments and get this merged?

app/assets/javascripts/components/repos_container.js.jsx
purinkle: AFAIK `done` will only be called when the response is successful. What do you propose for the failed state?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -214,17 +237,20 @@ class ReposContainer extends React.Component {
     });
   }
 
-  handleSync() {
-    $.ajax({
+  getUser() {
+    return $.ajax({
```
purinkle: 👍

app/assets/javascripts/components/repo_allowance.js.jsx
purinkle: The JSX seems to want ignore the newlines as whitespace between the individual elements. I have done this so that we don't end up with "Private Repos1/1".

app/assets/javascripts/components/repos_container.js.jsx
purinkle: What is the preference for one over the other?

app/assets/javascripts/components/repo_activation_button.js.jsx
gylaz: What does this function do then? To me it looks like it returns "Activate" which becomes the text between `<button></button>` tags.

app/assets/javascripts/components/repo_allowance.js.jsx
gylaz: Ah, I see. Would it be more readable with `Private Repos { " " }`? Or perhaps `{ "Private Repos " }`?

app/assets/javascripts/components/repo_allowance.js.jsx
```diff
@@ -0,0 +1,19 @@
...
+        <strong>
+          <span data-role="subscribed-repo-count">{subscribedRepoCount}</span>
+          { " " }
+          /
+          { " " }
```
gylaz: Same here, if possible `{ " /  " }`?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -127,12 +128,34 @@ class ReposContainer extends React.Component {
     }
   }
 
+  updateSubscribedRepoCount() {
+    this.getUser().done(user => {
+      const subscribedRepoCount = user.subscribed_repo_count;
+      const tierAllowance = user.tier_allowance;
+
+      if (subscribedRepoCount === 0) {
+        $(".allowance").parent("li").remove();
```
gylaz: Thoughts on using `data-role` instead to target the exact element we need to remove?

app/assets/javascripts/components/repos_container.js.jsx
```diff
@@ -127,12 +128,34 @@ class ReposContainer extends React.Component {
...
+      } else if (subscribedRepoCount === 1 && $(".allowance").length === 0) {
+        ReactDOM.render(
+          <RepoAllowance
+            subscribedRepoCount={subscribedRepoCount}
+            tierAllowance={tierAllowance}
+          />,
+          $("header ul").last().prepend("<li></li>").children()[0]
```
gylaz: This is a bit difficult to understand what exactly is happening (where the `RepoAllowance` is getting inserted). Can we make this clearer somehow. Perhaps giving the last `ul` a data role?

PR #1259, opened 2016-12-03
Fix credit card update flow

gylaz:
Need to `bind` to the React object due to the callback. Otherwise, it
tries to find `onUpdateCreditCard` on the window object.

Update the Stripe gem, while we're here.


app/assets/javascripts/components/update_account_credit_card.js.jsx
```diff
@@ -35,7 +35,7 @@ class UpdateAccountCreditCard extends React.Component {
       return (
         <h3>
           Monthly Billing
-          <a href="#" className="update-card" onClick={this.onUpdateCreditCard}>
+          <a href="#" className="update-card" onClick={this.onUpdateCreditCard.bind(this)}>
```
gylaz: This was the actual bug fix.

PR #1262, opened 2016-12-17
Use GitHub's new review API to make comments

gylaz:
We can now make all the comments in one request.

Big win for our API rate limit.


spec/services/build_report_spec.rb
```diff
@@ -58,52 +56,34 @@
       end
 
       context "when the build is not complete" do
```
teoljungberg: Thoughts on writing this as `when the build is incomplete`?

app/models/pull_request.rb
```diff
@@ -6,21 +6,22 @@ class PullRequest
...
-  def comment_on_violation(violation)
-    hound_github.add_pull_request_comment(
-      pull_request_number: number,
-      comment: violation.messages.join(COMMENT_LINE_DELIMITER),
-      commit: head_commit,
-      filename: violation.filename,
-      patch_position: violation.patch_position
-    )
+  def comment_on_violations(violations)
+    comments = violations.map do |violation|
```
teoljungberg: Thoughts on extracting a private method along the likes of `build_comments_from_violations`?

app/services/build_report.rb
```diff
@@ -7,20 +7,23 @@ def initialize(pull_request:, build:, token:)
...
     @pull_request = pull_request
     @token = token
-    @commenter = Commenter.new(pull_request)
+    @commenting_policy = CommentingPolicy.new(pull_request)
   end
 
   def run
     if build.completed?
-      commenter.comment_on_violations(priority_violations)
+      new_violations = priority_violations.select do |violation|
```
teoljungberg: Thoughts on extracting this to a private method? Since `#run` here only calls out to other private methods, I like that symmetry.

teoljungberg: I tried to find a testcase for this, but couldn't. Can you link me to one?

gylaz:
Seems that only collaborators can submit/create reviews. This is problematic because we don't add houndci-bot as a collaborator on public repos.

teoljungberg:
Is there a way to migrate us into being a collaborator for public repos?

gylaz:
Not totally, as it requires an invitation on the user's behalf -- we can do it for some repos using the tokens that we store in the DB, but we won't have tokens for all of the repos. We could also build-in a strategy where we try a review and if it fails then comment as we do now.

teoljungberg:
@gylaz That sounds like a fair approach to take.

gylaz:
I was wrong (or they changed it) but it looks like the commenting bot user doesn't have to be a collaborator to create reviews: https://github.com/gylaz/wormbots/pull/13#pullrequestreview-17783750

app/models/pull_request.rb
```diff
@@ -6,21 +6,16 @@ class PullRequest
   pattr_initialize :payload, :token
 
   def comments
-    @comments ||= user_github.pull_request_comments(full_repo_name, number)
+    @comments ||= user_github.pull_request_comments(repo_name, number)
```
gylaz: Sure, I can do that.

spec/support/helpers/github_api_helper.rb
```diff
@@ -129,18 +128,12 @@ def stub_repos_requests(token)
...
-      "https://api.github.com/repos/#{full_repo_name}/pulls/#{pull_request_number}/comments"
-    ).with(
-      body: {
-        body: comment,
-        commit_id: commit_sha,
-        path: file,
-        position: line_number
-      }.to_json
-    ).to_return(status: 200)
+  def stub_review_request(repo_name, pr_number, comments)
```
gylaz: We still use the WebMock stubs at more of an unit test-level, to ensure the request went out. It costs less than booting up a Sintra app. The FakeGithub feels like a good tool for more integration-level tests, but less for unit tests. Thoughts?

teoljungberg: I'm unsure if there's any real performance hit of booting up the Sinatra app up versus setting up a webmock. I for one prefer consistency, and would like us to have one way of handling stubbing external requests. And my money is on the FakeGithub approach.

lib/github_api.rb
```diff
@@ -63,14 +65,12 @@ def pull_request_comments(full_repo_name, pull_request_number)
...
-      options[:pull_request_number],
-      options[:comment],
-      options[:commit].sha,
-      options[:filename],
-      options[:patch_position],
+  def create_pull_request_review(repo_name, pr_number, comments)
+    client.post(
+      "#{Octokit::Repository.path(repo_name)}/pulls/#{pr_number}/reviews",
+      accept: PREVIEW_API_HEADER,
+      event: "COMMENT",
```
blackjid: Sorry for jumping in.. 
Just an idea, this could be configurable on the `.hould.yml` with to use `REQUEST_CHANGES` or `COMMENT` as an alternative or replacement to `fail_on_violations: true`.

Also, it would be nice when no violations are found, a review with `APPROVE` could be created.

teoljungberg: Interesting idea.

I think we should change `fail_on_violations` to comment with the state of `REQUEST_CHANGES`, and if it is not enabled - we comment with `COMMENT` instead.

What do you think @gylaz ?

gylaz: I think this is a great idea. Thoughts on merging this as-is then making that change in another PR?

app/services/complete_build.rb
```diff
@@ -5,20 +5,23 @@ def initialize(pull_request:, build:, token:)
...
     @pull_request = pull_request
     @token = token
-    @commenter = Commenter.new(pull_request)
+    @commenting_policy = CommentingPolicy.new(pull_request)
   end
 
   def call
     if build.completed?
-      commenter.comment_on_violations(priority_violations)
+      new_violations = priority_violations.select do |violation|
```
thorncp: Not necessarily in this PR, but should we consider removing the comment cap with this change (`priority_violations`)? We included that originally to not trip the abuse detector, but with this being a single request I assume that's not an issue anymore.

gylaz: We can consider this. Another reason for the comment limit, is that users complained of getting 100's of messages in one PR -- might still be an issue, even with a PR. We can definitely bump this number up to something more reasonable, and maybe move it into a config?

gylaz:
@thorncp Would you have time today to merge this, deploy and monitor? (I've already sanity tested on staging, but might be good to do it again for another pair of eyes).

jayroh:
I'll be testing out the reek stuff on staging today as well so if both this feature and the reek stuff were going on in parallel it would be interesting to review how things looked - considering the concerns about duplication of messaging (per comments in #1304).

thorncp:
Yep, I can devote some time today to getting this out.

jayroh:
🎉

PR #1264, opened 2016-12-20
Replace GitHub stubs with a fake

purinkle:
Before, all the calls to the GitHub API in the test environment were using WebMock stubs. This was adding lots of noise to the tests. Added a fake GitHub to replace the WebMock stubs.

* Removed redundant helper methods and fixtures.


sardaukar:
I much prefer how this reads. Good one @purinkle

lib/github_api.rb
```diff
@@ -5,6 +5,9 @@
 class GithubApi
   ORGANIZATION_TYPE = "Organization".freeze
 
+  cattr_accessor :client
+  self.client = Octokit::Client
```
gylaz: Why do this? What is the benefit? If we are overriding this somewhere, why not inject this into the constructor with a default?

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -81,53 +81,57 @@
...
-      plan = tier.current.id
-      repo = membership.repo
-      token = "letmein"
-      stub_customer_find_request(user.stripe_customer_id)
-      stub_sign_in(user)
-      stub_hook_creation_request(repo.name, hook_url, token)
-      stub_subscription_create_request(plan: plan, repo_ids: repo.id)
-      stub_subscription_update_request(plan: new_plan, repo_ids: repo.id)
+      repo = instance_double(
+        Repo,
```
gylaz: If we use the string version of this (i.e. `instance_double("Repo"...)`) then the constant does not need to be loaded. That's what we prefer.

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -81,53 +81,57 @@
...
+      user = instance_double(
+        User,
+        email: "TEST_USER_EMAIL",
+        repos: repos,
+        token: "TEST_USER_TOKEN",
+      )
+      users = class_double(User, first: user)
+      allow(RepoActivator).to receive(:new).and_return(repo_activator)
+      allow(RepoSubscriber).to receive(:subscribe).and_return(repo_subscriber)
+      allow(User).to receive(:where).and_return(users)
```
gylaz: This test does a lot of stubbing, but does it provide any value? Does the feature spec (if there is one) cover this code and the ability to update a subscription? Maybe we should just remove it...

spec/lib/github_api_spec.rb
```diff
@@ -4,7 +4,12 @@
 require "json"
 
 describe GithubApi do
-  before { stub_const("Hound::GITHUB_TOKEN", "token") }
+  before do
+    stub_const("Hound::GITHUB_TOKEN", "token")
+    GithubApi.client = Octokit::Client
+  end
+
+  after { GithubApi.client = FakeGithub }
```
gylaz: Why do we set the client to `FakeGithub` after the test?

spec/requests/builds_spec.rb
```diff
@@ -12,95 +12,36 @@
...
-        file_path: filename,
-        commit_id: commit_id_from_fixture,
-      )
 
       post builds_path, payload: payload
 
-      expect(new_comment_request).to have_been_requested
+      comments = FakeGithub.comments
+      comment = comments.first
+      expect(comments.length).to eq 1
```
gylaz: `length` -> `size`.

spec/requests/builds_spec.rb
```diff
@@ -12,95 +12,36 @@
...
+      expect(comment.repo).to eq repo_name
+      expect(comment.pull_id).to eq pr_number
+      expect(comment.body).to eq "Trailing whitespace detected."
+      expect(comment.commit_id).to eq pr_sha
+      expect(comment.path).to eq "spec/models/style_guide_spec.rb"
+      expect(comment.position).to eq 9
```
gylaz: Can we use the `have_attributes` matcher for these?

spec/support/fake_github.rb
```diff
@@ -0,0 +1,99 @@
...
++
++after_fork do |server, worker|
++  Signal.trap 'TERM' do
++    puts 'Wait for master to send QUIT'
++  end
++
++  defined?(ActiveRecord::Base) and
++    ActiveRecord::Base.establish_connection
++end
+      PATCH
```
gylaz: How about extracting this heredoc to a fixture?

.ruby-style.yml
```diff
@@ -573,6 +573,8 @@ Rails/HasAndBelongsToMany:
   Enabled: true
   Include:
   - app/models/**/*.rb
+Rails/HttpPositionalArguments:
```
teoljungberg: Thoughts on extracting this to a separate PR?

lib/github_api.rb
teoljungberg: I'd 👍 using dependency injection here instead. (which is also more thread-safe)

lib/github_api.rb
```diff
@@ -13,7 +16,7 @@ def initialize(token)
   end
 
   def client
-    @client ||= Octokit::Client.new(access_token: token, auto_paginate: true)
+    @client ||= self.class.client.new(access_token: token, auto_paginate: true)
```
teoljungberg: Thoughts on using memoization with an underscore for consistency with the rest of the app? Or more specifically the new parts of the app?

spec/controllers/sessions_controller_spec.rb
```diff
@@ -4,12 +4,16 @@
...
+        github_api = instance_double(
+          GithubApi,
+          scopes: "public_repo,user:email",
+        )
         request.env["omniauth.auth"] = stub_oauth(
           username: "jimtom",
           email: "jimtom@example.com",
           token: "letmein",
         )
+        allow(GithubApi).to receive(:new).and_return(github_api)
```
teoljungberg: Shouldn't this use the DI pattern, or use the `cattr_attribute`?

spec/controllers/subscriptions_controller_spec.rb
teoljungberg: I agree, I normally find controller specs not to add alot of use. I lean on feature specs for when a user traverses an application, or request specs when it is an API.

spec/lib/github_api_spec.rb
teoljungberg: It looks like we need to do it to get the state back on track after the tests. The `rails_helper` is setting `GithubApi.client = FakeGithub`, and that state needs to be reset.

teoljungberg: Could this be an `around` filter instead?

```ruby
around do |example|
  stub_const("Hound::GITHUB_TOKEN", "token")
  GithubApi.client = Octokit::Client
  
  example.run

  GithubApi.client = FakeGithub
end
```

spec/requests/builds_spec.rb
teoljungberg: I like that too!

spec/support/fake_github.rb
```diff
@@ -0,0 +1,99 @@
+class FakeGithub
+  Comment = Struct.new(:repo, :pull_id, :body, :commit_id, :path, :position)
+  Content = Struct.new(:content)
+  File = Struct.new(:filename, :patch, :status)
+  Hook = Struct.new(:id)
+
+  cattr_accessor :comments
```
teoljungberg: I generally try to shy away from `cattr_accessor`, as it's generally not thread-safe.

spec/support/fake_github.rb
```diff
@@ -0,0 +1,99 @@
...
+    self.class.comments << Comment.new(*args)
+  end
+
+  def pull_request_comments(repo, number)
+    [
+      Comment.new(
+        repo,
+        number,
+        "Line is too long.",
+        "81ee93414c230f71603265bd791bff5fbb8dba3e",
```
teoljungberg: Can this SHA be a bit more generic?

spec/support/fake_github.rb
```diff
@@ -0,0 +1,99 @@
...
+        "spec/models/style_guide_spec.rb",
```
teoljungberg: Same as this filename.

spec/support/fake_github.rb
```diff
@@ -0,0 +1,99 @@
...
+        5,
+      ),
+    ]
+  end
+
+  def pull_request_files(*_args)
+    [
+      File.new("spec/models/style_guide_spec.rb", <<-PATCH, "added")
+@@ -0,0 +1,22 @@
++worker_processes Integer(ENV[\"WEB_CONCURRENCY\"] || 3)
```
teoljungberg: Same as this filecontents.

spec/support/fake_github.rb
teoljungberg: I like that too.

spec/support/fake_github.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  def remove_hook(*_args)
+    true
+  end
+
+  def repos
+    [
+      {
+        full_name: "jimtom/My-Private-Repo",
```
teoljungberg: Can we have a more anonymous name?

spec/support/fake_github.rb
```diff
@@ -0,0 +1,99 @@
...
+        id: 1296269,
```
teoljungberg: Does this id need to be this specifically, or can it be a generic one such as `1`?

spec/controllers/sessions_controller_spec.rb
purinkle: For this level of testing, we do not want the test to drop into the lower classes. The test should only care that `GithubApi` is being called and not what `GithubApi` is doing. I would argue that we should not be accessing `GithubApi` direct from the controller. This is a story for another day.

teoljungberg: Sounds correct to me!

teoljungberg:
I like this alot! Great work @purinkle 

I want to hear what @gylaz thinks about this Sinatra-fake approach. I for one like it alot better

PR #1265, opened 2016-12-20
Update to Rails 5

jurre:
This passes all the tests locally ~~there's one deprecation warning left that I'll need to look into~~


bin/update
```diff
@@ -0,0 +1,29 @@
...
+
+  puts "\n== Updating database =="
+  system! 'bin/rails db:migrate'
+
+  puts "\n== Removing old logs and tempfiles =="
+  system! 'bin/rails log:clear tmp:clear'
+
+  puts "\n== Restarting application server =="
+  system! 'bin/rails restart'
+end
```
gylaz: Is this a generated file?

config/application.rb
```diff
@@ -1,20 +1,22 @@
...
+require File.expand_path("../boot", __FILE__)
 
-require 'rails/all'
+require "rails/all"
 
 Bundler.require(:default, Rails.env)
 
 module Houndapp
   class Application < Rails::Application
+    require "redirect_to_configuration"
```
gylaz: How about movign this with the other requires at the top of the file?

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
```
gylaz: Should this come from the env? `REDIS_URL`?

config/environment.rb
```diff
@@ -1,5 +1,5 @@
-# Load the rails application
-require File.expand_path('../application', __FILE__)
+# Load the Rails application.
+require_relative "application"
```
gylaz: Is this preferred now over the `File.expand_path` way?

spec/controllers/accounts_controller_spec.rb
```diff
@@ -21,7 +21,8 @@
           invalid_email = "newemail"
           stub_sign_in(user)
 
-          patch :update, billable_email: invalid_email, format: :json
+          patch :update,
+                params: { billable_email: invalid_email }, format: :json
```
gylaz: We use shallow nesting, meaning:
```rb
patch :update,
  params: { billable_email: invalid_email }, format: :json
```
but at this point, we prefer to put each argument on its own line.

bin/update
jurre: It is

config/environment.rb
jurre: I think so, this is what the update script does

config/cable.yml
jurre: Good point!

jurre: 04af9f4

bin/update
jurre: I think we can remove it actually

config/application.rb
jurre: Ah, we need `require File.expand_path("lib/redirect_to_configuration")` in that case. I'm ok with either, do you have a preference?

gylaz: I think I prefer a `require` at the top, and I'm fine with `File.expand_path`.

jurre:
Rebased and squashed, tests are passing and it seems to run ok for me locally.

The only thing I'm not 100% sure about is I'm getting this output when running the tests:

```bash
W, [2016-12-21T20:29:54.316259 #56619]  WARN -- sentry: ** [Raven] Unable to load raven/integrations/rack-timeout: cannot load such file -- rack/timeout/base
Expected string default value for '--serializer'; got true (boolean)
```

teoljungberg:
@jurre I found another warning while running the tests:

```
W, [2016-12-22T10:15:51.974445 #53793]  WARN -- sentry: ** [Raven] Unable to load raven/integrations/rack-timeout: cannot load such file -- rack/timeout/base
```

That can be fixed by running `bundle update rack-timeout`.

The second warning(which you brought up) seems to be a regression in `thor`, locking `thor` to `0.19.1` removes the warning.

Here's my total diff:

```diff
diff --git a/Gemfile b/Gemfile
index 453d7c9..b9b1dab 100644
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem "rails-assets-classnames", source: "https://rails-assets.org"
 gem "rails-assets-es5-shim", source: "https://rails-assets.org"
 gem "rails-assets-lodash", source: "https://rails-assets.org"
 gem "react-rails"
+gem "record_tag_helper"
 gem "resque", ">= 1.25.0"
 gem "resque-scheduler"
 gem "resque-sentry"
@@ -36,8 +37,8 @@ gem "sass-rails"
 gem "sinatra", "2.0.0.beta2"
 gem "split", require: "split/dashboard"
 gem "stripe"
+gem "thor", "0.19.1"
 gem "uglifier", ">= 2.7.2"
-gem "record_tag_helper"
 
 group :staging, :production do
   gem "rack-timeout"
diff --git a/Gemfile.lock b/Gemfile.lock
index 2bc816f..3e2cb24 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -198,7 +198,7 @@ GEM
       rack
     rack-test (0.6.3)
       rack (>= 1.0)
-    rack-timeout (0.2.4)
+    rack-timeout (0.4.2)
     rails (5.0.0.1)
       actioncable (= 5.0.0.1)
       actionmailer (= 5.0.0.1)
@@ -315,7 +315,7 @@ GEM
       sprockets (>= 3.0.0)
     stripe (1.57.1)
       rest-client (>= 1.4, < 4.0)
-    thor (0.19.4)
+    thor (0.19.1)
     thread_safe (0.3.5)
     tilt (2.0.5)
     tzinfo (1.2.2)
@@ -391,6 +391,7 @@ DEPENDENCIES
   sinatra (= 2.0.0.beta2)
   split
   stripe
+  thor (= 0.19.1)
   uglifier (>= 2.7.2)
   webmock
 
```

I took the liberty of applying that patch and merging this PR as eb5ded4

Thank you so much for working on this!

jurre:
👍 awesome, thanks

PR #1267, opened 2016-12-23
Move several files from lib to app

gylaz:
Rails 5 does not autoload directories from `autoload_paths` when
`eager_load` is set (it is in production env).

One solution is to add `lib/` to `eager_load` paths, but that is deemed
non thread-safe (and also eager loads all the other code in `lib/` like
tasks, etc).

The other solution, and the one I went with, is to move the necessary
code to `app/`. Thus, I moved the there files to `app/models/`. You
could think of them as models, or other domain logic of our app, but
that's where I put them for now.

Read more about this Rails 5 issue:

* https://github.com/rails/rails/issues/13142
* http://collectiveidea.com/blog/archives/2016/07/22/solutions-to-potential-upgrade-problems-in-rails-5/


teoljungberg:
Wouldn't another way of resolving the issue be to require the files where they are used. So i.e `app/controllers/sessions_controller.rb` could require `lib/github_api.rb`.

gylaz:
We decided to go the https://github.com/houndci/hound/pull/1269 route instead.

PR #1270, opened 2017-01-03
Use owner configuration for SCSS linting

purinkle:
Before, the SCSS linting always used Hound's default or the repository's own configuration. This meant we had to specify organisation-wide configurations in each individual repository. Updated the SCSS configuration to allow configuration at an organisational level.

https://trello.com/c/HY2hoTdq

### DIFF
--- a/app/models/config/scss.rb
+++ b/app/models/config/scss.rb
@@ -1,10 +1,25 @@
 module Config
   class Scss < Base
+    def initialize(hound_config, owner: MissingOwner.new)
+      super(hound_config)
+      @owner = owner
+    end
+
+    def content
+      owner_config.deep_merge(super)
+    end
+
     def serialize(data = content)
       Serializer.yaml(data)
     end
 
     private
+
+    attr_reader :owner
+
+    def owner_config
+      owner.hound_config
+    end
 
     def parse(file_content)
       Parser.yaml(file_content)
--- a/spec/models/config/scss_spec.rb
+++ b/spec/models/config/scss_spec.rb
@@ -3,21 +3,81 @@ require "app/models/config/base"
 require "app/models/config/scss"
 require "app/models/config/parser"
 require "app/models/config/serializer"
+require "app/models/missing_owner"
+require "app/services/build_owner_hound_config"
 
 describe Config::Scss do
   describe "#content" do
-    it "parses the configuration using YAML" do
-      raw_config = <<-EOS.strip_heredoc
-        linters:
-          BangFormat:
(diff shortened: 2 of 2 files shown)


spec/models/config/scss_spec.rb
```diff
@@ -3,21 +3,70 @@
 require "app/models/config/scss"
 require "app/models/config/parser"
 require "app/models/config/serializer"
+require "app/services/build_owner_hound_config"
 
 describe Config::Scss do
   describe "#content" do
-    it "parses the configuration using YAML" do
-      raw_config = <<-EOS.strip_heredoc
+    context "when there is no owner" do
```
teoljungberg: I normally like to order the happy paths above the sad-paths, and I would order this the same way as the conditional in `owner_config_content` (which I think is how we've done it in the other Linters, but I'm unsure).

spec/models/config/scss_spec.rb
```diff
@@ -3,21 +3,70 @@
...
+      it "parses the configuration using YAML" do
+        raw_config = <<-EOS.strip_heredoc
```
teoljungberg: Thoughts on using `<<~EOS` instead of `<<-EOS.strip_heredoc` since we're on Ruby 2.3?

spec/models/config/scss_spec.rb
```diff
@@ -3,21 +3,70 @@
...
 
 describe Config::Scss do
   describe "#content" do
-    it "parses the configuration using YAML" do
-      raw_config = <<-EOS.strip_heredoc
+    context "when an owner is provided" do
+      it "merges the configuration into the owner's configuration" do
+        owner = instance_double("Owner")
+        raw_owner_config = <<~EOS
+        linters:
```
teoljungberg: Should this be indented?

spec/models/config/scss_spec.rb
```diff
@@ -3,21 +3,70 @@
...
+            },
+          },
+        )
+      end
+    end
+
+    context "when there is no owner" do
+      it "parses the configuration using YAML" do
+        raw_config = <<~EOS
+        linters:
```
teoljungberg: Should this be indented too?

app/models/config/scss.rb
```diff
@@ -1,11 +1,26 @@
 module Config
   class Scss < Base
+    def initialize(hound_config, owner: MissingOwner.new)
```
gylaz: Nice, I like that `MissingOwner`!

PR #1271, opened 2017-01-03
Use owner configuration for JavaScript linting

purinkle:
Before, the JavaScript linting always used Hound's default or the repository's own configuration. This meant we had to specify organisation-wide configurations in each individual repository. Updated the JavaScript configuration to allow configuration at an organisational level.

https://trello.com/c/HY2hoTdq

### DIFF
--- a/app/models/config/eslint.rb
+++ b/app/models/config/eslint.rb
@@ -1,10 +1,25 @@
 module Config
   class Eslint < Base
+    def initialize(hound_config, owner: MissingOwner.new)
+      super(hound_config)
+      @owner = owner
+    end
+
+    def content
+      owner_config.deep_merge(super)
+    end
+
     def serialize(data = content)
       Serializer.json(data)
     end
 
     private
+
+    attr_reader :owner
+
+    def owner_config
+      owner.hound_config
+    end
 
     def parse(file_content)
       json_with_comments = JsonWithComments.new(file_content)
--- a/spec/models/config/eslint_spec.rb
+++ b/spec/models/config/eslint_spec.rb
@@ -5,19 +5,66 @@ require "app/models/config/parser"
 require "app/models/config/parser_error"
 require "app/models/config/serializer"
 require "app/models/config/json_with_comments"
+require "app/models/missing_owner"
+require "app/services/build_owner_hound_config"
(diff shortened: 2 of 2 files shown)


blackjid:
Looks like the title of the PR shoulds say **"Use owner configuration for ESLinter linting"**

app/models/config/eslint.rb
```diff
@@ -1,11 +1,30 @@
 module Config
   class Eslint < Base
+    def initialize(hound_config, owner: missing_owner)
```
teoljungberg: Thought's on inlining `MissingOwner.new` here?

spec/models/config/eslint_spec.rb
```diff
@@ -5,19 +5,66 @@
...
+            ],
+          },
+        )
+      end
+    end
+
+    context "when the owner is missing" do
+      it "parses the configuration using YAML" do
+        raw_config = <<~EOS
         rules:
```
teoljungberg: This should be indented a level.

spec/models/missing_owner_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "spec_helper"
+require "app/models/missing_owner"
+
+RSpec.describe MissingOwner do
+  describe "#hound_config" do
+    it "is an empty hash" do
+      expect(MissingOwner.new.hound_config).to be_empty
```
teoljungberg: This test will still pass if `#hound_config` is `""` or `[]`. Thoughts on using `expect().to eq({})` instead?

app/models/config/eslint.rb
```diff
@@ -1,11 +1,26 @@
 module Config
   class Eslint < Base
+    def initialize(hound_config, owner: MissingOwner.new)
+      super(hound_config)
+      @owner = owner
+    end
```
gylaz: Can we move this into the base class, since most (all?) of the children are doing the same owner logic?

purinkle: I will do this in a separate PR

teoljungberg: (done in https://github.com/houndci/hound/pull/1274)

PR #1273, opened 2017-01-03
Point the FAQ to Intercom's Educate product

r00k:
This commit removes our home-grown FAQ section and instead points users
at Intercom's new Educate product (basically a CMS for support
articles).

Our support person, Micah, has moved over the existing FAQs and will be
adding more as he encounters them.

[Here's the help site](https://intercom.help/hound).


app/views/application/_footer.haml
```diff
@@ -3,7 +3,7 @@
     %li
       = link_to "Source Code", "https://github.com/houndci/hound"
     %li
-      = link_to "FAQ", faq_path
+      = link_to "FAQ", "https://intercom.help/hound"
```
gylaz: What do you think about pointing it to https://intercom.help/hound/frequently-asked-questions so users don't have to click again?

r00k:
Feedback addressed, and all checks passing. Good to merge @gylaz ?

r00k:
> One other thing: what do you you think about slapping target="_blank" on these links, since they are navigating away from Hound?

My vote would be to let the users control their own browsing experience.

Personally, I always feel a bit annoyed when sites open new tabs for me.

I don't feel super strongly, though, so I'll add it if you do.

gylaz:
> I don't feel super strongly, though, so I'll add it if you do.

Me neither. Your call.

r00k:
Okay, merged!

PR #1275, opened 2017-01-04
Pass the owner through to the configuration

purinkle:
Before, we set up several configurations to accept an owner. This was all well and good, but we did not actually pass an owner through. Updated the SCSS, ESLint, and Haml linters to provide an owner to their configurations.

https://trello.com/c/HY2hoTdq

### DIFF
--- a/app/models/linter/eslint.rb
+++ b/app/models/linter/eslint.rb
@@ -9,6 +9,10 @@ module Linter
 
     private
 
+    def config
+      Config::Eslint.new(hound_config, owner: owner)
+    end
+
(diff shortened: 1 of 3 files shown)


app/models/linter/base.rb
```diff
@@ -63,6 +63,10 @@ def job_class
       job_name.constantize
     end
 
+    def owner
+      build.repo.owner || MissingOwner.new
```
teoljungberg: This looks like a leftover from https://github.com/houndci/hound/pull/1274.

PR #1276, opened 2017-01-04
Use owner configuration for Swift linting

purinkle:
Before, the Swift linting always used Hound's default or the repository's own configuration. This meant we had to specify organisation-wide configurations in each individual repository. Updated the Swift configuration to allow configuration at an organisational level.

https://trello.com/c/HY2hoTdq

### DIFF
--- a/app/models/config/swift.rb
+++ b/app/models/config/swift.rb
@@ -1,5 +1,9 @@
 module Config
   class Swift < Base
+    def content
+      owner_config.deep_merge(super)
+    end
+
     def serialize(data = content)
       Serializer.yaml(data)
     end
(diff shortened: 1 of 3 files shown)


app/models/linter/swift.rb
```diff
@@ -1,5 +1,11 @@
 module Linter
   class Swift < Base
     FILE_REGEXP = /.+\.swift\z/
+
+    private
+
+    def config
+      Config::Swift.new(hound_config, owner: owner)
+    end
```
gylaz: Time to move this into base?

app/models/config/swift.rb
```diff
@@ -1,5 +1,9 @@
 module Config
   class Swift < Base
+    def content
+      owner_config.deep_merge(super)
+    end
```
gylaz: Could this be moved to base?

PR #1277, opened 2017-01-05
Update base linter config

purinkle:
Before, we had started to pass the owner through to the individual configurations. This meant we were overriding the base class with similar methods in the children. Updated the configuration builder to accept an owner and updated the base lint to pass an owner.

https://trello.com/c/HY2hoTdq

### DIFF
--- a/app/models/config_builder.rb
+++ b/app/models/config_builder.rb
@@ -1,20 +1,21 @@
 class ConfigBuilder
-  def self.for(hound_config, name)
-    new(hound_config, name).config
+  def self.for(hound_config:, name:, owner:)
+    new(hound_config: hound_config, name: name, owner: owner).config
   end
 
-  def initialize(hound_config, name)
+  def initialize(hound_config:, name:, owner:)
     @hound_config = hound_config
     @name = name
+    @owner = owner
   end
 
   def config
-    config_class.new(hound_config)
+    config_class.new(hound_config, owner: owner)
   end
 
   private
 
-  attr_reader :hound_config, :name
+  attr_reader :hound_config, :name, :owner
 
   def config_class
     "Config::#{name.classify}".constantize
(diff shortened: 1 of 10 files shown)


app/models/config_builder.rb
```diff
@@ -1,20 +1,21 @@
 class ConfigBuilder
-  def self.for(hound_config, name)
-    new(hound_config, name).config
+  def self.for(hound_config:, name:, owner:)
+    new(hound_config: hound_config, name: name, owner: owner).config
```
teoljungberg: This can be simplified to:

```ruby
def self.for(*args)
  new(*args).config
end
```

If you're into that sorta thing.

app/models/linter/base.rb
```diff
@@ -68,7 +68,11 @@ def owner
     end
 
     def config
-      @config ||= ConfigBuilder.for(hound_config, name)
+      @config ||= ConfigBuilder.for(
```
teoljungberg: We're slowly moving over to prefixing memoization with `_` to denote that the ivars are private.

What do you think about doing this here too?

app/models/config_builder.rb
purinkle: I want to keep the named parameters in the method signature. It makes it easier for any current or future developers to see how to call the method.

app/models/linter/base.rb
purinkle: Resolved in dbd14e2

app/models/linter/base.rb
```diff
@@ -68,7 +68,11 @@ def owner
     end
 
     def config
-      @config ||= ConfigBuilder.for(hound_config, name)
+      @_config ||= ConfigBuilder.for(
```
gylaz: The way this reads has always irked me. It can be implied that `ConfigBuilder.for` returns a `ConfigBuilder`, when it actually returns a config for a specific linter. What if we moved it to services and named it something like `BuildConfig` or `BuildLinterConfig` (maybe in a separate PR)?

purinkle: Can do.

PR #1279, opened 2017-01-05
Move loading of content into base config

purinkle:
Before, we had started to override the loading of config content in each of the child classes. This meant there had become some divergence from the base definition. Created a config content loader and moved the loading and merging into the parent class.

https://trello.com/c/HY2hoTdq

### DIFF
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -6,7 +6,9 @@ module Config
     end
 
     def content
-      @content ||= ensure_correct_type(safe_parse(load_content))
+      @content ||= owner_config.deep_merge(load)
+    rescue ConfigContent::ContentError => exception
+      raise_parse_error(exception.message)
     end
 
     def serialize(data = content)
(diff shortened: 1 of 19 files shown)


app/models/config/credo.rb
```diff
@@ -7,5 +7,33 @@ def content
     def default_content
       ""
     end
+
+    private
+
+    def load_content
+      if file_path
+        if url?
```
gylaz: I think we're safe on not supporting url config file paths for Credo (since it's new). In general, we should move away from accepting url as a config file and move towards org-level configs.

app/models/config/ruby.rb
```diff
@@ -30,6 +30,12 @@ def parse_inherit_from(config)
       inherited_config.merge(config.except("inherit_from"))
     end
 
+    def safe_parse(content)
+      parse(content)
+    rescue JSON::ParserError, Psych::Exception => exception
```
gylaz: RuboCop is using a Yaml formatted config. Is `JSON::ParserError` relevant here?

app/models/config_content/remote.rb
```diff
@@ -0,0 +1,31 @@
+class ConfigContent
+  class Remote
```
gylaz: Nice abstraction! I do think we'll soon drop this support, to move towards org-level config rather than remote file urls.

app/models/config_content.rb
```diff
@@ -0,0 +1,59 @@
...
+  def parse
+    if incorrect_format?
+      raise_error %{"#{file_path}" must be a Hash}
+    else
+      parsed_content
+    end
+  end
+
+  def parsed_content
+    @_parsed_content ||= parser.call(content)
```
gylaz: Can we use a YAML parser for all/most linters (since JSON is a subset of YAML)?

app/models/config_content/remote.rb
purinkle: 😊

app/models/config_content.rb
purinkle: I have defined a default parse method in the base class as several child classes need the ability to override.

app/models/config/base.rb
```diff
@@ -20,58 +22,21 @@ def linter_name
...
     attr_reader :hound_config, :owner
-    attr_implement :parse, [:file_content]
 
-    def owner_config
-      owner.hound_config
+    def load
+      ConfigContent.new(
+        commit: commit,
+        file_path: file_path,
+        parser: ->(content) { parse(content) },
```
gylaz: Does this need to be a lambda? Can we just pass in the result of the `parse` method?

purinkle: At the time we create the object, we don't know what `content` will be. This is why we have to pass it in inside a lambda.

PR #1282, opened 2017-01-06
Fix webpack build in Heroku issue

sardaukar:


gylaz:
If this let us deploy master to production that I'm all for it!

PR #1283, opened 2017-01-10
Restoring production builds with webpack 2.x

sardaukar:
For some reason, webpack 1.14's production build has issues - upping the version to this beta fixes it. The final dev build sizes:
```
app-bundle.js (662 kB)
vendor-bundle.js (1.5 MB)
```
And for production:
```
app-bundle.js  (117 kB)
vendor-bundle.js  (352 kB)
```

The `keywords if/then/else require v5 option` message during the build is a known issue, and is fixed in webpack 2.2.0-rc.3, but `babel-loader` is not compatible with it yet.


sardaukar:
Yeah, sorry @gylaz I tried not having them but webpack is having none of it :|

PR #1284, opened 2017-01-10
Build owner config

purinkle:
Before, we retrieved the configuration from the owner but not doing anything with it. This meant we were not merging the correct configurations. Update the base config to create a config object from the retrieved file.

### DIFF
--- a/spec/models/config/eslint_spec.rb
+++ b/spec/models/config/eslint_spec.rb
@@ -5,7 +5,9 @@ require "app/models/config/parser"
 require "app/models/config/parser_error"
 require "app/models/config/serializer"
 require "app/models/config/json_with_comments"
+require "app/models/config_content"
 require "app/models/missing_owner"
+require "app/services/build_config"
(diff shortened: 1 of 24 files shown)


spec/models/config/eslint_spec.rb
```diff
@@ -36,7 +28,21 @@
           }
         EOS
         commit = stubbed_commit("config/.eslintrc" => raw_config)
+        hound_config = instance_double("HoundConfig")
```
sardaukar: lots of these... any way to DRY them up?

sardaukar:
Minor quirks here and there, but overall LGTM :shipit:

spec/models/config/eslint_spec.rb
purinkle: To be honest, I think it is better to be more verbose in the tests to tell a clearer story of what is actually happening. I lost a decent amount of time trying to traverse the helper methods that have been introduced to DRY code up in the past.

PR #1285, opened 2017-01-11
Don't fetch config when there is no owner

gylaz:
There was a infinite-loop situation, where we would attempt to keep
calling `content` (and subsequently `owner_config`) for each owner.
Since we moved to `MissingOwner`, we'd pass that in when there was no
owner, and thus loop infinitely passing around `MissingOwner` instance.

This prevents the above situation and returns empty config for
`MissingOwner`.

### DIFF
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -13,7 +13,11 @@ class Owner < ApplicationRecord
     config_enabled? && config_repo.present?
   end
 
-  def hound_config
-    BuildOwnerHoundConfig.run(self)
+  def config_content(linter_name)
+    BuildConfig.for(
+      hound_config: BuildOwnerHoundConfig.run(self),
+      name: linter_name,
+      owner: MissingOwner.new,
+    ).content
   end
 end
--- a/spec/models/owner_spec.rb
+++ b/spec/models/owner_spec.rb
@@ -63,13 +63,29 @@ describe Owner do
     end
   end
 
-  describe "#hound_config" do
-    it "is the content of the owner's Hound configuration" do
-      config = instance_double("HoundConfig")
-      owner = create(:owner)
-      allow(BuildOwnerHoundConfig).to receive(:run).and_return(config)
+  describe "#config_content" do
+    it "returns the content for a specific linter" do
+      owner_hound_config = {
+        "ruby" => {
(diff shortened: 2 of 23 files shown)


app/models/owner.rb
```diff
@@ -13,7 +13,11 @@ def has_config_repo?
     config_enabled? && config_repo.present?
   end
 
-  def hound_config
-    BuildOwnerHoundConfig.run(self)
+  def config_content(linter_name)
+    BuildConfig.for(
+      hound_config: BuildOwnerHoundConfig.run(self),
+      name: linter_name,
+      owner: MissingOwner.new,
```
teoljungberg: This is weird to me, shouldn't the `owner` here be `self` since we are in the owner model?

spec/models/owner_spec.rb
```diff
@@ -63,13 +63,29 @@
...
+      }
+      owner_ruby_config = {
+        "LineLength" => { "Max" => 90 },
+      }
+      owner_hound_contents = owner_hound_config.to_yaml
+      owner_ruby_contents = owner_ruby_config.to_yaml
+      owner = create(:owner, config_repo: "foo/bar", config_enabled: true)
+      github_api = instance_double("GithubApi", repo: true)
+      allow(github_api).to receive(:file_contents).
+        with("foo/bar", ".hound.yml", "HEAD").
```
teoljungberg: Thoughts on splitting this `with` part of the stub out to the expectation instead?

spec/models/owner_spec.rb
```diff
@@ -63,13 +63,29 @@
...
+        and_return(double(content: Base64.encode64(owner_hound_contents)))
+      allow(github_api).to receive(:file_contents).
+        with("foo/bar", ".rubocop.yml", "HEAD").
```
teoljungberg: Same goes for this stubbing.

app/models/owner.rb
gylaz: No, because the `owner` of self is none/nothing. `BuildConfig` treats the `owner` field, as the entity to get the high-level configuration from.

spec/models/owner_spec.rb
gylaz: It has to be here to return the correct result, which is different from the other `allow` below.

PR #1286, opened 2017-01-11
Webpack 2.x production build fixes (II)

sardaukar:
Hoping not creating source maps in production will help the bundle behave normally (also, only build said bundle once in Heroku).

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -36,17 +36,21 @@ if (isProd) {
         if_return: true,
         join_vars: true,
       },
+      // Mangling would reduce the size of the bundle, but it made the deploys
+      // crash on Heroku. The reason is yet unknown.
+      // So, in lieu of that - let's disable it.
+      mangle: false,
       output: {
         comments: false
       },
-      sourceMap: false,
+      sourceMap: false
     })
   );
 }
 
 module.exports = {
   context: __dirname,
-  devtool: isProd ? 'hidden-source-map' : 'cheap-module-eval-source-map',
+  devtool: isProd ? false : 'inline-source-map',
   entry: {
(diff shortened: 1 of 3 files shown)


sardaukar:
Final bundle sizes without mangling:
```
app-bundle.js  71.4 kB
vendor-bundle.js   750 kB
```

I'll keep trying to bring it down in further PRs.

webpack.config.js
```diff
@@ -57,6 +58,7 @@ module.exports = {
       'react',
       'es5-shim/es5-shim',
       'es5-shim/es5-sham',
+      'lodash'
```
gylaz: Is this related to this fix?

webpack.config.js
```diff
@@ -36,17 +36,18 @@ if (isProd) {
         if_return: true,
         join_vars: true,
       },
+      mangle: false, // mangling would reduce the bundle's size, but it also makes it not work!
```
gylaz: Perhaps worth mentioning that it doesn't work on Heroku specifically?

webpack.config.js
teoljungberg: It looks like it was omitted from the vendor bundle. So my bet is that its needed.

webpack.config.js
teoljungberg: I'll go ahead and update that

webpack.config.js
teoljungberg: And by omitted I mean that it was forgotten.

sardaukar: Nah, it was in the 'total bundle' this just moves it from the app one to the vendor one to make the app one just have our React components.

PR #1288, opened 2017-01-11
Consolidate the external API of all service objects

teoljungberg:
- Don't duplicate the methods signature in static facades

  Since the method signature, including its parameters are already
  present in the initializer. We can safely remove it from the static
  facade without hurting readability.

- Rename `#run` to `#call`.
- Rename `.run` to `.call`.
- Add static facades where they are missing.

### DIFF
--- a/app/queries/recent_builds_by_repo_query.rb
+++ b/app/queries/recent_builds_by_repo_query.rb
@@ -1,15 +1,13 @@
 class RecentBuildsByRepoQuery
   NUMBER_OF_BUILDS = 20
 
-  def self.run(*args)
-    new(*args).run
-  end
+  static_facade :call
 
   def initialize(user:)
     @user = user
   end
 
-  def run
+  def call
     Build.find_by_sql([<<-SQL, user_id: @user.id, limit: NUMBER_OF_BUILDS])
       WITH user_builds AS (
         SELECT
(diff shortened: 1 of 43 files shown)


app/services/build_config.rb
```diff
@@ -1,6 +1,6 @@
 class BuildConfig
-  def self.for(hound_config:, name:, owner:)
-    new(hound_config: hound_config, name: name, owner: owner).config
+  def self.for(*args)
```
gylaz: This `for` really should just be a `call`.

app/queries/recent_builds_by_repo_query.rb
```diff
@@ -1,15 +1,15 @@
 class RecentBuildsByRepoQuery
   NUMBER_OF_BUILDS = 20
 
-  def self.run(*args)
-    new(*args).run
+  def self.call(*args)
+    new(*args).call
```
gylaz: What are your thoughts on using [`static_facade`](https://github.com/barsoom/attr_extras#static_facade) from `attr_extras` gem?

teoljungberg: I like it!

app/services/build_config.rb
teoljungberg: Good catch!

PR #1289, opened 2017-01-11
Remove Stripe related migrations

teoljungberg:
These files were used to migrate towards the tiered pricing that was
introduced, since all customers have been migrated. These no longer
serve a purpose.


sardaukar:
LGTM :shipit:

PR #1291, opened 2017-01-11
Run react tests in `rake`'s default task

teoljungberg:
- Add `js:spec` task to wrap `npm run build && npm run test`
- Add `builder:audit` task to the `default` task from the get go.

### DIFF
--- a/lib/tasks/js.rake
+++ b/lib/tasks/js.rake
@@ -1,0 +1,5 @@
+namespace :js do
+  task spec: "webpack:build" do
+    sh "npm run test"
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/tasks/js.rake
```diff
@@ -0,0 +1,5 @@
+namespace :js do
+  task spec: "webpack:build" do
+    sh "npm run test"
```
gylaz: You mention `npm run build` in the commit message, but I don't see that here. Do we need it?

teoljungberg: It's being done in `webpack:build` (which is a dependent task on this task).

According to my conversation with @sardaukar, you need to build using webpack before running the test suite. So I guess so.

gylaz:
Might want to update the `circleci.yml` file to just run `rake`.

teoljungberg:
@gylaz Good catch. Thank you!

PR #1293, opened 2017-01-12
Fix coffeelint json config parsing

dzader:
https://github.com/houndci/hound/commit/4632b20bf5c4fb00723ad4002547c445ddf5c7ab#diff-31c9aca594d7a0025b941e49336e8cc5R39 moved all parsing to be yaml - but coffeelint config is stored as json

### DIFF
--- a/spec/models/config/coffee_script_spec.rb
+++ b/spec/models/config/coffee_script_spec.rb
@@ -4,6 +4,7 @@ require "app/models/config/coffee_script"
 require "app/models/config/parser"
 require "app/models/config/parser_error"
 require "app/models/config_content"
+require "app/models/config/json_with_comments"
 require "app/models/missing_owner"
 
 describe Config::CoffeeScript do
@@ -27,6 +28,26 @@ describe Config::CoffeeScript do
       it "returns the content from GitHub as a hash" do
         raw_config = <<~EOS
           { "arrow_spacing": { "level": "error" } }
+        EOS
+        config = build_config(raw_config)
+
+        result = config.content
+
+        expect(result).to eq(
+          "arrow_spacing" => { "level" => "error" },
+        )
+      end
+    end
+
+    context "config with comments and trailing commas" do
+      it "returns the content from Github as a hash" do
+        raw_config = <<~EOS
+          {
+            "arrow_spacing": {
+                // this should be stripped out
+                "level": "error",
+            }
+          }
         EOS
         config = build_config(raw_config)
 
--- a/app/models/config/coffee_script.rb
+++ b/app/models/config/coffee_script.rb
@@ -3,5 +3,13 @@ module Config
     def serialize(data = content)
       Serializer.json(data)
     end
+
+    private
+
+    def parse(file_content)
+      json_with_comments = JsonWithComments.new(file_content)
+      content_without_comments = json_with_comments.without_comments
+      super(content_without_comments)
+    end
   end
 end


gylaz:
YAML is technically a superset of JSON. So, you can parse the JSON config using a YAML parser.

dzader:
@gylaz seems like it does have issues handling comments parsing via YAML parser though - worth moving it back to Parser.json ?

dzader:
and TIL about YAML being a superset of JSON - interesting

gylaz:
> seems like it does have issues handling comments parsing via YAML parser though - worth moving it back to Parser.json ?

That is true. `JSON.parse` seems to handle (strip) comments just fine.

Any reason you closed this? Should we re-open this?

dzader:
@gylaz reopened - was a little trigger happy on closing it

dzader:
still need to add/fix specs

thorncp:
One of the reasons we are using the YAML parser is because it supports trailing commas, which is not uncommon for the JS linters' config files to have. 

```ruby
p YAML.load('{ "the-key": "the-value", }')
p JSON.parse('{ "the-key": "the-value", }')
```

```text
% ruby json.rb
{"the-key"=>"the-value"}
/Users/chris/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/json/common.rb:156:in `parse': 784: unexpected token at '{ "the-key": "the-value", }' (JSON::ParserError)
        from /Users/chris/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/json/common.rb:156:in `parse'
        from json.rb:5:in `<main>'
```

thorncp:
Although it's entirely possible that was a problem specific to ESLint because it supports multiple [config file types](http://eslint.org/docs/user-guide/configuring). Can we get away with a strict JSON parser for coffeelint?

gylaz:
It all comes down to -- is it standard JSON, or is it a JSON-like custom syntax (like ESLint uses, I believe).

We also have [`Config::JsonWithComments`](https://github.com/houndci/hound/blob/master/app/models/config/json_with_comments.rb) which we use with ESLint and TSLint. But now I'm wondering if simply parsing it as JSON would've been enough. Though @thorncp might be right, ESLint allows trailing commas.

dzader:
```
{
    // This name MUST match the default configuration of the rule being loaded.
    "newline_at_eof": {
        // NodeJS module to load. It can also be a path to the rule (useful for devleopment)
        "module": "coffeelint-newline-at-eof",
        // Maybe the rule author set it to error by default and you only want a warning.
        "level": "warn"
    }
}
```

Taken from their website - so seems they are pro-comments

I did switch this PR to use Config::JsonWithComments to be consistent

thorncp:
It looks like the approach is this PR matches what [coffeelint does](https://github.com/clutchski/coffeelint/blob/a5babd23aa094132c0bdf97d47e639650aec72a5/src/configfinder.coffee#L36). This seems good to me.

@dzader Do you mind squashing this down to one commit?

spec/models/config/coffee_script_spec.rb
```diff
@@ -38,6 +39,26 @@
       end
     end
 
+    context "config with comments" do
+      it "returns the content from Github as a hash" do
+        raw_config = <<~EOS
+          { "arrow_spacing":
```
gylaz: Mind putting `"arrow_spacing":` on its own line?

spec/models/config/coffee_script_spec.rb
```diff
@@ -38,6 +39,26 @@
...
+            {
+              // this should be stripped out
+              "level": "error"
+            }
```
gylaz: Worth throwing a trailing comma on this (to make sure YAML works, by JSON won't handle it)?

dzader:
squashed and added trailing comma

dzader:
@gylaz anything else needed here or is this good?

gylaz:
Merged. Thanks @dzader!

PR #1294, opened 2017-01-13
Update config docs

gylaz:
* Remove default config docs for JSHint, since we're now starting new customers with JSHint defaults.
* Use safe new tab opening options

### DIFF
--- a/app/views/application/_header.haml
+++ b/app/views/application/_header.haml
@@ -36,7 +36,7 @@
         %li
           = link_to "Pricing", root_path(anchor: "pricing"), class: "pricing"
         %li
-          - options = { target: "_blank", class: "tweet-us" }
+          - options = new_window_options(class: "tweet-us")
           = link_to "https://twitter.com/houndci", options do
             %span @houndci
         %li.auth-container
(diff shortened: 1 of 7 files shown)


app/views/application/_header.haml
```diff
@@ -36,7 +36,7 @@
         %li
           = link_to "Pricing", root_path(anchor: "pricing"), class: "pricing"
         %li
-          - options = { target: "_blank", class: "tweet-us" }
+          - options = { class: "tweet-us" }.merge(new_window_options)
```
tabfugnic: What are your thoughts on be able to pass a hash to `new_window_options`? That way you can merge it there.

app/views/home/index.haml
```diff
@@ -176,5 +176,5 @@
 
       .actions
         = link_to "https://github.com/houndci/hound/blob/master/doc/SECURITY.md",
-          target: "_blank", class: "security-policy" do
+          { class: "security-policy" }.merge(new_window_options) do
```
tabfugnic: same as above

app/views/application/_header.haml
gylaz: That's an interesting idea. I'll give it a try.

PR #1295, opened 2017-01-13
Ensure inherit config is merged before owner's

gylaz:
There was a logical error, where owner's (org-level) config was merged
with the repo config first, and then merged with the repo's inherited
config. Thus, the last step of merging with inherited allowed for the
owner cofig to override the inherited (which is the repo's config still).

### DIFF
--- a/app/models/config/ruby.rb
+++ b/app/models/config/ruby.rb
@@ -4,7 +4,7 @@ module Config
       if legacy?
         hound_config.content
       else
-        parse_inherit_from(super)
+        super
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/models/config/ruby.rb
```diff
@@ -14,16 +14,20 @@ def serialize(data = content)
 
     private
 
+    def load
+      config = super
+      inherited_config = parse_inherit_from(super)
```
thorncp: Use the `config` variable here instead of calling `super` again? I don't have a good reason, but it feels weird to call `super` twice.

gylaz: Totally!

spec/models/config/ruby_spec.rb
```diff
@@ -74,21 +74,28 @@
...
+          "EnforcedStyle" => "double_quotes",
+        },
+        "Style/HashSyntax" => {
+          "EnforcedStyle" => "ruby19",
+        },
+      }
+      owner = instance_double("Owner", config_content: owner_config)
       hound_config = build_hound_config(commit, "config/rubocop.yml")
-      config = described_class.new(hound_config)
+      config = described_class.new(hound_config, owner: owner)
```
teoljungberg: We're not actively using `described_class` anywhere in the codebase, so for consistency. I'd use `Config::Ruby` here.

gylaz: Done.

PR #1296, opened 2017-01-17
Show which repos are private in the list

LkeMitchll:
Right now there's no way to identify which repos in the list are private or not, after toggling 'include private repos'

This PR adds a visual indicator to help with this, on hover there is a label:

screen shot 2017-01-17 at 11 31 10 am

### DIFF
--- a/app/assets/javascripts/components/__tests__/__snapshots__/repo-test.jsx.snap
+++ b/app/assets/javascripts/components/__tests__/__snapshots__/repo-test.jsx.snap
@@ -7,6 +7,15 @@ exports[`test renders a repo appropriately 1`] = `
   </div>
   <div
     className="repo-activation-toggle">
+    <label
+      className="repo-label tooltip">
+      <i
+        className="fa fa-lock" />
+      <span
+        className="tooltip--message">
+        Private Repo
+      </span>
+    </label>
     <RepoActivationButton
       isProcessingId={null}
       onRepoClicked={[Function]}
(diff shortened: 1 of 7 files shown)


sardaukar:
LGTM :shipit:

r00k:
At first glance, the lock icon made it seem to be that I _could not_ enable Hound on those repos.

They seem _locked_, rather than _private_.

r00k:
Finally, does this info add value? Does anything change when a repo is private? If not, why include it in the UI?

gylaz:
> Does anything change when a repo is private?

Private repos count against paid tier quota. Thus, enabling a private repo that would put you in the next bucket will show an upgrade banner, where a public repo won't. Not sure if visually that needs to be differentiated.

r00k:
> Private repos count against paid tier quota.

Oh! Duh. I forgot :).

In that case, I think it's worth displaying in the UI.

LkeMitchll:
[Refine repos page UX](https://trello.com/c/uTdmx2Ga/769-refine-repos-page-ux)

smharley:
I agree with Ben above. Would it make more sense if the [Private Repo] label was next to the repo name, instead of next to the Activate button? Also seems like there's plenty of room to show [Private Repo] next to all of the private repos instead of only showing the label on hover. Might help alleviate some of that confusion and keep it simple

tysongach:
I agree with the points and also think that this whole thing could be simplified a bit:

- What about removed the tooltip altogether? They add roadblocks for assistive technology, so some people may not be able to access that content. Instead just a simple “private” label next to the repo name could be a good approach?
- If we do the above, no icon is need, so no confusion there.
- Instead of `repo-label`, can we make this a more reusable component by just naming it `label` or `badge`?

tysongach:
Closing this, as it appears https://github.com/houndci/hound/pull/1383 will get merged, instead.

PR #1298, opened 2017-01-18
Always merge configurations

purinkle:
Before, the Python and Go linters had their own specific configurations when they did not need to. Updated both of the linters to inherit the configuration implementation from their bases.

https://trello.com/c/PMwYwgPL

### DIFF
--- a/app/models/linter/python.rb
+++ b/app/models/linter/python.rb
@@ -4,10 +4,6 @@ module Linter
 
     private
 
-    def config
-      Config::Python.new(hound_config, owner: owner)
-    end
-
     def enqueue_job(attributes)
       Resque.push(
         "python_review",
--- a/spec/models/config/go_spec.rb
+++ b/spec/models/config/go_spec.rb
@@ -2,14 +2,25 @@ require "spec_helper"
 require "app/models/config/base"
 require "app/models/config/go"
 require "app/models/config/parser"
+require "app/models/config_content"
+require "app/models/missing_owner"
 
 describe Config::Go do
   describe "#content" do
-    it "returns an empty string" do
-      hound_config = double("HoundConfig")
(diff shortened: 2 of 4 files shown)


app/models/linter/python.rb
```diff
@@ -4,10 +4,6 @@ class Python < Base
 
     private
 
-    def config
-      Config::Python.new(hound_config, owner: owner)
```
teoljungberg: I don't see a modified test for this for Python. Is one needed?

spec/models/config/go_spec.rb
```diff
@@ -2,14 +2,25 @@
...
-      config = Config::Go.new(hound_config)
+    it "returns an empty hash" do
+      config_content = instance_double("ConfigContent", load: {})
+      content = { go: { enabled: true } }
+      raw_content = <<~EOS
+        go:
+          enabled: true
+      EOS
+      commit = instance_double("Commit", file_content: raw_content)
+      hound_config = double("HoundConfig", commit: commit, content: content)
```
teoljungberg: This could be an `instance_double` too.

app/models/linter/python.rb
purinkle: Because this is a private method, I thought that there should be no need to update the existing tests as the functionality should stay the same.

teoljungberg: Fair reasoning. I was interested in if there was any tests that needed to be changed (since CI was failing).

But if this is all green, then this is fine.

purinkle: The test was failing because the unit test had overstretched itself into areas of the code it should not have been touching. All sorted now.

spec/models/config/go_spec.rb
purinkle: Resolved in 04dc610

PR #1299, opened 2017-01-18
[WiP] Add org-level config component

purinkle:


app/assets/stylesheets/components/_switch.scss
```diff
@@ -0,0 +1,60 @@
+$switch-height: 32px;
+$switch-padding: 2px;
+$knob-size: $switch-height - ($switch-padding * 2) - 2px;
+$switch-width: 52px;
```
gylaz: Maybe move these to `app/assets/stylesheets/base/_variables.scss`?

spec/factories.rb
```diff
@@ -91,8 +91,8 @@
   end
 
   factory :owner do
-    github_id
-    name { generate(:github_name) }
+    github_id 1
+    name "TEST_GITHUB_LOGIN"
```
gylaz: Why was this change necessary? Both `github_id` and `name` are unique, so creating two `owner` factories in one test would fail.

app/assets/javascripts/components/organization_configuration.jsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+    Ajax.updateOwner(this.props.id, Object.assign({}, currentState, options));
+  }
+
+  renderRepoSelect() {
+    if (this.state.enabled) {
+      return (
+        <div className="organization-header-config">
+          <span className="organization-header-label">
+            Use hound.yml from
```
gylaz: `User .hound.yml from`?

app/assets/javascripts/components/organization_configuration.jsx
```diff
@@ -0,0 +1,111 @@
...
+          <label className="toggle-switch" htmlFor={switchId} />
+        </div>
+
+        {this.renderRepoSelect()}
+      </div>
+    );
+  }
+}
+
+module.exports = OrganizationConfiguration;
```
gylaz: This is certainly a lot of code for a dropdown and a toggle. Part of the problem I have with JS frameworks. Just my 2 cents.

purinkle:
[Support org-level styleguide.](https://trello.com/c/HY2hoTdq/691-support-org-level-styleguide)

app/assets/javascripts/components/organization_configuration.jsx
purinkle: If the main repo list was not rendered using AJAX but using standard Rails then it would be easier to add these sort of components. For me, it would be better if we could get the whole application to work without JS and then add enhanced functionality with JS. For example, this would just be a small `form` for each organisation that would submit normally. We would then "enhance" the functionality to allow the form to submit via AJAX.

PR #1300, opened 2017-01-19
Removed duplicate private modifier

olimart:


gylaz:
Nice catch! Thanks @olimart.

PR #1301, opened 2017-01-24
Set Hound up to Flog Ruby changes

purinkle:
Before, we only checked Ruby files using Rubocop. This only allowed us to comment on the style of the files. Added a Flog integration which will comment on Ruby methods that are too complex.

### DIFF
--- a/spec/models/linter/flog_spec.rb
+++ b/spec/models/linter/flog_spec.rb
@@ -1,0 +1,85 @@
+require "rails_helper"
+
+module Linter
+  RSpec.describe Flog do
+    describe ".can_lint?" do
+      context "when given a '.rb' file" do
+        it "is true" do
+          expect(Flog.can_lint?("foo.rb")).to be(true)
+        end
+      end
+
+      context "when given a '.rake' file" do
+        it "is true" do
+          expect(Flog.can_lint?("foo.rake")).to be(true)
+        end
+      end
+
+      context "when given a non-Ruby file" do
+        it "is false" do
+          expect(Flog.can_lint?("foo.js")).to be(false)
+        end
+      end
+    end
+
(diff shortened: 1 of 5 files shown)


spec/models/linter/flog_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      context "when given a non-Ruby file" do
+        it "is false" do
+          expect(Flog.can_lint?("foo.js")).to be(false)
+        end
+      end
+    end
+
+    describe "#enabled?" do
+      it "is if the linter is enabled in the Hound configuration" do
```
jayroh: Would you consider changing this to `it "is true if the linter is enabled ...` or placing this within a context?

```ruby
context "when Flog linter is enabled" do
  it "is true" do
```

purinkle: Resolved in 774582d

jayroh:
Over in my work with reek I'm running into some issues with `Config::Base` due to my not having a `Config::Reek` class. How are your tests passing if, in `app/models/config/base.rb` it looks for the linter name here

```end
def linter_name
  self.class.name.demodulize.underscore
end
```

I'm getting 

```
[3] pry(#<Config::Unsupported>)> linter_name
=> "unsupported"
[4] pry(#<Config::Unsupported>)> self.class
=> Config::Unsupported
```

purinkle:
@jayroh, I managed to get around your config issue because Flog does not support any configuration. Thus I have not had to add a config class. I overwrote the serialisation in my `Linter::Flog` class.

jayroh:
@purinkle I, too, think this should be good to land! Would love to get this up to staging for some testing along with the reek branch

PR #1302, opened 2017-01-24
Dry up testing common linter behavior

jayroh:
While reviewing the specs for the suite of linters there is common
behavior that is repeatedly tested that is ripe for extraction to
something like a shared example.

To reduce boilerplate I've started with one quick win - a small test to
assert that linters allow only certain types of files and/or extensions.

### DIFF
--- a/spec/models/linter/base_spec.rb
+++ b/spec/models/linter/base_spec.rb
@@ -10,15 +10,10 @@ module Linter
   end
 end
 
-describe Linter::Base do
-  describe ".can_lint?" do
-    it "uses the FILE_REGEXP to determine the match" do
-      yes_lint = Linter::Test.can_lint?("foo.yes")
-      no_lint = Linter::Test.can_lint?("foo.no")
-
-      expect(yes_lint).to eq true
-      expect(no_lint).to eq false
-    end
+describe Linter::Test do
+  it_behaves_like "a linter" do
+    let(:lintable_files) { %w(foo.yes) }
+    let(:not_lintable_files) { %w(foo.no bar.nope) }
   end
 
   describe "#file_included?" do
--- a/spec/models/linter/coffee_script_spec.rb
+++ b/spec/models/linter/coffee_script_spec.rb
@@ -1,38 +1,9 @@
 require "rails_helper"
 
 describe Linter::CoffeeScript do
-  describe ".can_lint?" do
-    context "given a .coffee file" do
-      it "returns true" do
-        result = Linter::CoffeeScript.can_lint?("foo.coffee")
-
-        expect(result).to eq true
-      end
-    end
-
-    context "given a .coffee.erb file" do
-      it "returns true" do
-        result = Linter::CoffeeScript.can_lint?("foo.coffee.erb")
-
-        expect(result).to eq true
-      end
-    end
-
-    context "given a .coffee.js file" do
-      it "returns true" do
-        result = Linter::CoffeeScript.can_lint?("foo.coffee.js")
-
-        expect(result).to eq true
-      end
-    end
-
-    context "given a non-coffee file" do
-      it "returns false" do
-        result = Linter::CoffeeScript.can_lint?("foo.js")
-
-        expect(result).to eq false
-      end
-    end
+  it_behaves_like "a linter" do
+    let(:lintable_files) { %w(foo.coffee foo.coffee.erb foo.coffee.js) }
+    let(:not_lintable_files) { %w(foo.js) }
   end
 
   describe "#file_review" do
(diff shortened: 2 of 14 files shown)


jayroh:
This is, for all intents and purposes, a preliminary effort to help reduce some boilerplate for the entire collection of linter models we have in hound currently. Additionally this helps future linter authors more quickly test the common assertion(s) more succinctly.

As someone coming in (pretty blind) to write a few new linters it took me a little time to determine what the common methods and interfaces are and what I should be testing.

If I saw something like this I could very quickly see - "oh, I need to ensure my linter checks for certain file types. Those types are X, Y, and not Z".

spec/models/linter/base_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "spec_helper"
+require "shared_linter_examples"
```
jayroh: I explicitly placed this file in `./spec` and required it so as not to hide it in the (pardon the phrase) junk-drawer'ness that is `spec/support`. I prefer to know where things like a shared example are being pulled *from*

Of course, I could be convinced otherwise

spec/models/linter/coffee_script_spec.rb
```diff
@@ -69,13 +41,13 @@
   end
 
   def stub_coffeelint_config(config = {})
-    stubbged_config = instance_double(
+    stubbed_config = instance_double(
```
jayroh: fixed typos here, and below

spec/models/linter/base_spec.rb
```diff
@@ -10,15 +11,10 @@ class Test < Base
...
-  describe ".can_lint?" do
-    it "uses the FILE_REGEXP to determine the match" do
-      yes_lint = Linter::Test.can_lint?("foo.yes")
-      no_lint = Linter::Test.can_lint?("foo.no")
-
-      expect(yes_lint).to eq true
-      expect(no_lint).to eq false
-    end
+describe Linter::Test do
+  it_behaves_like "a linter" do
```
jayroh: Thoughts on how descriptive this could be?

`it_behaves_like "a linter"` is more generic and could encompass further/other types of behavior. For now, however, it only contains tests for `#can_lint?`.

As a result this could be updated to something like `it_behaves_like "a linter checking for matching file names"`

spec/models/linter/base_spec.rb
teoljungberg: The thoughtbot convention seems to be `spec/support/shared_examples` - https://github.com/thoughtbot/suspenders/blob/2a955cb49a909dc10b5c9437334b5c4330bc8cc7/lib/suspenders/app_builder.rb#L344.

Personally, I'm fine with it being placed in there.

spec/models/linter/base_spec.rb
teoljungberg: I like how general the string is, since it allows it to be extended further in the future.

spec/models/linter/coffee_script_spec.rb
teoljungberg: Nice catch!

spec/shared_linter_examples.rb
```diff
@@ -0,0 +1,15 @@
+RSpec.shared_examples "a linter" do
+  describe ".can_lint?" do
+    it "uses the FILE_REGEXP to determine the match" do
+      lintable_files.each do |file|
```
teoljungberg: What do you think about extracting these into two separate test cases? One for the `yes` part, and one for the `no` part?

spec/shared_linter_examples.rb
```diff
@@ -0,0 +1,15 @@
...
+        yes_lint = described_class.can_lint?(file)
+        expect(yes_lint).to eq true
```
teoljungberg: To follow inline with the four-phase test. I'd add a newline above the expectation.

spec/shared_linter_examples.rb
jayroh: Excellent idea, my good man.

spec/shared_linter_examples.rb
jayroh: Good catch 👍

spec/models/linter/base_spec.rb
purinkle: I would happy with a mixture of the two solutions. Sling the file in `spec/support` and then have an explicit `require` in the test file.

jayroh: @purinkle I think that's a fair approach.

To be more clear as to why I prefer the explicitness of requiring it - shared examples aren't something I run into often amongst all projects I've worked on. So when I do, it catches me a little off guard, as there is rarely a convention as to where I can find where they are defined.

The `require` in this case is more like a hint, than it is a necessity.

teoljungberg: I for one would omit the require, and count on it being loaded when RSpec "boots". As that is the convention I have seen in other applications that have used shared examples.

spec/models/linter/base_spec.rb
gylaz: Chiming in post factum, but I've been generally shying away from RSpec examples (they don't really provide anything a method can't handle). In `linters` we just used a method to extract out common behavior (instead of a shared example): https://github.com/houndci/linters/blob/master/spec/support/helpers/linters_helper.rb#L2

teoljungberg: Hm, I forgot that was the setup we are using in this project.

I do agree with @gylaz, we should be consistent, so I'd move this into a module.

PR #1303, opened 2017-01-24
By default, do not enable HTTPS when running specs

jayroh:
Until this commit the specs hitting app endpoints would have their
behavior altered depending on what was set in your `.env.local` file for
`ENABLE_HTTPS`. It should always be set to "no" when the specs are being
run.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,2 @@
+RAILS_ENV=test
+ENABLE_HTTPS=no
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,4 +1,5 @@
 ENV["RAILS_ENV"] ||= "test"
+ENV["ENABLE_HTTPS"] = "no"
 
 require "config/environment"
 require "rspec/rails"
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,5 +1,4 @@
 ENV["RAILS_ENV"] ||= "test"
-ENV["ENABLE_HTTPS"] = "no"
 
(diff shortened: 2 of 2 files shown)


spec/rails_helper.rb
```diff
@@ -1,4 +1,5 @@
 ENV["RAILS_ENV"] ||= "test"
+ENV["ENABLE_HTTPS"] = "no"
```
teoljungberg: Would this make more sense to have in `.env.test`?

teoljungberg: Which tests require us to dissallow `HTTPS`? Shouldn't they instead leverage `stub_const("Hound:: HTTPS_ENABLED", false)` instead?

https://github.com/houndci/hound/blob/44a5039bcb3326d052a744ea40e78599b0ac7217/config/initializers/constants.rb#L3

purinkle: I agree with @teoljungberg that this should reside in a `.env` file of some sort. I'd resist `stub_const`, though.

teoljungberg: @purinkle We're already using `stub_const` quite heavily.

.env.test
```diff
@@ -0,0 +1,2 @@
+RAILS_ENV=test
+ENABLE_HTTPS=no
```
jayroh: Moved the configuration here per your suggestions @teoljungberg and @purinkle.

the RAILS_ENV is set here too - because, if not, it would somehow pick up another environment and the tests would fail in a similar fashion.

PR #1304, opened 2017-01-25
Add reek as a linter to main Hound app

jayroh:
In addition to the soon-to-be merged flog linter, this commit enables using reek to surface potential code smells.

More information on reek is available here: https://github.com/troessner/reek

### DIFF
--- a/app/jobs/reek_review_job.rb
+++ b/app/jobs/reek_review_job.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+class ReekReviewJob
+  @queue = :reek_review
+end
--- a/app/models/config/reek.rb
+++ b/app/models/config/reek.rb
@@ -1,0 +1,7 @@
+module Config
+  class Reek < Base
+    def content
+      ""
+    end
+  end
+end
--- a/app/models/hound_config.rb
+++ b/app/models/hound_config.rb
@@ -10,6 +10,7 @@ class HoundConfig
     Linter::Jshint,
     Linter::Remark,
     Linter::Python,
+    Linter::Reek,
     Linter::Ruby,
     Linter::Scss,
     Linter::Swift,
@@ -19,6 +20,7 @@ class HoundConfig
     credo
     eslint
     flog
+    reek
     remark
     python
     tslint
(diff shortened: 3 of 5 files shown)


croaky:
Possible to summarize the new capabilities Reek provides over what Hound currently checks, and which things overlap?

https://github.com/troessner/reek/blob/master/docs/Code-Smells.md
https://github.com/bbatsov/rubocop/tree/master/lib/rubocop/cop

Will Hound comment twice on the same violation found by Reek and Rubocop? Three times if found by Reek, Rubocop, and Flog? Do we need to write and maintain some kind of "flattening" mechanism for the overlap?

jayroh:
@croaky you're right in that there's some overlap. I think a good approach with this particular linter - I'd consider it a very beta experiment  - is to see how it behaves in conjunction with both rubocop and flog and tune as necessary.

To answer your question - I don't think we want a duplication of similar messages across both rubocop/reek/flog (Of those I think rubocop and reek would be the most likely to overlap). For that reason, some way to flatten it might be ideal.

I think @gylaz had started some work to consolidate N comments into one review comment/set. Which, if it does some of that comment consolidation, could be used to facilitate this.

jayroh:
That being said, would anyone object to me pulling this in?

jayroh:
Ping @croaky. What you think, Dan?

cc @gylaz

croaky:
Is it disabled by default?

jayroh:
It is disabled by default. Yes.

teoljungberg:
I think it's worth investigating how we like the combination Rubocop + Reek.
If it turns out that they generally comment on the same violations, it might be
worth investigating how we can flatten the messages.

gylaz:
I don't think there is anything wrong with supporting both. It may be a use case that someone doesn't care for RuboCop but does enjoy what Reek has to say. Either way, Reek is off by default, so someone *has* to opt-in.

jayroh:
Ok. It sounds like because it's disabled by default this is something we can feel ok continuing to explore. If there's duplicate noise, then we'll figure out how to best handle that with something like a merge+flatten strategy.

I'm going to merge this and move it along the pipeline.

PR #1305, opened 2017-01-26
Adds info on beta linters, re-orders the markup to match TOC layout

sardaukar:
- more info on Go linting
- adds info on TS and Elixir linting

### DIFF
--- a/app/views/pages/configuration.haml
+++ b/app/views/pages/configuration.haml
@@ -20,9 +20,13 @@
       %li
         = link_to "Swift", "#swift", class: "docs-nav-link"
       %li
-        = link_to "ESLint (beta)", "#eslint", class: "docs-nav-link"
+        = link_to "ESLint", "#eslint", class: "docs-nav-link"
       %li
-        = link_to "Python (beta)", "#python", class: "docs-nav-link"
+        = link_to "Python", "#python", class: "docs-nav-link"
+      %li
+        = link_to "TSLint", "#tslint", class: "docs-nav-link"
+      %li
+        = link_to "Elixir", "#elixir", class: "docs-nav-link"
 
   %section.docs-content
     %article.docs-article
@@ -42,8 +46,17 @@
     %article.docs-article#configuration
       %h3 Hound Configuration
 
+      %p The following linters are enabled by default:
       %p
-        All supported linters, except the ones in beta, are enabled by default.
+        %ul
+          %li RuboCop (Ruby)
+          %li CoffeeLint (CoffeeScript)
+          %li JavaScript (JSHint)
+          %li SCSS (SCSSLint)
+          %li Haml (HAMLLint)
+          %li Go (golint)
+          %li Swift (SwiftLint)
+      %p All other linters must be explicitly enabled.
 
       %p
         Hound will look for a custom configuration file named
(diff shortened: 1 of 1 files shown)


sardaukar:
Pinging @jakecraige and @jayroh for input on this, not 100% sure on the new linter config keys

jayroh:
The new linters can remain off the radar for the time being while we put them through their paces with a small subset of customers and test repos. So no worries on those two in particular (for now). Once they're ready to go we'll get a PR for them up 😄

croaky:
The "document Golint" part looks good to merge at your discretion!

app/views/pages/configuration.haml
```diff
@@ -399,47 +452,80 @@
...
+          = link_to "Credo docs",
+            "http://credo-ci.org/",
+            new_window_options
+        and reference it in
+        %em.code-inline .hound.yml
+
+        %code.code-block
+          :preserve
+            elixir:
+              config_file: .credo.exs
```
jakecraige: The config file is good here, but you would also need to specify `enabled: true` for it to work while in beta.

jakecraige:
Config stuff looks good for Elixir. Had one comment about making sure it's enabled. It's been on a few internal repos for the past few weeks without issue so I'd be comfortable adding it as a public beta

app/views/pages/configuration.haml
```diff
@@ -23,6 +23,10 @@
         = link_to "ESLint (beta)", "#eslint", class: "docs-nav-link"
       %li
         = link_to "Python (beta)", "#python", class: "docs-nav-link"
+      %li
+        = link_to "TSLint (beta)", "#tslint", class: "docs-nav-link"
+      %li
+        = link_to "Elixir (beta)", "#elixir", class: "docs-nav-link"
```
teoljungberg: We haven't even added support for Elixir in Hound proper, which means we haven't even tested it out.

I'd personally prefer if we tested it out internally first before releasing it into the public.

Thoughts?

jakecraige: What do you mean? It is in Hound proper, https://github.com/houndci/hound/pull/1268. As briefly discussed in Slack we have been using it internally on a few projects with no issues so far, so I'd be okay with having this public beta. But not opposed to waiting a bit longer if we want. I don't predict any changes in that time though

croaky:
I agree Elixir seems like it's been used a bunch on thoughtbot projects and we should feel comfortable releasing documentation on it now.

croaky:
Do we need the "beta" distinction at all? Should "beta" for Hound mean simply undocumented and unpromoted on the website, but able to be enabled by thoughtbot or people following along on the OSS projects?

gylaz:
I think we should rename "beta" linters to "disabled by default", since we'll need that now. For example, ESLint is no longer "beta" but it's disabled to favor JSHint out-of-the-box.

croaky:
> rename "beta" linters to "disabled by default"

Nice. More actionable

app/views/pages/configuration.haml
teoljungberg: > What do you mean? It is in Hound proper, #1268.

You're totally right, I missed out upon that.

If you're happy with how stable it is, I am too!

gylaz:
Made the requested changes here, plus removed the "beta" meaning. Going to merge this.

PR #1307, opened 2017-01-26
Document deployment

teoljungberg:
To avoid confusion for new contributors on how the deployment of the
application works, an update to the documentation was required.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -222,6 +222,15 @@ outbound `ReportInvalidConfigJob` queue:
 * `pull_request_number` - The GitHub pull request number. This is provided by
   the inbound queue.
 
+## Deploying
+If you have previously run the `bin/setup` script, you can deploy to staging
+and production with:
+
+```sh
+% bin/deploy staging
+% bin/deploy production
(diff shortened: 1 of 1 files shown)


teoljungberg:
I opened a PR for hound/linters too for its documentation - https://github.com/houndci/linters/pull/92

jayroh:
After deploying is there anything one should, or could, monitor as a bit of a sanity check? Tail the logs on heroku for the main rails app? View the jobs queue for the linter app / workers?

I'm not suggesting doing those automatically but it might be helpful to echo something out saying something to the effect of: 

> To verify health of the system after deployment, tail logs with `heroku logs --tail ...` and/or view the jobs queue at the following URL.

teoljungberg:
@jayroh Sorry for dropping the ball on this, I had totally forgotten I had opened this PR.

Tailing the logs sadly doesn't tell much, and the background jobs queue is finicky at best and tends to cough a while after the deploy. Meaning the deployer probably will have closed the tab by then.
Since we are reporting the size of the queue once it goes over a certain threshold, I think we can leave it at that.

PR #1314, opened 2017-02-06
Use yarn instead of npm to manage node dependencies

blackjid:
Heroku should run yarn instead of npm install on the presence of this file


gylaz:
Not sure about this. We're not using yarn in Hound, are we? How does this lock file get generated, how do we update it?

blackjid:
Yarn should be a drop in replacement to `npm` in the complete workflow, but it creates the `yarn.lock` file and maintain the dependencies version resolutions.

https://yarnpkg.io

The `yarn.lock` file is created the first time just running `yarn` on a project where you have `package.json` file.

After that, if you run `yarn` on a project with the `yarn.lock` you can be sure wich versions of the dependencies are installed.

To add dependencies `yarn add <package-name>` and they will be added to the `package.json` and locked in `yarn.lock`
To update just `yarn update <package-name>`

The nodejs buildpack for heroku use `yarn` instead of `npm` if the project has a `yarn.lock` file.

gylaz:
I think we should probably decide if we want to switch to Yarn, and then replace all the references to `npm` with `yarn`.

Merging just `yarn.lock` will create a out-of-sync issues if we continue using `npm` and add/update some libraries.

blackjid:
Makes sense, maybe this is a good place to start that discussion.

gylaz:
Sounds good to me. I'm not opposed to the switch.

Do you want to track down `npm` references and switch them over to `yarn` in this PR?

blackjid:
Ok, I also opened a PR to the linters repo, https://github.com/houndci/linters/pull/94
I will take a look at that one too..

gylaz:
Thanks @blackjid!

PR #1317, opened 2017-02-10
Don't submit a review for zero violations

thorncp:
Some jobs are failing with what appears to be an error message from
GitHub complaining about an empty review. We don't have great logging
capabilities into the job failures, so I'm making an assumption that not
submitting a review with empty comments with fix this.

> undefined method `map' for "You need to leave a comment indicating the
> requested changes."


jcmorrow:
So the review thing is totally separate from the CI status?

thorncp:
It is. And because the review piece happens first, it blowing up causes the status to never be set again (seen in this PR where the status is still pending).

PR #1318, opened 2017-02-15
SEO updates for homepage (title, meta, heading)

jayroh:


app/views/home/index.haml
```diff
@@ -1,4 +1,5 @@
-- content_for :page_title, "Hound"
+- content_for :page_title, "Github Code Review Tool for JS, Ruby, Swift, Elixir, and more"
```
croaky: "GitHub" (capital "H") everywhere

croaky: Not sure if we want "JavaScript" instead of "JS". If we do, we'd want to drop Swift or Elixir to stay within the 50-60 character SEO length guideline.

```
> "Github Code Review Tool for JS, Ruby, Swift, Elixir, and more".size
=> 61
```

app/views/home/index.haml
```diff
@@ -1,4 +1,5 @@
...
+- content_for :meta_description, "Hound is the GitHub code review tool the thoughtbot team wanted but never got, so they built it. You can use it too. Now checking Ruby, JS, Swift, CoffeeScript, Go and more. "
```
croaky: Could go `do` block style or move this to locales to hard wrap at 80 characters

app/views/home/index.haml
jayroh: Going to drop Swift and keep Elixir, since customers for this are more likely to fall within that Elixir camp. Just a hunch

teoljungberg:
For future reference, this was merged as https://github.com/houndci/hound/commit/3d3b26c45df83be12630e6bf4f78ba32b5ce48a4

PR #1322, opened 2017-03-03
Masquerading as a user

jakecraige:
This adds a route that admins can go to that will put them in a mode where they
masquerade as the user they provided. They can then browse hound as if they were
that user, and then click "Stop Masquerading" to return to their own account.

screen shot 2017-03-03 at 15 41 05

https://trello.com/c/h9VtPK6F/800-as-an-admin-i-would-like-to-masquerade-as-a-user-so-that-i-may-help-debug-issues-that-they-are-running-into

### DIFF
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -2,7 +2,15 @@ module Admin
   class ApplicationController < Administrate::ApplicationController
     before_action :authenticate_admin
 
+    helper_method :sidebar_resources
+
     private
+
+    def sidebar_resources
+      Administrate::Namespace.new(namespace).resources.select do |resource|
+        DashboardManifest::DASHBOARDS.include?(resource)
+      end
+    end
 
     def authenticate_admin
       unless github_admin?
--- a/app/controllers/admin/masquerades_controller.rb
+++ b/app/controllers/admin/masquerades_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+module Admin
+  class MasqueradesController < Admin::ApplicationController
+    def show
+      requested_user = User.find_by(username: params[:username])
+      session[:masqueraded_user_id] = requested_user.id
+
+      redirect_to repos_path
+    end
+
+    def destroy
+      session.delete(:masqueraded_user_id)
+
+      redirect_to repos_path
+    end
+  end
+end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
   before_action :capture_campaign_params
   before_action :authenticate
 
-  helper_method :current_user, :signed_in?
+  helper_method :current_user, :signed_in?, :masquerading?
 
   private
 
@@ -40,16 +40,28 @@ class ApplicationController < ActionController::Base
   end
 
   def current_user
-    @current_user ||= User.where(remember_token: session[:remember_token]).first
+    @_current_user ||= find_user_or_masqerade
   end
 
   def analytics
-    @analytics ||= Analytics.new(current_user, session[:campaign_params])
+    @_analytics ||= Analytics.new(current_user, session[:campaign_params])
+  end
+
+  def masquerading?
+    session[:masqueraded_user_id]
   end
 
   protected
 
   def verified_request?
     super || valid_authenticity_token?(session, request.headers["X-XSRF-TOKEN"])
+  end
+
+  def find_user_or_masqerade
+    if masquerading?
+      User.find_by(id: session[:masqueraded_user_id])
+    else
+      User.find_by(remember_token: session[:remember_token])
+    end
   end
 end
--- a/spec/features/admin_masquerades_spec.rb
+++ b/spec/features/admin_masquerades_spec.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+feature "Admin masquerades as user" do
+  scenario "admin sees user repos and can stop masquerading" do
+    repo = create(:repo)
+    admin = create(:user, username: "admin", token: "admin")
+    user = create(:user, repos: [repo], username: "admin", token: "user")
+    stub_const("Hound::ADMIN_GITHUB_USERNAMES", ["admin"])
+
+    sign_in_as(admin, "admin")
+    visit admin_masquerade_path(user.username)
+
+    expect(current_path).to eq(repos_path)
+    within "header .account" do
+      expect(page).to have_text(user.username)
+    end
+
+    click_on "Stop Masquerading"
+
+    within "header .account" do
+      expect(page).to have_text(admin.username)
+    end
+  end
+end
--- a/app/views/admin/application/_sidebar.html.haml
+++ b/app/views/admin/application/_sidebar.html.haml
@@ -1,0 +1,6 @@
+%ul.sidebar__list
+  - sidebar_resources.each do |resource|
+    %li
+      = link_to display_resource_name(resource),
+        [namespace, resource],
+        class: "sidebar__link sidebar__link--#{nav_link_state(resource)}"
--- a/app/views/application/_header.haml
+++ b/app/views/application/_header.haml
@@ -31,7 +31,12 @@
               .avatar{ style: "background-image: url('#{avatar}')" }
             %span= current_user.username
         %li
-          = link_to t("sign_out"), sign_out_path, class: "sign-out"
+          - if masquerading?
+            = link_to t("stop_masquerading"),
+              admin_masquerade_path(current_user.username),
+              method: :delete
+          - else
+            = link_to t("sign_out"), sign_out_path, class: "sign-out"
       - else
         %li
           = link_to "Pricing", root_path(anchor: "pricing"), class: "pricing"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -4,6 +4,7 @@ en:
   syncing_repos: "Loading..."
   authenticate: "Sign In with GitHub"
   sign_out: "Sign Out"
+  stop_masquerading: "Stop Masquerading"
   active_repos: "Active repos"
   search_placeholder: "Search by repo"
   pending_status: "Hound is busy reviewing changes..."
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,6 +4,7 @@ Houndapp::Application.routes.draw do
       resources dashboard_resource
     end
 
+    resources :masquerades, param: :username, only: [:show, :destroy]
     root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
   end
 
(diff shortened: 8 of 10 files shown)


app/controllers/admin/masquerades_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+module Admin
+  class MasqueradesController < ApplicationController
+    def show
+      requested_user = User.find_by(username: params[:username])
+      session[:masqueraded_user_id] = requested_user.id
+      redirect_to repos_path
+    end
+
+    def destroy
+      session[:masqueraded_user_id] = nil
```
thorncp: What do you think about using `#delete` here so the key is removed altogether?

thorncp:
Is the route a `show` for the east of using the feature (no ui)?

app/controllers/admin/masquerades_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+module Admin
+  class MasqueradesController < ApplicationController
+    def show
+      requested_user = User.find_by(username: params[:username])
+      session[:masqueraded_user_id] = requested_user.id
+      redirect_to repos_path
```
teoljungberg: Thoughts on adding a newline above here to denote that the redirect is the last "step" of the action?

app/controllers/admin/masquerades_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+    end
+
+    def destroy
+      session[:masqueraded_user_id] = nil
+      redirect_to repos_path
```
teoljungberg: Same suggestion as above.

app/controllers/application_controller.rb
```diff
@@ -40,13 +40,23 @@ def signed_in?
   end
 
   def current_user
-    @current_user ||= User.where(remember_token: session[:remember_token]).first
+    @current_user ||= begin
```
teoljungberg: Thought's on extracting the actual finding of the user to another method?

teoljungberg: Thought's on prefixing the instance variable with `_` to denote that it's private? (and since instance variables leak in controller).

app/controllers/application_controller.rb
```diff
@@ -40,13 +40,23 @@ def signed_in?
...
+      if masquerading?
+        User.find_by(id: session[:masqueraded_user_id])
+      else
+        User.find_by(remember_token: session[:remember_token])
+      end
+    end
   end
 
   def analytics
     @analytics ||= Analytics.new(current_user, session[:campaign_params])
```
teoljungberg: I'm make the same suggestion here too.

app/controllers/application_controller.rb
```diff
@@ -40,13 +40,23 @@ def signed_in?
...
   end
 
+  def masquerading?
+    session[:masqueraded_user_id]
+  end
+
   protected
```
teoljungberg: What's the purpose of having a `protected` modifier in a controller?

app/views/admin/application/_sidebar.html.erb
```diff
@@ -0,0 +1,11 @@
+<ul class="sidebar__list">
```
teoljungberg: Should this be HAML instead of ERB?

spec/controllers/admin/masquerades_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe Admin::MasqueradesController, "#show" do
```
teoljungberg: Thought's on moving this to a request spec instead of a controller spec?

spec/features/admin_masquerades_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
teoljungberg: Normally - we name the feature specs after the role:

> Use names like ROLE_ACTION_spec.rb, such as user_changes_password_spec.rb, for feature spec file names.

From https://github.com/thoughtbot/guides/tree/master/style/testing

spec/features/admin_masquerades_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+require "rails_helper"
+
+feature "Admin masquerades as user" do
+  scenario "admin sees user repos and can stop masquerading" do
+    admin = create(:user, token: "admin")
+    repo = create(:repo)
+    user = create(:user, repos: [repo], token: "user")
+    stub_const("Hound::ADMIN_GITHUB_USERNAMES", [admin.username])
```
teoljungberg: Can we give the user a username instead of relying on the fixture?

(This suggestion goes for this entire test file).

spec/controllers/admin/masquerades_controller_spec.rb
jayroh: Hey Teo! Question for you, sir - what's your rationale for having this particular type of test in a request spec instead of controller? Because there are redirects occurring and that correlating directly to a REQUEST's 301 behavior?

I've fallen into the familiar pattern of using request specs to test things like api requests.

spec/features/admin_masquerades_spec.rb
jakecraige: What's your thinking behind doing that? I view this as the job of fixtures to correctly define values as we'd expect them to be in the app, so I'm comfortable expecting this to me a unique username for the admin user.

app/controllers/application_controller.rb
jakecraige: Happy to do this and other suggestions about naming as I agree with them, but they weren't part of this PR so I try to keep changes to a minimum and not clean up unrelated things to the main changes if possible.

app/controllers/application_controller.rb
jakecraige: This line has been here [for 3 years](https://github.com/houndci/hound/commit/8573cbf1d86c0ee5adac8f4c7d776213e250edb2#diff-55c5b7aecfb519d0e4880eaf2788eb6eR29)

app/controllers/application_controller.rb
teoljungberg: I don't think counting diff stats that way is very productive. And I do think we should fix up code to be inline with the "best practices" as we go along. This falls into that area.

I find a diff like this to be much easier to reason about:

```ruby
def current_user
  @_current_user ||= find_user_or_masquerade
end

def find_user_ormasquerade
  # ...
end
```

Another thing worth noting, vim's built-in indentation will not indent this `begin` block as it is in this PR.

app/controllers/application_controller.rb
teoljungberg: I know, I was wondering why as it's not somthing I've seen before.

spec/controllers/admin/masquerades_controller_spec.rb
teoljungberg: Good question, I normally never write controller specs as they don't provide much value in my mind. And, they are in effect deprecated by Rails itself.

Request specs will actually render the view, whilst controller specs won't. I also like the explicitness of having `follow_redirect!` in there.

spec/features/admin_masquerades_spec.rb
teoljungberg: I view a feature spec as the flow a user takes, so I like to explictly fill in values as the user does. And a user fills in `my-username` instead of `user.username`.

gylaz:
This got me thinking about use cases this PR is helpful for...

1. User reports they can't see a certain repo when they load their repos. We can masquerade and verify that what they are reporting is true.
2. User can't activate a private repo -- we can't really do much here since we'd need their credit card information.

So, one use case that I could find useful. Is it worth this code change?

app/controllers/application_controller.rb
```diff
@@ -40,13 +40,23 @@ def signed_in?
   end
 
   def current_user
-    @current_user ||= User.where(remember_token: session[:remember_token]).first
+    @current_user ||= begin
+      if masquerading?
+        User.find_by(id: session[:masqueraded_user_id])
```
gylaz: `User.find(session[:masqueraded_user_id])`?

app/controllers/application_controller.rb
gylaz: It's back from when `protected` was cool, and everyone was doing it.

spec/features/admin_masquerades_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    sign_in_as(admin, "admin")
+    visit admin_masquerade_path(user.username)
+
+    expect(current_path).to eq(repos_path)
+    within "header .account" do
+      expect(page).to have_content(user.username)
```
gylaz: `have_content` -> `have_text`

spec/features/admin_masquerades_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    click_on "Stop Masquerading"
+
+    within "header .account" do
+      expect(page).to have_content(admin.username)
```
gylaz: `have_content` -> `have_text`

app/controllers/admin/application_controller.rb
```diff
@@ -2,8 +2,16 @@ module Admin
...
 
+    helper_method :sidebar_resources
+
     private
 
+    def sidebar_resources
+      Administrate::Namespace.new(namespace).resources.select do |resource|
+        DashboardManifest::DASHBOARDS.include?(resource)
+      end
+    end
```
gylaz: I don't understand this method. What is the sidebar and where does it display? What relevance does dashboards including a given resource has?

spec/features/admin_masquerades_spec.rb
jakecraige: Is this not in that format? `admin` as role, `masquerades` as action? What would you recommend?

app/controllers/application_controller.rb
jakecraige: I think `find_by` might be preferable here since we probably don't want the app to blow up with a not found error if the ID is somehow wrong. 

Though I don't see a case where that should happen so maybe we should since it being invalid must be some sort of bug?

app/controllers/admin/application_controller.rb
jakecraige: I describe this a bit in [the commit](https://github.com/houndci/hound/pull/1322/commits/b520233b7e4f8feb775300abc6a9adc3feb7ce91), but basically this override adminstrate's default behavior of assuming everything in `/admin` is a resource and makes it a whitelist of only things we specify as part of our dashboard. 

Without this it attempts to load `Masquerade` as a model and add a tab to the sidebar, which we don't want, because of the addition of the new route

jakecraige:
Tests all green, feedback implemented and squished. I have no context on the need of this features so I can't really comment on https://github.com/houndci/hound/pull/1322#issuecomment-285500252. I think @jayroh had some interest in it?

jayroh:
Oh hell yeah this is GREAT

PR #1324, opened 2017-03-21
Fix Markdown rendering

tysongach:
GitHub recently updated their Markdown parser which appears to have
broken some of the formatting we had. For example, we use ordered list
markers that always begin with "1" and when rendered, the ordered
numbers are automatically created. Due to the parsing changes, some
lists had indentation errors that were causing the order of the markers
to not be linear.

See: https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown

- - -

screen shot 2017-03-20 at 20 21 50


teoljungberg:
Solid!

PR #1326, opened 2017-03-23
Store unique repo ids in Stripe's metadata

gylaz:
We've got some dups in there, this should clean it up.


teoljungberg:
Is there a test we can add to make sure this doesn't turn into a regression in the future?

gylaz:
Modified an existing test that would fail without this change in place.

PR #1327, opened 2017-03-23
Fix Webpack config

gylaz:
Tests on CI were failing due to Webpack config being broken somehow.
Did some JS library get updated somehow?

Anyway, updating webpack and the output path seems to make it happy again.


teoljungberg:
I'd do a sanity deploy to staging to make sure this is running smoothly, as webpack has been finicky for us in the past.

But if all looks good, I see no issues with this besides that the tests are still failing.

gylaz:
Staging looks good.

PR #1331, opened 2017-03-26
Store linters and their default config in a hash

gylaz:
This makes it straight forward to see which linter is enabled by
default.
This gets away from the legacy "beta" concept and makes it easier to
enable/disable future linters.

### DIFF
--- a/app/models/hound_config.rb
+++ b/app/models/hound_config.rb
@@ -1,31 +1,21 @@
 # frozen_string_literal: true
 class HoundConfig
-  LINTERS = [
-    Linter::CoffeeScript,
-    Linter::Credo,
-    Linter::Eslint,
-    Linter::Flog,
-    Linter::Go,
-    Linter::Haml,
-    Linter::Jshint,
-    Linter::Remark,
-    Linter::Python,
-    Linter::Reek,
-    Linter::Ruby,
-    Linter::Scss,
-    Linter::Swift,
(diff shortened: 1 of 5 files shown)


app/models/hound_config.rb
```diff
@@ -48,11 +38,16 @@ def fail_on_violations?
...
   def default_config
-    LINTER_NAMES.each.with_object({}) do |name, config|
-      config[name] = { "enabled" => !BETA_LINTERS.include?(name) }
+    LINTERS.each.with_object({}) do |(linter_class, config), result|
+      name = linter_class.name.demodulize.underscore
+      result[name] = { "enabled" => config[:default] }
     end
   end
 
+  def enabled_by_default?(linter_class)
```
teoljungberg: This doesn't seem to be used anywhere, as far as I can see.

app/models/hound_config.rb
```diff
@@ -1,31 +1,21 @@
...
-    credo
-    eslint
-    flog
-    reek
-    remark
-    python
-    tslint
-  ).freeze
+  LINTERS = {
+    Linter::CoffeeScript => { default: true },
```
teoljungberg: I really like this approach!

app/models/hound_config.rb
gylaz: Good call.

PR #1333, opened 2017-03-31
Fix duplicate Owner#name issue

jayroh:
An issue came to light for one of our users where his repos were not
being refreshed. After digging in a bit I found that the
`RepoSynchronization` service was calling `Owner.upsert` across all the
user's repositories and hitting a `PG::UniqueViolation` exception. The
following example illustrates it:

```ruby
user = User.find(####)
repo_synch = RepoSynchronization.new(user)
repo_synch.start

PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_owners_on_name"
DETAIL:  Key (name)=(organization-name) already exists.
: INSERT INTO "owners" ("created_at", "updated_at", "github_id", "name", "organization") VALUES ($1, $2, $3, $4, $5) RETURNING "id"
```

Since both `github_id` and `name` have unique constraints on those columns,
we should guard against trying to insert new records with existing names
as well.

This commit will check for existing `github_id`s first, then the `name`,
defaulting to a brand new record.

### DIFF
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -6,7 +6,7 @@ class Owner < ApplicationRecord
     owner.update!(
       github_id: github_id,
       name: name,
-      organization: organization
+      organization: organization,
     )
     owner
   end
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -2,10 +2,12 @@ class Owner < ApplicationRecord
   has_many :repos
 
   def self.upsert(github_id:, name:, organization:)
-    owner = find_or_initialize_by(github_id: github_id)
-    owner.name = name
-    owner.organization = organization
-    owner.save!
+    owner = find_by(github_id: github_id) || find_by(name: name) || Owner.new
+    owner.update!(
+      github_id: github_id,
+      name: name,
+      organization: organization
+    )
     owner
   end
 
--- a/spec/models/owner_spec.rb
+++ b/spec/models/owner_spec.rb
@@ -13,7 +13,7 @@ describe Owner do
         new_owner = Owner.upsert(
           github_id: github_id,
           name: name,
-          organization: organization
+          organization: organization,
         )
 
         expect(new_owner).to be_persisted
@@ -28,7 +28,7 @@ describe Owner do
         updated_owner = Owner.upsert(
           github_id: owner.github_id,
           name: new_name,
-          organization: true
+          organization: true,
         )
 
         expect(updated_owner.name).to eq new_name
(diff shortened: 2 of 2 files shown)


thorncp:
I'm wondering if we need a bit more fanfare around this.

A scenario:

1. User signs up with Hound with `github_id=1` and `name=one`
1. They use Hound a bit, amassing `Repo`, `Build`, etc records
1. They delete their account on GitHub
1. User signs up with Hound with `github_id=2` and `name=one`
1. We update the original owner record, based on name, to be `github_id=2` and `name=one`
1. All the old `Repo`, `Build`, etc records are now attached to the new repo

What's the best course of action from there? Should we care about those records? The new user won't have access to the old repos on GitHub (they don't exist), so Hound will never have `Membership` records for them, but the data will be an invalid state.

jayroh:
I was thinking something similar, Chris.

What do you think about possibly switching the unique constraints from:

* `owners name` unique
* `owners github_id` unique

To

* `owners [name, github_id]` unique

?

thorncp:
That could make sense. I think we'd still want to make sure we don't duplicate `github_id` records, though, so maybe keep the unique constraint on it?

We could detect when this happens, then delete the old owner and associated records. Would that be a complete solution?

gylaz:
I worry about deleting owners and records. What if our assumptions here are wrong?

thorncp:
> I worry about deleting owners and records. What if our assumptions here are wrong?

Yeah, valid concern. Is there any way to verify the assumption? It seems hard to impossible.

jayroh:
What are we talking about in terms of size of the production database?
Could we test against that?

I was thinking that a good first stop gap might be to rescue that exception
and report it to slack - just to get a handle on how often it happens?

gylaz:
@jayroh That might be a good idea. My guess is (and from what I've seen) this is an edge case. If we could identify/surface the problem somehow and fix it by hand (i.e console) then that might be best.

jayroh:
👍

app/models/owner.rb
```diff
@@ -2,10 +2,12 @@ class Owner < ApplicationRecord
   has_many :repos
 
   def self.upsert(github_id:, name:, organization:)
-    owner = find_or_initialize_by(github_id: github_id)
-    owner.name = name
-    owner.organization = organization
-    owner.save!
+    owner = find_by(github_id: github_id) || find_by(name: name) || Owner.new
```
teoljungberg: I think we're missing a test-case when the last branch here gets triggered.

jayroh:
Closing this in favor of #1336

PR #1334, opened 2017-03-31
Add support for stylelint

blackjid:
We worked with @gmq to add support for https://stylelint.io/

It works quite similar to tslint and eslint.
We've rolled this in our infrastructure and it's looking good

this should close https://github.com/houndci/hound/issues/1037

### DIFF
--- a/spec/models/linter/stylelint_spec.rb
+++ b/spec/models/linter/stylelint_spec.rb
@@ -4,7 +4,7 @@ require "rails_helper"
 describe Linter::Stylelint do
   it_behaves_like "a linter" do
     let(:lintable_files) { %w(foo.scss) }
-    let(:not_lintable_files) { %w(foo.xxx) }
+    let(:not_lintable_files) { %w(foo.sass) }
   end
 
   describe "#file_review" do
@@ -41,7 +41,7 @@ describe Linter::Stylelint do
   end
 
   def stub_scss_config(config = {})
-    stubbed_scss_config = double(
+    stubbed_scss_config = instance_double(
       "StylelintConfig",
       content: config,
       serialize: config.to_s,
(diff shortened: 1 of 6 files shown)


spec/models/linter/stylelint_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe Linter::Stylelint do
+  it_behaves_like "a linter" do
+    let(:lintable_files) { %w(foo.scss) }
+    let(:not_lintable_files) { %w(foo.xxx) }
```
gylaz: Does it lint `sass` files? If not, maybe use that as the `not_lintable_files` extension?

spec/models/linter/stylelint_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        pull_request_number: build.pull_request_number,
+        patch: commit_file.patch,
+        content: commit_file.content,
+        config: "{}",
+      )
+    end
+  end
+
+  def stub_scss_config(config = {})
+    stubbed_scss_config = double(
```
gylaz: `double` to `instance_double`?

spec/models/linter/stylelint_spec.rb
blackjid: It doen't. Maybe with some extra work it could.

spec/models/linter/stylelint_spec.rb
blackjid: sounds like a `instance_double` is a best option here

gylaz:
Thanks again @blackjid!

blackjid:
Awesome!
@gmq great job!

PR #1335, opened 2017-04-02
Sanitize Flake8 config

gylaz:
The configuration that Flake8 accepts from the config file can have
certain options on multiple lines, like:
```
[flake8]
ignore =
  # some comment
  E221,
  # another comment
  E222
```

The library we use to parse the ini file isn't able to handle that case
and errors out.

This solution puts the multi-line options on one line,
as well as removes all the comments from the config file.

### DIFF
--- a/app/services/sanitize_ini_file.rb
+++ b/app/services/sanitize_ini_file.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+class SanitizeIniFile
+  static_facade :call
+
+  def initialize(config)
+    @config = config
+  end
+
+  def call
+    lines_without_comments.
+      map { |line| normalize_line(line) }.
+      join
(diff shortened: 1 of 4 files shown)


app/services/sanitize_ini_file.rb
```diff
@@ -0,0 +1,35 @@
...
+    lines_without_comments.
+      map { |line| normalize_line(line) }.
+      join
+  end
+
+  private
+
+  def normalize_line(line)
+    case line.rstrip
+    when /.+=\s*$/
```
thorncp: With `line.rstrip` as the `case` argument, will the `\s*` catch anything? Can it be removed?

gylaz: 👍

PR #1336, opened 2017-04-04
Rescue & handle RecordNotUnique error in repo sync

jayroh:
The `RepoSynchronization` service object has recently run into an issue
where a user has had a repo owner with a different github_id, but a name
for the Owner/Org that is the same as a previous Owner record. This has
caused an exception during the synchronization process that was not
recovered.

This commit will rescue that exception in the RepoSynchronization service
object, and will continue the `each` loop if it encounters it.

Down in the Owner model class, however, the upsert that creates this
exception will send the error to Raven along with some additional info
on the offending repo - the github_id and the name.

### DIFF
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -7,6 +7,17 @@ class Owner < ApplicationRecord
     owner.organization = organization
     owner.save!
     owner
+
+  rescue ActiveRecord::RecordNotUnique => exception
+    Raven.capture_exception(
+      exception,
+      extra: {
+        github_id: github_id,
+        name: name,
+      },
+    )
+
+    raise exception
   end
 
   def has_config_repo?
--- a/app/services/repo_synchronization.rb
+++ b/app/services/repo_synchronization.rb
@@ -8,17 +8,25 @@ class RepoSynchronization
 
     Repo.transaction do
       repos.each do |resource|
-        attributes = repo_attributes(resource.to_hash)
-        repo = Repo.find_or_create_with(attributes)
-        user.memberships.create!(
-          repo: repo,
-          admin: resource.to_hash[:permissions][:admin],
-        )
+        begin
+          create_user_membership_from!(resource)
+        rescue ActiveRecord::RecordNotUnique
+          next
+        end
       end
     end
   end
 
   private
+
+  def create_user_membership_from!(resource)
+    attributes = repo_attributes(resource.to_hash)
+    repo = Repo.find_or_create_with(attributes)
+    user.memberships.create!(
+      repo: repo,
+      admin: resource.to_hash[:permissions][:admin],
+    )
+  end
 
   def api
     @api ||= GithubApi.new(user.token)
--- a/spec/models/owner_spec.rb
+++ b/spec/models/owner_spec.rb
@@ -4,6 +4,31 @@ describe Owner do
   it { should have_many(:repos) }
 
   describe ".upsert" do
+    context "when name exists" do
+      it "captures exception, notifies raven, and raises to caller" do
+        existing_name = "ralphbot"
+        new_id = 567
+        create(:owner, github_id: 1234, name: existing_name)
+        allow(Raven).to receive(:capture_exception)
+
+        expect do
+          Owner.upsert(
+            github_id: new_id,
+            name: existing_name,
+            organization: true,
+          )
+        end.to raise_exception(ActiveRecord::RecordNotUnique)
+
+        expect(Raven).to have_received(:capture_exception).with(
+          instance_of(ActiveRecord::RecordNotUnique),
+          extra: {
+            github_id: new_id,
+            name: existing_name,
+          },
+        )
+      end
+    end
+
     context "when owner does not exist" do
       it "creates owner" do
         github_id = 1234
--- a/spec/services/repo_synchronization_spec.rb
+++ b/spec/services/repo_synchronization_spec.rb
@@ -22,6 +22,20 @@ describe RepoSynchronization do
       expect(user.reload.repos.first).to be_in_organization
     end
 
+    context "when there is an 'ActiveRecord::RecordNotUnique' exception" do
+      it "finishes the synchronization" do
+        repo = stub_github_api_repos(repo_id: 456, repo_name: "user/newrepo")
+        user = create(:user)
+        synchronization = RepoSynchronization.new(user)
+        allow(synchronization).to receive(:repo_attributes).
+          and_raise(ActiveRecord::RecordNotUnique)
+
+        result = synchronization.start
+
+        expect(result).to eq [repo]
+      end
+    end
+
     context "when the user is a repo admin" do
       it "the memberships admin flag is true" do
         stub_github_api_repos(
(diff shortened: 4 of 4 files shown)


spec/models/owner_spec.rb
```diff
@@ -4,6 +4,25 @@
...
+
+        expect do
+          Owner.upsert(
+            github_id: new_id,
+            name: existing_name,
+            organization: true,
+          )
+        end.to raise_exception(ActiveRecord::RecordNotUnique)
+
+        expect(Raven).to have_received(:capture_exception)
```
jayroh: Would have tested that it received the method with the appropriate message parameter but the message is a complicated mess of the quer and newlines and didn't want to waste time noodling around with formatting.

teoljungberg: You could write something like:

```ruby
expect(Raven).to have_recevied(:capture_exception).with(
  an_instance_of(ActiveRecord::RecordNotUnique),
  extra: {
    # ...
  },
)
```

spec/services/repo_synchronization_spec.rb
```diff
@@ -22,6 +22,19 @@
       expect(user.reload.repos.first).to be_in_organization
     end
 
+    context "when there is an 'ActiveRecord::RecordNotUnique' exception" do
+      it "finishes the synchronization" do
+        stub_github_api_repos(repo_id: 456, repo_name: "user/newrepo")
+        user = create(:user)
+
```
teoljungberg: I'd remove this newline since the stubbing is a part of the setup phase of the test.

spec/services/repo_synchronization_spec.rb
```diff
@@ -22,6 +22,19 @@
...
+        synchronization = RepoSynchronization.new(user)
+        allow(synchronization).to receive(:repo_attributes).
+          and_raise(ActiveRecord::RecordNotUnique)
+
+        expect { synchronization.start }.not_to raise_error
```
teoljungberg: This is my least favorite assertion ever. Can we test that the synchronization was finished without stopping?

spec/services/repo_synchronization_spec.rb
```diff
@@ -22,6 +22,19 @@
       expect(user.reload.repos.first).to be_in_organization
     end
 
+    context "when there is an 'ActiveRecord::RecordNotUnique' exception" do
+      it "finishes the synchronization" do
+        stub_github_api_repos(repo_id: 456, repo_name: "user/newrepo")
+        user = create(:user)
+
+        synchronization = RepoSynchronization.new(user)
+        allow(synchronization).to receive(:repo_attributes).
```
teoljungberg: We're lacking an assertion that this method was actually called, which sometimes isn't what you want to do. But, since it's important that it is called and raises for this test - I would add that assertion

spec/services/repo_synchronization_spec.rb
jayroh: Ever?

Give me time. I'll top it soon enough ;)

app/services/repo_synchronization.rb
```diff
@@ -8,12 +8,16 @@ def start
...
-        )
+        begin
+          attributes = repo_attributes(resource.to_hash)
+          repo = Repo.find_or_create_with(attributes)
+          user.memberships.create!(
+            repo: repo,
+            admin: resource.to_hash[:permissions][:admin],
+          )
+        rescue ActiveRecord::RecordNotUnique
+          next
```
gylaz: Should we also be capturing this situation with Raven?

spec/models/owner_spec.rb
gylaz: Could you do something like:
```rb
expect(Raven).to have_received(:capture_exception) do |args|
  expect(args[0]).to be_an_isntance_of ActiveRecord::RecordNotUnique
  expect(args[1]).to eq(extra: { ... })
end
```
Not sure what `args` will be, but something like this might work.

app/services/repo_synchronization.rb
```diff
@@ -8,12 +8,16 @@ def start
...
-          repo: repo,
-          admin: resource.to_hash[:permissions][:admin],
-        )
+        begin
+          attributes = repo_attributes(resource.to_hash)
+          repo = Repo.find_or_create_with(attributes)
+          user.memberships.create!(
+            repo: repo,
+            admin: resource.to_hash[:permissions][:admin],
+          )
```
gylaz: Thoughts on pulling these lines into a private method to simplify the outer method?

app/services/repo_synchronization.rb
jayroh: Since we capture the exception one level down (in the model) and then bubble up the exception it would be doing the work of reporting it twice. Order of operations is:

1. Model - captures, reports, and raises to caller.
2. Repo synch - rescues same exception and continues the loop.

app/services/repo_synchronization.rb
jayroh: Good idea, Greg. I didn't pull just the membership creation, but the entire setup here. attributes, repo, user membership, etc - into a private method.

spec/models/owner_spec.rb
jayroh: Good call, guys. Very helpful - thank you!

app/services/repo_synchronization.rb
gylaz: Which model? I see that the `User` model rescues `ActiveRecord::RecordInvalid`, and the `Owner` model doesn't come into play here, does it?

jayroh: The `Owner` model.

app/models/owner.rb
```diff
@@ -7,6 +7,17 @@ def self.upsert(github_id:, name:, organization:)
...
+  rescue ActiveRecord::RecordNotUnique => exception
+    Raven.capture_exception(
+      exception,
+      extra: {
+        github_id: github_id,
+        name: name,
+      },
+    )
+
+    raise exception
```
jayroh: @gylaz here's where it runs into the PG unique constraint and goes 💥

app/services/repo_synchronization.rb
```diff
@@ -8,18 +8,26 @@ def start
...
       repos.each do |resource|
-        attributes = repo_attributes(resource.to_hash)
-        repo = Repo.find_or_create_with(attributes)
-        user.memberships.create!(
-          repo: repo,
-          admin: resource.to_hash[:permissions][:admin],
-        )
+        begin
+          create_user_membership_from!(resource)
+        rescue ActiveRecord::RecordNotUnique
```
gylaz: Previous thread got collapsed, but I fee like I'm missing something `create_user_membership_from!` doesn't call `Owner.upsert`, so how does this capture the exception `Owner.upsert` raises?

jayroh: It's a little buried in the private methods, but it eventually does get called, here:

https://github.com/houndci/hound/blob/60294248e0d4d6657489bf3b28893e22049d0fae/app/services/repo_synchronization.rb#L36

gylaz: Ah, I see. Yeah, a bit surprising that `repo_attributes` could create an `Owner`, didn't expect that reading that block. Anyways, that's for working through this with me and helping me see the 🚥

jayroh: Anytime, brother. 👊

jayroh:
@gylaz @teoljungberg squashed/rebased.

👍,👎❓

gylaz:
🚢

jayroh:
:shipit:  🎉

PR #1338, opened 2017-04-07
Rename Python linter classes to Flake8

gylaz:
This is the first step in renaming our classes after the linters rather
than languages.
This will make it more reasonable to add additional linters for same
language.


gylaz:
This matches the naming https://github.com/houndci/linters/pull/101 expects.

gylaz:
Hmm, I wonder if something else is going on. [This line](https://github.com/houndci/hound/blob/master/app/services/resolve_config_aliases.rb#L6) maps legacy `python` key to the new `flake8` key. And we do have a [test](https://github.com/houndci/hound/blob/master/spec/models/hound_config_spec.rb#L114) to make sure `flake8` works as a key.

gedankenstuecke:
Ah, I think the problem was on my side in terms of understanding of how Hound runs with `flake8`. 

I've used Hound with Ruby projects in the past and I think rubocop checks every single line in files that have changed, even if the line itself was untouched? 

Coudl it be that for `flake8` only those lines are checked that have actually been changed?

gylaz:
In general, Hound *only* comments/reports on lines that changed in a given PR, independent of which linter. Let us know if you see a particular unexpected behavior (linking to the PR with the issue helps a lot).

PR #1342, opened 2017-04-21
List all supported languages in README

jaysonvirissimo:
Add Go, HAML, and Swift


gylaz:
Thanks @jaysonvirissimo!

PR #1347, opened 2017-05-11
Report linter errors as the main PR review comment

gylaz:
When a linter encounters an unexpected error, like invalid config
syntax, we will receive it in the `error` attribute in the payload.
Then we'll report it to the user via the main comment that goes into the
PR review body.

### DIFF
--- a/app/models/pull_request.rb
+++ b/app/models/pull_request.rb
@@ -13,9 +13,14 @@ class PullRequest
     @commit_files ||= modified_commit_files
   end
 
-  def comment_on_violations(violations)
+  def make_comments(violations, errors)
     comments = violations.map { |violation| build_comment(violation) }
-    hound_github.create_pull_request_review(repo_name, number, comments)
+    hound_github.create_pull_request_review(
+      repo_name,
+      number,
+      comments,
+      format_errors(errors),
+    )
   end
 
   def repository_owner_name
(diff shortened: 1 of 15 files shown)


salbertson:
@gylaz do you have a screenshot, or a test PR, to get a better idea what this would look like?

app/models/pull_request.rb
```diff
@@ -62,6 +67,21 @@ def build_comment(violation)
...
+  def format_errors(errors)
+    if errors.any?
+      error_details = errors.map { |error| build_error_details(error) }
+      ["Some files could not be reviewed due to errors:"].
+        concat(error_details).join
+    end
+  end
+
+  def build_error_details(error)
+    lines = error.split("\n")
```
thorncp: `String#lines`?

spec/models/build_spec.rb
```diff
@@ -37,4 +37,21 @@
       end
     end
   end
+
+  describe "#review_errors" do
+    it "returns and list of unique linter errors" do
```
thorncp: and -> a?

gylaz:
@salbertson @thorncp from this PR: https://github.com/gylaz/wormbots/pull/13#pullrequestreview-37914344
screenshot 2017-05-12 11 42 03

It will surface nearly identical errors, because it will be an error per file (and their body differs in the temp directory path). We could probably add some logic to do a 90% match on the error body.

PR #1349, opened 2017-05-12
Support Sass-lint

whyderrick:
This PR adds Hound Support for Sass-lint using the single `linters` queue.

It includes an update to `/configurations` letting people know how to enable and configure Sass-lint.

### DIFF
--- a/spec/models/config/sass_lint_spec.rb
+++ b/spec/models/config/sass_lint_spec.rb
@@ -1,0 +1,82 @@
+require "app/models/config/base"
+require "app/models/config/sass_lint"
+require "app/models/config/parser"
+require "app/models/config/parser_error"
+require "app/models/config/serializer"
+require "app/models/config_content"
+require "app/models/missing_owner"
+
+describe Config::SassLint do
+  describe "#content" do
+    context "when an owner is provided" do
+      it "merges the configuration into the owner's configuration" do
+        raw_config = <<~EOS
+          rules:
+            indentation:
+              - 2
+              -
+                size: 2
+        EOS
+        owner_config = {
+          "rules" => {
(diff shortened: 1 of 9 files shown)


spec/models/config/sass_lint_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+              foo: bar
+      EOS
+      config = build_config(raw_config)
+
+      expect(config.serialize).to eq <<~EOS
+        ---
+        rules:
+          indentation:
+          - 2
+          - size: 2
```
whyderrick: The indentations look different here, but I tested in IRB and `YAML.load` called on either of these strings returns the same hash: `{"rules"=>{"indentation"=>[2, {"size"=>2, "foo"=>"bar"}]}}`

spec/models/config/sass_lint_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+require "app/models/config/parser"
+require "app/models/config/parser_error"
+require "app/models/config/serializer"
+require "app/models/config_content"
+require "app/models/missing_owner"
+
+describe Config::SassLint do
+  describe "#content" do
+    context "when an owner is provided" do
+      it "merges the configureation into the owner's configuration" do
```
thorncp: configureation -> configuration

spec/models/config/sass_lint_spec.rb
thorncp: YAML is weird

spec/models/config/sass_lint_spec.rb
whyderrick: 👍

PR #1350, opened 2017-05-17
Allow bulk subscribers to enable repos

gylaz:
We accidentally removed this feature as part of a previous commit that
did some load times optimizations.

### DIFF
--- a/app/serializers/repo_serializer.rb
+++ b/app/serializers/repo_serializer.rb
@@ -13,10 +13,10 @@ class RepoSerializer < ActiveModel::Serializer
   )
 
   def price_in_cents
-    if object.private?
-      scope.tier_price * 100
-    else
+    if object.public? || object.bulk?
       0
+    else
+      scope.tier_price * 100
     end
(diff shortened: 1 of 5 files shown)


app/serializers/repo_serializer.rb
```diff
@@ -13,10 +13,10 @@ class RepoSerializer < ActiveModel::Serializer
   )
 
   def price_in_cents
-    if object.private?
-      scope.tier_price * 100
-    else
+    if object.public? || object.bulk?
```
thorncp: This will re-introduce an N+1 -- it will query `BulkCustomer` for each repo in the collection. It's probably not worth trying to deal with it in this PR, since it's a bug fix, but it'd be nice to do that soon.

gylaz: Good point. I'll make a card on Trello for this.

gylaz: https://trello.com/c/i4m7XIVB

PR #1354, opened 2017-05-29
Limit the size of GitHub review body comment

gylaz:
GitHub limits the review body (main comment) to be 65,536 characters or
less. Otherwise, an error gets thrown.
This change will truncate our errors to fit within that limit, while
still tryign to show the error.

### DIFF
--- a/app/models/review_body.rb
+++ b/app/models/review_body.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+class ReviewBody
+  MAX_BODY_LENGTH = 65536
+  SUMMARY_LENGTH = 80
+  LINE_DELIMITER = "<br>"
+  HEADER = "Some files could not be reviewed due to errors:"
+  DETAILS_FORMAT = "<details><summary>%s</summary><pre>%s</pre></details>"
+
+  def initialize(errors)
+    @errors = errors
+  end
+
+  def to_s
+    if errors.any?
+      error_details = errors.map { |error| build_error_details(error) }
+      [HEADER].concat(error_details).join
+    else
+      ""
+    end
+  end
+
+  private
+
+  attr_reader :errors
+
+  def build_error_details(error)
+    summary = error_summary(error)
+    details = error[0...(allowed_error_length - summary.length)].
+      lines.
+      map(&:rstrip).
+      join(LINE_DELIMITER)
+
+    sprintf(DETAILS_FORMAT, summary, details)
+  end
+
+  def error_summary(error)
+    error.lines.first.strip.truncate(SUMMARY_LENGTH)
+  end
+
+  def allowed_error_length
+    (MAX_BODY_LENGTH - formatting_characters_size) / errors.size
+  end
+
+  def formatting_characters_size
+    HEADER.length + all_detail_formats_length + all_lines_delimiters_length
+  end
+
+  def all_detail_formats_length
+    DETAILS_FORMAT.size * errors.size
+  end
+
+  def all_lines_delimiters_length
+    (errors.flat_map(&:lines).size - errors.size) * LINE_DELIMITER.length
+  end
+end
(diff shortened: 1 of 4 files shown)


app/models/review_body.rb
```diff
@@ -0,0 +1,50 @@
...
+      join(COMMENT_LINE_DELIMITER)
+
+    sprintf(details_format, summary, details)
+  end
+
+  def per_error_length
+    (MAX_BODY_LENGTH - extra_characters_size) / errors.size
+  end
+
+  def extra_characters_size
```
teoljungberg: From reading this method, I don't understand what it does.

I think it may be worth extracting methods of the calculations performed in it.

app/models/review_body.rb
```diff
@@ -0,0 +1,50 @@
...
+    header.size +
+      (details_format.size * errors.size) +
+      (errors.flat_map(&:lines).size * COMMENT_LINE_DELIMITER.size)
+  end
+
+  def header
```
teoljungberg: Thought's on extracting this into a constant?

app/models/review_body.rb
```diff
@@ -0,0 +1,50 @@
...
+    "Some files could not be reviewed due to errors:"
+  end
+
+  def details_format
```
teoljungberg: Same suggestion as above.

spec/models/pull_request_spec.rb
```diff
@@ -60,57 +60,31 @@
...
+        payload.pull_request_number,
+        [
+          {
+            path: violation.filename,
+            position: violation.patch_position,
+            body: violation.messages.join,
+          },
+        ],
+        "Some files could not be reviewed due to errors:" \
+        "<details><summary>invalid config</summary>" \
```
teoljungberg: Maybe my eyes are tired, but shouldn't these "joined" lines be indented a level more?

I think this is a case of vim's indentation being iffy.


```ruby
"here's a string " \
  "that spans over multiple lines"
```

If that's right, you may need to extract a local variable to hold the string instead to correct the faulty indentation.

spec/models/review_body_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    context "with errors" do
+      it "returns formatted errors" do
+        review_body = ReviewBody.new(["invalid config", "foo\n  bar"])
+
+        result = review_body.to_s
+
+        expect(result).to eq(
+          "Some files could not be reviewed due to errors:" \
+          "<details><summary>invalid config</summary>" \
```
teoljungberg: Same comment regarding this indentation being faulty.

spec/models/review_body_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      context "when errors are too long" do
+        it "truncates the errors" do
+          stub_const("ReviewBody::MAX_BODY_LENGTH", 153)
+          review_body = ReviewBody.new(["invalid config", "foo\n  bar"])
+
+          result = review_body.to_s
+
+          expect(result).to eq(
+            "Some files could not be reviewed due to errors:" \
+            "<details><summary>invalid config</summary>" \
```
teoljungberg: Indentation, yet again.

app/models/review_body.rb
gylaz: 👍

app/models/review_body.rb
gylaz: 👍

spec/models/pull_request_spec.rb
gylaz: You're probably right about the indentation, thought I thought it looked nice aligned as-is -- it looks closer to a heredoc block. But I'm okay indenting the rest.

gylaz:
@teoljungberg ready for another look.

PR #1358, opened 2017-05-30
Remove unused CSS

tysongach:
The `ol` with a class of `getting-started` does not exist in the code base, nor does an `img` with a class of `logo`.

Most of this was added in https://github.com/houndci/hound/commit/6f2f450a065eda592824ab0fd09a76f309beadb1


tysongach:
I happened to come across it while doing work for a future PR to update Bourbon.

PR #1360, opened 2017-05-30
Update Bourbon

tysongach:
This updates Bourbon to 5.0 (currently in a stable beta). To do that we need to remove usage of any removed mixins, which were all prefixing mixins. We have Autoprefixer do prefixing for us now, so the mixins aren't really needed. Updating Autoprefixer to the latest also ensures we are getting the latest prefixes.

Administrate had a hard Bourbon dependency in many of its early releases. So to update to Bourbon 5, we need to update Administrate (which fully _removed_ the Bourbon dependency in 0.7.0).

We also have locked Neat into the 1.x cycle, because 2.0 is out, but it's a _major_ difference to 1.x and upgrading will take a fair amount of manual effort.


app/assets/stylesheets/pages/repos/_onboarding.scss
```diff
@@ -1,8 +1,8 @@
 .onboarding {
-  @include animation(pulse-in 0.4s ease-in-out);
-  @include animation-delay(0.2s);
-  @include animation-fill-mode(both);
-  @include background-image(linear-gradient(160deg, rgb(196, 115, 209), rgb(153, 115, 209)));
+  animation-delay: 0.2s;
+  animation-fill-mode: both;
+  animation: pulse-in 0.4s ease-in-out;
+  background-image: linear-gradient(160deg, rgb(196, 115, 209), rgb(153, 115, 209));
```
tysongach: This was existing, so I’m going to leave as-is for now and keep this PR focused on updating Bourbon.

tysongach:
It appears that this PR might need a bit of tweaking to Administrate with its update.

app/views/admin/application/_navigation.html.haml
```diff
@@ -0,0 +1,5 @@
+%nav.navigation{ role: "navigation" }
+  - navigation_resources.each do |resource|
+    = link_to display_resource_name(resource),
+      [namespace, resource.path],
```
gylaz: Is the `.path` required here now?

tysongach: Oh, I don’t know…

tysongach: @gylaz Looks like `.path` was added in https://github.com/thoughtbot/administrate/commit/e928379c6eae5c721134b7d0d11c713e6150cd40 so that it works with namespaced models. Should we leave it?

gylaz: I brought this up, because before we were only calling `resource` here. If that doesn't work and `path` is needed, then I'm 👍

tysongach: I’m honestly unsure…I noted in Slack, that I don’t see any navigation items listed locally. I assume this is because I don't have the prod. database locally? Wasn't quite sure how to test besides pushing to staging, which I think I'll do and see.

PR #1365, opened 2017-05-31
Refactor base typography Sass

tysongach:
Use global variables in a more expected fashion and group them together.


tysongach:
Thanks, @conchan!

PR #1366, opened 2017-05-31
Refactor button system

tysongach:
- Use classes instead of Sass mixins
- House all button styles in one partial

## Before

screen shot 2017-06-01 at 09 39 34

## After

screen shot 2017-06-01 at 09 38 16


app/assets/stylesheets/components/_button.scss
```diff
@@ -0,0 +1,59 @@
+.button {
```
tysongach: I don’t agree with this linter. It doesn’t offer concrete evidence as to how it makes the code worse beyond “you might not be using the cascade,” which I disagree with. We have this disabled in thoughtbot’s guides.

app/assets/stylesheets/components/_button.scss
```diff
@@ -0,0 +1,59 @@
...
+    color: $white;
+  }
+}
+
+.button--large {
+  font-size: modular-scale(1);
+  padding: modular-scale(-1) modular-scale(2);
+}
+
+.button__icon {
```
tysongach: I went BEM here because there are some BEM-esque double hyphens already in the codebase. Surprisingly, we don't have SCSS-Lint configured within this repo, so this feels like  a default linter config warning being thrown, rather than one we agree with.

app/assets/stylesheets/components/_button.scss
gylaz: Since `houndci` is its own org, we can't use thoughtbot's master config. We could update HoundCI's master config to disable this.

app/views/home/index.haml
```diff
@@ -14,7 +14,10 @@
           you and your team to better review and maintain a clean codebase.
 
         .home-hero-cta
-          = render "application/auth_button"
+          = link_to "/auth/github",
+                    class: "button button--primary button--large" do
```
gylaz: Should this nesting be?
```
= link_to "/auth/github",
  class: "button button--primary button--large" do
```

tysongach:
@conchan In this case, I see the use of mixins as a needless abstraction. We can accomplish the same goal with vanilla CSS (and less code). I noticed that we were calling the button mixins, but then tweaking them slightly each time (i.e. `@include` the button mixin, but tweak its padding in this case). I find that hard to follow and track down (the tweaks would happen across many partials) where styles are coming from and it is also a breakdown of a consistent button system.

conchan:
Great—always hesitant about cases like this, so thanks for sharing your reasoning!

tysongach:
Thanks for the reviews!

PR #1369, opened 2017-06-01
Update rubocop config names and options for new version

ruddzw:
After the May 31 update of the version of rubocop that Hound uses, I found that I needed to update these configurations.

This goes a few steps further than #1367, re-sorting the names alphabetically, and also removing/updating option names that caused warnings.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,33 +10,6 @@ AllCops:
   StyleGuideCopsOnly: false
 Rails:
   Enabled: true
-Style/AccessModifierIndentation:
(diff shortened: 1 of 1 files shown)


gylaz:
Nice! Does this mean #1367 can be closed in favor of this?

ruddzw:
As far as I'm concerned, yeah! I would have built mine on top of that, but I didn't notice the PR until I was filing mine.

gylaz:
Thanks @ruddzw!

ruddzw:
Thank you for the quick turnaround! 🙌

PR #1371, opened 2017-06-01
Redesign app navigation

tysongach:
- Remove sticky header; it caused shifting in the nav's height which can be distracting.
- Wrap the navigation in a `nav` element (rather than `header`) which is more semantically correct and maps to the `navigation` role in the accessibility tree.
- Component-ize CSS: `avatar`, `logo` and `icon` are now reusable components.
- Begin using SVG `symbol` for icon and SVG assets, instead of Font Awesome or using the SVG through an `img` tag. This allows for CSS styles (e.g. `fill` color) and _greatly_ improves accessibility.
- Remove Twitter link (it is still listed in the footer).
- Maintain Pricing and Sign In with GitHub links across all breakpoints instead of removing them on mobile.
- Vastly simplify CSS, including removal of over-qualified selectors (e.g. `ul li a.auth`).
- Remove the need for page-based classes via `content_for :header_class`.

## Before

screen shot 2017-06-01 at 17 07 03

screen shot 2017-06-01 at 17 07 32

screen shot 2017-06-01 at 17 07 27

screen shot 2017-06-01 at 17 06 23

## After

screen shot 2017-06-01 at 17 04 48
screen shot 2017-06-01 at 17 04 17
screen shot 2017-06-01 at 17 04 39
screen shot 2017-06-01 at 17 06 00


app/assets/stylesheets/components/_app-nav.scss
```diff
@@ -0,0 +1,60 @@
...
+  a {
+    color: $base-font-color;
+
+    &:hover {
+      color: $base-accent-color;
+    }
+  }
+}
+
+.app-nav__container {
```
tysongach: It’s BEM, though. I need to update you to understand.

PR #1373, opened 2017-06-02
Whitelist serialized symbols when parsing YAML

### LINKED ISSUE #1370: New Rubocop Cop Breaks Hound
With hound's latest upgrade of rubocop, a new cop was introduced called `Style/InverseMethods` that requires config that has serialized symbols, as part of this PR: https://github.com/bbatsov/rubocop/pull/4000

The default config in rubocop for this new cop looks like this (note the method names that are symbols, e.g. `:any?`, `:even?`, etc):

```yaml
Style/InverseMethods:
  Enabled: true
  # `InverseMethods` are methods that can be inverted by a not (`not` or `!`)
  # The relationship of inverse methods only needs to be defined in one direction.
  # Keys and values both need to be defined as symbols.
  InverseMethods:
    :any?: :none?
    :even?: :odd?
    :==: :!=
    :=~: :!~
    :<: :>=
    :>: :<=
    # `ActiveSupport` defines some common inverse methods. They are listed below,
    # and not enabled by default.
    #:present?: :blank?,
    #:include?: :exclude?
  # `InverseBlocks` are methods that are inverted by inverting the return
  # of the block that is passed to the method
  InverseBlocks:
    :select: :reject
    :select!: :reject!
```

Because of this, your call to `Yaml.safe_load` in `Parser` raises since you only whitelist serialization of `Regexp` and not `Symbol`.  Rubocop's own config loader also added `Symbol` to its own whitelisting in the same PR that introduced the cop: https://github.com/bbatsov/rubocop/pull/4000/files#diff-e2fc41e8fa0504d203570a8d822495afR171

The ramification of all of this is that if your rubocop config includes this cop, hound will fail to run on your pull request and display an error message: `Error parsing config for: ruby. Click "details" for assistance.`

gylaz:
RuboCop config can now contain serialized symbols in its YAML.

Closes #1370

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -345,6 +345,9 @@ Metrics/AbcSize:
     conditions.
   Enabled: true
   Max: 15
+Metrics/BlockLength:
+  Exclude:
+    - "spec/**/*"
 Metrics/BlockNesting:
   Description: Avoid excessive block nesting
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#three-is-the-number-thou-shalt-count
--- a/app/models/config/parser.rb
+++ b/app/models/config/parser.rb
@@ -1,7 +1,7 @@
 module Config
   class Parser
     def self.yaml(content)
-      YAML.safe_load(content, [Regexp])
+      YAML.safe_load(content, [Regexp, Symbol])
     end
 
     def self.json(content)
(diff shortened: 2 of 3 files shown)


gylaz:
Best reviewed with [`w=1`](https://github.com/houndci/hound/pull/1373/files?w=1) param.

PR #1379, opened 2017-06-02
Clean up Sass variables

tysongach:
- Remove unused variables
- Group typography variables
- Move partial-specific variables into their partial, rather than
  defining them alongside global variables
- Remove `$large-border-radius` which was only used once and only added
  `1px` to `$base-border-radius`

### DIFF
--- a/app/assets/stylesheets/base/_tables.scss
+++ b/app/assets/stylesheets/base/_tables.scss
@@ -1,3 +1,5 @@
+$_table-heading-background-color: rgb(245, 245, 245);
+
 table {
   border-collapse: collapse;
   width: 100%;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/base/_tables.scss
```diff
@@ -1,3 +1,5 @@
+$_table-heading-background-color: rgb(245, 245, 245);
```
gylaz: Why underscore here? Also, why put this outside of `_variables.scss`?

tysongach: This indicates a private variable. `_variables.scss` is for global variables, whereas partial-specific or "private" variables are defined in their own partial.

tysongach: The idea is that over time `_variables.scss` tends to get _massive_, but often times many of the variables in there are only used for one purpose, or they are specific to a single component. This helps to organize things a bit and maintain the length of `_variables.scss`. the thoughtbot.com `_variables.scss` is 216 lines long!

PR #1383, opened 2017-06-02
Add private badge to repos list

tysongach:
This is an alternative approach to https://github.com/houndci/hound/pull/1296.

**Note:** This is just the design implementation. Functionality for the following items is still needed:

- The `ariaDescribedby` attribute in the React component doesn’t yet render anything.
- The `idName ` attribute in the React component doesn’t yet render anything.
- The logic for only rendering this badge when a repo is indeed private is not implemented (i.e. “Private” is showing on every single repo right now).

screen shot 2017-06-02 at 16 38 04

### DIFF
--- a/app/assets/javascripts/components/repo.jsx
+++ b/app/assets/javascripts/components/repo.jsx
@@ -28,8 +28,7 @@ class Repo extends React.Component {
   render() {
     const { isProcessingId, repo } = this.props;
     const { active, id, name, price_in_cents } = repo;
-
-    const showPrivate = !active && price_in_cents > 0;
+    const showPrivate = price_in_cents > 0;
 
     return (
       <li
@@ -45,10 +44,13 @@ class Repo extends React.Component {
           {name}
         </div>
 
-        <div className={classNames(
-          "repo-activation-toggle",
-          {"repo-activation-toggle--private": showPrivate}
-        )}>
+        { showPrivate &&
+          <span className="badge margin-left-small">
+            Private
+          </span>
+        }
+
+        <div className="repo-activation-toggle">
           {this.renderButton()}
         </div>
       </li>
--- a/app/assets/stylesheets/components/_badge.scss
+++ b/app/assets/stylesheets/components/_badge.scss
@@ -1,0 +1,10 @@
+.badge {
+  border: 1px solid $light-gray;
+  border-radius: 3px;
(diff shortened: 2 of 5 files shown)


tysongach:
[Indicate private repos](https://trello.com/c/LR5FzEhd/818-indicate-private-repos)

gylaz:
Any concerns that positioning of the `private` element isn't consistent -- based on the length of the repo name? On advantage having private next to the `Activate` button is that it would have the same positioning, but not sure if that's better.

tysongach:
@gylaz That’s a great point. Sometimes inconsistent alignment like that can create a really uneven looking design, but in this case since the badge is not very bold and doesn't impose a lot of visual weight, I think it's fine. I more like that we associate the repo name and its private status via proximity (being so close to each other).

Here's a larger view of a bigger list:

screencapture-localhost-5000-repos-1496437313683

gylaz:
Good points! I sort of liked that `private` was close to the action button too, since it visually helps to confirm that you're activating a repo that you need to pay for. But I think either way is probably fine.

BTW, looks like snapshots need to be updated.

gylaz:
@tysongach should we merge this once tests are green?

tysongach:
@gylaz So far this PR is design-only. There’s not yet any logic to properly label private repos as private.

app/assets/javascripts/components/repo.jsx
```diff
@@ -41,10 +41,14 @@ class Repo extends React.Component {
...
           {name}
         </div>
 
+        <span className="badge margin-left-small" idName="repo-{name}">
+          Private
+        </span>
+
         <div className={classNames(
           "repo-activation-toggle",
           {"repo-activation-toggle--private": showPrivate}
```
gylaz: Do we still need this class? Maybe it can be removed/cleaned up now, since we're going the label route.

gylaz:
@tysongach I've made this functional. I also removed `ariaDescribedby` and `idName` as they don't seem to be supported by React. Should we merge?

tysongach:
@gylaz Great! Yeah, merge away! 👍 

> I also removed ariaDescribedby and idName as they don't seem to be supported by React.

I added these as hopefuls, but let’s go ahead and merge for now. In the future, we can circle back and make these work; it will greatly improve accessibility.

PR #1385, opened 2017-06-05
Update helper to use Resource objects

jackvnimble:
The Administrate update changed the return value from
`namespace.resources` to return an array of `Resource` objects instead
of resource names. This updates the helpers and ensures that if the
admin navigation gets clobbered in the future a test fails.


jackvnimble:
[Bug: Administrate nav links not showing up](https://trello.com/c/DoswFoMB/816-bug-administrate-nav-links-not-showing-up)

PR #1394, opened 2017-06-09
Remove Font Awesome

tysongach:
Some light research indicates that 53% of our CSS on production is from Font Awesome. I’d like to investigate removing it with the hope of improving performance, as well as accessibility, because font icons are very inaccessible.

Overall, I think it’s a good time to evaluate icons that are providing us value, and re-implement them where we see fit.

- The icons on the home page within the “Why does style matter?” section, as well as the two icons in the “Pricing” section are quite abstract. I think we can get by without them.
- There is a `spin` icon on the `repo--processing` class. I'm not sure how or where this is used. 
- There is a `exclamation-circle` icon used within a flash message when your email is updated. Again, I'm not sure how or where this is used. When I'm signed in, I see nowhere to let me update my email; it's tied to my GitHub.
- The `credit-card` is shown next to “Update Credit Card”. Probably fine without an icon.


thorncp:
The `spin` one is intended to be a progress indicator after you click the activate button for a repo. It seems to not be working anymore. I'm guessing it was broken during the move to React. Seems fine to remove it.

You can change the email address that's included on receipts on the account page: https://houndci.com/account. I think we can safely remove that icon:

image

gylaz:
@tysongach You can edit your email from the "Account" page, which can be accessed by clicking on your avatar. Maybe we could improve the discoverability of that page at some point.

tysongach:
> You can edit your email from the "Account" page, which can be accessed by clicking on your avatar. Maybe we could improve the discoverability of that page at some point.

I don’t see that…maybe because I’m on a special pricing plan on production, and locally as well?

tysongach:
This is ready for review again.

app/assets/javascripts/components/repo.jsx
```diff
@@ -36,8 +36,7 @@ class Repo extends React.Component {
         className={
           classNames(
             "repo",
-            {"repo--active": active},
-            {"repo--processing": isProcessingId === id}
```
gylaz: Was this not working? We probably do want to show something if the repo is being activated/deativated, like a spinner, but if this wasn't working right now, so I'm okay removing it in such a case.

tysongach: I couldn’t get it to work, myself, and @thorncp’s comments above lead me to believe he couldn’t either. We can add a card to tackle this feature in another way in the future.

PR #1397, opened 2017-06-12
Use PathSpec gem to handle ignore files

gylaz:
`File.fnmatch` differs from the method used in most ignorefiles, and
thus does not behave the same as most linters locally.
(PathSpec)[https://github.com/highb/pathspec-ruby] gem brings ignore
path matching to Ruby.


lib/js_ignore.rb
```diff
@@ -1,19 +1,21 @@
 class JsIgnore
   DEFAULT_EXCLUDED_PATHS = %w(
-    vendor/*
-    node_modules/*
+    vendor/**/*
```
phyrog: Might want to replace this with `bower_components/*` to be more consistent with the default eslint behavior (see http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories).

I believe they use `bower_components/*` so you can whitelist single packages. Is that also possible with `bower_components/**/*`?

gylaz: Rather than replacing, we can add `bower_components/**/*` here. I think some folks (in Rails apps) have JS in `vendor/**/*`.

phyrog: Yeah, I guess that works too. Might be surprising to people who actually want to lint their `vendor` directory (which is probably a rare edge case), however.

PR #1401, opened 2017-06-21
Upgrade dotenv gem

gylaz:
Fixes an issue where `.env.test` was not being picked up during local
tests.


gylaz:
Handled in #1402.

PR #1402, opened 2017-06-22
Upgrade Rails to 5.1.1

gylaz:


gylaz:
Bumped `rspec-rails`.

PR #1403, opened 2017-06-23
Change primary typeface

tysongach:
I find Lato—the current typeface used across the app—to be a tad too friendly and wobbly (the flow and alignment between characters is inconsistent); lacking structure. Given those characteristics, this makes text feel disjointed from the strong, hefty Hound logo. I also find Lato to be quite thin and hard to read sometimes.

I recently came across the [Tofino](http://tofino.losttype.com/) typeface and tried it out on the Hound app. I think improves readability, but also pairs really well with the Hound logo. Even better, we can remove a bunch of CSS that tweaked letter spacing because Tofino is better built.

**Note:** I only have a free trial/personal version of Tofino, which I used to try it out on the app. We’d need to purchase a commercial license which is a flat $80, and also add it as a webfont to the app (which is why I labeled this as a WIP).

---

**With Tofino:**

houndci com-home-tofino

### DIFF
--- a/app/assets/stylesheets/base/_base.scss
+++ b/app/assets/stylesheets/base/_base.scss
@@ -3,6 +3,7 @@
 @import "animations";
 @import "buttons";
 @import "extends";
+@import "fonts";
 @import "forms";
 @import "helpers";
 @import "tables";
--- a/app/assets/stylesheets/base/_fonts.scss
+++ b/app/assets/stylesheets/base/_fonts.scss
@@ -1,0 +1,55 @@
+$_font-host-url: "http://font.thoughtbot.com/hound/";
+
+@include font-face(
+  "Tofino",
+  "#{$_font-host-url}Tofino-Book",
+  $asset-pipeline: false
+) {
+  font-style: normal;
+  font-weight: $font-weight-book;
+}
+
+@include font-face(
+  "Tofino",
+  "#{$_font-host-url}Tofino-BookItalic",
+  $asset-pipeline: false
+) {
+  font-style: italic;
+  font-weight: $font-weight-book;
+}
+
+@include font-face(
+  "Tofino",
+  "#{$_font-host-url}Tofino-Regular",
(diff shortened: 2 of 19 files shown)


smharley:
Overall, this is a great change! I think the homepage needs a little more love though (which could definitely be done separately). The italic "Currently style checking" feels more out of place and the language names below are a bit more condensed and harder to read. I have a similar feeling with the "Why does style matter" headings, maybe the bold is just throwing me off? Maybe the caps keep the letter-spacing?

In the docs, it feels like the line-height could be increased - most noticeable with the list in your example.

tysongach:
All good suggestions! I’ll do some tweaking. I do plan to do more work on the home page and others beyond the typeface change, which I think will go a long way in making it feel more honed-in.

tysongach:
@smharley Tweaked:

localhost-5000- laptop with hidpi screen

smharley:
Looking great!

tysongach:
Thanks for the sharp eyes, @smharley!

tysongach:
I’ve not yet purchased font and contacted them to make sure we’d get WOFF and/or WOFF2 font files for the web. We’ll also need to figure out where to host those. I don’t think it’d be great if we added those font files to this public repo.

tysongach:
I received WOFF files from the font maker and have purchased the license. Need to figure out where to host the font files. Perhaps our thoughtbot S3 bucket?

gylaz:
S3 would be a good place. I think there is a standard place somewhere where thoughtbot stores its marketing assets, maybe that's where this one should go?

tysongach:
I’ve added the fonts to our `thoughtbot-fonts` S3 buckets, which is served up at `font.thoughtbot.com`. I’ve also removed the Google Fonts import. This should be ready for final review and testing. Is there any thing we need to check/do because the fonts are served from a different domain?

gylaz:
If it looks good on staging, then good to merge.

PR #1418, opened 2017-07-20
Avoid querying bulk_customers for each repo being serialized

salbertson:


gylaz:
@salbertson Thanks! We were actually thinking that a bulk customers can now be represented by an `Owner` that is `whitelisted`. Thus we can preload a `repo`'s `owner` and check that flag. What do you think about this approach?

salbertson:
@gylaz hmm, not sure. Is "whitelisting" a thing we do now in Hound?

salbertson:
[In RepoSerializer repo.bulk? causes N+1. Let's fix that.](https://trello.com/c/i4m7XIVB/813-in-reposerializer-repobulk-causes-n1-lets-fix-that)

gylaz:
Creating a bulk customer record is basically what we call whitelisting. We could keep the bulk customer table, but make a `belongs_to` relationship to an `Owner`. Then we could do `repo.owner.bulk_customer.present?` to mean it's bulk/whitelisted.

app/controllers/repos_controller.rb
```diff
@@ -10,8 +10,17 @@ def index
...
+          bulk_customers_by_org: find_bulk_customers(repos).index_by(&:org),
+        )
       end
     end
   end
+
+  private
+
+  def find_bulk_customers(repos)
+    BulkCustomer.where(org: repos.map(&:organization).uniq)
```
gylaz: `map` -> `pluck`?

app/serializers/repo_serializer.rb
```diff
@@ -37,4 +37,16 @@ def has_admin_membership?
   def has_subscription?
     object.subscription&.user_id == scope.id
   end
+
+  def bulk?
+    if bulk_customers_by_org
```
gylaz: In what situation will this be nil?

spec/controllers/repos_controller_spec.rb
```diff
@@ -26,5 +26,25 @@
         expect(user.repos.size).to eq(1)
       end
     end
+
+    context "when current user is a bulk member of a repo" do
+      it "returns 0 price in cents for that repo" do
+        repo = create(
+          :repo,
+          in_organization: true,
+          private: true,
```
gylaz: Would our `in_private_org` trait come-in handy here?

spec/controllers/repos_controller_spec.rb
```diff
@@ -26,5 +26,25 @@
...
+          name: "yeehaw/wat",
+        )
+        create(:bulk_customer, org: "yeehaw")
+        user = create(:user)
+        user.repos << repo
```
gylaz: Could combine these two lines as `user = create(:user, repos: [repo])`.

gylaz:
I've opened #1433 to basically do this, but in a simpler "whitelist" approach. Going to close this PR in favor of the other one.

PR #1422, opened 2017-08-03
Move to 100% Haml views

sardaukar:
Getting rid of the few ERB views we have.


teoljungberg:
I'm personally in the ERB camp, but I know the rest of the team prefer HAML.

sardaukar:
@teoljungberg same, but having both is kind of odd. And Haml had a recent release, so it doesn't seem to be dead anymore.

app/views/shared/_plan_vertical.html.haml
```diff
@@ -0,0 +1,16 @@
+%div{ :class => "plan plan-vertical #{current ? 'current' : ''}" }
```
croaky: `class:` ?

sardaukar: Argh, I used `erb2haml` and this showed up. I need to read up on Haml to do this the 'proper' way.

PR #1426, opened 2017-08-16
Upgrade Ruby to 2.4.1

gylaz:
Update a few gems in order to get the tests to pass.


.env
```diff
@@ -16,7 +16,7 @@ RACK_TIMEOUT=20
 REDIS_URL=redis://localhost:6379
 RESQUE_ADMIN_PASSWORD=password
 RESQUE_JOB_TIMEOUT=120
-SECRET_KEY_BASE=some_really_really_long_super_secret_key_base
+SECRET_KEY_BASE=some_32byte_long_secret_key_base
```
teoljungberg: I'm curious, why did this need to change?

gylaz: There is something about Ruby 2.4.1 (bug?) that forces the `SECRET_KEY_BASE` to be *exactly* 32 bytes.

croaky: Maybe `exactly_32_bytes_long_string_key`? Also, is it byte size or character length that matters? Think it's the same in this example but not sure what the restriction is, whether the name should reflect characters instead of bytes.

tersiasarendse:
pull request

PR #1427, opened 2017-08-24
Remove unused route

m1foley:
Looks like the code for `repo_syncs#index` was removed in 95edf32.


gylaz:
Good eye! Thanks @m1foley

PR #1428, opened 2017-08-26
Remove unused PullRequest method

m1foley:
Looks like `PullRequest#repository_owner_name` was removed in 68e3136 and is no longer referenced.


gylaz:
Thank you!

PR #1430, opened 2017-08-31
Revert "Upgrade Ruby to 2.4.1"

gylaz:
This reverts commit b77a3e364fbf17023c4d920f3406248666f1bc0c.

Because Ruby 2.4.0+ does not allow keys longer than `32` bytes, as per
https://github.com/rails/rails/pull/25192

On staging and production we are currently using longer than 32 strings,
and we can't just shorten them (grab the first 32 chars) since we have
tokens stored in the DB encrypted with those longer strings.


gylaz:
It is the `secret_key_base` and this line is affected in particular: https://github.com/houndci/hound/blob/9bafdd49100f9a122cc5165742e9eae82f5a349b/app/models/user.rb#L53, which calls line 85 and `Rails.application.secrets.secret_key_base`.

teoljungberg:
What's the upgrade roadmap to get us to 2.4.*?

gylaz:
@teoljungberg We could wait to see if future Ruby releases resolve this issue.

However, if we want to do something proactive, we could migrate (re-encrypt) all the tokens with a `secret_key_base` that takes the first 32 characters of the current, long key base.

PR #1431, opened 2017-08-31
Accept invitation when added as collaborator

gylaz:
When collaboratore is added to a repo, GitHub now requires the
collaborator to accept the invitation.

This is keeping us from allowing users to activate their repos.

### DIFF
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -108,21 +108,22 @@ class GithubApi
     )
   end
 
+  def accept_invitation(repo_name)
+    repo_invitation = find_invitation_for_repo(repo_name)
+
+    if repo_invitation
+      accept_repository_invitation(repo_invitation)
+    else
+      raise "Invitation for Hound to #{repo_name} not found"
+    end
+  end
+
   def add_collaborator(repo_name, username)
-    client.add_collaborator(
-      repo_name,
-      username,
-      accept: "application/vnd.github.ironman-preview+json",
-    )
+    client.add_collaborator(repo_name, username)
   end
 
   def remove_collaborator(repo_name, username)
-    # not sure if we need the accept header
-    client.remove_collaborator(
-      repo_name,
-      username,
-      accept: "application/vnd.github.ironman-preview+json",
-    )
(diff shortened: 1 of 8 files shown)


lib/github_api.rb
```diff
@@ -108,21 +108,31 @@ def create_error_status(full_repo_name, sha, description, target_url = nil)
...
+    end
+
+    if repo_invitation
+      response = client.accept_repository_invitation(
+        repo_invitation.id,
+        accept: "application/vnd.github.swamp-thing-preview",
+      )
+      response == ""
+    else
+      raise "Invitation for Hound to #{repo_name} not found"
```
thorncp: Will this happen if a user activates a second repo?

gylaz: This got collapsed, but this is mean to execute for each repo.
So, when the user activates a second repo, they will hit this again.
Does that make sense?

thorncp: Yeah. So that means we expect GitHub to send an invitation for each repo?

lib/github_api.rb
```diff
@@ -147,4 +148,21 @@ def create_status(repo:, sha:, state:, description:, target_url: nil)
...
+      invitation.repository.full_name == repo_name
+    end
+  end
+
+  def accept_repository_invitation(invitation)
+    response = client.accept_repository_invitation(
+      invitation.id,
+      accept: "application/vnd.github.swamp-thing-preview",
+    )
+    response == ""
```
thorncp: This seems odd. What does it look like if it fails?

gylaz: It's actually a `204` (No Response), but Octokit just returns the blank/empty body as a string.

> What does it look like if it fails?

It raises an error `404 - Not Found`.

gylaz:
Tested on staging -- looked good.

lib/github_api.rb
gylaz: Yep. It's because the collaborator is added to each individual repo now instead of an org, so GitHub will (should!) issue an invitation for each.

PR #1433, opened 2017-09-11
Move BulkCustomers to whitelisted owners

gylaz:
Owners/organizations now have a flag to mark them as whitelisted.
This means that a user can enable any number of repos under that
owner/org for free.

This also simplifies, and speeds up, how we check if a repo is
whitelisted (previously bulk).

I've changed the admin side to display all of the owners now, with
interesting data, instead of just bulk customers.


thorncp:
Will administrate paginate the new dashboard?

gylaz:
Seems that Administrate will paginate at 20 by default.

app/models/owner.rb
```diff
@@ -20,6 +20,10 @@ def self.upsert(github_id:, name:, organization:)
     raise exception
   end
 
+  def private_active_repos_count
+    repos.active.where(private: true).count
```
teoljungberg: Worth adding a private scope to the repos?

db/schema.rb
```diff
@@ -99,26 +95,24 @@
...
-  add_index "repos", ["name"], name: "index_repos_on_name", using: :btree
-  add_index "repos", ["owner_id"], name: "index_repos_on_owner_id", using: :btree
-
   create_table "subscriptions", force: :cascade do |t|
-    t.datetime "created_at",                                                               null: false
-    t.datetime "updated_at",                                                               null: false
-    t.integer  "user_id",                                                                  null: false
-    t.integer  "repo_id",                                                                  null: false
-    t.string   "stripe_subscription_id", limit: 255,                                       null: false
+    t.datetime "created_at",                                                                 null: false
```
teoljungberg: Maybe we should disable the alignment in the schema by setting `ActiveRecord::SchemaDumper.standardized_argument_widths` and `ActiveRecord::SchemaDumper.standardized_type_widths`.

app/models/owner.rb
gylaz: We've probably resisted adding it by now because of the Ruby keyword `private`. It is likely just semantic though.

db/schema.rb
gylaz: I think Rails 5.1+ does away with it, so maybe we should upgrade?

app/models/owner.rb
gylaz: I'm going to try and do this in a follow-up PR.

PR #1435, opened 2017-09-12
Upgrade Ruby to 2.4.1 and Rails to 5.1.4

gylaz:
Found the solution to the `secret_base_key` problem, which we've
encountered in previous attempts to upgrade to Ruby 2.4.1:
https://github.com/rails/rails/issues/28401#issuecomment-286375944


Gemfile
```diff
@@ -22,7 +22,7 @@ gem "paranoia", "~> 2.2"
 gem "pathspec"
 gem "pg"
 gem "puma"
-gem "rails", "5.0.0.1"
+gem "rails", "~> 5.1.4"
```
teoljungberg: What do you think about dropping the patch-level here? That way we can run `bundle update rails` to update a patch release safely

gylaz: In general, I'm fine with not pinning in Gemfile. Rails feels like one of those, you would want to specifically upgrade (read the docs, migrate some code, etc). So, this is probably okay. But if we wanted to remove all pins in a PR, I'd probably thumbs it up.

PR #1440, opened 2017-09-16
Ignore GitHub status update errors

gylaz:
Sometimes the user's or Hound's token isn't privileged to set the
GitHub PR status for a given repo. Status isn't critical, and shouldn't
fail the rest of the build -- as long as we can access the files and
make comments.

We're actually seeing a lot of status "Not Found" errors that prevent
builds from finishing, thus this is measure to mitigate that.
This will reduce the job failure noise and allow us to identify which
repos we can't _actually_ review.

### DIFF
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -7,28 +7,38 @@ class CommitStatus
 
   def set_pending
     github.create_pending_status(repo_name, sha, I18n.t(:pending_status))
+  rescue Octokit::NotFound
+    # noop
   end
 
   def set_success(violation_count)
     message = I18n.t(:complete_status, count: violation_count)
     github.create_success_status(repo_name, sha, message)
+  rescue Octokit::NotFound
+    # noop
   end
 
   def set_failure(violation_count)
     message = I18n.t(:complete_status, count: violation_count)
-    github.create_error_status(repo_name, sha, message)
+    create_error_status(repo_name, sha, message)
   end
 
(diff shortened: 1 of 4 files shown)


app/models/commit_status.rb
```diff
@@ -7,29 +7,39 @@ def initialize(repo_name:, sha:, token:)
 
   def set_pending
     github.create_pending_status(repo_name, sha, I18n.t(:pending_status))
+  rescue Octokit::NotFound
+    # noop
```
thorncp: Is it worth tracking these?

spec/services/build_runner_spec.rb
```diff
@@ -299,6 +312,7 @@ def configuration_url
         repo.name,
         "somesha",
         I18n.t(:hound_error_status),
+        nil,
```
thorncp: `nil` is being passed around quite a bit now. Anything we can do to clean that up, short of refactoring the whole thing (probably out of scope here)?

gylaz: In `CommitStatus#create_error_status` we could check if `configuration_url` is `nil` and call it without the last argument. Extra, weird conditional, but it wouldn't lead to this diff noise. Thoughts?

app/models/commit_status.rb
gylaz: My feeling is no. They've been nothing but noise when I look at job failures/errors.

PR #1445, opened 2017-09-25
AccountDock invoicing

sardaukar:
This PR also fixes the styling for the "Update Email" button, and doesn't try to update the email if the input is empty.

Missing styling for the AccountDock popup at the moment.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -80,6 +80,10 @@ class User < ApplicationRecord
       order("LOWER(name) ASC")
   end
 
+  def card_exists?
+    stripe_customer_id.present?
+  end
+
   private
 
   def crypt
--- a/app/views/accounts/show.haml
+++ b/app/views/accounts/show.haml
@@ -54,9 +54,9 @@
                   )
 
       - if current_user.stripe_customer_id.present?
-        %script{src:"https://static.accountdock.com/app.min.js"}
-        %div{class:"billing-container"}
-          %button{id:"billing", class:"repo-toggle billing-button"} View Billing
+        %script{ src:"https://static.accountdock.com/app.min.js" }
+        .billing-container
+          %button.repo-toggle.billing-button{ id: "billing" } View Billing
         :javascript
           var handler = AccountDock.configure({
             key: '#{ENV['ACCOUNTDOCK_API_KEY']}',
--- a/app/views/accounts/show.haml
+++ b/app/views/accounts/show.haml
@@ -54,23 +54,7 @@
                   )
 
       - if current_user.stripe_customer_id.present?
-        %script{ src:"https://static.accountdock.com/app.min.js" }
-        .billing-container
-          %button.repo-toggle.billing-button#billing View Billing
-        :javascript
-          var handler = AccountDock.configure({
-            key: "#{ENV['ACCOUNTDOCK_API_KEY']}",
-            customer: "#{current_user.stripe_customer_id}"
-          });
-          $(function(){
-            $("#billing").on(
-              "click",
-              function(event) {
-                event.preventDefault();
-                handler.open();
-              }
-            );
-          });
+        = render(partial: "shared/account_dock", locals: { account_key: ENV["ACCOUNTDOCK_API_KEY"], stripe_customer_id: current_user.stripe_customer_id})
 
     - if current_user.stripe_customer_id.present?
       = javascript_pack_tag "update_account_email"
--- a/app/views/accounts/show.haml
+++ b/app/views/accounts/show.haml
@@ -53,11 +53,32 @@
                     precision: 0,
                   )
 
+      - if current_user.stripe_customer_id.present?
+        %script{src:"https://static.accountdock.com/app.min.js"}
+        %div{class:"billing-container"}
+          %button{id:"billing", class:"repo-toggle billing-button"} View Billing
+        :javascript
+          var handler = AccountDock.configure({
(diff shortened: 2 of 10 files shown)


sardaukar:
[Add support for invoices](https://trello.com/c/7fD6UUpD/788-add-support-for-invoices)

sardaukar:
screen shot 2017-09-25 at 15 31 13

app/views/accounts/show.haml
```diff
@@ -53,11 +53,32 @@
                     precision: 0,
                   )
 
+      - if current_user.stripe_customer_id.present?
+        %script{ src:"https://static.accountdock.com/app.min.js" }
+        .billing-container
+          %button.repo-toggle.billing-button{ id: "billing" } View Billing
```
gylaz: Should we use `#billing` HAML tag for id?

app/views/accounts/show.haml
```diff
@@ -53,11 +53,32 @@
...
+        :javascript
+          var handler = AccountDock.configure({
+            key: '#{ENV['ACCOUNTDOCK_API_KEY']}',
+            customer: '#{current_user.stripe_customer_id}'
+          });
+          $(function(){
+            $('#billing').on(
+              'click',
```
gylaz: Worth using double quotes on the strings in this JS snippet?

app/views/accounts/show.haml
```diff
@@ -53,11 +53,32 @@
...
+              function(event) {
+                event.preventDefault();
+                handler.open();
+              }
+            );
+          });
+
+
```
gylaz: Remove one of these newlines?

app/views/accounts/show.haml
sardaukar: the AccountDock JS payload uses this id to trigger the modal :|

app/views/accounts/show.haml
sardaukar: this is straight from their docs - I can tweak it

sardaukar:
@gylaz now we need to tweak the AccountDock modal (maybe) to look a bit more in place within Hound

app/views/accounts/show.haml
gylaz: Maybe my comment was misunderstood, but HAML syntax of `%button.repo-toggle.billing-button#billing` is preferred to generate an `id` attribute.

sardaukar: Ah ok, I thought you meant we should change the id, my bad.

app/views/accounts/show.haml
```diff
@@ -53,11 +53,31 @@
                     precision: 0,
                   )
 
+      - if current_user.stripe_customer_id.present?
+        %script{ src:"https://static.accountdock.com/app.min.js" }
```
teoljungberg: What do you think about extracting a partial out of this?

sardaukar: 👍

app/views/accounts/show.haml
```diff
@@ -53,11 +53,31 @@
                     precision: 0,
                   )
 
+      - if current_user.stripe_customer_id.present?
```
teoljungberg: Checkingfor the `stripe_customer_id` feels very much on an implementation detail. What do you think of changing this to something along the lines of `if current_user.customer?`

sardaukar: I went with `#card_exists?` that was already introduced into `UserSerializer` before.

teoljungberg: Awesome.

app/views/accounts/show.haml
```diff
@@ -53,11 +53,15 @@
...
+        = render(partial: "shared/account_dock", locals: { account_key: ENV["ACCOUNTDOCK_API_KEY"], stripe_customer_id: current_user.stripe_customer_id})
```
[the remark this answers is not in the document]
teoljungberg: What do you think about adding the `account_key` as a constant in `Consants::*` ?

app/views/accounts/show.haml
```diff
@@ -53,11 +53,15 @@
                     precision: 0,
                   )
 
-    - if current_user.stripe_customer_id.present?
+      - if current_user.card_exists?
+        = render(partial: "shared/account_dock", locals: { account_key: ENV["ACCOUNTDOCK_API_KEY"], stripe_customer_id: current_user.stripe_customer_id})
```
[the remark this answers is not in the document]
sardaukar: I ain't gonna fix this, because Haml is so annoying about line breaks that I can't even :D

app/views/shared/_account_dock.html.haml
```diff
@@ -0,0 +1,17 @@
+%script{ src:"https://static.accountdock.com/app.min.js" }
+.billing-container
+  %button.repo-toggle.billing-button#billing View Billing
```
salbertson: What do you think about "Billing History" instead of "View Billing"? This entire page is billing, sort of.

app/models/user.rb
```diff
@@ -80,6 +80,10 @@ def repos_by_activation_ability
       order("LOWER(name) ASC")
   end
 
+  def card_exists?
```
salbertson: 🎉

sardaukar:
@salbertson I'm out of context here, don't really remember what it looked like sorry :|

salbertson:
@sardaukar no problem, thanks.

PR #1447, opened 2017-09-26
update yarn version

davidimoore:
The purpose of this PR is to update the yarn version specified in the package.json file

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "engines": {
     "node": "^8.4.0",
-    "yarn": "^1.0.1"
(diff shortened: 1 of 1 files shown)


gylaz:
LGTM, but looks like Circle doesn't have yarn `1.1.0` yet?

davidimoore:
I wonder if requiring a yarn version is necessary?

gylaz:
Do you mean removing it from the `engines` specifications?

davidimoore:
for yarn, yeah...feels equivalent to mandating a bundler version.

PR #1451, opened 2017-10-03
Handle Stripe hook for deleted subscription events

gylaz:
We sometimes delete/cancel user subscriptions directly from Stripe.
Stripe is nice enough to notify us of this via a webhook.
This ensures we cleanup properly on our end -- delete all the
subscriptions by the give stripe token and deactivate the related repos.

### DIFF
--- a/app/controllers/deleted_subscriptions_controller.rb
+++ b/app/controllers/deleted_subscriptions_controller.rb
@@ -1,0 +1,9 @@
+class DeletedSubscriptionsController < ApplicationController
+  skip_before_action :authenticate, only: :create
+
+  def create
+    DeleteSubscriptions.call(params)
+
+    head :ok
+  end
+end
--- a/spec/requests/subscriptions_spec.rb
+++ b/spec/requests/subscriptions_spec.rb
@@ -1,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe "Subscriptions" do
+  context "when event is for subscription deleted" do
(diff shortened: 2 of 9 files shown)


app/controllers/deleted_subscriptions_controller.rb
```diff
@@ -0,0 +1,7 @@
+class DeletedSubscriptionsController < ApplicationController
+  skip_before_action :authenticate, only: :create
+
+  def create
+    DeleteSubscriptions.call(params)
+  end
```
thorncp: There doesn't seem to be a response for this action. Perhaps add a `head :ok`?

spec/requests/subscriptions_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+          },
+        },
+      }
+
+      post "/deleted_subscriptions", params: params
+
+      subscription.reload
+      expect(subscription).to be_deleted
+      expect(subscription.repo).not_to be_active
+    end
```
thorncp: Add an expectation for a 200 response?

app/controllers/deleted_subscriptions_controller.rb
gylaz: Good call.

spec/requests/subscriptions_spec.rb
gylaz: 👍

PR #1452, opened 2017-10-04
Add aliases for ScssLint and SassLint

gylaz:
Sometimes users will use the actual linter names in the config file.
We should support these to avoid confusion for our users.


valentin-h:
Approved Greg !

gylaz:
Merged as 46d5d834230b4a60da59a4d266d77c41b9f5274a.

PR #1457, opened 2017-11-01
Protect organization configuration.

robkilby:
This change protects the organization-wide configuration settings from
changes by anyone who is not an owner in the organization. Changes to an
user's own personal organization are allowed at all times.


robkilby:
Hi @HoundCI is there any chance of getting this PR looked at? The issue I'm fixing here is causing a number of problems at the company I work for. Thanks!

PR #1459, opened 2017-11-04
Show linter versions in the configuration docs

gylaz:
Parsing most linter versions from `Gemfile.lock` and `yarn.lock`.

Golint has no semver-like version, so we don't display any version.
Credo, Flake8 and SwiftLint are hard-coded for now, because they don't
often change.

Re-ordered the linters on the configuration doc page for a better flow.

### DIFF
--- a/app/models/linter_version.rb
+++ b/app/models/linter_version.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+class LinterVersion
+  LINTERS_REPO_NAME = "houndci/linters"
+  LOCKFILE_TO_LINTERS = {
+    gemfile: %i(
+      flog
+      haml_lint
+      reek
+      rubocop
+      scss_lint
+      slim_lint
+    ),
+    yarn: %i(
+      coffeelint
+      eslint
+      jshint
+      sass-lint
+      stylelint
+      tslint
+    ),
+  }.freeze
+
+  def self.all
+    new.all
+  end
+
+  def all
+    LOCKFILE_TO_LINTERS.flat_map do |lockfile, linters|
+      linters.map do |linter|
+        [linter, send("version_from_#{lockfile}", linter)]
+      end
+    end.to_h
+  end
+
+  private
+
+  def version_from_gemfile(linter_name)
+    gemfile_content.
+      scan(/^ {4}#{linter_name} \((.*)\)/).
+      flatten.first || "N/A"
+  end
+
+  def version_from_yarn(linter_name)
+    yarn_content.
+      scan(/^#{linter_name}@(?:[^\n]+)\s+version\s"([\d\.]+)/).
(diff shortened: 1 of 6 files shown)


app/views/pages/configuration.haml
```diff
@@ -228,53 +208,141 @@
...
+      %p
+        To ignore certain files and directories, add
+        %em.code-inline .eslintignore
+        file to your project,
+        include in it path patterns for files/directories that you want ignored,
+        and reference the ignore file in your
+        %em.code-inline .hound.yml
+        \. See
+        = link_to "linter docs",
+          "http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories",
```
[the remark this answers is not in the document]
gylaz: Nein.

app/models/linter_version.rb
```diff
@@ -0,0 +1,70 @@
...
+        [linter, send("version_from_#{lockfile}", linter)]
+      end
+    end.to_h
+  end
+
+  private
+
+  def version_from_gemfile(linter_name)
+    gemfile_content.
+      scan(/^ {4}#{linter_name} \((.*)\)/).
```
thorncp: I didn't know `#scan` worked this way with capture groups. TIL

app/models/linter_version.rb
```diff
@@ -0,0 +1,70 @@
...
+      flatten.first || "N/A"
+  end
+
+  def version_from_yarn(linter_name)
+    yarn_content.
+      scan(/^#{linter_name}@(?:[^\n]+)\s+version\s"([\d\.]+)/).
+      flatten.max || "N/A"
```
thorncp: Is `max` doing a string comparison here? Do we need special handling for version numbers?

E.g

```
irb(main):001:0> ["1.23.1", "1.3.1"].max
=> "1.3.1"
```

gylaz: Yep, it's doing string comparison. I think the biggest one is always the one we need. In your example, it looks like `1.3.1` is the desired result?

thorncp: If it's semantic versioning, 1.23.1 would be the latest.

teoljungberg: I agree, it should be `1.23.1`. We could use `Gem::Version` since that handles that case for us:

```
% irb
>> [Gem::Version.new("1.23.1"), Gem::Version.new("1.3.1")].max
=> #<Gem::Version "1.23.1">
```

tysongach:
[Show linter versions in our docs and config page](https://trello.com/c/b4OiREVY/773-show-linter-versions-in-our-docs-and-config-page)

app/models/linter_version.rb
salbertson: That should work, @gylaz what do you think?

PR #1469, opened 2018-01-30
fix migrations

acrogenesis:
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateDieselClearanceUsers < ActiveRecord::Migration[4.2]
/app/vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/migration.rb:525:in `inherited'

### DIFF
--- a/db/migrate/20121229003013_create_diesel_clearance_users.rb
+++ b/db/migrate/20121229003013_create_diesel_clearance_users.rb
@@ -1,4 +1,4 @@
-class CreateDieselClearanceUsers < ActiveRecord::Migration
+class CreateDieselClearanceUsers < ActiveRecord::Migration[4.2]
   def self.up
     create_table(:users) do |t|
       t.string   :email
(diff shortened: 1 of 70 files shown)


db/migrate/20170911205516_add_whitelisted_to_owners.rb
```diff
@@ -1,4 +1,4 @@
-class AddWhitelistedToOwners < ActiveRecord::Migration[5.0]
+class AddWhitelistedToOwners < ActiveRecord::Migration[4.2][5.0]
```
gylaz: Why does this have both `[4.2]` and `[5.0]`? Should all of these be `[5.0]` instead?

acrogenesis: My bad @gylaz I remade the commit with those fixed, strangely the migrations ran with `[4.2][5.0]`. I left them as `[4.2]`.

Only two migrations specified the version.

acrogenesis:
@gylaz I believe it's ready for merge. 👍

gylaz:
Thanks @acrogenesis!

PR #1471, opened 2018-02-16
Update gems

tysongach:


nickcharlton:
Rebased so we could pick up the CI fix in #1474 and now I'm going to merge this so we get rid of the other warnings.

tysongach:
Nice, thanks Nick!

PR #1475, opened 2018-02-18
Capitalize the h in GitHub

tarebyte:
Ok so this a Pull Request full of pageantry. I'll totally understand if this change isn't something the project wants to make, as it touches a lot of files in a rather unnecessary way.

This adds `GitHub` as an acronym in the `config/initializers/inflections.rb` file so that a file like `github_api.rb` translates to `class GitHubApi`.

### DIFF
--- a/lib/github_api.rb
+++ b/lib/github_api.rb
@@ -3,7 +3,7 @@ require "attr_extras"
 require "octokit"
 require "base64"
 
-class GithubApi
+class GitHubApi
   ORGANIZATION_TYPE = "Organization"
   PREVIEW_API_HEADER = "application/vnd.github.black-cat-preview+json"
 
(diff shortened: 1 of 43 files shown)


lib/github_api.rb
```diff
@@ -3,7 +3,7 @@
 require "octokit"
 require "base64"
 
-class GithubApi
+class GitHubApi
```
gylaz: Generally, we'd have to name this file as `git_hub_api.rb` because of this change, since every camel cased word is considered a separated word in Ruby. GitHub is actually one word not two, so we intentionally named this file `GithubApi` due to what Ruby considers a word.

salbertson:
Looks good to me, thanks @tarebyte! Would you mind resolving the conflicts and squashing your commits?

tarebyte:
> Would you mind resolving the conflicts and squashing your commits?

Done!

PR #1477, opened 2018-02-19
Remove unused PostCSS plugins

tysongach:
I don't believe any of this is needed, because all of our application styles are on the Rails side.


gylaz:
Does the app look good without this locally? Worth checking out on staging? Just curious if React is doing anything with styles.

tysongach:
@gylaz I see no visual differences between `master` and this branch locally. I will note, though, that the `/repos` page doesn't load repos for me locally, not even on `master`.

screen shot 2018-02-19 at 17 29 48

I successfully have had the local dev GitHub app setup; should I be seeing my typical list of repos I am part of here?

gylaz:
It should work locally, yes. Probably worth digging into those 500 responses.

PR #1478, opened 2018-02-19
Remove package.json engines

tysongach:
Declaring Node as an engine is causing Yarn to error when running
`bin/setup`.

https://docs.npmjs.com/files/package.json#engines

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,10 +7,6 @@
     "test:watch": "jest --watch app/javascript/__tests__/*-test.jsx",
     "test:coverage": "jest --coverage app/javascript/__tests__/*-test.jsx"
   },
-  "engines": {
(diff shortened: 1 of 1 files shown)


gylaz:
@tysongach Try upgrading yarn to the latest. It shouldn't care about engines anymore. Having said that, I'm not sure that specifying `engines` has helped us in any way, so I'm 👍 on this.

tysongach:
@gylaz 👋 I'm on Yarn v1.3.2, which is the latest stable version. I was able to get around the issue by manually running `yarn install --ignore-engines`, instead of simply `yarn` as defined in `bin/setup`. But yeah, I don't think having this in `package.json` is doing anything for us, so I figured it's fine to remove and fix the problem that way.

PR #1480, opened 2018-02-20
Switch to chrome webdriver for JS feature tests

gylaz:
The future of QT and capybara-webkit is unknown.
Selenium with Chrome (headless) seems like a great alternative.

Ensure the repos list landing page (after sign in) finishes loading the
list of repose (via Ajax) for JS tests. Otherwise, a JS alert popup with
a failure causes tests to fail. This actually happens in Chrome on
production when navigating from the repos page before the AJAX request
 to get the repos finishes (the `Your repos failed to load.` alert).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/hound
     docker:
-      - image: circleci/ruby:2.5.0-node
+      - image: circleci/ruby:2.5.0-node-browsers
         environment:
           RAILS_ENV: test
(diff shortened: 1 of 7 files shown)


circle.yml
```diff
@@ -26,16 +26,6 @@ jobs:
         name: Restore yarn cache
         key: houndci-yarn-{{ checksum "yarn.lock" }}
 
-      - type: cache-restore
```
salbertson: 🎉

PR #1482, opened 2018-02-25
Activate repo when there is no invitation

gylaz:
If Hound already can access the repo, we don't need to accept an
invitation -- when invitation doesn't exist the activation process blows
up.


salbertson:
Looks good to me. 👍

PR #1486, opened 2018-03-03
Restore jquery-ujs to support delete links

salbertson:
This fixes the [Stop Masquerading](https://github.com/houndci/hound/blob/150d791c056d870c80feb5b7f609be45499ae5d0/app/views/application/_app_nav.haml#L39) link.


gylaz:
Which delete links? What issue does this fix?

PR #1487, opened 2018-03-03
Enable Style/FrozenStringLiteralComment and update Ruby files

tarebyte:
This just enforces that the entire application is using the magic `frozen_string_literal: true` comment.


gylaz:
Hi @tarebyte. We follow the philosophy of not re-writing the entire codebase to follow a particular (new) rule, but to update it overtime, updating files/lines as we touch them. Given, that this PR seems to be against our philosophy. Is there any particular reason we should bend our rule in this case?

tarebyte:
> Is there any particular reason we should bend our rule in this case?

Not really ¯\\_(ツ)_/¯

I'm fine with closing this is something you don't want.

PR #1493, opened 2018-03-18
Don't cut off repo names

salbertson:
Before
screenshot 2018-03-17 18 01 19

After
screenshot 2018-03-17 18 01 35

### DIFF
--- a/app/assets/stylesheets/pages/repos/_index.scss
+++ b/app/assets/stylesheets/pages/repos/_index.scss
@@ -28,7 +28,7 @@
   @include ellipsis;
   color: $medium-font-color;
   font-size: 1.5em;
-  line-height: 1;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/pages/repos/_index.scss
```diff
@@ -28,7 +28,7 @@
   @include ellipsis;
   color: $medium-font-color;
   font-size: 1.5em;
-  line-height: 1;
+  line-height: 1.2;
```
gylaz: what are the units here? percent? is it worth being specific?

salbertson: It's not percent, it's regular ole unitless `line-height`.

PR #1496, opened 2018-03-23
Handle nil options when resolving config conflicts

salbertson:
* Avoid `NoMethodError: undefined method `[]' for nil:NilClass`

### DIFF
--- a/app/services/resolve_config_conflicts.rb
+++ b/app/services/resolve_config_conflicts.rb
@@ -21,7 +21,7 @@ class ResolveConfigConflicts
       else
         raise Config::ParserError.new(
           "Missing options or invalid config format",
-          linter_name: linter
+          linter_name: linter,
         )
       end
       resolved_config
--- a/app/services/resolve_config_conflicts.rb
+++ b/app/services/resolve_config_conflicts.rb
@@ -15,7 +15,7 @@ class ResolveConfigConflicts
     @config.reduce({}) do |resolved_config, (linter, options)|
       if options.nil?
         raise Config::ParserError.new(
-          "Invalid #{linter} config or options are missing.",
+          "#{linter} options in your .hound.yml are invalid",
           linter_name: linter,
         )
       else
--- a/app/services/resolve_config_conflicts.rb
+++ b/app/services/resolve_config_conflicts.rb
@@ -18,6 +18,11 @@ class ResolveConfigConflicts
           resolved_config[CONFLICTS[linter]] = { "enabled" => false }
         end
         resolved_config[linter] = options
+      else
+        raise Config::ParserError.new(
+          "Missing options or invalid config format",
+          linter_name: linter
+        )
       end
       resolved_config
     end
--- a/app/services/resolve_config_conflicts.rb
+++ b/app/services/resolve_config_conflicts.rb
@@ -13,16 +13,16 @@ class ResolveConfigConflicts
 
   def call
     @config.reduce({}) do |resolved_config, (linter, options)|
-      if options
+      if options.nil?
+        raise Config::ParserError.new(
+          "Missing configuration options or invalid format",
+          linter_name: linter,
+        )
+      else
         if CONFLICTS.has_key?(linter) && options["enabled"] == true
           resolved_config[CONFLICTS[linter]] = { "enabled" => false }
         end
         resolved_config[linter] = options
-      else
-        raise Config::ParserError.new(
-          "Missing options or invalid config format",
-          linter_name: linter,
-        )
       end
       resolved_config
     end
--- a/app/services/resolve_config_conflicts.rb
+++ b/app/services/resolve_config_conflicts.rb
@@ -15,7 +15,7 @@ class ResolveConfigConflicts
     @config.reduce({}) do |resolved_config, (linter, options)|
       if options.nil?
         raise Config::ParserError.new(
-          "Missing configuration options or invalid format",
+          "Invalid #{linter} config or options are missing.",
           linter_name: linter,
         )
       else
--- a/app/services/resolve_config_conflicts.rb
+++ b/app/services/resolve_config_conflicts.rb
@@ -13,10 +13,12 @@ class ResolveConfigConflicts
 
   def call
     @config.reduce({}) do |resolved_config, (linter, options)|
-      if CONFLICTS.has_key?(linter) && options["enabled"] == true
-        resolved_config[CONFLICTS[linter]] = { "enabled" => false }
+      if options
+        if CONFLICTS.has_key?(linter) && options["enabled"] == true
+          resolved_config[CONFLICTS[linter]] = { "enabled" => false }
+        end
+        resolved_config[linter] = options
       end
-      resolved_config[linter] = options
       resolved_config
     end
   end
--- a/spec/services/build_runner_spec.rb
+++ b/spec/services/build_runner_spec.rb
@@ -152,7 +152,7 @@ describe BuildRunner do
         expect(github_api).to have_received(:create_error_status).with(
           repo.name,
           payload.head_sha,
-          '"config/rubocop.yml" must be a Hash',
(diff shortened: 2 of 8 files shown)


salbertson:
[Handle nil config options when checking for conflicts](https://trello.com/c/w6tEjDjY/836-handle-nil-config-options-when-checking-for-conflicts)

app/services/resolve_config_conflicts.rb
```diff
@@ -13,6 +13,8 @@ def initialize(config)
 
   def call
     @config.reduce({}) do |resolved_config, (linter, options)|
+      return {} unless options
```
gylaz: Don't we want to skip the iteration, and handle any other linters, if any?
What do you think about something like?
```
@config.reduce({}) do |resolved_config, (linter, options)|
  if options
    if CONFLICTS.has_key?(linter) && options["enabled"] == true
     resolved_config[CONFLICTS[linter]] = { "enabled" => false }
    end
    resolved_config[linter] = options
  end
  resolved_config
end
```

spec/services/resolve_config_conflicts_spec.rb
```diff
@@ -23,5 +23,15 @@
         expect(resolved_config["sass_lint"]).to eq("enabled" => true)
       end
     end
+
+    context "given nil config options" do
+      it "doesn't raise and returns empty Hash" do
+        config = { "sass_lint" => nil }
```
gylaz: Curious what is the situation where this happens.
Is it a config like:
```yaml
# .hound.yml

sass_lint:
```

salbertson: Looks like lack of indentation.
```
eslint:
enabled: true
config_file: ./eslintrc.json
```

salbertson: @gylaz should we validation these configs and report the errors to the team, instead of skipping them?

spec/services/resolve_config_conflicts_spec.rb
```diff
@@ -23,5 +23,16 @@
...
     end
+
+    context "given nil config options" do
+      it "skips nil options" do
+        config = { "sass_lint" => nil, "eslint" => { "enabled" => true }}
+
+        resolved_config = ResolveConfigConflicts.call(config)
+
+        expect(resolved_config["jshint"]).to eq("enabled" => false)
+        expect(resolved_config["eslint"]).to eq("enabled" => true)
```
gylaz: Just an FYI, could use `include` matcher, like:
```rb
expect(resolved_config).to include(
  "jshint" => { "enabled" => false },
  "eslint" => { "enabled" => true },
)
```

app/services/resolve_config_conflicts.rb
```diff
@@ -13,10 +13,17 @@ def initialize(config)
...
     @config.reduce({}) do |resolved_config, (linter, options)|
-      if CONFLICTS.has_key?(linter) && options["enabled"] == true
-        resolved_config[CONFLICTS[linter]] = { "enabled" => false }
+      if options
+        if CONFLICTS.has_key?(linter) && options["enabled"] == true
+          resolved_config[CONFLICTS[linter]] = { "enabled" => false }
+        end
+        resolved_config[linter] = options
+      else
+        raise Config::ParserError.new(
```
salbertson: @gylaz something like this?

gylaz: yes

app/services/resolve_config_conflicts.rb
```diff
@@ -13,10 +13,17 @@ def initialize(config)
...
+          "Missing options or invalid config format",
```
gylaz: Do you think `Missing options` would be helpful to the end user?

How about something like, `Invalid config. Please check your .hound.yml format.`?

salbertson: I don't think users ever see that message, do they?

https://github.com/houndci/hound/blob/25380a02a9069d204e79500d01311a1dd06f1f6d/app/services/build_runner.rb#L94-L100

app/services/resolve_config_conflicts.rb
```diff
@@ -13,10 +13,17 @@ def initialize(config)
 
   def call
     @config.reduce({}) do |resolved_config, (linter, options)|
-      if CONFLICTS.has_key?(linter) && options["enabled"] == true
-        resolved_config[CONFLICTS[linter]] = { "enabled" => false }
+      if options.nil?
```
salbertson: Great catch by [tests](https://circleci.com/gh/houndci/hound/3522). I updated it so we raise only when options are truly missing.

app/services/resolve_config_conflicts.rb
gylaz: They do. If you look inside `ReportInvalidConfig`, you can see that we make a call to GitHub to set the status message with the error.

salbertson: How, we never pass `exception` to `.call`, just `exception.linter_name`?

gylaz: oh, ha! this particular call doesn't pass in the message, but `app/jobs/report_invalid_config_job.rb` does. I'm not sure which path we take in this situation, or if the job is used at all.

salbertson: I'm going to leave it for now to address the issue and report the errors. I guess we can have a bigger discussion about what messages to show. Sound good?

gylaz: The original intention was to pass through the message. I don't see the harm in making the message more user friendly now and pass it through. Is there some reason not to?

BTW, Looks like `ReportInvalidConfigJob` is not used anymore.

salbertson: Which message would we rather use?

"Missing configuration options or invalid format"
or
"Error parsing config for: sass_lint. Click details for assistance."

gylaz: Ha, neither.

Something like this would be better IMO:
```
Invalid config or missing options for sass_lint.
```
or
```
sass_lint config error or missing options.
```
(so the linter is up front, and won't get cutoff).

salbertson: Original intention was to show an error, any error, so the build wouldn't fail when an invalid config was used. I'd like to keep this PR focussed, that's the only reason.

gylaz: > Original intention was to show an error, any error, so the build wouldn't fail

I'm not sure I'm convinced. Why did we even include the linter name in the generic error, and not say something like `Oops, something went wrong?`.

I mean, since we're touching this code, I thought it would good to update the message to something more explanatory to the user (so, they know where to look to fix).

I'm okay if you want to make a quick followup PR, to keep this one focused. But I think if we backlog it, we'll not likely to come back to this soon.
I believe that a slightly better commit message, can reduce the amount of customer support requests. (That's the hope anyway).

You call though.

spec/services/build_runner_spec.rb
```diff
@@ -152,7 +152,7 @@
         expect(github_api).to have_received(:create_error_status).with(
           repo.name,
           payload.head_sha,
-          I18n.t(:config_error_status, linter_name: "rubocop"),
+          '"config/rubocop.yml" must be a Hash',
```
gylaz: Huh... to the end user, that's not very helpful. Will everyone know that YAML is basically a hash?

How about `"config/rubocop.yml" must be valid YAML format`?

salbertson: Does the message need to be more general to fit all linters config descriptions?

https://github.com/houndci/hound/blob/5803a8e86c968e278b992ce4c19d78d66870ecc1/app/models/config_content.rb#L36

I didn't add this message, it's being used because we are now providing `ReportInvalidConfig.call` with an optional message.

https://github.com/houndci/hound/blob/238f4952b9f835df4604488552f84142d05de467/app/services/build_runner.rb#L95-L100

PR #1497, opened 2018-03-23
Upgrade with new payment information

salbertson:
Prompt user for their payment info when their upgrade fails.

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -7,31 +7,26 @@ class SubscriptionsController < ApplicationController
   def create
     if current_user.plan_upgrade?
       render json: {}, status: :payment_required
-    elsif activator.activate && create_subscription
-      render json: repo, status: :created
     else
-      activator.deactivate
-
-      head 502
+      activate_and_create_subscription
     end
   end
 
   def update
-    if activator.activate && create_subscription
-      render json: repo, status: :created
-    else
-      activator.deactivate
-      head 502
-    end
+    activate_and_create_subscription
   end
 
   def destroy
-    if activator.deactivate && delete_subscription
-      analytics.track_repo_deactivated(repo)
+    if activator.deactivate
+      if delete_subscription
+        analytics.track_repo_deactivated(repo)
 
-      render json: repo, status: :created
+        render json: repo, status: :created
+      else
+        render_error("There was an issue deleting the subscription")
+      end
     else
-      head 502
+      render_error("There was an issue deactivating the repo")
     end
   end
 
@@ -71,5 +66,32 @@ class SubscriptionsController < ApplicationController
     if current_user.email.blank?
       current_user.update(email: params[:email])
     end
+  end
+
+  def activate_and_create_subscription
+    if activator.activate
+      if create_subscription
+        render json: repo, status: :created
+      else
+        activator.deactivate
+        render_error("There was an issue creating the subscription")
+      end
+    else
+      render_error("There was an issue activating the repo")
+    end
+  end
+
+  def render_error(error)
+    Raven.capture_message(
+      error,
+      extra: {
+        repo: repo.name,
+        repo_id: repo.id,
+        user_email: current_user.email,
+        username: current_user.username,
+      },
+    )
+
+    render json: { errors: [error] }, status: :bad_gateway
   end
 end
--- a/app/javascript/components/NotifyTierChange/components/UpgradeSubscriptionLink.jsx
+++ b/app/javascript/components/NotifyTierChange/components/UpgradeSubscriptionLink.jsx
@@ -5,6 +5,12 @@ import * as Ajax from '../../../modules/Ajax';
 import { getCSRFfromHead } from '../../../modules/Utils';
 
 export default class UpgradeSubscriptionLink extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { disabled: false };
+  }
+
   componentWillMount() {
     $.ajaxSetup({
       headers: {
@@ -33,7 +39,7 @@ export default class UpgradeSubscriptionLink extends React.Component {
     const { price, title } = this.props.nextTier;
 
     this.checkout().open({
-      name: title,
+      name: `Upgrade to ${title}`,
       amount: price * 100,
       email: userEmailAddress,
       panelLabel: "{{amount}} per month",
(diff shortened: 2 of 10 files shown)


salbertson:
[When a card is expired and user tries to upgrade the plan, we should surface the error from Stripe in the UI](https://trello.com/c/IlCA1XGF/831-when-a-card-is-expired-and-user-tries-to-upgrade-the-plan-we-should-surface-the-error-from-stripe-in-the-ui)

salbertson:
[Users are unable to re-activate Hound if they had let a subscription in the past expire.](https://trello.com/c/hFmQeb47/833-users-are-unable-to-re-activate-hound-if-they-had-let-a-subscription-in-the-past-expire)

app/javascript/components/NotifyTierChange/components/UpgradeSubscriptionLink.jsx
```diff
@@ -33,19 +39,42 @@ export default class UpgradeSubscriptionLink extends React.Component {
...
       allowRememberMe: false
     });
   }
 
+  handleFailedUpgrade(response) {
+    const { intercom } = this.props;
+
+    const subscriptionFailed = response &&
+      response.errors &&
+      response.errors.includes("There was an issue creating the subscription")
```
gylaz: Semicolon at the end?

app/javascript/components/NotifyTierChange/components/UpgradeSubscriptionLink.jsx
```diff
@@ -33,19 +39,42 @@ export default class UpgradeSubscriptionLink extends React.Component {
...
+
+    if (subscriptionFailed) {
+      this.showCreditCardForm();
+    } else {
+      this.setState({ disabled: false });
+
+      intercom(
+        "showNewMessage",
+        "I cannot upgrade and activate my repo. Please help!"
```
gylaz: Does this start a thread conversation? Does the user know this is being sent on their behalf? Would it be surprising to the user?

spec/services/repo_subscriber_spec.rb
```diff
@@ -22,14 +22,45 @@
...
+          RepoSubscriber.subscribe(repo, user, nil)
 
           expect(subscription_request).to have_been_requested
           expect(subscription_update_request).to have_been_requested
           expect(update_request).not_to have_been_requested
           expect(repo.subscription.stripe_subscription_id).
             to eq(stripe_subscription_id)
         end
+
+        it "creates a new Stripe subscription with new card, and repo subscription" do
```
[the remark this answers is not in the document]
gylaz: Maybe remove a few words? Something like:
```
creates a Stripe and repo subscriptions with new card
```

spec/services/repo_subscriber_spec.rb
```diff
@@ -22,14 +22,45 @@
...
+          repo = create(:repo, private: true)
+          user = create(
+            :user,
+            stripe_customer_id: stripe_customer_id,
```
gylaz: Where does `stripe_customer_id` variable come from?

spec/services/repo_subscriber_spec.rb
```diff
@@ -22,14 +22,45 @@
...
+            repos: [repo],
+          )
+          stub_customer_find_request
+          update_request = stub_customer_update_request
```
gylaz: What's the difference between this one and `customer_update_request`?
Do we even need this assertion? If `customer_update_request` i called once, then we know that no other `customer update` requests were made, right?

app/javascript/components/NotifyTierChange/components/UpgradeSubscriptionLink.jsx
salbertson: It will open the Intercom messenger and pre-fill the message. The user will have a chance to edit the message, or include info, and click "send".

spec/services/repo_subscriber_spec.rb
salbertson: Looks like it's a spec helper method.

https://github.com/houndci/hound/blob/986aec32a1b3aef5df38aea079b5bcd05168808d/spec/support/helpers/stripe_api_helper.rb#L2-L4

spec/services/repo_subscriber_spec.rb
salbertson: Good call, we don't need this, you're right.

app/controllers/subscriptions_controller.rb
```diff
@@ -72,4 +67,34 @@ def update_email
...
+      else
+        activator.deactivate
+        render_error("There was an issue creating the subscription")
+      end
+    else
+      render_error("There was an issue activating the repo")
+    end
+  end
+
+  def render_error(error)
```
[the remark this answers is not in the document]
gylaz: Should we change this to 15?

app/jobs/accept_github_invitation_job.rb
```diff
@@ -0,0 +1,15 @@
+class AcceptGitHubInvitationJob < ApplicationJob
+  queue_as :high
+
+  def perform(repo_name)
+    return if hound_github.repository?(repo_name)
```
gylaz: We've been avoiding these early returns, instead favoring:
```rb
unless hound_github.repository?(repo_name)
  hound_github.accept_invitation(repo_name)
end
```

Thoughts?

app/jobs/accept_github_invitation_job.rb
```diff
@@ -0,0 +1,15 @@
+class AcceptGitHubInvitationJob < ApplicationJob
```
gylaz: This doesn't seem related to this PR.

spec/jobs/accept_github_invitation_job_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+        expect(github).not_to have_received(:accept_invitation)
+      end
+    end
+
+    context "when Hound doesn't have access to the repo" do
+      it "accepts the invitation" do
+        github = double(GitHubApi, repository?: false, accept_invitation: nil)
+        allow(GitHubApi).to receive(:new).and_return(github)
+
+        job = subject.perform("hound/test")
```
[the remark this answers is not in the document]
salbertson: 🦅 👁

app/jobs/accept_github_invitation_job.rb
salbertson: I don't mind the "guard" style, but I'm cool with it if that's been the convention. Any good reason other than convention?

app/jobs/accept_github_invitation_job.rb
salbertson: You got me. Should I pull this into a separate PR?

app/jobs/accept_github_invitation_job.rb
gylaz: > Any good reason other than convention?

The only good argument that I've heard is one-liners hide complexity:

It's more explicit that there is clearly branching logic/complexity when using `if/else`. This really comes into play if you have multiple early `return` lines -- looks simple, but really there's a bit of complexity to the method.

app/jobs/accept_github_invitation_job.rb
salbertson: https://github.com/houndci/hound/pull/1501

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -101,6 +102,7 @@
           email: "TEST_USER_EMAIL",
           repos: repos,
           token: "TEST_USER_TOKEN",
+          username: "jimtom",
```
gylaz: Jimtom lives!

PR #1501, opened 2018-03-24
Accept GitHub invitations in a background job

salbertson:


app/jobs/accept_github_invitation_job.rb
```diff
@@ -0,0 +1,15 @@
+class AcceptGitHubInvitationJob < ApplicationJob
+  queue_as :high
+
+  def perform(repo_name)
+    return if hound_github.repository?(repo_name)
```
gylaz: Did you decide to keep this early return?

app/jobs/accept_github_invitation_job.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    hound_github.accept_invitation(repo_name)
+  end
+
+  private
+
+  def hound_github
+    @hound_github ||= GitHubApi.new(Hound::GITHUB_TOKEN)
```
gylaz: We've been also using `@_hound_github` naming convention for memoized ivars -- just a naming technique to signify that it's only for memoization.

app/services/repo_activator.rb
```diff
@@ -10,7 +10,9 @@ def initialize(github_token:, repo:)
   def activate
     change_repository_state_quietly do
       if repo.private?
-        add_hound_to_repo && create_webhook && repo.activate
+        add_hound_to_repo
+
+        create_webhook && repo.activate
```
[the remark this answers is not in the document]
gylaz: Oh interesting, the `else` is doing the same thing now. Nice catch Hound!

spec/jobs/accept_github_invitation_job_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "rails_helper"
+
+describe AcceptGitHubInvitationJob do
```
gylaz: Rspec now wants us to prefix top-level `describe`s like `RSpec.describe`. They're trying to deprecate adding `describe` to the global object.

spec/jobs/accept_github_invitation_job_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  it "is retryable" do
+    expect(AcceptGitHubInvitationJob.new).to be_a(Retryable)
+  end
+
+  it "queue_as high" do
+    expect(AcceptGitHubInvitationJob.new.queue_name).to eq("high")
+  end
+
+  describe "perform" do
```
gylaz: Should this be `#perform`?

spec/jobs/accept_github_invitation_job_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    context "when Hound has access the repo" do
```
gylaz: `has access to the repo`?

spec/jobs/accept_github_invitation_job_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      it "doesn't need to accept the invitation" do
+        github = double(GitHubApi, repository?: true, accept_invitation: nil)
```
gylaz: Using `instance_double` has the advantage to verify the double when the whole suite runs -- it's a verifying double.

Also, using the string version (like `instance_double("GitHubApi", ...)`) means the constant `GitHubApi` doesn't need to be loaded when running solo test, but will still get properly verified when the suite runs.

salbertson:
[See what we can do about missing invitations](https://trello.com/c/wikNleaU/837-see-what-we-can-do-about-missing-invitations)

app/jobs/accept_github_invitation_job.rb
salbertson: I cherry picked what I had in the other branch, so I'll update this to use `unless`.

spec/jobs/accept_github_invitation_job_spec.rb
salbertson: Are there deprecation warnings yet?

spec/jobs/accept_github_invitation_job_spec.rb
salbertson: Yes, done.

spec/jobs/accept_github_invitation_job_spec.rb
salbertson: Nice, I'll use `instance_double`. I'm going to leave the constant since we need it on the next line to stub `new`.

spec/jobs/accept_github_invitation_job_spec.rb
gylaz: Not yet, but I think we've been trying to do that on the project.

Here's more on it: https://relishapp.com/rspec/rspec-core/v/3-7/docs/configuration/global-namespace-dsl

PR #1502, opened 2018-03-24
Inherit .hound.yml from org-level repo

gylaz:
If org-level configuration is enabled, then all enabled repos that fall
under that org should respect and use its `.hound.yml` file.

### DIFF
--- a/app/models/hound_config.rb
+++ b/app/models/hound_config.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 class HoundConfig
+  CONFIG_FILE = ".hound.yml"
   LINTERS = {
     Linter::CoffeeScript => { default: true },
     Linter::Credo => { default: false },
@@ -19,17 +20,17 @@ class HoundConfig
     Linter::Swift => { default: true },
     Linter::Tslint => { default: false },
   }.freeze
-  CONFIG_FILE = ".hound.yml"
 
-  attr_reader_initialize :commit
+  attr_initialize [:commit!, :owner!]
+  attr_reader :commit
+  attr_private :owner
 
   def content
-    @_content ||= default_config.deep_merge(resolved_conflicts_config)
+    @_content ||= default_config.deep_merge(merged_config)
   end
 
   def linter_enabled?(name)
-    key = name.downcase
-    config = options_for(key)
+    config = options_for(name)
 
     !!config["enabled"]
   end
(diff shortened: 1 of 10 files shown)


gylaz:
[Org-level doesn't work without `.hound.yml` in the repo -- but should](https://trello.com/c/W7sx7U9Y/803-org-level-doesnt-work-without-houndyml-in-the-repo-but-should)

app/models/hound_config.rb
```diff
@@ -19,17 +20,17 @@ class HoundConfig
     Linter::Swift => { default: true },
     Linter::Tslint => { default: false },
   }.freeze
-  CONFIG_FILE = ".hound.yml"
 
-  attr_reader_initialize :commit
+  attr_initialize [:commit!, :owner!]
```
salbertson: I'm still not used to these helpers.

app/models/owner.rb
```diff
@@ -2,21 +2,13 @@ class Owner < ApplicationRecord
...
   rescue ActiveRecord::RecordNotUnique => exception
-    Raven.capture_exception(
-      exception,
-      extra: {
-        github_id: github_id,
-        name: name,
-      },
-    )
-
+    capture_exception(exception, name, github_id)
```
salbertson: `Raven.capture_exception` was distracting, or what?

spec/models/hound_config_spec.rb
```diff
@@ -39,101 +38,133 @@
...
+        commit = stub_commit(".hound.yml" => repo_hound_config)
+        owner = owner_stub(owner_hound_config)
+        hound_config = build_hound_config(commit, owner)
+
+        expect(hound_config).to be_enabled_for("flake8")
+      end
+    end
+
+    context "when linter enabled at owner-level but disabled at repo-level" do
+      it "returns true" do
```
salbertson: Returns true or false?

spec/models/hound_config_spec.rb
```diff
@@ -39,101 +38,133 @@
...
+        repo_hound_config = ""
+        commit = stub_commit(".hound.yml" => repo_hound_config)
+        owner = owner_stub(owner_hound_config)
+        hound_config = build_hound_config(commit, owner)
+
+        expect(hound_config).to be_enabled_for("flake8")
+      end
+    end
+
+    context "when linter enabled at owner-level but disabled at repo-level" do
```
salbertson: `when linter is enabled...`?

app/models/hound_config.rb
salbertson: What do you think about default argument values of `EmptyCommit.new` and `MissingOwner.new`?

app/models/owner.rb
gylaz: Yeah, big block of unimportant code...

app/models/hound_config.rb
gylaz: I don't think the default arguments would do us any good. Most of the time, we want the caller to pass to use something other than `EmptyCommit`. Just calling `HoundConfig.new` isn't really useful. WDYT?

spec/models/hound_config_spec.rb
gylaz: I did this consciously, since adding `is` pushes the line over 80 chars.

PR #1503, opened 2018-03-25
Update user's email on login

gylaz:


gylaz:
[Update user email when authenticating](https://trello.com/c/gQIQeqgJ/385-update-user-email-when-authenticating)

PR #1504, opened 2018-03-25
Rename Ruby related classes/files to RuboCop

gylaz:
As part of the move to name files after the linters.


gylaz:
[Upgrade Golint](https://trello.com/c/BK1ej5ux/822-upgrade-golint)

spec/services/resolve_config_aliases_spec.rb
```diff
@@ -6,11 +6,12 @@
       it "renames them to the appropriate linter" do
         config = {
           "javascript" => { "enabled" => false },
-          "ruby" => { "enabled" => false },
+          "ruby" => { "enabled" => true },
+          "flog" => { "enabled" => false },
```
salbertson: Why is `flog` included in the test?

gylaz: Just an arbitrary key I used in the test.

gylaz:
@salbertson I think `linters` will just pick it up off that queue.

PR #1505, opened 2018-03-26
Handle parsing empty JSON config

gylaz:
This can happen if a user provided a `config_file` entry, but the file
cannot be found at that path (in the repo on the branch).
Hound shouldn't error in such a case.

Addresses: https://sentry.io/hound-1/production/issues/400232909

### DIFF
--- a/spec/models/config/jshint_spec.rb
+++ b/spec/models/config/jshint_spec.rb
@@ -7,25 +7,35 @@ require "app/models/missing_owner"
 
 describe Config::Jshint do
   describe "#content" do
-    it "parses the configuration using JSON" do
-      raw_config = <<~EOS
-        {
-          "maxlen": 80
-        }
-      EOS
-      config = build_config(raw_config)
+    context "when configuration is valid JSON" do
+      it "parses the configuration using JSON" do
+        raw_config = <<~JSON
+          {
+            "maxlen": 80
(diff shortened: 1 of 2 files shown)


spec/models/config/jshint_spec.rb
```diff
@@ -7,25 +7,35 @@
...
 describe Config::Jshint do
   describe "#content" do
-    it "parses the configuration using JSON" do
-      raw_config = <<~EOS
-        {
-          "maxlen": 80
-        }
-      EOS
-      config = build_config(raw_config)
+    context "when configuration is valid JSON" do
```
salbertson: What if the config is present, but isn't actually valid JSON? What should we do?

salbertson:
[Look into frequent exceptions in Sentry](https://trello.com/c/kXazfotV/838-look-into-frequent-exceptions-in-sentry)

spec/models/config/jshint_spec.rb
gylaz: The way it works now is:
- an exception will be raised
- the exception will be caught and and we'll call `report_config_file_as_invalid` which will set the GitHub status with a failure mesage

salbertson: Nice, ok.

PR #1506, opened 2018-03-26
Use Intercom directly on window

salbertson:


salbertson:
Yea, not sure why, but I think it's because Intercom is loaded by Segment and not ready when the props are set up for the component.

PR #1509, opened 2018-03-28
Ensure we handle a no/empty config file

gylaz:
If the config file isn't found, or it's blank, for some reason, we
should treat it as an empty config and continue.

### DIFF
--- a/app/models/config/parser.rb
+++ b/app/models/config/parser.rb
@@ -1,7 +1,11 @@
 module Config
   class Parser
     def self.yaml(content)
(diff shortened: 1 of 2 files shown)


salbertson:
:shipit:

app/models/config/parser.rb
```diff
@@ -15,9 +15,5 @@ def self.json(content)
     def self.ini(content)
       IniFile.new(content: content).to_h
     end
-
-    def self.raw(content)
-      content
-    end
```
gylaz: This wasn't used anywhere.

PR #1510, opened 2018-03-29
Remove unused ReportInvalidConfigJob and documentation

salbertson:


CONTRIBUTING.md
```diff
@@ -229,9 +229,6 @@ Once linting is complete, resulting violations should be posted to the outbound
 * `patch` - The patch content from GitHub for the file being reviewed. This is
   provided by the inbound queue.
 
-If the given `config` is invalid, the invalid file should be posted to the
```
salbertson: @gylaz what should we replace this with, any ideas?

gylaz: I think we can just remove this. `linters` handles any config errors through normal job reporting means.

PR #1515, opened 2018-04-04
Report config file parse errors

salbertson:
* Be specific and set status error referencing the file name.

### DIFF
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -7,7 +7,9 @@ module Config
 
     def content
       @content ||= owner_config.deep_merge(load)
-    rescue ConfigContent::ContentError, Config::ParserError
+    rescue ConfigContent::ContentError => exception
+      raise_parse_error(exception.message)
+    rescue Config::ParserError
       raise_parse_error("#{file_path} format is invalid")
     end
 
@@ -40,7 +42,7 @@ module Config
     end
 
     def raise_parse_error(message)
-      raise Config::ParserError.new(message, linter_name: linter_name)
+      raise Config::ParserError, message
     end
 
     def file_path
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -39,6 +39,12 @@ module Config
       Parser.yaml(content)
     end
 
+    def rescue_and_raise_parse_error
+      yield if block_given?
+    rescue StandardError
+      raise_parse_error("#{file_path} format is invalid")
+    end
+
     def raise_parse_error(message)
       raise Config::ParserError.new(message, linter_name: linter_name)
     end
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -7,9 +7,7 @@ module Config
 
     def content
       @content ||= owner_config.deep_merge(load)
-    rescue ConfigContent::ContentError => exception
-      raise_parse_error(exception.message)
-    rescue Config::ParserError
+    rescue ConfigContent::ContentError, Config::ParserError
       raise_parse_error("#{file_path} format is invalid")
     end
 
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -9,6 +9,8 @@ module Config
       @content ||= owner_config.deep_merge(load)
     rescue ConfigContent::ContentError => exception
       raise_parse_error(exception.message)
+    rescue Config::ParserError
+      raise_parse_error("#{file_path} format is invalid")
     end
 
     def serialize(data = content)
@@ -36,15 +38,7 @@ module Config
     end
 
     def parse(content)
-      rescue_and_raise_parse_error do
-        Parser.yaml(content)
-      end
-    end
-
-    def rescue_and_raise_parse_error
-      yield
-    rescue Config::ParserError
-      raise_parse_error("#{file_path} format is invalid")
+      Parser.yaml(content)
     end
 
     def raise_parse_error(message)
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -42,8 +42,8 @@ module Config
     end
 
     def rescue_and_raise_parse_error
-      yield if block_given?
-    rescue StandardError
+      yield
+    rescue Config::ParserError
       raise_parse_error("#{file_path} format is invalid")
     end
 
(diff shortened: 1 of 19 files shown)


salbertson:
[Look into frequent exceptions in Sentry](https://trello.com/c/kXazfotV/838-look-into-frequent-exceptions-in-sentry)

app/models/config/base.rb
```diff
@@ -36,7 +36,15 @@ def owner_config
...
 
     def parse(content)
-      Parser.yaml(content)
+      rescue_and_raise_parse_error do
+        Parser.yaml(content)
+      end
+    end
+
+    def rescue_and_raise_parse_error
+      yield if block_given?
```
gylaz: Looks like block is always given. What do you think about removing `if block_given?` to make it always required?

app/models/config/base.rb
```diff
@@ -36,7 +36,15 @@ def owner_config
...
+    rescue StandardError
```
gylaz: What do you think about catching errors in the `Config::Parser` class, in each method? That way each method can define what parse error to catch, rather than a generic `StandardError`? Then we could raise `ConfigContent::ContentError`, since it's already being caught and handled in `Config::Base`.

spec/models/config/flake8_spec.rb
```diff
@@ -64,6 +64,26 @@
...
+            docs,
+            migrations
+            tests
+            management
+        CONFIG
+        config = build_config(raw_config)
+
+        expect { config.content }.to(
+          raise_error(Config::ParserError),
+        )
```
gylaz: Haven't seen this style of line breaking on this project (`to` with parens). We would typically do:
```rb
expect { config.content }.to raise_error(
  Config::ParserError
)
# or
expect { config.content }.
  to raise_error(Config::ParserError)
```

app/models/config/base.rb
salbertson: I thought about that, but we don't know the filename for message if we raise in `Config::Parser`.

spec/models/config/flake8_spec.rb
salbertson: Works for me. 👍

app/models/config/base.rb
salbertson: We could rescue the specific errors in `Config::Parser` and raise `Config::ParserError` without the message and filename, then rescue and re-raise the same exception with the message and filename? That would avoid using `StandardError`.

salbertson: Check out the last [commit](https://github.com/houndci/hound/pull/1515/commits/7b887b30b9160fcc820977f462a67d077f688545) and let me know if that makes more sense. Neither approache is great, maybe there's a better approach.

app/models/config/base.rb
```diff
@@ -36,7 +36,15 @@ def owner_config
...
     def parse(content)
-      Parser.yaml(content)
+      rescue_and_raise_parse_error do
+        Parser.yaml(content)
+      end
+    end
+
+    def rescue_and_raise_parse_error
+      yield
+    rescue Config::ParserError
```
gylaz: What do you think about, instead of having this method, adding `Config::ParserError` to the `rescue` clause of line 10 of this file, and using this updated message there? This way we don't need to wrap our calls with `rescue_and_raise_parse_error`, and any parse errors will be caught and reported as invalid config.

salbertson: Great idea, I like that.

app/models/config/base.rb
```diff
@@ -9,6 +9,8 @@ def content
       @content ||= owner_config.deep_merge(load)
     rescue ConfigContent::ContentError => exception
       raise_parse_error(exception.message)
+    rescue Config::ParserError
+      raise_parse_error("#{file_path} format is invalid")
```
gylaz: Maybe just do this?
```rb
rescue ConfigContent::ContentError, Config::ParserError
  raise_parse_error("#{file_path} format is invalid")
```
I think this message is much better than anything we raised before, and both situations means config is invalid format.

salbertson: @gylaz it looks like `ContentError` is used for other cases, not just parse errors.

https://github.com/houndci/hound/blob/bf2c9a67150861d1d73781b68791b7fd59522292/app/models/config_content/remote.rb#L15

If we rescue the same way for both I think we lose messages like "404 Could not find resource".

gylaz: Honestly, I don't think anyone is using that. We added it, but didn't advertise it, then moved to org-wide repos. I'm fine with doing as-is. Then maybe killing this remote business and doing what I proposed later?

salbertson: Ok, I think I've straightened out the use of `ContentError` and `ParserError`, it feels like they are being used appropriately. I removed the requirement for `linter_name` since it wasn't being used anymore, and we want to use the exception for things without a linter name, like `HoundConfig`.

@gylaz let me know what you think.

app/models/config/parser.rb
```diff
@@ -6,6 +6,8 @@ def self.yaml(content)
       else
         {}
       end
+    rescue Psych::SyntaxError
+      raise Config::ParserError
```
salbertson: We could do something with the original message, but I'm not sure what.

app/models/config/base.rb
salbertson: Will there be a `file_path` when `HoundConfig` raises `ParserError`?

gylaz: Fine for now. I still think there should only be one exception -- we cannot parse this format -- but we can get to it later.

app/models/hound_config.rb
```diff
@@ -70,6 +70,8 @@ def parsed_config
 
   def parse(file_content)
     Config::Parser.yaml(file_content) || {}
+  rescue Config::ParserError
+    raise Config::ParserError.new("#{CONFIG_FILE} format is invalid")
```
salbertson: These should be surfaced in the same way, reported to the user as a GitHub status message.

app/models/config/parser.rb
gylaz: Maybe, "Invalid YAML format" or "Invalid JSON format", or ini, etc?

salbertson: Yep, totally. We can iterate on it.

PR #1520, opened 2018-04-07
Enable Reek, Flog and Flay

gylaz:
We should eat our own dog food.


salbertson:
🐕

gylaz:
Ha, j/k, we don't even use Flay. I'll remove that.

PR #1521, opened 2018-04-07
Rename Go to Golint

gylaz:
Point Golint at the `linters` queue, which is handled by the `linters`
service.

This is a companion to https://github.com/houndci/linters/pull/188


salbertson:
Looks good to me. 👍

PR #1524, opened 2018-04-10
Verify Stripe webhook requests

salbertson:


app/controllers/deleted_subscriptions_controller.rb
```diff
@@ -2,8 +2,18 @@ class DeletedSubscriptionsController < ApplicationController
...
+    signature = request.env["HTTP_STRIPE_SIGNATURE"]
 
+    Stripe::Webhook.construct_event(
+      payload,
+      signature,
+      ENV.fetch("GITHUB_WEBHOOK_SECRET"),
+    )
+  rescue JSON::ParserError, Stripe::SignatureVerificationError
+    head 400
+  else
```
gylaz: Haven't seen that much use of `else` with `rescue`. What do you think about putting the contents on this `else` in to the happy-path flow instead?

Issue #1511, opened 2018-03-30
Allow setting `public_file_server` env value in production


zuzannast:
Right now, the `config/environments/production.rb` hardcodes `public_file_server` value to false. I'd like to be able to set this value with the environment variable as in some cases I want it to be true. 
Let me know what you think about it I'll make a PR for that.

gylaz:
Sure, I don't mind. We could do something like `config.public_file_server.enabled = ENV.fetch('SERVE_STATIC', false)` to allow for not setting the env value.

PR #1525, opened 2018-04-11
Allow setting serve static files with env

zuzannast:
I'm sending the PR as discussed in #1511.


gylaz:
Merged. Thanks @zuzannast.

PR #1541, opened 2018-05-21
Add support for linting .erb.html files.

yxhuvud:
Note: Something that would greatly help developing new linters is to have a dry-run mode which doesn't require github integration. Something as simple as a raketask that you can invoke with a path to a gitrepo and a refspec/branch, and see that the integration work would be golden. Sadly I don't have enough understanding of this codebase to implement that.


gylaz:
Thanks @yxhuvud! We'll take a look at this to get this merged in.

salbertson:
@gylaz are we planning to merge this, or should I close it?

gylaz:
We could add this as linter, but first we need to get it working over in `linters`.

PR #1549, opened 2018-06-14
Use status to notify customers when their payment fails

salbertson:
screenshot 2018-06-14 13 59 13

### DIFF
--- a/app/services/build_runner.rb
+++ b/app/services/build_runner.rb
@@ -29,14 +29,12 @@ class BuildRunner
 
     if repo.private? && owner.past_due?
       commit_status.set_past_due_status(
-        owner.most_recent_invoice.hosted_invoice_url
+        owner.most_recent_invoice.hosted_invoice_url,
       )
     else
       build = create_build
       review_files(build)
-      if build.file_reviews.empty?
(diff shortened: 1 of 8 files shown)


app/services/build_runner.rb
```diff
@@ -25,11 +25,16 @@ def set_internal_error
...
+    owner = upsert_owner
+
+    if repo.private? && owner.past_due?
+      commit_status.set_past_due_status(
+        owner.most_recent_invoice.hosted_invoice_url,
+      )
+    else
+      build = create_build
+      review_files(build)
+      set_no_violations_status if build.file_reviews.empty?
```
gylaz: What's up with this? We've preferred explicit `if` branches over condition modifiers. Any reason to break that consistency?

salbertson: I thought Hound was suggesting a single line. Maybe we should disable the rule.

Should I change it back?

> Favor modifier if usage when having a single-line body.

gylaz: Was there a RuboCop violation with that text? Yeah, I think we should rollback and configure the rule.

PR #1551, opened 2018-06-16
Update webpacker, selenium-driver, and JS test libs

gylaz:
Experienced some issues locally with Webpacker, and all resources that I
could find were recommending upgrading. Seemed to solve my issue.


gylaz:
Thanks. Will test it out on staging next.

PR #1557, opened 2018-07-13
Get the latest past due invoice

gylaz:


app/models/owner.rb
```diff
@@ -52,6 +52,7 @@ def recent_invoice_url
     if stripe_subscription_id
       Stripe::Invoice.
         list(subscription: stripe_subscription_id).
+        sort_by { |invoice| -invoice.date }.
```
salbertson: Is it worth testing this?

gylaz: Well, since _someone_ didn't write any tests for this method at all, I'd say not worth it 😄

PR #1559, opened 2018-07-18
Update links to our docs

salbertson:
Configuration instructions have been moved to our help docs.


app/views/application/_app_nav.haml
```diff
@@ -6,7 +6,7 @@
         %use{ "xlink:href" => "/images/icons.svg#hound" }
 
     .app-nav__app-items
-      = link_to "Docs", configuration_path, class: "app-nav__item"
+      = link_to "Docs", ENV.fetch("DOCS_URL"), class: "app-nav__item"
       = link_to "Help", ENV.fetch("HELP_URL"), class: "app-nav__item"
```
gylaz: What do you think about just having the `Docs` link and pointing it at `https://intercom.help/hound`? This way, we don't have unnecessary, redundant links in our menu?

spec/services/build_runner_spec.rb
```diff
@@ -378,6 +378,6 @@ def invalid_config_file(pull_request, stubs = {})
   end
 
   def config_url
```
gylaz: How about just combining these two methods in this this test file?

salbertson: Good call, one wasn't even used.

app/views/application/_app_nav.haml
salbertson: I'd rather leave both, at least for now. Docs goes to our configuration articles and help is more general.

gylaz: I just don't see a difference in meaning of "Docs" and "Help". Maybe if we split them up as "Config" and "FAQ". Otherwise, it doesn't really tell the user where they should be navigating to find an answer to what they're looking for.

If we just linked them to Docs/Help, then it's just another click-through to get to where they need (with helpful subtitles).

PR #1560, opened 2018-07-23
Surface GitHub collaborator seat error

salbertson:


lib/error_message_translation.rb
```diff
@@ -7,6 +7,8 @@ def self.from_error_response(error)
 
     if matches.present? && matches.captures[0] == FORBIDDEN_ERROR_CODE
       matches.captures[1]
+    elsif error.message.include? "You must purchase at least one more seat"
+      "You need to add at least one more seat to GitHub before enabling Hound. https://help.github.com/articles/adding-seats-to-your-organization"
```
gylaz: Maybe shorten this to just `Please add a seat to GitHub to enable Hound. [link]`

spec/lib/error_message_translation_spec.rb
```diff
@@ -45,4 +57,14 @@ def octokit_403_error_message
   def octokit_400_error_message
     "PUT https://api.github.com/teams/3675/memberships/houndci: 400 - Problems parsing JSON. // See: https://developer.github.com/v3"
   end
+
+  def octokit_422_error_message
+<<-ERROR
```
gylaz: How about using `<<~` heredoc, which allows you to indent everything properly?

lib/error_message_translation.rb
gylaz: Also maybe use a hereblock.

salbertson:
@gylaz I went through and addressed the style comments. This should be ready now.

spec/lib/error_message_translation_spec.rb
```diff
@@ -34,6 +34,18 @@
...
       end
     end
+
+    context "when error related to adding more GitHub seats" do
+      it "returns a helpful message" do
+        error = double("error", message: octokit_422_error_message)
+
+        result = ErrorMessageTranslation.from_error_response(error)
+
+        expect(result).to(
```
gylaz: Have you been doing this recently? I haven't seen much usage of `.to(` in RSpec. Typically I see this type of line splitting:
```rb
expect(result).
  to eq(...)
```
Thoughts?

spec/lib/error_message_translation_spec.rb
```diff
@@ -34,6 +34,18 @@
...
+          eq("Please add a GitHub seat to enable Hound. https://help.github.com/articles/adding-seats-to-your-organization"),
```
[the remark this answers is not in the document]
gylaz: Want to break this line up to appease Hound?

PR #1566, opened 2018-08-09
Redirect Marketplace users to GitHub when viewing Account

salbertson:


app/controllers/accounts_controller.rb
```diff
@@ -1,4 +1,5 @@
 class AccountsController < ApplicationController
+  before_action :redirect_marketplace_users
```
gylaz: How about throwing an `only: :show` on here?

PR #1568, opened 2018-08-16
Show GitHub Marketplace plans on Account page

salbertson:
Support upgrading and downgrading GitHub Marketplace plans in Hound.

screenshot 2018-08-15 16 27 40
screenshot 2018-08-15 16 27 40 2


app/controllers/plans_controller.rb
```diff
@@ -1,16 +1,38 @@
 class PlansController < ApplicationController
+  MARKETPLACE_URL = ENV.fetch(
+    "MARKETPLACE_URL",
+    "https://www.github.com/marketplace/hound"
```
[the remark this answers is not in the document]
gylaz: Do you imagine this might change?

app/controllers/plans_controller.rb
```diff
@@ -1,16 +1,38 @@
...
+
   def index
     @plans = ActiveModel::ArraySerializer.new(
-      Plan.all,
+      plan_selector.all,
       each_serializer: PlanSerializer,
       scope: current_user,
     )
-    @repo = Repo.find(plan_params[:repo_id])
+    @repo = repo
```
gylaz: Kinda awkward that `repo` sets the ivar too, but I understand.

app/models/github_plan.rb
```diff
@@ -0,0 +1,60 @@
...
+  #   {
+  #     id: 1063,
+  #     price: 149,
+  #     range: 5..20,
+  #     title: "Octodog",
+  #     slug: "MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW4xMDYz"
+  #   },
+  # ].freeze
+
+  # staging
```
gylaz: This looks like want to use a YAML config file with `staging, prod, etc` plans.

app/views/application/_app_nav.haml
```diff
@@ -6,6 +6,11 @@
         %use{ "xlink:href" => "/images/icons.svg#hound" }
 
     .app-nav__app-items
+      - if signed_in?
+        = link_to "Repos",
+          repos_path,
+          class: "app-nav__item"
```
gylaz: One line?

app/controllers/plans_controller.rb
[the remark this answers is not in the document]
salbertson: I doubt it.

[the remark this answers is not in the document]
salbertson: Actually, it will need to change when developing or testing on staging.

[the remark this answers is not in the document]
gylaz: Maybe config file (per env) then? Doesn't seem like sensitive data, and not likely to change within a specific env.

app/controllers/plans_controller.rb
```diff
@@ -1,16 +1,42 @@
...
+    @marketplace_upgrade_url = marketplace_upgrade_url
+
+    # Should we be finding the current owner's plan based on repo and
+    # basing upgrade/downgrade on that?
```
gylaz: Yes

app/controllers/plans_controller.rb
```diff
@@ -1,16 +1,42 @@
...
   def plan_params
     params.permit(:repo_id)
   end
+
+  def plan_selector
+    PlanSelector.new(current_user, repo: repo)
+  end
+
+  def marketplace_upgrade_url
+    # Should we include this in the serialized plans instead?
```
gylaz: That'd be good.

app/controllers/subscriptions_controller.rb
```diff
@@ -5,24 +5,21 @@ class FailedToActivate < StandardError; end
...
-        render(
-          json: { upgrade_url: repo.owner.upgrade_url },
-          status: :forbidden,
-        )
-      else
-        activate
-        render json: repo, status: :created
-      end
-    elsif current_user.plan_upgrade?
+    plan_selector = PlanSelector.new(current_user, repo: repo)
```
gylaz: How about making both of these kwargs?

app/controllers/subscriptions_controller.rb
```diff
@@ -5,24 +5,21 @@ class FailedToActivate < StandardError; end
...
       render json: {}, status: :payment_required
+    elsif plan_selector.marketplace_plan?
+      activate
     else
       activate_and_create_subscription
     end
   end
 
   def update
+    # Do we need Marketplace logic here too? Just #activate?
```
gylaz: Yes, I think we need to do something special here too. At least, make sure that upgrading a Marketplace plan works as it should.

app/models/plan_selector.rb
```diff
@@ -1,39 +1,91 @@
...
   end
 
   def current_plan
-    find_plan(repo_count)
+    if marketplace_plan?
+      plans.detect do |plan|
+        plan.id == marketplace_plan_id
+      end
+    else
+      plans.detect do |plan|
```
gylaz: Use `find_plan_by_active_repo_count` here?

app/models/plan_selector.rb
```diff
@@ -1,39 +1,91 @@
...
   end
 
   def previous_plan
-    find_plan(repo_count.pred)
+    find_plan_by_active_repo_count(active_repo_count.pred)
   end
 
-  def upgrade?
-    current_plan != next_plan
+  def all
```
gylaz: I think `plans` is a better name since this class is called `PlanSelector` (`all` implies you might be returning all plan selectors).

app/models/plan_selector.rb
```diff
@@ -1,39 +1,91 @@
...
+    owner_marketplace_plan_id || first_user_marketplace_plan_id
+  end
+
+  def owner_marketplace_plan_id
+    repo&.owner&.marketplace_plan_id
+  end
 
-  def repo_count
-    @_repo_count ||= repos.size
+  def first_user_marketplace_plan_id
```
gylaz: Do we need this? Won't we always have a repo (that we're trying to activate) and thus an owner?

app/models/plan_selector.rb
```diff
@@ -1,39 +1,91 @@
...
+    find_plan_by_active_repo_count(active_repo_count.pred)
   end
 
-  def upgrade?
-    current_plan != next_plan
+  def all
+    plans
+  end
+
+  def marketplace_plan?
```
gylaz: This should be as simple as `!!repo.owner.marketplace_plan_id`

app/models/plan_selector.rb
```diff
@@ -1,39 +1,91 @@
 class PlanSelector
   BULK_ID = "bulk".freeze
 
-  def initialize(user)
+  def initialize(user, repo: nil)
```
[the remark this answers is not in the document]
gylaz: Thinking again on this, `repo` might be all we need. A user takes an action to activate a repo to reach this page. The only other place is the `Account` page, in which case we can just pass in the first subscribed private repo.

app/models/plan_selector.rb
salbertson: In some cases we don't have a `repo` when using `PlanSelector`, so I think we have to fall back to checking if the user is associated with any repos that have "marketplace" owners.

app/models/plan_selector.rb
salbertson: We won't always have a repo, `PlanSelector` is used in some cases without a `repo`.

app/models/plan_selector.rb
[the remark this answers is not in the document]
salbertson: If we can depend on a repo always being provided that would be really nice. I'll take a swing at it.

app/controllers/plans_controller.rb
```diff
@@ -1,16 +1,42 @@
...
+    if plan_selector.marketplace_plan?
+      # Should we include the username or org name for upgrading on Marketplace?
```
gylaz: I think it's always the owner's name.

app/javascript/components/NotifyTierChange/components/UpgradeSubscriptionLink.jsx
```diff
@@ -73,16 +73,22 @@ export default class UpgradeSubscriptionLink extends React.Component {
   }
 
   handleClick() {
-    this.setState({ disabled: true });
+    if (this.props.marketplaceUpgradeUrl) {
+      // improve this flow so we end up back at houndci.com
```
gylaz: Did we ever ask GitHub if there was a way to do this?

app/models/plan_selector.rb
```diff
@@ -1,39 +1,72 @@
...
   end
 
   private
 
   attr_reader :user
 
-  def repo_count
-    @_repo_count ||= repos.size
+  def repo
+    @repo || user.repos.active.first || user.repos.first
```
salbertson: @gylaz I moved this logic into `PlanSelector` instead of accessing the first repo when initializing. Repo subscription records aren't being used because Marketplace users don't have those records, which might be an issue.

app/controllers/subscriptions_controller.rb
salbertson: I take care of this when testing upgrade, downgrade and cancel flows for Marketplace.

app/models/user.rb
```diff
@@ -11,16 +11,9 @@ class User < ApplicationRecord
 
   before_create :generate_remember_token
 
+  # Use PlanSelector whenever possible!
```
gylaz: Remove?

app/models/user.rb
```diff
@@ -84,6 +73,19 @@ def card_exists?
...
+    if plan_selector.marketplace_plan?
+      # This assumes a user manages one Marketplace purchase.
+      first_available_repo.owner.repos.active.where(private: true)
+    else
+      super
+    end
+  end
+
+  def first_available_repo
+    repos.active.first || repos.first
```
gylaz: `repos.order(:active).first`?

spec/models/plan_selector_spec.rb
```diff
@@ -1,70 +1,139 @@
...
+      )
+      repo = instance_double(
+        "Repo",
+        owner: double.as_null_object
+      )
+      plan_selector = PlanSelector.new(user: user, repo: repo)
 
-      expect(plan_selector.current_plan).to eq Plan.new(Plan::PLANS[1])
+      expect(plan_selector.current_plan).to(
+        eq StripePlan.new(StripePlan::PLANS[1])
```
gylaz: I don't think I've ever seen `.to(...)` usage before. How about:
```rb
expect(plan_selector.current_plan).
  to eq StripePlan.new(StripePlan::PLANS[1])
```

spec/presenters/account_page_spec.rb
```diff
@@ -46,10 +46,13 @@
...
       presenter = instance_double("PlanPresenter")
-      plan = instance_double("Plan")
+      plan = instance_double("StripePlan")
       user = instance_double("User")
       page = AccountPage.new(user)
-      allow(Plan).to receive(:all).once.with(no_args).and_return([plan])
+      plan_selector = instance_double("PlanSelector", plans: [plan])
+      allow(PlanSelector).to(
+        receive(:new).once.with(user: user).and_return(plan_selector)
+      )
```
gylaz: Same here, maybe break up on a period?

app/models/plan_selector.rb
```diff
@@ -1,39 +1,68 @@
 class PlanSelector
   BULK_ID = "bulk".freeze
 
-  def initialize(user)
-    @user = user
+  def initialize(user:, repo: nil)
+    @user = user || User.new
```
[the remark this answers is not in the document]
salbertson: I think tests were initializing `PlanSelector` with a nil user for some reason. I can look into it to get rid of the `User.new`.

app/models/plan_selector.rb
```diff
@@ -1,39 +1,68 @@
 class PlanSelector
   BULK_ID = "bulk".freeze
 
-  def initialize(user)
-    @user = user
+  def initialize(user:, repo: nil)
```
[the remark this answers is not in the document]
salbertson: Sure, we could. Do you think it's unclear the way it's set up?

app/models/plan_selector.rb
```diff
@@ -1,39 +1,68 @@
...
+    @user = user || User.new
+    @repo = repo
   end
 
   def current_plan
-    find_plan(repo_count)
+    if marketplace_plan?
+      plans.detect do |plan|
+        plan.id == marketplace_plan_id
+      end
```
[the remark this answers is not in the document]
salbertson: I like it, done. 🍰

app/models/user.rb
salbertson: I figured that we would see this comment when reaching for delegate methods and use `PlanSelector` instead.

app/models/plan_selector.rb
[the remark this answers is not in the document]
gylaz: I just think it's clearer in the constructor what the default value is if nothing is passed.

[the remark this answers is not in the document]
salbertson: Cool, sounds good to me.

app/models/plan_selector.rb
```diff
@@ -1,39 +1,62 @@
 class PlanSelector
   BULK_ID = "bulk".freeze
 
-  def initialize(user)
+  def initialize(user:, repo: nil)
     @user = user
+    @repo = repo || user&.first_available_repo
```
gylaz: Is there a situation where `user` can be `nil`?

app/models/plan_selector.rb
```diff
@@ -1,39 +1,62 @@
...
+      GitHubPlan
+    else
+      StripePlan
+    end
   end
 
-  def find_plan(count)
-    Plan.find_by(count: count)
+  def marketplace_plan_id
+    repo&.owner&.marketplace_plan_id
```
gylaz: Do we need to be this defensive here? Aren't we always going to either have a repo passed in, or grab the first user's repo?

app/models/plan_selector.rb
salbertson: Yes, there are older test cases where `user` is `nil`.

app/models/plan_selector.rb
salbertson: I think we do, to support users without any repos.

app/models/plan_selector.rb
gylaz: No, I mean are there practical situations where we would not have a user?
My guess is there aren't as we'll only ever need the plan once a user is signed in.
If so, the old test cases should be updated or removed.

salbertson: Yes, there are cases where `user` is `nil`. For example, the "home" page and `Home` class are used for unauthenticated users and instances of `Home` use the `PlanSelector`. I added [this](https://github.com/houndci/hound/pull/1568/commits/3c049fe80d9a8759789885275b902bf589eeabfa) to make things more clear, let me know if it helps.

gylaz: Gotcha. Yeah, that's nice!

PR #1570, opened 2018-09-01
Use SwiftLint linter service

gylaz:
Use our Linux/Heroku service instead of MacStadium


gylaz:
Verified on staging via https://github.com/houndci/test-sandbox/pull/18

PR #1571, opened 2018-09-10
Show searchbar on its own line

gylaz:
The two buttons aren't related to the search, and they don't need to be
pressed for the search to happen. Thus, I think it makes sense to have
the searchbar live on its own line.

Also, don't hide "Include private repos" button on small screens
(phones).

toolbar


salbertson:
@gylaz one question, what happens when private repos are added and that button isn't shown any longer?

gylaz:
@salbertson Then "Refresh Repo List" takes up the whole width.

PR #1575, opened 2018-09-15
Ensure presence of current org-wide config option

gylaz:
If the current org-wide config repository is not in the list of the
repos (for whatever reason), it won't be included in the dropdown list
of possible repos for org-wide configuration. This will cause us to
select another/different repo for org-wide config automatically,
effectively changing the org-wide config on the user.

This ensures the current org-wide config is always in that list.

### DIFF
--- a/app/javascript/components/ReposContainer/components/Organization/OrganizationConfiguration.jsx
+++ b/app/javascript/components/ReposContainer/components/Organization/OrganizationConfiguration.jsx
@@ -10,11 +10,7 @@ export default class OrganizationConfiguration extends React.Component {
   constructor(props) {
     super(props);
 
-    let repo = "";
-
-    if (props.repo) {
-      repo = props.repo;
-    }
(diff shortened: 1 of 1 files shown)


app/javascript/components/ReposContainer/components/Organization/OrganizationConfiguration.jsx
```diff
@@ -10,11 +10,7 @@ export default class OrganizationConfiguration extends React.Component {
   constructor(props) {
     super(props);
 
-    let repo = "";
-
-    if (props.repo) {
-      repo = props.repo;
-    }
+    const repo = props.repo ? props.repo : "";
```
salbertson: What do you think about this?

`const repo = props.repo || '';`

gylaz: Ah, yes!

PR #1576, opened 2018-09-15
Save installation_id on setup

gylaz:
We need to configure our GitHub app to navigate to `/setup` path, after
they install our app (GitHub provides this). Then we will save their
`installation_id` and redirect them to `/repos`.


app/controllers/setups_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      current_user.update(installation_ids: ids, repos: [])
+    end
+
+    redirect_to repos_path
+  end
+
+  private
+
+  def save_installation_id
+    if params[:installation_id]
```
salbertson: When will the param be present? Do we expect the redirect to have the `installation_id` param?

Gemfile.lock
```diff
@@ -194,7 +194,7 @@ GEM
       multi_json (~> 1.3)
       multi_xml (~> 0.5)
       rack (>= 1.2, < 3)
-    octokit (4.9.0)
+    octokit (4.12.0)
```
salbertson: What does the new version give us? Anything good?

app/controllers/setups_controller.rb
```diff
@@ -0,0 +1,20 @@
+class SetupsController < ApplicationController
+  prepend_before_action :save_installation_id
```
salbertson: I wouldn't mind all of this happening in `SessionsController`.

app/controllers/setups_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def show
+    if session[:installation_id]
+      ids = current_user.installation_ids | [session[:installation_id].to_i]
+      current_user.update(installation_ids: ids, repos: [])
```
salbertson: Should we raise if this update doesn't succeed?

app/controllers/setups_controller.rb
gylaz: This is the initial redirect from GitHub, which will have the `installation_id`. This is why we save it before any other before-actions.

Gemfile.lock
gylaz: Nothing interesting that I noticed, but more up-to-date endpoints.

app/controllers/setups_controller.rb
gylaz: I don't see how it would work. There are two paths -- one a user is re-directed to a login, which gets `POST`ed to our callback. Another path is GitHub redirects to`/setup` via `GET`.

app/controllers/setups_controller.rb
gylaz: Yes, I originally was deciding between `/setup` and `/installation`. Setup seemed friendlier and matched what GitHub calls the field. Is there any particular way you prefer `Installation` over `Setup`?

salbertson: Just thought it was a little vague. What about `/github_setup`?

PR #1578, opened 2018-09-21
Disable repos when a Marketplace subscription is cancelled

salbertson:


spec/models/github_event_spec.rb
```diff
@@ -48,5 +48,24 @@
         end
       end
     end
+
+    context "for a cancellation" do
+      it "disables the owner's active private repos" do
+        owner = create(:owner, github_id: 123456, name: "salbertson")
+        public_repo = create(:repo, :active, owner: owner)
+        private_repo = create(:repo, :active, :private, owner: owner)
+
```
gylaz: ✂️ ?

PR #1579, opened 2018-09-21
Hide private repos button for App users

salbertson:


app/models/user.rb
```diff
@@ -50,8 +50,10 @@ def token
   end
 
   def has_access_to_private_repos?
-    if token_scopes
-      token_scopes.split(",").include? "repo"
+    if token_scopes && token_scopes.split(",").include?("repo")
+      true
+    elsif installation_ids.any?
+      true
```
gylaz: Seems like this method should just be:
```rb
token_scopes&.split(",")&.include?("repo") || installation_ids.any?
```

PR #1583, opened 2018-09-26
Add a rake job to periodically sync GitHub plans

gylaz:
This is a requirement by GitHub, so we stay in sync with them if we miss
a webhook.


lib/tasks/plan.rake
```diff
@@ -0,0 +1,6 @@
+namespace :plan do
+  desc "Reset refreshing_repos flag for stuck users"
```
salbertson: Is this an outdated task description?

lib/tasks/plan.rake
```diff
@@ -0,0 +1,6 @@
...
+  task update_all: :environment do
+    UpdatePlans.call
```
salbertson: Should this be `UpdateGitHubPlans`?

lib/tasks/plan.rake
gylaz: Yes, copy pasta.

lib/tasks/plan.rake
gylaz: 👍

PR #1587, opened 2018-10-02
Listen for correct Marketplace cancellation event

salbertson:
https://developer.github.com/v3/activity/events/types/#marketplacepurchaseevent


app/models/github_event.rb
```diff
@@ -47,6 +41,11 @@ def update_purchase
         marketplace_plan_id: body["marketplace_purchase"]["plan"]["id"],
       )
     when "cancelled"
+      owner = Owner.find_by!(
+        github_id: body["marketplace_purchase"]["account"]["id"],
+      )
```
gylaz: We probably don't need to do a `find_by` again. The `upsert` should already return the right owner.

app/models/github_event.rb
```diff
@@ -47,6 +41,11 @@ def update_purchase
...
+      owner.active_private_repos.each(&:deactivate)
```
gylaz: Do we maybe want to deactivate all the repos? Including public? Or does this event only apply to cancelling paid subscriptions?

salbertson: Good question. Nope, this event is sent for any plan.

I think it's safe for us to assume all repos active under the owner would be tied to the purchase. If they cancel, regardless of plan, they probably want to disable Hound for everything.

app/models/github_event.rb
```diff
@@ -47,6 +41,7 @@ def update_purchase
         marketplace_plan_id: body["marketplace_purchase"]["plan"]["id"],
       )
     when "cancelled"
+      owner.active_private_repos.update_all(active: false)
```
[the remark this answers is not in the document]
salbertson: @gylaz thoughts on `update_all`? Think it's worth avoiding individual updates?

PR #1588, opened 2018-10-02
Handle App uninstallation and updates

salbertson:
https://developer.github.com/v3/activity/events/types/#installationrepositoriesevent

### DIFF
--- a/app/models/github_event.rb
+++ b/app/models/github_event.rb
@@ -64,10 +64,11 @@ class GitHubEvent
       repos = Repo.where(installation_id: body["installation"]["id"])
       repos.update_all(active: false, installation_id: nil)
 
-      User.where(
+      users = User.where(
         "? = ANY(installation_ids)",
-        body["installation"]["id"]
-      ).find_each do |user|
+        body["installation"]["id"],
+      )
+      users.find_each do |user|
         user.installation_ids.delete(body["installation"]["id"])
         user.save!
       end
--- a/app/models/github_event.rb
+++ b/app/models/github_event.rb
@@ -26,6 +26,8 @@ class GitHubEvent
       owner.active_private_repos.each(&:deactivate)
     when INSTALLATION
       update_installation
+    when INSTALLATION_REPOSITORIES
+      update_installation
     else
       Rails.logger.info("Unhandled GitHub event: #{type} -- #{action}")
     end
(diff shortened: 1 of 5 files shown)


spec/models/github_event_spec.rb
```diff
@@ -66,5 +66,105 @@
...
+          type: GitHubEvent::INSTALLATION,
+          body: body,
+        )
+
+        event.process
+
+        expect(repo.reload.installation_id).to be_nil
+      end
+
+      it "removes the installation id from user"
```
salbertson: @gylaz should we remove the installation id from the user record too?

app/models/github_event.rb
```diff
@@ -22,6 +24,10 @@ def process
         github_id: body["marketplace_purchase"]["account"]["id"],
       )
       owner.active_private_repos.each(&:deactivate)
+    when INSTALLATION
+      update_installation
+    when INSTALLATION_REPOSITORIES
+      update_installation
```
gylaz: `when INSTALLATION, INSTALLATION_REPOSITORIES` since they do the same thing?

app/models/github_event.rb
```diff
@@ -53,6 +59,25 @@ def update_purchase
...
+      repos = Repo.where(
+        installation_id: body["installation"]["id"],
+        github_id: body["repositories_removed"].map { |repo| repo["id"] },
+      )
+      repos.update_all(active: false, installation_id: nil)
+    when "added"
+      repos = Repo.where(
+        github_id: body["repositories_added"].map { |repo| repo["id"] },
+      )
+      repos.update_all(installation_id: body["installation"]["id"])
```
gylaz: Do you think the users would expect these repos to "just be enabled", when they give permissions over on GitHub's end, or no?

spec/models/github_event_spec.rb
```diff
@@ -66,5 +66,111 @@
...
+        repo = create(:repo, :active, owner: owner, installation_id: 2)
+        body = JSON.parse(read_fixture("github_app_uninstall.json"))
+        event = described_class.new(
+          type: GitHubEvent::INSTALLATION,
+          body: body,
+        )
+
+        event.process
+
+        expect(repo.reload.installation_id).to be_nil
```
gylaz: What do you think about doing this check in the first test as well? Seems like a large setup, just for one expectation. Could do it in one check via:
```rb
expect(repo.reload).to have_attributes(
  active: false,
  installation_id: nil,
)
```

spec/models/github_event_spec.rb
gylaz: Yes, I think so.

app/models/github_event.rb
salbertson: Are you asking if Hound should be enabled when users provide additional repo access? I don't think so, especially if they choose an entire org's repos.

spec/models/github_event_spec.rb
salbertson: I like it. Done.

app/models/github_event.rb
```diff
@@ -48,6 +58,34 @@ def update_purchase
...
+      repos.update_all(active: false, installation_id: nil)
+
+      users = User.where(
+        "? = ANY(installation_ids)",
+        body["installation"]["id"],
+      )
+      users.find_each do |user|
+        user.installation_ids.delete(body["installation"]["id"])
+        user.save!
+      end
```
gylaz: How about pulling this part out into a method? `update_installation` is getting pretty massive.

PR #1592, opened 2018-10-12
Notify Sentry when there's an issue updating any GitHub status

salbertson:


app/models/commit_status.rb
```diff
@@ -57,7 +65,13 @@ def set_past_due_status(invoice_url)
   def create_error_status(repo_name:, sha:, message:, url: nil)
     github.create_error_status(repo_name, sha, message, url)
   rescue Octokit::NotFound
-    # noop
+    notify_sentry(
+      "Failed to set error status",
+      repo_name: repo_name,
+      sha: sha,
```
gylaz: Since these are instance variables, how about merging them into `metadata` in `notify_sentry`? So we're not passing them around.

PR #1597, opened 2018-10-17
Include configured linter version with build job

salbertson:


app/models/linter/base.rb
```diff
@@ -47,7 +47,11 @@ def build_review_job_attributes(commit_file)
         linter_name: name,
         patch: commit_file.patch,
         pull_request_number: build.pull_request_number,
-      }
+      }.tap do |attributes|
+        if version.present?
+          attributes[:linter_version] = version
```
salbertson: @gylaz should we include `linter_version` value even when it's `nil`, so the "schema" doesn't change?

spec/models/linter/base_spec.rb
```diff
@@ -1,21 +1,69 @@
...
+        hound_config = double("HoundConfig", linter_version: nil)
+        linter = build_linter(build: build, hound_config: hound_config)
+        commit_file = build_commit_file(filename: "wat.txt")
+        build_config = double("BuildConfig", serialize: "config")
+        allow(BuildConfig).to receive(:call).and_return(build_config)
+        allow(FileReview).to receive(:create!)
+
+        expect(Resque).to receive(:enqueue).with(
+          TestReviewJob,
+          hash_excluding(:linter_version)
```
gylaz: I think it's probably fine to send an empty value here if `linter_version` is not defined in the repo's config. It's probably easier to check against some known empty value than play the add/remove game.

spec/models/linter/base_spec.rb
```diff
@@ -1,21 +1,69 @@
...
+          pull_request_number: 123
+        )
+        hound_config = double("HoundConfig", linter_version: nil)
+        linter = build_linter(build: build, hound_config: hound_config)
+        commit_file = build_commit_file(filename: "wat.txt")
+        build_config = double("BuildConfig", serialize: "config")
+        allow(BuildConfig).to receive(:call).and_return(build_config)
+        allow(FileReview).to receive(:create!)
+
+        expect(Resque).to receive(:enqueue).with(
```
gylaz: We try to not put expectations before the exercise phase. How about we use `allow` here, then `expect` after the actual call?

spec/models/linter/base_spec.rb
```diff
@@ -1,21 +1,69 @@
...
+          hash_excluding(:linter_version)
+        )
+
+        linter.file_review(commit_file)
+      end
+    end
+
+    context "when a linter version is configured" do
+      it "enqueues a job with the linter version" do
+        build = double(
```
gylaz: `instance_double`s everywhere?

app/models/linter/base.rb
gylaz: Yes, I think that it's simpler that way.

PR #1598, opened 2018-10-18
Choose token that has permissions to set PR status

gylaz:


app/models/github_auth.rb
```diff
@@ -15,20 +15,19 @@ def token
...
 
-  def can_reach_repository?(user)
-    if GitHubApi.new(user.token).repository?(repo.name)
-      true
-    else
-      repo.remove_membership(user)
-      false
-    end
+  def has_pr_status_permissions?(user)
+    GitHubApi.new(user.token).statuses(repo.name, "master")
```
salbertson: Does reading statuses mean they can write?

app/models/github_auth.rb
```diff
@@ -15,20 +15,19 @@ def token
   def user
     if repo.installation_id.nil?
       @_user ||= users_with_token.shuffle.detect(-> { hound_user }) do |user|
-        can_reach_repository?(user)
+        has_pr_status_permissions?(user)
```
salbertson: What do you think about `can_access_github_statuses?`?

app/models/github_auth.rb
gylaz: Dang, good point. I did some more testing, and it appears not to be viable in some situations.

app/models/github_auth.rb
gylaz: Maybe even better `can_set_pr_status?` (since this class is already called GitHubStatus mentioning GitHub again is probably not necessary).

app/models/github_auth.rb
salbertson: 😞

PR #1601, opened 2018-10-19
Remove user from repo if they can't set pr status

gylaz:
This will allow us to cleanup users who no longer have access to set the
status on a repo. Since we retry a failed job, we should get a good
token eventually.

There is a bit of a drawback, that we'll end-up using Hound's token more
for public repos, but that seems like a fair trade off.


app/models/commit_status.rb
```diff
@@ -1,84 +1,72 @@
...
-    )
+    create_error_status(I18n.t(:past_due_status), invoice_url)
   end
 
   private
 
-  def create_error_status(repo_name:, sha:, message:, url: nil)
-    github.create_error_status(repo_name, sha, message, url)
+  def remove_user_from_repo
+    github_auth.user && repo.remove_membership(github_auth.user)
```
salbertson: This just remove the `memberships` record, right?

app/models/commit_status.rb
```diff
@@ -1,84 +1,72 @@
...
+    github.create_error_status(repo.name, sha, message, url)
   rescue Octokit::NotFound
-    notify_sentry(
-      "Failed to set error status",
-      message: message,
-      url: url
-    )
+    notify_sentry("Failed to set error status", message: message, url: url)
+    remove_user_from_repo
+    raise
```
salbertson: What do you think about moving the similar rescue logic to a method?

app/models/commit_status.rb
gylaz: Yes

app/models/commit_status.rb
gylaz: I suppose, we could move `notify_sentry` and `remove_user_from_repo` into a method.
But I do think that having `raise` in the `rescue` block is valuable -- it's makes it hard to miss that we're re-raising. Otherwise, if we hide `raise` into a method, then someone reading this can easily mistake this for preventing the error from being raised. Thoughts?

salbertson: If you think it's clearer to leave the raise then I'd leave it. Was just a thought, It's pretty clear what's going.

PR #1607, opened 2018-10-20
Rename CoffeeScript to CoffeeLint

gylaz:


spec/models/hound_config_spec.rb
```diff
@@ -123,13 +123,13 @@
       it "converts them and returns true" do
         commit = stub_commit(
           ".hound.yml" => <<~EOS
-            coffeescript:
+            sass-lint:
```
salbertson: Why `sass-lint`?

PR #1608, opened 2018-10-21
Rename BuilRunner to StartBuild

gylaz:
It follows more of our convestion of naming services after what they do,
using a verb first. For example, `CreateRepo` or `SubmitReview`.

`StartBuild` fits nicely with the `CompleteBuild` service that we have.

### DIFF
--- a/app/services/start_build.rb
+++ b/app/services/start_build.rb
@@ -1,4 +1,4 @@
-class BuildRunner
+class StartBuild
   static_facade :call
   pattr_initialize :payload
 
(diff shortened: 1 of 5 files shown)


app/services/start_build.rb
```diff
@@ -110,4 +106,8 @@ def report_config_file_as_invalid(exception)
   def set_no_violations_status
     commit_status.set_success(0)
   end
+
+  def set_internal_error
+    commit_status.set_internal_error
+  end
```
salbertson: 🙌

PR #1610, opened 2018-10-21
Rename Haml to HamlLint

gylaz:
To follow the convention of naming linters after their action names.


app/models/linter/haml_lint.rb
```diff
@@ -0,0 +1,11 @@
+module Linter
+  class HamlLint < Base
+    FILE_REGEXP = /.+\.haml\z/
+
+    private
+
+    def job_class
+      LintersJob
```
salbertson: Does it make sense to return `LintersJob` from `Base` method and implement `job_class` for linters that still use a dedicated queue?

gylaz: Yes, that's better than what we have. But I was thinking of doing a PR, to switch the remaining ones over to `LintersJob`.

PR #1611, opened 2018-10-21
Rename Scss to ScssLint

gylaz:


app/models/linter/scss_lint.rb
```diff
@@ -0,0 +1,11 @@
+module Linter
+  class ScssLint < Base
+    FILE_REGEXP = /.+\.scss\z/
+
+    private
+
+    def job_class
+      LintersJob
```
salbertson: Is this our new "base" behavior?

gylaz: Yes, I'll take care of this in a followup PR.

PR #1613, opened 2018-10-24
Add companies to landing page and make minor changes to content

salbertson:


app/controllers/home_controller.rb
```diff
@@ -5,6 +5,144 @@ class HomeController < ApplicationController
...
+      },
+      {
+        name: "Lumosity",
+        logo: "companies/lumosity.png",
+      },
+      {
+        name: "Blendtec",
+        logo: "companies/blendtec.png",
+      },
+    ]
```
gylaz: What do you think about pulling these from a yaml file?

salbertson: Sure, might make it easier to manage. I did the same for supported languages too.

PR #1615, opened 2018-10-24
Use GitHub App to access repos

salbertson:
@gylaz what do you think about doing something like this? We should have less issues with invalid tokens and users will have more control over access, plus it's not as scary as giving Hound full private access.


gylaz:
Seems nice! Have you tried if GitHub will redirect back to us on installation?

PR #1625, opened 2018-10-30
Use ActiveJob to process CompletedFileReviewJob

salbertson:


spec/jobs/completed_file_review_job_spec.rb
```diff
@@ -14,25 +22,12 @@
...
-    end
-
-    context "when Resque process is killed" do
-      it "enqueues job" do
-        allow(CompleteFileReview).to receive(:call).
-          and_raise(Resque::TermException.new(1))
-        allow(Resque).to receive(:enqueue)
 
-        CompletedFileReviewJob.perform(attributes)
+        expect_any_instance_of(CompletedFileReviewJob).to(
```
gylaz: I see you keep doing this way. Do you prefer it?
I've always broke it up like this in tests:
```rb
expect_any_instance_of(CompletedFileReviewJob).
  to receive(:retry_job).with(wait: 30.seconds)
```

salbertson: I like it because it's following our standard way to break up a chain of method calls, not a special case for tests and readability.

salbertson:
@gylaz should we use `HashWithIndifferentAccess` or be strict when accessing job arguments?

https://github.com/houndci/hound/pull/1625/commits/8658e5f3501708dcae6aea74468ecc4d1135931c

gylaz:
@salbertson Strict is probably fine. The code that pulls out the attributes doesn't change much.

spec/jobs/completed_file_review_job_spec.rb
gylaz: But... I've never seen parens after `.to` in tests. There will always ever be only one argument.

> it's following our standard way to break up a chain of method calls

Our standard way to break up chain of method calls is to break after the period. Are we talking about different things here? To break up on parens like that, is to break up method calls with many/long arguments.

salbertson: > Are we talking about different things here?

I don't think so, but I you're right. Take a look at the recent commit and let me know if that's what you'd expect. I'll try to stick to what you are recommending.

spec/jobs/completed_file_review_job_spec.rb
```diff
@@ -14,36 +22,22 @@
...
-
-    context "when Resque process is killed" do
-      it "enqueues job" do
-        allow(CompleteFileReview).to receive(:call).
-          and_raise(Resque::TermException.new(1))
-        allow(Resque).to receive(:enqueue)
 
-        CompletedFileReviewJob.perform(attributes)
+        expect_any_instance_of(CompletedFileReviewJob).
+          to(receive(:retry_job).with(wait: 30.seconds))
```
gylaz: Parens after `to` are very un rspec-like :)

PR #1629, opened 2018-11-01
Default LintersJob as the default job for linters

gylaz:
Only a handful of linters still queue to their specific queues.
Everything else is moving to the `linters` queue.

### DIFF
--- a/app/models/linter/credo.rb
+++ b/app/models/linter/credo.rb
@@ -1,5 +1,11 @@
 module Linter
   class Credo < Base
     FILE_REGEXP = /.+(\.ex|\.exs)\z/
+
+    private
+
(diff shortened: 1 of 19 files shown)


app/models/linter/credo.rb
```diff
@@ -1,5 +1,11 @@
 module Linter
   class Credo < Base
     FILE_REGEXP = /.+(\.ex|\.exs)\z/
+
+    private
+
+    def job_class
+      CredoReviewJob
```
salbertson: Did you use `job_name.constantize` to generate these?

gylaz: I just typed them out to match the job class name.

PR #1638, opened 2018-11-15
Add support for PHP_CodeSniffer

gylaz:


app/models/hound_config.rb
```diff
@@ -11,6 +11,7 @@ class HoundConfig
     Linter::HamlLint => { default: true },
     Linter::Jshint => { default: true },
     Linter::Reek => { default: false },
+    Linter::Phpcs => { default: true },
```
salbertson: What do you think about disabling new linters by default and allowing users to opt in? Anyone with Hound enabled will immediately see PHP reviews when this is released, whether they want them or not. It might be fine, just curious what you think.

gylaz: I thought about along the same lines too. How many of our current users are also committing PHP changes? Maybe ones that are using JS or CSS linters. My intentions were to have this enabled as it might become one of the core linters. However, we can always make it the default later, and I think it does make sense to keep it *off* by default for now.

salbertson: Sounds good. I've come across a few open source projects using Hound that also have PHP files, so it's not rare.

PR #1664, opened 2019-01-23
Process high jobs if medium and low queues are empty

salbertson:
This should help keep up with spikes in `high` jobs.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,7 +1,7 @@
 web: bundle exec puma -C config/puma.rb
 webpack_dev_server: ./bin/webpack-dev-server
-resque_high: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=high,medium bundle exec rake resque:work
-resque_medium: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=medium bundle exec rake resque:work
-resque_low: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=low,medium bundle exec rake resque:work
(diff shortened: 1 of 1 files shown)


Procfile
```diff
@@ -1,7 +1,7 @@
 web: bundle exec puma -C config/puma.rb
 webpack_dev_server: ./bin/webpack-dev-server
-resque_high: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=high,medium bundle exec rake resque:work
-resque_medium: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=medium bundle exec rake resque:work
-resque_low: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=low,medium bundle exec rake resque:work
+resque_high: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=high bundle exec rake resque:work
+resque_medium: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=medium,high bundle exec rake resque:work
+resque_low: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=low,high,medium bundle exec rake resque:work
```
gylaz: Does the order of these matter? `low,high,medium` seems like a weird order, if it doesn't.

salbertson: Yes, the order matters. `low` jobs will be processed first, then `high`, and then `medium`.

I basically want `high` jobs processed when `low` or `medium` jobs are clear.

PR #1666, opened 2019-01-24
Don't delete Stripe subscription when creation fails

salbertson:
A uniqueness constraint prevented a `subscriptions` record from being created, which lead to the Stripe subscription being deleted. This change adjusts the way we deal with cleaning up Stripe subscriptions if there's an issue on our side (uniqueness constraint, etc.). Now we don't attempt to clean up the Stripe subscription when there's an exception.

### DIFF
--- a/app/services/repo_subscriber.rb
+++ b/app/services/repo_subscriber.rb
@@ -35,13 +35,13 @@ class RepoSubscriber
       repo_id: repo.id,
     )
 
+    payment_gateway_subscription.subscribe(repo.id)
+
     repo.create_subscription!(
       user_id: user.id,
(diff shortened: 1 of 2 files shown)


gylaz:
I would maybe re-phrase the commit message a bit. `This is not what we want` isn't exactly correct. 99% of the time, we would want to clean-up the new subscription that was created on Stripe -- which is why we had this code. In the recent issue we ran into, it deleted an existing, long-running subscription, which has never happened before. I think we're just being more careful now, to avoid the situation again (however, potentially fine with leaving around a subscription on Stripe, that failed on our side).

PR #1673, opened 2019-02-15
Limit retry attempts of complete file reviews

gylaz:
If there's an `ActiveRecord::RecordNotFound` the jobs will retry
forever. We should only retry the `retry_attempts` times


salbertson:
Looks good to me. 👍

PR #1684, opened 2019-03-16
Move from Resque to Sidekiq for all jobs

gylaz:


Procfile
```diff
@@ -1,7 +1,6 @@
 web: bundle exec puma -C config/puma.rb
 webpack_dev_server: ./bin/webpack-dev-server
-resque_high: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=high bundle exec rake resque:work
-resque_medium: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=medium,high bundle exec rake resque:work
-resque_low: env TERM_CHILD=1 RESQUE_TERM_TIMEOUT=8 QUEUE=low,high,medium bundle exec rake resque:work
-resque_scheduler: bundle exec rake resque:scheduler
+jobs_high: bundle exec sidekiq -q high
+jobs_medium: bundle exec sidekiq -q medium -q high
+jobs_low: bundle exec sidekiq -q low -q high -q medium
```
salbertson: We can now process all queues using a single configuration and one process type on Heroku.

`bundle exec sidekiq -C config/sidekiq.yml`

and

```
---
:concurrency: 2
staging:
  :concurrency: 4
production:
  :concurrency: <%= Integer(ENV.fetch("SIDEKIQ_CONCURRENCY", 5)) %>
:queues:
  - ["high", 10]
  - ["medium", 2]
  - ["low", 1]
```

app/controllers/admin/job_failures_controller.rb
```diff
@@ -1,14 +1,13 @@
 module Admin
   class JobFailuresController < Admin::ApplicationController
```
salbertson: Will we need this after switching to Sidekiq?

app/jobs/application_job.rb
```diff
@@ -1,21 +1,3 @@
...
-
-  around_perform do |_, block|
-    begin
-      block.call
-    rescue Resque::TermException
-      retry_job
-    end
-  end
+class ApplicationJob
+  include Sidekiq::Worker
```
salbertson: 👏

app/jobs/application_job.rb
```diff
@@ -1,21 +1,3 @@
-class ApplicationJob < ActiveJob::Base
-  include Retryable
-
-  cattr_accessor :timeout
-
-  rescue_from(Octokit::Unauthorized) do |exception|
```
salbertson: Are we going to allow these to fail multiple times? Fine with me if that's simpler.

app/jobs/buildable.rb
```diff
@@ -1,4 +1,14 @@
 module Buildable
+  extend ActiveSupport::Concern
+
+  included do
+    sidekiq_retries_exhausted do |job|
```
salbertson: Won't this wait until almost a month before setting the error status?

app/services/complete_file_review.rb
```diff
@@ -2,7 +2,7 @@ class CompleteFileReview
   static_facade :call
 
   def initialize(attributes)
-    @attributes = attributes
+    @attributes = attributes.deep_symbolize_keys
```
salbertson: Why do we now need to symbolize keys?

config/initializers/sidekiq.rb
```diff
@@ -1,10 +0,0 @@
-require "sidekiq"
-require "redis-namespace"
-
-Sidekiq.configure_client do |config|
-  config.redis = { namespace: "resque" }
-end
-
-Sidekiq.configure_server do |config|
-  config.redis = { namespace: "resque" }
```
salbertson: @gylaz I believe we'll have to coordinate a similar change to linters as well.

config/routes.rb
```diff
@@ -17,7 +17,6 @@
     root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
   end
 
-  mount Resque::Server, at: "/resque"
   mount Sidekiq::Web, at: "/sidekiq"
```
salbertson: We could rename this route to `/queue`.

config/sidekiq.yml
```diff
@@ -0,0 +1,3 @@
+:timeout: <%= ENV.fetch("JOB_TIMEOUT", 120).to_i %>
+:concurrency: <%= ENV.fetch("SIDEKIQ_CONCURRENCY", 2).to_i %>
+:max_retries: <%= ENV.fetch("JOB_RETRY_ATTEMPTS", 5).to_i %>
```
salbertson: Ah, we could add queue configuration I mentioned to this.

Procfile
gylaz: That only says which queue to work on and the priority, but no way to specify priority per process/worker.
What we currently have is a dedicated `low` worker that works on low jobs first, but then when idle will help out with medium and hight. Same for `medium` queue (which is 99% of our build jobs), it was a dedicated worker (which helps out with high).

app/controllers/admin/job_failures_controller.rb
gylaz: I think so. I use this page to get an better idea of why things are failing.

app/jobs/application_job.rb
gylaz: For now, yes. If it gets too noisy we can catch these on single-case basis.

app/jobs/buildable.rb
gylaz: No, we have set `max_retries` to 3.

app/services/complete_file_review.rb
gylaz: Because the keys from Sidekiq come back as strings: `{ 'violations' => ...}`, and it's easier to work with symbols, plus less git noise. Thoughts?

config/initializers/sidekiq.rb
gylaz: Yep, done in https://github.com/houndci/linters/pull/271

config/routes.rb
gylaz: What do you think of `/jobs`? Queue can be tricky to type.

salbertson: That's fine with me. 👍

app/services/complete_file_review.rb
salbertson: Sounds good, as long as it's not risky.

app/jobs/buildable.rb
salbertson: How fast could those retries be run?

app/controllers/admin/job_failures_controller.rb
salbertson: Have you used the Retries tab much in Sidekiq UI? We should be able to get most of what we need from that or Sentry.

Procfile
salbertson: What I'm saying is I don't think we need different types of worker processes. We could have one type of worker (dyno) in Heroku and adjust the weights for each queue. This simplifies our setup, makes sure workers always have something to do, and still gives us flexibility to process more high or medium jobs if necessary.

gylaz: Should we worry about being blocked? Imagine something like this:

- We get a bunch of jobs in `high` and `medium` (which have higher priority than `low`)
- All of the workers are busy working off those two queues
- `low` queue is neglected due to having lower priority

Is that a situation we should anticipate? I'm okay if we don't, since it makes config simpler, but I was trying to preserve historical functionality.

app/controllers/admin/job_failures_controller.rb
gylaz: This would be the Dead tab in resque -- where failed jobs show up. Yes, Resque UI had that functionality. However, This controller groups errors by similar messages, which is easier to see, and remove groups/batches. When there are 3,000 some failed jobs, it's hard to see failure trends when starting at a really long list.

app/jobs/buildable.rb
gylaz: From Sidekiq docs:
```
 # | Next retry backoff | Total waiting time
 -------------------------------------------
 1 |     0d  0h  0m 30s |     0d  0h  0m 30s
 2 |     0d  0h  0m 46s |     0d  0h  1m 16s
 3 |     0d  0h  1m 16s |     0d  0h  2m 32s
```
There's also a way to set a flat wait time with `sidekiq_retry_in`.

Procfile
gylaz: Basically, the current setup ensures there's always a worker to work on a given queue, but when that worker is idle it works on other queues as well (rather than sitting idle). I'm not sure if it's needed though -- the `medium` queue is for most normal builds (plus repo refreshes and invite acceptances), the `high` queue is to process the results of linters, and `low` queue is for large/slow builds. The `medium` queue can't really get blocked by `high` because in order for `high` to get backed up, it would need to process a bunch of `medium` build jobs first (so that's okay). The `low` can be potentially blocked, but I'm guessing that's okay -- those jobs are for huge PRs that we don't favor anyway?

Additionally, does it make sense to use `medium` for `CompletedFileReviewJob` for parity with `SmallBuildJob`?

app/jobs/buildable.rb
salbertson: Cool, is a couple minutes enough time? Just want to make sure things like invitation jobs and rate limits don't fail because we retried too quickly.

app/controllers/admin/job_failures_controller.rb
salbertson: Failed jobs show up in Retries whenever they fail and are rescheduled. This is usually where I go to find issues. Just making sure we aren't rebuilding the Sidekiq UI and saving all this job data if we don't have to. I'm cool with it if it's useful to you.

gylaz: Retries happen when we aren't watching the queue (we only retry 3 times), then die.
The Dead tab is where jobs go after all retries have exhausted. I'd say that's what we'd be checking the most (when debugging something or seeing why jobs fail/die).
We get quite a bit of "dead" jobs over a few days.
Over a week or two there are thousands.
It's hard to scan that list to see if there are some common trends.
In this regard, Sidekiq UI doesn't have anything "new" that Resque didn't have.

This admin page, groups the errors to make trend identification easier (if a bunch of failures are coming from one repo). It was a pain-point that I had when working on Hound's stability, and adding this page helped me.

It doesn't store any data. It uses Sidekiq API to fetch the data from Redis (same place Sidekiq pulls the data from). There's not much code. I don't think we need to remove this, as I find it useful for now.

app/jobs/buildable.rb
gylaz: Previously, we retried every thirty seconds. So, this is even a longer wait.

PR #1688, opened 2019-03-22
Rename Sidekiq UI password variable

gylaz:
To follow the `/jobs` path and other variables that start with `JOB_`.


gylaz:
Already added a variable to production with this name.

PR #1690, opened 2019-03-23
Group dead jobs that failed due to setting status

gylaz:
This reducers the noise on the `/admin/job_failures` page, and let's us
focus on more important failures.


salbertson:
@gylaz looks good to me. 👍

PR #1698, opened 2019-04-01
Use the GitHub app for all users

gylaz:


gylaz:
@salbertson This one is ready for review.

app/models/app_token.rb
```diff
@@ -6,7 +6,7 @@ def initialize
 
   def generate
     issue_time = Time.now.to_i
-    expiration_time = issue_time + (10 * 60)
+    expiration_time = issue_time + (10 * 60 - 10)
```
salbertson: What's the adjustment to the expiration for?

gylaz: There's a small time frame window where a token might be expired on GitHub's end, but not our end -- a few seconds (request in transit, etc). This is just a 10 second window to allow for those inconsistencies.

PR #1699, opened 2019-04-02
Added .DS_Store to .gitignore

jhsu802701:
This prevents Mac users from accidentally adding .DS_Store to the source code.  (Yes, I've been on a project that had several instances of the .DS_Store file.)


gylaz:
Not all of us use Macs. This would be better ignored in your top-level, home directory gitignore.

PR #1702, opened 2019-04-02
Ran database migration,  updated db/schema.rb

jhsu802701:


gylaz:
I got this same diff. Thanks, merging.

PR #1705, opened 2019-04-05
Don't build webhook fore repo with active installation_id

gylaz:
We're processing some repo PRs twice because they have an active webhook
and they've installed our GitHub app, thus we're getting notified twice.

This ensures we don't process the webhooks if a repo has an active
`installation_id`.


gylaz:
[Hound reports an invalid number of violations](https://trello.com/c/Pf8c3vpB/891-hound-reports-an-invalid-number-of-violations)

PR #1712, opened 2019-04-25
Serialize empty configs as empty strings

gylaz:
Otherwise, the config was being serialized as a literal empty hash,
while most of the configs are servialized to text (then parsed on the
`linters` side).

This fixes an error reported by the `linters`.

### DIFF
--- a/app/models/config/base.rb
+++ b/app/models/config/base.rb
@@ -14,7 +14,7 @@ module Config
     end
 
     def serialize
-      content
+      raise NotImplementedError, "implement ##{method} in your config class"
     end
 
     def linter_name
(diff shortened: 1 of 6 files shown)


app/models/config/base.rb
```diff
@@ -14,7 +14,7 @@ def content
     end
 
     def serialize
-      content
+      raise NotImplementedError, "implement ##{method} in your config class"
```
salbertson: I haven't used this `method` trick. Are you sure it works?

https://ruby-doc.org/core-2.2.0/Object.html#method-i-method

gylaz: This comes from our `linters` repo, where we use it in the `.../options/base.rb`.

PR #1716, opened 2019-04-27
Updated success copy

TomasBarry:


gylaz:
While I love the theme, I think this one might be a bit confusing. @salbertson thoughts?

hlpatrick93:
Sounds like a really nice idea!

gylaz:
@TomasBarry Sorry it took this long to get back to it. We do love the idea of Hound messaging having character. We talked about it and had some concerns about `No bones to pick` being explanatory enough. We decided that `Smells good to me. Woof!` would be a better fit (a play on LGTM and goes nicely together with pending review message). If you're able to make that change, we can merge your PR. Otherwise, we can make the change ourselves.

salbertson:
Thanks @TomasBarry!

TomasBarry:
@gylaz, @salbertson - I've updated the PR based on your comments. Thanks for looking!

gylaz:
Thank you, merged!

PR #1730, opened 2019-07-10
Don't get the email from Stripe on every page

gylaz:
We currently are calling Stripe on each page load/reload to fetch the
user billable email.

This email is only needed during Plan purchase/upgrade.

Currently, Stripe API is having issues and our pages won't load.


salbertson:
🚢

PR #1731, opened 2019-07-13
Refactored actions_controller_spec

serjpetrenko:
Overview:

- Refactored actions_controller_spec to avoid multiple create action to the database.

### DIFF
--- a/spec/controllers/activations_controller_spec.rb
+++ b/spec/controllers/activations_controller_spec.rb
@@ -75,7 +75,7 @@ RSpec.describe ActivationsController do
             properties: {
               name: repo.name,
               private: false,
-            }
+            },
           )
       end
     end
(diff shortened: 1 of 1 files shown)


gylaz:
Thanks, but we don't use `let` or `before` in this codebase.
For reasoning, you can check out: https://thoughtbot.com/blog/lets-not

If there are specific techniques you utilized here "to avoid multiple create action to the database", I'd love to discuss them, and see how they can be applied.

serjpetrenko:
@gylaz Thanks for your response. I think there is nothing to discuss because you don't use `let` and `before`. The main idea of my refactoring avoids multiple create action for `repo` and `membership` variables and just create it once at the start of the test. I think this PR can be closed.

PR #1738, opened 2019-08-24
Add .ts and .tsx support

brendanoffer:
Add TS and TSX support to ESLint.

Typescript linting has been done in the past by TSLint, which is now being deprecated. The TSLint team [is pointing users to use @typescript-eslint ](https://medium.com/palantir/tslint-in-2019-1a144c2317a9)for typescript support; the Microsoft Typescript team, themselves, have been [transferring their linting capabilities to use ESLint](https://github.com/microsoft/TypeScript/issues/30553).

If users do not want their typescript files to be linted support they can simply add `*.ts` and `*.tsx` to their `.eslintignore` file.

### DIFF
--- a/app/models/linter/eslint.rb
+++ b/app/models/linter/eslint.rb
@@ -1,6 +1,6 @@
 module Linter
   class Eslint < Base
-    FILE_REGEXP = /.+(\.js|\.es6|\.jsx|\.vue)\z/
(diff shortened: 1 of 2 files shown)


gylaz:
Thanks @brendanoffer! Merging this as-is. We'll see if we get any feedback about potentially linting TS files with both ESLint and TSLint (and how often that arises).

PR #1741, opened 2019-09-09
Configure OmniAuth to get the GitHub user's email

gylaz:
GitHub Apps don't use scopes, but OmniAuth GitHub library still needs
this option, to know to hit the GitHub API to fetch the email for the
user that is logging in.


salbertson:
🥇

PR #1748, opened 2019-10-20
Fix spelling error on README.md

trillyon:

PR #1749, opened 2019-10-20
fix spelling error on README.md

trillyon:
so i can't do research (as shown on #1748) so i did a slash instead.


trillyon:
this is so useless.. im just gonna close it now

PR #1752, opened 2019-11-08
Suggested changes

salbertson:


spec/services/suggest_changes_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      COMMENT
+    end
+
+    it "makes suggestion for missing space after comma" do
+      violation = instance_double(
+        "Violation",
+        messages: [
+          "A space is required after ',' comma-spacing"
+        ],
+        source: "function wat(one,two, three) {"
```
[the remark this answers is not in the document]
salbertson: @gylaz it would be nice to have a suggestion here. 😄 

```suggestion
        source: "function wat(one,two, three) {",
```

app/services/complete_file_review.rb
```diff
@@ -27,7 +27,13 @@ def complete_file_review
   def build_file_review_violations
     attributes.fetch(:violations).each do |violation|
       line = commit_file.line_at(violation.fetch(:line))
-      file_review.build_violation(line, violation.fetch(:message))
+
+      # we need to make sure all jobs are complete before relying on source
+      file_review.build_violation(
+        line,
+        violation.fetch(:message),
+        violation.fetch(:source)
```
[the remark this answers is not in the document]
gylaz: Looks like these blocks are whitespace aware, so you have to add matching amount of spacing.
```suggestion
        violation.fetch(:source),
```

app/services/suggest_changes.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  attr_reader :violation
+  attr_accessor :suggestion
+
+  def apply_suggestion(message)
+    case message
+    when /Missing semicolon/
+      self.suggestion = violation.source << ";"
+    when /A space is required after ','/
+      self.suggestion = violation.source.gsub(/(,)([^ ])/, ', \2')
```
gylaz: Hmm, I think if we want to apply multiples of these to the same "fixed" version, we need to do something like:
```rb
self.suggestion = (suggestion.presence || violation.source).gsub(/(,)([^ ])/, ', \2')
```
Prob makes sense to extract `(suggestion.presence || violation.source)` to a method?

spec/requests/builds_spec.rb
```diff
@@ -12,9 +12,21 @@
 
   context "with violations" do
     it "makes a new comment and cleans up resolved one" do
-      existing_comment_violation = { line: 5, message: "Line is too long." }
-      new_violation1 = { line: 3, message: "Trailing whitespace detected." }
-      new_violation2 = { line: 9, message: "Avoid empty else-clauses." }
+      existing_comment_violation = {
+        line: 5,
+        message: "Line is too long.",
+        source: "debugger",
```
gylaz: Does it makes sense to make this a bit more code-like? (e.g. `foo = bar(1, 2)`)
When I saw `source: "debugger"` it confused me for a sec.

app/services/complete_file_review.rb
```diff
@@ -27,7 +27,13 @@ def complete_file_review
   def build_file_review_violations
     attributes.fetch(:violations).each do |violation|
       line = commit_file.line_at(violation.fetch(:line))
-      file_review.build_violation(line, violation.fetch(:message))
+
+      # we need to make sure all jobs are complete before relying on source
```
gylaz: What do you mean by this? Why does it matter if all jobs are complete or not?

salbertson: Never mind, I think we are safe if linters is released first and `source` is always available. I was thinking some scheduled jobs wouldn't have `source`.

app/services/complete_file_review.rb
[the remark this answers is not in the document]
salbertson: 😢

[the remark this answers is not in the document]
salbertson: @gylaz is there any reason the leading whitespace wouldn't be preserved when saving `source`?

app/services/suggest_changes.rb
salbertson: I ran into a case that makes me think we should only make one suggestion at a time. Here's an example where combining suggestions gets tricky:

```
[
  "Put a comma after the last parameter of a multiline method call.",
  "Layout/TrailingWhitespace: Trailing whitespace detected.",
]
```

and source is `violation.fetch(:source) `

The order of our applied suggestions starts to make a big difference. If we aren't careful we end up suggesting `violation.fetch(:source) ,`.

app/models/violation.rb
```diff
@@ -14,4 +14,23 @@ def add_message(message)
...
   end
+
+  def source=(value)
+    encrypted_source = crypt.encrypt_and_sign(value)
+    self[:source] = encrypted_source
+  end
+
+  def source
+    encrypted_source = self[:source]
+    unless encrypted_source.nil?
```
gylaz: How about `if encrypted_source`?

app/services/suggest_changes.rb
```diff
@@ -0,0 +1,42 @@
...
+    @violation = violation
+    @messages = violation.messages
+    @suggestion = ""
+  end
+
+  def call
+    violation.messages.each do |message|
+      apply_suggestion(message)
+      break unless suggestion.empty?
+    end
```
gylaz: Does it make sense to loop over these, if we're going to try and only fix one of the violation messages?

app/services/suggest_changes.rb
gylaz: Hmm, okay. Should we only make suggestions on lines that only have one violation? Or should we only make a suggestion on first (or last) violation?

db/migrate/20191108222026_add_source_to_violations.rb
```diff
@@ -0,0 +1,7 @@
+class AddSourceToViolations < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
```
gylaz: What is this for?

doc/SECURITY.md
```diff
@@ -218,9 +218,10 @@ to comment about the violations on the pull request.
 
 `CompleteFileReview` saves each violation, the line number, and patch position,
 in the `violations` table of our Postgres database.
-We do not save any of your code in Postgres.
-We do store the contents of the files to review temporarily in Sidekiq, and it
-gets cleared out as soon as the job is processed.
+We store single lines of your code in Postgres, which is encrypted and encoded
```
gylaz: How about explaining this a bit more:
```
We only store single lines of code that we detected a linter violation
```
Or something? So we explain exactly which lines we store.

doc/SECURITY.md
```diff
@@ -218,9 +218,10 @@ to comment about the violations on the pull request.
...
+similar to tokens using `ActiveSupport::MessageEncryptor`.
+We also store the contents of the files to review temporarily in Sidekiq, but
```
gylaz: `Sidekiq` -> `Redis`?

app/services/complete_file_review.rb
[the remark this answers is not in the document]
gylaz: I'm actually not sure if the whitespace is preserved or not.

app/models/violation.rb
salbertson: What do you think about using a trailing `if`?

`crypt.decrypt_and_verify(self[:source]) if self[:source]`

app/services/suggest_changes.rb
salbertson: I think so because we don't know which violation we have a suggestions for until we iterate. Is there a better way?

app/services/suggest_changes.rb
salbertson: I think the first line we have a suggestion for would be a good start.

db/migrate/20191108222026_add_source_to_violations.rb
salbertson: Never mind, it was for the `algorithm: :concurrently` option but not sure why I added that to begin with. The table update should be fast since there's no default value.

app/models/violation.rb
gylaz: thoughtbot taught me that was a bad idea 😄 

Generally, I'm against it, as it's easier to miss the `if` on longer lines.

app/services/suggest_changes.rb
gylaz: I see. Ok.

`break if suggestion.present?`

app/models/violation.rb
```diff
@@ -14,4 +14,20 @@ def add_message(message)
   def messages
     self[:messages].uniq
   end
+
+  def source=(value)
+    encrypted_source = crypt.encrypt_and_sign(value)
+    self[:source] = encrypted_source
```
gylaz: Do we need this intermediate variable `encrypted_source`?

app/models/user.rb
```diff
@@ -39,14 +39,11 @@ def has_active_repos?
...
-    write_attribute(:token, encrypted_token)
+    self[:token] = encrypted_token
   end
 
   def token
-    encrypted_token = read_attribute(:token)
-    unless encrypted_token.nil?
-      crypt.decrypt_and_verify(encrypted_token)
-    end
+    crypt.decrypt_and_verify(self[:token]) if self[:token]
```
gylaz: Why did you make changes to this file?

salbertson: To keep it consistent with `Violation` class.

app/models/violation.rb
salbertson: Nope, updated.

app/models/user.rb
gylaz: The logic is changing here though. Does `crypt.decrypt_and_verify(nill)` return `nil`?

salbertson: I changed the logic back to be safe, it's been working this long without any issues. I did leave the change from `#read_attribute` to `self[]`.

app/services/suggest_changes.rb
```diff
@@ -0,0 +1,42 @@
...
+  static_facade :call
+
+  def initialize(violation)
+    @violation = violation
+    @messages = violation.messages
+    @suggestion = ""
+  end
+
+  def call
+    violation.messages.each do |message|
```
gylaz: Worth using the `messages` var here? Or maybe kill line 6?

app/services/suggest_changes.rb
```diff
@@ -0,0 +1,42 @@
...
+  def apply_suggestion(message)
+    case message
+    when /Trailing whitespace detected/
+      self.suggestion = violation.source.gsub(/(.*)\s$/, '\1')
+    when /A space is required after ','/
+      self.suggestion = violation.source.gsub(/(,)([^ ])/, ', \2')
+    when /Put a comma after the last parameter of a multiline method call/
+      self.suggestion = violation.source << ","
+    when /Missing semicolon/
+      self.suggestion = violation.source << ";"
```
gylaz: Dos `violation.source << ";"` mutate the `violation.source`? Maybe `+` or interpolation is safer?

doc/SECURITY.md
```diff
@@ -218,9 +218,11 @@ to comment about the violations on the pull request.
...
 `CompleteFileReview` saves each violation, the line number, and patch position,
 in the `violations` table of our Postgres database.
-We do not save any of your code in Postgres.
-We do store the contents of the files to review temporarily in Sidekiq, and it
-gets cleared out as soon as the job is processed.
+We also store single lines of code that are used to report violations and
+suggested changes. The data is encrypted and encoded similar to tokens using
+`ActiveSupport::MessageEncryptor`. We also store the contents of the files to
+review temporarily in Redis, but that is removed as soon as the Sidekiq job is
+processed.
```
gylaz: 👍

spec/services/suggest_changes_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+            "Put a comma after the last parameter of a multiline method call.",
+          ],
+          source: "    violation.fetch(:source)",
+        )
+        suggest_changes = described_class.new(violation)
+
+        result = suggest_changes.call
+
+        expect(result).to eq <<~COMMENT.chomp
+          Put a comma after the last parameter of a multiline method call.<br>```suggestion
```
gylaz: What do you think about adding `\n` before or after `<br>` in the code, so these assertions don't look so whacky?

app/services/suggest_changes.rb
salbertson: Yes, good call. Done.

spec/services/suggest_changes_spec.rb
salbertson: Not sure what you mean. If I do that then the tests will fail because the messages won't match. 😕

app/services/suggest_changes.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  private
+
+  attr_reader :violation, :messages
+  attr_accessor :suggestion
+
+  def apply_suggestion(message)
+    case message
+    when /Trailing whitespace detected/
```
salbertson: @gylaz we need to figure out a better way to generate these suggestions.

spec/services/suggest_changes_spec.rb
gylaz: I'm referring to this line: https://github.com/houndci/hound/pull/1752/files#diff-326c371081e0ec9a5f273d1c45b64541R20

If we add an `\n`, like:
```ruby
messages + "\n<br>```suggestion\n#{suggestion}\n```"
```
then this test line would look like:
```rb
Put a comma after the last parameter of a multiline method call.
<br>```suggestion
```

app/services/suggest_changes.rb
gylaz: Yep, though not sure how "smart" we want to get about these.

spec/services/suggest_changes_spec.rb
salbertson: @gylaz works for me, done.

PR #1766, opened 2019-12-17
Delete outdated comments without new violations

gylaz:
Currently, we will only delete outdated violations if other violations
were found. Moving this logic out of that condition, will ensure we
cleanup outdated comments regardless of finding new violations.


salbertson:
🚢

PR #1775, opened 2020-02-04
Only delete Hound's comments

gylaz:
There surfaced a situation where we deleted another bot's comments
because we detected them to be our own. If two bots evaluate a PR and
Hound finds nothing, it will attempt to cleanup other bot's comments.


app/policies/commenting_policy.rb
```diff
@@ -11,7 +11,8 @@ def comment_on?(violation)
 
   def outdated_comments(found_violations)
     comments.select do |comment|
-      comment.user.type == "Bot" && outdated_comment?(comment, found_violations)
+      comment.user.login == "hound[bot]" &&
```
salbertson: Is there any reason to check for a partial match? Some like `comment.user.login.downcase.include? "hound"`.

gylaz: Hmm, good point. This is purely just to avoid deleting someone else's comment, but you're right that our staging app will be `hound-staging[bot]`.

Not sure on the best approach here. `*hound*` seems too broad -- I bet we'll run into a situation where a user can have that in their username (e.g. `codehound` 😆 ).

How about either:
```rb
comment.user.type == "Bot" && comment.user.login.start_with?("hound")
```
or 
```rb
comment.user.login =~ /hound.*\[bot\]/
```

salbertson: Either sounds good, whatever you think.

PR #1778, opened 2020-02-19
Upgrade codebase to Rails 6

gylaz:
Ran `rails app:update`, then had to update a few gems and modify a few
places in code, to get this to work.


gylaz:
Sanity tested this on staging -- both the UI and PR reviews. ✅

PR #1794, opened 2020-05-28
Replace normalize-rails with npm package

tysongach:
The normalize-rails gem hasn’t been updated since 2016 and the
maintainer has ghosted.

The npm package is maintained by the person behind normalize.css, so it
see updates when necessary.

https://rubygems.org/gems/normalize-rails/


config/initializers/assets.rb
```diff
@@ -6,6 +6,4 @@
 # Precompile additional assets.
 Rails.application.config.assets.precompile += %w( repos/*.js )
 
-# webpack
-webpack_output_path = Rails.root.join("app", "assets", "webpack")
-Rails.application.config.assets.paths << webpack_output_path
```
tysongach, line 11: This appears to have been left from an older config for webpack and it’s no longer needed.

gylaz:
Looks like a bunch of CVEs are blocking this from passing.

PR #1795, opened 2020-05-29
Replace scss-lint for stylelint

tysongach:
scss-lint is no longer being actively worked on and stylelint is a more full-featured utility.

See: https://github.com/sds/scss-lint


tysongach:
[Replace scss-lint with stylelint in the Hound app](https://trello.com/c/upDeaBfO/927-replace-scss-lint-with-stylelint-in-the-hound-app)

PR #1807, opened 2020-07-24
Swap Haml for ERB

tysongach:
What do we think about starting to use ERB instead of Haml from here on out? I find Haml hard to work with and Rails’ new-ish `tag` method can help a lot with view density…


gylaz:
Personally, I don't have a problem with HAML, and like that I don't have to worry about closing tags. Any particular reasons you find HAML hard to work with?

tysongach:
@gylaz Two reasons primarily:

1. Dealing with Haml’s indentation requirements and getting attributes & values to output correctly always lands me on Haml’s documentation trying to figure out how to do it
2. I often find that semantic HTML is considered less in Haml because the elements are abstracted away, so unless one is looking at the rendered DOM it’s easy to wind up with less meaningful content structures

Have you seen Rails’ new-ish `tag` method? It’s wonderful and, often, makes for a nice compact template file (no need to close the tag, either, if you don’t open a block):

- Haml:

    ```haml
    .flash{ class: key }
      = value
    ```

- ERB:

    ```erb
    <%= tag.div value, class: "flash #{key}" %>
    ```

gylaz:
Thanks for explanation @tysongach. I don't have a strong opinion here. @salbertson WDYT?

salbertson:
@tysongach @gylaz sorry for the slow response. I'm totally ok with ditching Haml and using ERB going forward.

PR #1808, opened 2020-07-29
Pull avatars from GitHub instead of Gravatar

tysongach:
`https://github.com/<username>.png` will return a user or organization’s
avatar. It’s been written about on the GitHub Blog:
https://github.blog/2020-04-16-github-protips-tips-tricks-hacks-and-secrets-from-jason-etcovitch/#get-a-user-or-organizations-avatar

The `preconnect` is added to establish a connection to github.com
early, increasing perceived image loading performance.

No visual changes (beyond the fact that someone’s avatar will switch to their 
GitHub one, if it’s not the same as their Gravatar)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,7 +1,6 @@
 module ApplicationHelper
   def avatar_url(user)
-    gravatar_id = Digest::MD5::hexdigest(user.email.downcase)
-    "https://www.gravatar.com/avatar/#{gravatar_id}"
+    "https://github.com/#{user.username}.png?size=64"
   end
 
   def display_onboarding?
--- a/app/views/application/_app_nav.haml
+++ b/app/views/application/_app_nav.haml
@@ -46,7 +46,7 @@
         = link_to account_path, class: "app-nav__item app-nav__account-link" do
           - if current_user.email.present?
             .avatar.margin-right-x-small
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,7 +1,6 @@
 module ApplicationHelper
   def avatar_url(user)
-    gravatar_id = Digest::MD5::hexdigest(user.email.downcase)
-    "https://www.gravatar.com/avatar/#{gravatar_id}"
+    "https://github.com/#{user}.png?size=64"
```
gylaz, line 3: Should we be consistent here and call call `user.username` (even though `user.to_s` will call it)?

app/views/application/_app_nav.haml
```diff
@@ -46,7 +46,10 @@
         = link_to account_path, class: "app-nav__item app-nav__account-link" do
           - if current_user.email.present?
             .avatar.margin-right-x-small
-              = image_tag avatar_url(current_user), alt: ""
+              = image_tag avatar_url(current_user.username),
```
gylaz, line 49: This is related to the previous comment. Either stick with just `current_user` and let (`to_s` do its thing) or use `user.username` everywhere?

tysongach: Ah, yes - makes sense!

PR #1835, opened 2021-03-09
Suggestion configuration

salbertson:


spec/models/hound_config_spec.rb
```diff
@@ -201,6 +201,56 @@
     end
   end
 
+  describe "#suggestions" do
```
gylaz, line 204:
```suggestion
  describe "#suggestions?" do
```

spec/models/hound_config_spec.rb
```diff
@@ -201,6 +201,56 @@
...
+    context "when suggestions are enabled" do
+      it "returns true" do
+        commit = stub_commit(
+          ".hound.yml" => <<~CONFIG
+            rubocop:
+              suggestions: true
```
gylaz, line 210: Wouldn't it be nice for this to be enabled by default?

salbertson: Let's figure out if it's on by default after we know it's working the way we expect.

gylaz (approved):

PR #1852, opened 2021-08-01
Activate repo with existing stripe customer

gylaz:
If an org (that a given repo belongs to) already has paid/subscribed
repos, we will now use that Stripe account to allow enabling another
paid repo under that Strip account. This should prevent users from same
GitHub orgs needing to create new Stripe accounts to enable more repos.

### DIFF
--- a/app/services/repo_subscriber.rb
+++ b/app/services/repo_subscriber.rb
@@ -40,11 +40,10 @@ class RepoSubscriber
       plan: plan.id,
       repo_id: repo.id,
     )
-
     payment_gateway_subscription.subscribe(repo.id)
 
     repo.create_subscription!(
-      user_id: user.id,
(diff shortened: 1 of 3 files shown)


app/services/repo_subscriber.rb
```diff
@@ -74,8 +71,19 @@ def find_and_update_stripe_customer
...
+  def stripe_user
+    @_stripe_user ||= if user.stripe_customer_id.present?
+      user
+    else
+      same_org_repos = repo.owner.repos.joins(:subscription)
+      if same_org_repos.any?
+        same_org_repos.first.subscription.user
+      end
+    end
+  end
```
gylaz, line 83: This is pretty much the main change/logic of this PR.

PR #1854, opened 2021-08-09
Look for existing paying Stripe user first

gylaz:
This fixes the logic where a user that activates a repo already has a
Stripe account, but it's not set-up for being charged (older users).
We now will try to find an existing paying user within the same org,
before checking if the current user has a Stripe account.

### DIFF
--- a/app/services/repo_subscriber.rb
+++ b/app/services/repo_subscriber.rb
@@ -72,13 +72,22 @@ class RepoSubscriber
   end
 
   def stripe_user
-    @_stripe_user ||= if user.stripe_customer_id.present?
-      user
-    else
-      same_org_repos = repo.owner.repos.joins(:subscription)
-      if same_org_repos.any?
-        same_org_repos.first.subscription.user
+    @_stripe_user ||= begin
+      org_payee = find_org_paying_user
+
+      if org_payee.present?
+        org_payee
+      elsif user.stripe_customer_id.present?
+        user
       end
+    end
+  end
+
+  def find_org_paying_user
+    same_org_repos = repo.owner.repos.joins(:subscription)
+
+    if same_org_repos.any?
+      same_org_repos.first.subscription.user
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/services/repo_subscriber.rb
```diff
@@ -72,16 +72,25 @@ def find_and_update_stripe_customer
...
 
   def stripe_user
-    @_stripe_user ||= if user.stripe_customer_id.present?
-      user
-    else
-      same_org_repos = repo.owner.repos.joins(:subscription)
-      if same_org_repos.any?
-        same_org_repos.first.subscription.user
+    @_stripe_user ||= begin
+      org_payee = find_org_paying_user
```
salbertson, line 76: Should this be "payee" or something like `org_paying_user`?

app/services/repo_subscriber.rb
```diff
@@ -72,16 +72,25 @@ def find_and_update_stripe_customer
...
+
+      if org_payee.present?
```
salbertson, line 78: So the logic is flipped? We first try to find a repo owner (org) that is paying and fall back to a user with a Stripe customer ID?

salbertson (approved):
Any test changes? Looks good to me. 👍

app/services/repo_subscriber.rb
gylaz: or payee == "org paying user" in my mind. but if you think `org_paying_user` would be clearer, I can make that change.

app/services/repo_subscriber.rb
gylaz: Yes. The issue I encountered in prod, is that a bunch of users from the same org had already `stipe-customer_id` setup, because our previous logic would create an Stripe customer for them. So, we'd use their Stripe user, but it lacked payment method, thus the repo activation would fail.

gylaz:
> Any test changes?

Good call. I've enhanced the test, so that both users have Stripe accounts (but only one has a valid Stripe subscription).

